PR #22, opened 2018-09-05
Removed redundant pr/push build behaviour

Zeitsperre:
Currently, pushes to PRs triggers two synonymous builds. This setting only triggers builds for pushes to PRs XOR Master.

Also adding Python3.7 is a good idea in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ language: python
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


Zeitsperre:
I've disabled the 3.7 build but left it there for when more 3.7 libraries become available. Now we only build against master (as PRs XOR direct commits).

PR #23, opened 2018-09-10
Added sebfeature

Zeitsperre:
I added something that could be useful. Likely no.

### DIFF
--- a/xclim/stats.py
+++ b/xclim/stats.py
@@ -5,6 +5,8 @@ import dask
 import xarray as xr
 from scipy import stats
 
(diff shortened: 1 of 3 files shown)


xclim/stats.py
```diff
@@ -5,6 +5,8 @@
 import xarray as xr
 from scipy import stats
 
+def newfunction:
+    pass
```
Zeitsperre: Why did you do this? Idiot!

Zeitsperre:
Please please merge this!

PR #25, opened 2018-09-13
Coverage+Coveralls CI

Zeitsperre:
Attempting to set this up. Ideally, unit-testing will be done within a Tox environment and, assuming that tests pass, coverage testing will be reported to coveralls.io and report metrics here.

If anyone wants to help out (like @Balinus :+1:), feel free to add to the branch.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,8 @@
+build:
+  image: latest
+
+python:
+  version: 3.6
+  setup_py_install: true
+  extra_requirements:
+      - docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,12 @@ matrix:
 #      sudo: required
 #      dist: xenial
     - env:
+        - TOXENV=docs
+      python: 3.6
+    - env:
+        - TOXENV=flake8
+      python: 3.6
+    - env:
         - TOXENV=py36
       python: 3.6
     - env:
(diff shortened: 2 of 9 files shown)


Balinus:
The python 3.4 build errors is still the same as last week? 

Nice work! It's really useful to have the coverage tested.

Zeitsperre:
Yeah, I'm not sure where the issue is. I'll try seeing if there's an issue with the availability of Python 3.4. The vms are built on Ubuntu so it's possible that 3.4 isn't readily available (ie: must be installed) prior to testing.

Zeitsperre:
Also, please don't merge this; still several testing issues to fix.

Balinus:
Python 3.4 is indeed downloaded and installed. The stalling occurs after, when requirements are installed. Tried locally and the package that takes the most time is Spinx and guzzle-sphinx-theme. Perhaps trying without them? Are they needed actually? I mean, I thought it was documentation-related.

Balinus:
Just restarted manually the build and it worked. I often has this problem with ClimateTools.jl when building on OSX.

Zeitsperre:
That's a good point. I'll try setting tox to only build docs on Python3.6. ReadTheDocs has its own webhooks so it doesn't really matter too much whether it builds or fails in Travis/Tox. I can even look into whether RtD can publish build checks here. If we want, I can start by compartmentalizing the docs-build.

PR #35, opened 2018-09-26
added fake icclim module to access icclim related functions directly

huard:
Documentation now includes an ICCLIM page where we'll find ICCLIM indicators under their own ICCLIM standard name.

### DIFF
--- a/xclim/__init__.py
+++ b/xclim/__init__.py
@@ -95,4 +95,5 @@ def __build_icclim(mode='warn'):
     sys.modules['xclim.icclim'] = mod
     return mod
 
+
 icclim = __build_icclim('ignore')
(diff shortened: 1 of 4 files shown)


xclim/__init__.py
```diff
@@ -89,7 +91,9 @@ def __build_icclim(mode='warn'):
                'DTR': indices.daily_temperature_range,
                }
 
-    return build_module('icclim', mapping, doc="""ICCLIM indices""", mode=mode)
+    mod = build_module('icclim', mapping, doc="""ICCLIM indices""", mode=mode)
+    sys.modules['xclim.icclim'] = mod
+    return mod
```
Zeitsperre: Why is this building a module? I've never seen this before and am curious. What's gained by this behaviour?

Zeitsperre: Woops, TEDLT. Approved.

PR #45, opened 2018-10-01
Error and Warning Logging

Zeitsperre:


xclim/stats.py
```diff
@@ -15,7 +18,9 @@ def fit(arr, dist='norm'):
     # Get the distribution object
     dc = getattr(stats, dist, None)
     if dc is None:
-        raise ValueError("Statistical distribution {} is not in scipy.stats.".format(dist))
+        e = "Statistical distribution {} is not in scipy.stats.".format(dist)
+        logging.exception(ValueError(e))
```
huard: Is that still necessary ?
```suggestion
```

huard: Sounds good. Let's try to get all these PR checked-in before the end of the week.

Zeitsperre:
I think this one is good as well (for now).

There's a limitation python2.7 with `unittest` not recognizing warnings produced via the `warnings` module (ie: `warnings.warn()`) . I haven't tested out all algorithms to ensure that warnings are raised for edge-cases, but they SHOULD work and examples can be found in the `test_modules.py` to replicate assertions for raised warnings if we wish to test others in the future.

PR #46, opened 2018-10-01
Check that array has daily frequency

huard:
Using infer_freq from pandas. 
Does not support non-standard calendars. I suggest we add that to the list of xarray improvements to develop.


Zeitsperre:
@huard can this branch be closed or are you still working on it?

huard:
Can be closed, we'll complete later.

Issue #41, opened 2018-10-01
Decorators - Daily time step and Units conversion


tlogan2000:
### Description

XCLIM assumes a daily time-step (currently) : 

- [ ] Create decorator  to test monotonic daily time

All functions should also be decorated with a units converter decorator:

- [ ] Temp units to Kelvin (from C and F(?))
- [ ]  Pr units to 'mm'  (from kg m-2 s-1  and mm s-1, others?)

In my mind these replace the various 'check_valid' decorators  (which currently do not convert and simply check) and are potentially too strict on CF-Compliance ?  CF-checking could eventually be done on WPS PAVICS side for non expert users. (i.e. we assume that internal XCLIM users know what they are doing)

Balinus:
Does non-standard calendars have monotonic daily time?

huard:
Within their own calendars, yes. But the issue is to implement this check for these non standard calendars.

Issue #43, opened 2018-10-01
Add warning loggers for field changes after setting "keep_attrs" to True


Zeitsperre:
A warning should be logged when calculations of an index results in changes to the variable/units outputted. For example:

* Tmax and Tmin (K / deg C / F) -->  Growing Degree Days (sum deg C days )

Logging is a better option than printing as warnings can be selectively suppressed.

huard:
If I understand correctly, this is already handled using the with_attrs decorator in indices.py. What remains to be done is to add the decorator to each index.

huard:
@Zeitsperre Ok to close ?

Zeitsperre:
Yep, looks good to me. Thanks!

PR #47, opened 2018-10-01
implemented runtime attribute assignment in with_attrs

huard:
Related to #41


xclim/indices.py
```diff
@@ -534,6 +566,8 @@ def tx_mean(tasmax, freq='YS'):
     return arr.mean(dim='time')
 
 
+@with_attrs(standard_name='tx_min', long_name='Minimum of daily maximum temperature',
+            cell_methods='time: minimum within {freq}')
```
Zeitsperre: I believe this is the source of Python2 errors. I don't think that fstrings with keywords can be passed within function arguments. Do you know of a workaround?

Zeitsperre: Confirming this. Hard coding 'years' in `cell_methods` passes py27 build.

huard:
But is the general solution proposed fixing the issue you had in mind ?

Zeitsperre:
I think so. I just need to implement the logger using the module you pointed out but for this issue I feel that this manages to maintain/change attrs as needed and isn' clunky boilerplate code. If the py27 build can be worked around, this would effectively close #41 and #43.

huard:
I think this can be merged.

PR #49, opened 2018-10-02
Definitions and formatting

Zeitsperre:
Adding some ReST docstrings to the indices and other evaluations/functions. Feel free to correct me where I may be misinterpreting the parameters/products.

### DIFF
--- a/xclim/indices.py
+++ b/xclim/indices.py
@@ -22,12 +22,17 @@ ftomm = np.nan
 # E.g. http://vocab.nerc.ac.uk/collection/P07/current/BHMHISG2/
 
 def first_paragraph(txt):
-    """Return the first paragraph of a text."""
+    r"""Return the first paragraph of a text
+
+    Parameters
+    ----------
+    txt : str
+    """
     return txt.split('\n\n')[0]
 
 
 def with_attrs(**func_attrs):
-    """Set attributes in the decorated function at definition time,
+    r"""Set attributes in the decorated function at definition time,
     and assign these attributes to the function output at the
     execution time.
 
@@ -62,16 +67,21 @@ def with_attrs(**func_attrs):
 @with_attrs(standard_name='water_volume_transport_in_river_channel', long_name='discharge', units='m3 s-1')
 @valid_daily_mean_discharge
 def base_flow_index(q, freq='YS'):
-    """Base flow index
+    r"""Base flow index
 
     Return the base flow index, defined as the minimum 7-day average flow divided by the mean flow.
 
     Parameters
     ----------
     q : xarray.DataArray
-      The river flow [m³/s]
+      Rate of iver discharge [m³/s]
     freq : str, optional
       Resampling frequency
+
+    Returns
+    -------
+    xarray.DataArrray
+      Base flow index.
     """
     m7 = q.rolling(time=7, center=True).mean(dim='time').resample(time=freq)
     mq = q.resample(time=freq)
@@ -83,7 +93,7 @@ def base_flow_index(q, freq='YS'):
 @with_attrs(standard_name='cold_spell_duration_index', long_name='', units='days')
 @valid_daily_min_temperature
 def cold_spell_duration_index(tasmin, tn10, freq='YS'):
-    """Cold spell duration index
+    r"""Cold spell duration index
 
     Re-samples the daily minimum temperature series by returning the number of days per
     period where the temperature is below the calendar day 10th percentile (calculated
@@ -100,6 +110,10 @@ def cold_spell_duration_index(tasmin, tn10, freq='YS'):
     freq : str, optional
       Resampling frequency
 
+    Returns
+    -------
+    xarray.DataArray
+      Cold spell duration index.
 
     See also
     --------
@@ -123,7 +137,7 @@ def cold_spell_duration_index(tasmin, tn10, freq='YS'):
 
 @valid_daily_mean_temperature
 def cold_spell_index(tas, thresh=-10, window=5, freq='AS-JUL'):
-    """Cold spell index.
+    r"""Cold spell index
 
     Days that are part of a cold spell (5 or more consecutive days with Tavg < -10°C)
     """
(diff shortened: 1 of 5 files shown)


xclim/indices.py
```diff
@@ -265,9 +264,23 @@ def daily_temperature_range(tasmax, tasmin, freq='YS'):
...
+def freshet_start(tas, thresh=0.0, window=5, freq='YS'):
+    r"""First day of consistent threshold temperature.
+
+    Return first day of year when a temperature threshold is exceeded
     over a given number of days.
+
+    Parameters
+    ----------
+    tas : xarray.DataArray
+      Mean daily temperature
```
Zeitsperre: I'm coming across function asking for deg C and others demanding K. Just for my reference, xclim performs the conversion handling for both, yes?

Zeitsperre: @huard

Zeitsperre:
Feel free to merge and close this branch. It would be good to get this into master so people can update descriptions from here.

xclim/indices.py
```diff
@@ -83,23 +93,27 @@ def base_flow_index(q, freq='YS'):
 @with_attrs(standard_name='cold_spell_duration_index', long_name='', units='days')
 @valid_daily_min_temperature
 def cold_spell_duration_index(tasmin, tn10, freq='YS'):
-    """Cold spell duration index.
+    r"""Cold spell duration index
 
-    Resample the daily minimum temperature series by returning the number of days per
+    Re-samples the daily minimum temperature series by returning the number of days per
```
huard: resampling is a verb.

xclim/indices.py
```diff
@@ -74,7 +74,7 @@ def base_flow_index(q, freq='YS'):
     Parameters
     ----------
     q : xarray.DataArray
-      Rate of iver discharge [m³/s]
+      Rate of river discharge [m³/s]
```
Zeitsperre: Thanks!

xclim/indices.py
```diff
@@ -83,23 +93,27 @@ def base_flow_index(q, freq='YS'):
...
+    Re-samples the daily minimum temperature series by returning the number of days per
     period where the temperature is below the calendar day 10th percentile (calculated
     over a centered 5-day window for values during the 1961–1990 period) for a minimum of
     at least six consecutive days.
 
     Parameters
     ----------
     tasmin : xarray.DataArray
-      Minimum daily temperature values [K].
+      Minimum daily temperature values [C] or [K]
```
huard: We have still no explicit unit conversion routine. That might be covered by ocgis, but I think it can get confusing if the function's doc assumes that it's being wrapped by an external library. I suggest we stick with one unit and provide complimentary unit conversion through ocgis.

xclim/indices.py
Zeitsperre: Indeed.

xclim/indices.py
```diff
@@ -83,23 +93,27 @@ def base_flow_index(q, freq='YS'):
...
     tn10 : xarray.DataArray
       The climatological 10th percentile of the minimum daily temperature computed over
-      the period 1961-1990, computed for each day of the year using a centered 5-day window.
```
huard: I suggest we use grammatically correct sentences whenever possible.

xclim/indices.py
huard: I see you assumed the unit choice everywhere. Let's keep it that way then and get unit conversion to work with or without ocgis.

xclim/indices.py
huard: Ie with punctuation. I suggest that if clause has a verb, use punctuation, otherwise not.

xclim/indices.py
```diff
@@ -378,22 +476,70 @@ def heating_degree_days(tas, freq='YS', thresh=17):
...
     hd = (tasmax > K2C + thresh) * 1
     return hd.resample(time=freq).sum(dim='time')
 
 
 @valid_daily_max_temperature
 def ice_days(tasmax, freq='YS'):
-    """Number of days where the daily maximum temperature is below 0℃."""
+    r"""Number of freezing days
+
+    Number of days where daily maximum temperature does not exceed 0℃.
```
huard: I feel negative clauses makes understanding more difficult (does not exceed).

xclim/indices.py
```diff
@@ -378,22 +476,70 @@ def heating_degree_days(tas, freq='YS', thresh=17):
...
+
+    Parameters
+    ----------
+    tasmax : xarrray.DataArray
+      Maximum daily temperature [℃] or [K]
+    freq : str, optional
+      Resampling frequency.
```
huard: dot or no dot after resampling frequency ?

xclim/indices.py
```diff
@@ -443,34 +588,74 @@ def tg_mean(tas, freq='YS'):
...
+      Resampling frequency
+    """
     arr = tasmin.resample(time=freq) if freq else tasmin
     return arr.mean(dim='time')
 
 
 @valid_daily_min_temperature
 def tn_min(tasmin, freq='YS'):
-    """Minimum of daily minimum temperature."""
+    """Extreme minimum temperature
```
huard: Here you're using extreme, whereas you used Highest earlier. I suggest we stick with the ICCLIM definitions whenever reasonable.

xclim/indices.py
```diff
@@ -527,35 +713,90 @@ def prcp_tot(pr, freq='YS', units='kg m-2 s-1'):
...
+      Resampling frequency
+    """
+
     return tasmax.resample(time=freq).max(dim='time')
 
 
 @valid_daily_max_temperature
 def tx_mean(tasmax, freq='YS'):
-    """Mean of daily maximum temperature."""
+    r""" Mean max temperature
```
huard: extra space

xclim/indices.py
Zeitsperre: Sounds good. I'll add a TODO at the top concerning that problem.

xclim/indices.py
Zeitsperre: Got it.

xclim/indices.py
Zeitsperre: Sounds good.

xclim/indices.py
Zeitsperre: I'm going to go with dot on param strings.

Zeitsperre: Scratch that. No dot unless it follows a sentence structure.

Issue #30, opened 2018-09-25
varia : calendars, time_bnds, ...


sbiner:
Here's a few info/considerations linked to xclim

* we should take car of passing xarray.DataArray to the different indices and not xarray.Dataset. It may work but there are problems downstream.

* Nothing get's done with xarray.resample until we actually read the data. To keep in mind for the benchmark.

* xarray.resample does not support 365_day and 360_day calendars. For a reason i ignore yet (raised an [issue](https://github.com/pydata/xarray/issues/2437) about it), using xr.open_dataset() on a 365_day netcdf works (but it should not).

* if we aim at producing standard netCDF files, we will eventually have to take care of the time_bnds. That implies that we have to compute them and maybe/probably attach them as coordinates to the output xarray.DataArray

Feel free to comment

huard:
Time_bnds should be handled by ocgis. I suggest defering this item to later. Calendar support is a problem. Suggestions?

Zeitsperre:
Do we know exactly which calendars are supported by xarray? Perhaps we can implement a check for non-standard calendars that will be handled exclusively in ocgis.

Would there be ways of mapping xclim functions within ocgis's OcgOperations?

sbiner:
xr supports all the calendars, but only standard calendar is supported now for xr.resample. Discussed this with Travis and we are thinking  of bypassing xr.resample and using xr.groupby to mimic xr.resample.

tlogan2000:
Talking to seb (if understand) the idea is to create a subfunction that tags the data (e.g. 'Year_month','Year_Season') that can be used in .groupby instead of reseample.  Would probably have to limit the number of frequencies available to be used (e.g. 'M','Y','QS-DEC') only for now.

huard:
Ok. Maybe resample implementation in xarray is something we could subcontract.

tlogan2000:
Yes a possibility.  @sbiner  is working on his labeling/groupby method (says to give him a couple of days) but really an xarray library capable of dealing with various calendars would be the ideal long term solution in my opinion...   

Note, other than the resample problem, simply opening a 360day nc file gives me a warning when using a 'open_dataset'.  Inspection of dates indicates that they are present but non-monotonic (affects resample but could mess up other things as well most likrly).. xarray seems to use the standard calendar any how which ends up giving me 3 values of march 1 in a row.... 

So the contract could potentially look at xarray implmentation of non-standard calendars in general?

[...]

PR #51, opened 2018-10-03
Replace resample method by daily_downsampler

huard:
Partial fix for #30

To use, decorate index function with @with_attrs. This will check if the index is a CFTimeIndex instance, and if yes, will replace the standard resample method by daily_downsampler, which I've modified a little to copy resample's signature. 

Using this, we can use ``resample`` in our indicators. When xarray has support for non-standard calendar resampling, we can just remove that bit of code in the decorator. 

The code is barely tested, but it's all I can do at the moment...

### DIFF
--- a/tests/test_indices.py
+++ b/tests/test_indices.py
@@ -67,11 +67,12 @@ class Test_max_1day_precipitation_amount():
         assert rx1day.time.dt.year == 2000
         assert len(rx1day) == 1
 
-    def non_standard_calendar(self):
+    @pytest.mark.skip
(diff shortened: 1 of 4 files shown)


huard:
The problem with this is that if we create another dataarray inside the index function, it will not inherit from the dds resampling method, so I doubt this is a good solution. I suggest we dump that PR and I'll work on adding resample support in xarray.

tlogan2000:
Sounds good. We will keep going on the indices as planned....

PR #55, opened 2018-10-03
Rx1day update

tlogan2000:
Added strict nan behavior : a single nan in a given period will result in a nan rx1day value.  Note nan presence triggers a warning from the reduce() method as well indicating there are non-valid values...


tlogan2000:
@huard yes definitely. Thinking is 'better safe than sorry' for now.... with plans to become more flexible in future

tlogan2000:
@Zeitsperre not sure why flake8 fails on py3.6 build....maybe you can take a look. My latest commit was supposed to take care of white space errors..

PR #58, opened 2018-10-09
Class to implement missing data handling, unit conversion, cf…

huard:
Okay, so it seems that we're stretching the limit of what can be done with a function and decorators. 
I've created a class to encapsulate checks, missing data handling, metadata attributes, etc. 

The idea would be to remove all decorators from indices.py, and replace the logic using subclasses (see example in temperature.py). I think the simpler design will be easier to understand. 

It's just a prototype, let me know what you think.

### DIFF
--- a/tests/test_temperature.py
+++ b/tests/test_temperature.py
@@ -1,0 +1,26 @@
+import pytest
+import xarray as xr
+import numpy as np
+import pandas as pd
+from xclim.temperature import TxMax
+
+@pytest.fixture
+def tas_series():
+
+    def _tas_series(values):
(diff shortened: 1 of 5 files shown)


sbiner:
Une classe par indice, c'est pas un peu overkill? Si on faisait une classe par type de variable (pr, tas, tasmin et tasmax) on couvrirait presque tous les indices pour faire les checks de base et la conversion d'unité.

huard:
Ça va pas mal ressembler à ça. On va pouvoir sous-classer Tx_Max. Je viens de le faire pour TxMin.

tlogan2000:
I think i need a 5min walkthrough of the code to understand completely. @huard are you at Ouranos wednesday pm (e.g. after your ETS meeting)

PR #62, opened 2018-10-11
Removed most decorators from indices.py. Added test for UnivariateInd…

huard:
I've removed most of the decorators in indices. 
Their functionality is now implemented in utils.UnivariateIndicator

I suggest we don't yet spend too much time on those classes as discussions with partners may bring changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ bottleneck
 netCDF4
 pandas
 funcsigs
+pint
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,0 +1,15 @@
+import pandas as pd
+import pytest
+import xarray as xr
+
+
+@pytest.fixture
+def tas_series():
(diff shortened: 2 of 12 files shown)


huard:
Resolving conflicts...

Zeitsperre:
Looks like I've botched a few reverts. In the process of fixing it now. Pay no mind.

huard:
Note that xarray does not support 3.4 officially anymore. No point for us to do so.

huard:
For the flake8 problems, would including tests into the package fix it ? This is what xarray does.

Zeitsperre:
@huard Including the tests in the package would be a potential solution. I can go about doing this this today. It shouldn't take long.

huard:
I'm concerned this will cause conflicts with a number of open PRs we have. 
It might be easier to merge all outstanding PRs, then do this change.

PR #75, opened 2018-10-22
vDTR

Zeitsperre:


xclim/indices.py
```diff
@@ -225,16 +225,25 @@ def daily_freezethaw_cycles(tasmax, tasmin, freq='YS'):
...
+
     ft = (tasmin < K2C) * (tasmax > K2C) * 1
     return ft.resample(time=freq).sum(dim='time')
 
 
 def daily_temperature_range(tasmax, tasmin, freq='YS'):
     r"""Mean of daily temperature range."""
-    dtr = tasmax - tasmin
+
+    dtr = abs(tasmax - tasmin)
```
huard: Why the abs ? Negative values would indicate a problem in input files that would be missed if taking the absolute value.

tests/test_indices.py
```diff
@@ -269,6 +269,13 @@ def test_static_daily_temperature_range(self):
         dtr = xci.daily_temperature_range(tasmax, tasmin, freq="YS")
         np.testing.assert_array_equal([dtr.mean()], [20])
 
+    def test_static_variable_daily_temperature_range(self):
```
huard: I'm not sure I understand the logic of that test. It does not seem very discriminating. I suggest using a synthetic input array that you know exactly what answer it should yield and compare against that.

Zeitsperre: All I'm looking for is that vDTR is between 0 and 10. Not very robust. I'll see if I can throw together a test now to catch edge cases. Not sure what else to test for.

tests/test_indices.py
```diff
@@ -251,38 +249,74 @@ def test_simple(self):
...
+        np.testing.assert_array_less([0], [dtr.mean()])
+        np.testing.assert_allclose([dtr.mean()], [20], atol=10)
+
+    def test_static_daily_temperature_range(self, tasmax_series, tasmin_series):
+        tasmax, tasmin = self.static_tmax_tmin_setup(tasmax_series, tasmin_series)
         dtr = xci.daily_temperature_range(tasmax, tasmin, freq="YS")
-        np.testing.assert_array_equal([dtr.mean()], [20])
 
-    def test_all_freeze_thaw_cycles(self):
+        np.testing.assert_almost_equal([dtr.mean()], [6.942], decimal=3)
```
huard: Ok, but ideally these examples could be checked by doing mental calculations. 
For example, max_values = min+values + 10
dtr is already the mean.

tests/test_indices.py
```diff
@@ -251,38 +249,74 @@ def test_simple(self):
...
+        tasmax = tasmax_series(max_values)
+        tasmin = tasmin_series(min_values)
+        return tasmax, tasmin
+
+    def test_random_daily_temperature_range(self, tasmax_series, tasmin_series):
+        days = 365
+        tasmax, tasmin = self.random_tmax_tmin_setup(days, tasmax_series, tasmin_series)
+        dtr = xci.daily_temperature_range(tasmax, tasmin, freq="YS")
+
+        np.testing.assert_array_less([0], [dtr.mean()])
```
huard: assert_array_almost_equal(x, y, decimal=6, err_msg='', verbose=True)
x : array_like
    The actual object to check.
y : array_like
    The desired, expected object.

tests/test_indices.py
```diff
@@ -251,38 +249,74 @@ def test_simple(self):
...
+    def test_static_variable_daily_temperature_range(self, tasmax_series, tasmin_series):
+        tasmax, tasmin = self.static_tmax_tmin_setup(tasmax_series, tasmin_series)
+        dtr = xci.variable_daily_temperature_range(tasmax, tasmin, freq="YS")
+
+        np.testing.assert_almost_equal([dtr.mean()], [2.667], decimal=3)
+
+    def test_uniform_freeze_thaw_cycles(self, tasmax_series, tasmin_series):
         temp_values = np.zeros(365)
-        tasmax, tasmin = self.tmax_tmin_time_series(temp_values + K2C)
+        tasmax, tasmin = tasmax_series(temp_values + 5 + K2C), tasmin_series(temp_values - 5 + K2C)
```
huard: Yes!

xclim/indices.py
```diff
@@ -225,16 +226,25 @@ def daily_freezethaw_cycles(tasmax, tasmin, freq='YS'):
...
 
 def daily_temperature_range(tasmax, tasmin, freq='YS'):
     r"""Mean of daily temperature range."""
+
     dtr = tasmax - tasmin
     return dtr.resample(time=freq).mean(dim='time')
 
 
+def variable_daily_temperature_range(tasmax, tasmin, freq="YS"):
+    r"""Mean absolute day-to-day variation in daily temperature range"""
```
huard: Complete docstring with description, example usage, etc. I've never seen this one used, so worthwhile to explain how it's used.

tests/test_indices.py
```diff
@@ -251,38 +255,77 @@ def test_simple(self):
...
-            tasmax, tasmin = self.tmax_tmin_time_series(temp_values + K2C)
-            ft = xci.daily_freezethaw_cycles(tasmax, tasmin, freq="YS")
-            runs = np.append(runs, ft)
-        np.testing.assert_allclose(np.mean(runs), 120, atol=20)
+    def test_static_freeze_thaw_cycles(self, tasmax_series, tasmin_series):
+        tasmax, tasmin = self.static_tmax_tmin_setup(tasmax_series, tasmin_series)
+        tasmin = np.subtract(tasmin, 15)
+        ft = xci.daily_freezethaw_cycles(tasmax, tasmin, freq="YS")
+
+        np.testing.assert_array_equal([np.sum(ft)], [4])
```
tlogan2000: why use np.sum(ft)?   ft is a single value in this case ...

PR #76, opened 2018-10-22
Python3.7 build

Zeitsperre:
First attempt at a manual travis Py37 build. WIP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ matrix:
       python: 3.7
       sudo: required
       dist: xenial
+      addons:
+        apt:
+            packages:
+                - libnetcdf-c++4
     - env:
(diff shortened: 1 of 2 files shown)


Zeitsperre:
Not yet. Due to the limitations of Travis (which is natively running Ubuntu 14.04), the environment needs to be manually specified. I think I might need to explicitly install the HDF5 libraries via apt. In theory, it SHOULD work.

I'm curious to see how 3.7 will be different for the libraries as there are some major departures (e.g. the base library `round()` function now follows IEEE protocols). I'll be focussing on this more next week.

huard:
My perception is that this is not a priority, and that work on 3.7 support should be deferred to a later milestone.

PR #83, opened 2018-10-25
Opg

huard:
Tests and docstrings for opg indices

### DIFF
--- a/xclim/indices.py
+++ b/xclim/indices.py
@@ -467,7 +467,7 @@ def heating_degree_days(tas, freq='YS', thresh=17):
     xarray.DataArray
       Heating degree days index.
     """
-    return tas.pipe(lambda x: thresh - x) \
+    return tas.pipe(lambda x: K2C + thresh - x) \
         .clip(0) \
         .resample(time=freq) \
         .sum(dim='time')
--- a/xclim/indices.py
+++ b/xclim/indices.py
@@ -129,6 +129,7 @@ def cold_spell_index(tas, thresh=-10, window=5, freq='AS-JUL'):
 
     return group.apply(rl.windowed_run_count_ufunc, window=window)
 
+
 # TODO: mix up in docsring for tas
 def cold_and_dry_days(tas, tgin25, pr, wet25, freq='YS'):
     r"""Cold and dry days.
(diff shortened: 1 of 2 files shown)


xclim/indices.py
```diff
@@ -424,9 +531,13 @@ def heat_wave_index(tasmax, thresh=25.0, window=5, freq='YS'):
     Parameters
     ----------
     tasmax : xarrray.DataArray
-      Maximum daily temperature [℃] or [K[
+      Maximum daily temperature [℃] or [K]
     thresh : float
+<<<<<<< HEAD
```
Zeitsperre: Merge conflict

xclim/utils.py
```diff
@@ -124,6 +128,192 @@ def daily_downsampler(da, freq='YS'):
     return buffer.groupby('tags')
 
 
+def get_ev_length(ev, verbose=True, method=2, timing=True, using_dask_loop=True):
```
Zeitsperre: This is throwing a PEP8 error (C901) due to cyclomatic complexity. Too many functions and conditional statements. We can either suspend that check (as we will likely be writing even more complex indices) or we can try refactoring this function to externalize functions that might be called by other indices in the future. Thoughts?

huard: This is @sbiner 's stuff ; ) I'll let him handle it.

xclim/indices.py
huard: Resolved in the next opg PR...

xclim/indices.py
```diff
@@ -8,6 +8,10 @@
 
 import numpy as np
 import xarray as xr
+from warnings import warn
+import logging
+from xclim.utils import get_ev_length
+from xclim.utils import get_ev_end
```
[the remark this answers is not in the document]
huard: Resolved in the next OPG PR.

xclim/utils.py
huard: I'm resolving since does not affect the current PR. I think I based my PR out of his branch to avoid conflicts.

huard:
@Zeitsperre I'll let you merge since you review is holding it. 
@tlogan2000 we'll need an executive decision on the max consecutive spell question, both on the concept and the code (get_ev_length vs run_length).

tlogan2000:
Short term executive decision is to reprodruce ICCLIM - so resample first then calculate consec days.  Just need to document this is the docstinrg perhaps. 

 I think most use cases of this will be with freq = YS or QS so I think it is not that common a problem....
Longterm we can have many heated discussions ...

PR #91, opened 2018-10-31
Wps hook

huard:
A few small changes and fixes in the process to create a WPS Process based on xclim classes. Master version of finch has now one WPS process based on xclim. 

I've started modifying the tg_mean class according to Lars' specs, hence the name change.

### DIFF
--- a/xclim/temperature.py
+++ b/xclim/temperature.py
@@ -48,12 +48,12 @@ class Tasmax(UnivariateIndicator):
 
 
 tmmean = Tas(identifier='tmmean',
-              units='K',
-              long_name="Mean daily mean temperature",
-              standard_name="air_temperature",
-              description="{freq} of daily mean temperature.",
-              keywords='',
-              compute=_ind.tg_mean,)
+             units='K',
+             long_name="Mean daily mean temperature",
+             standard_name="air_temperature",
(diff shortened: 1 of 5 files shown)


xclim/temperature.py
```diff
@@ -47,13 +47,13 @@ def cfprobe(self, da):
...
-              long_name="{freq} mean temperature",
-              standard_name="{freq} mean temperature",
-              description="{freq} of daily mean temperature.",
-              keywords='',
-              compute=_ind.tg_mean,)
+tmmean = Tas(identifier='tmmean',
+             units='K',
+             long_name="Mean daily mean temperature",
+             standard_name="air_temperature",
+             description="{freq} of daily mean temperature.",
```
tlogan2000: Will '{freq}' be translated to something more verbose using a dictionary?  Will need to apply both the freq and the method no? 'YS' yields annual but need to specify that used used .mean(dim='time') somehow so the description reads: 'Annual mean of daily mean temp' or equivalent

huard: Typo on my end

huard: Yes, see _attrs_mapping in utils.UnivariateIndicator.

tlogan2000: dacc

PR #99, opened 2018-11-06
Support for multivariate indicators

huard:
Converts UnivariateIndicator to Indicator, which can handle any number of driving variables. 

The logic is based on counting the number of required_units. If required_units is a string, then we assume the class is Univariate. If required_units is a tuple `('K', 'K')`, then we assume it's multivariate with n = len(required_units).

### DIFF
--- a/xclim/temperature.py
+++ b/xclim/temperature.py
@@ -47,16 +47,18 @@ class Tasmax(Indicator):
         checks.check_valid(da, 'cell_methods', 'time: maximum within days')
         checks.check_valid(da, 'standard_name', 'air_temperature')
 
+
 class TasminTasmax(BivariateIndicator):
     required_units = ('K', 'K')
-    
+
     def cfprobe(self, dan, dax):
         for da in (dan, dax):
             checks.check_valid(da, 'cell_methods', 'time: maximum within days')
             checks.check_valid(da, 'standard_name', 'air_temperature')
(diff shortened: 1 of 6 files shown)


xclim/temperature.py
```diff
@@ -7,20 +7,21 @@
...
 for each indicator.
 
 """
 
 from . import checks
 from . import indices as _ind
-from .utils import UnivariateIndicator
+from .utils import Indicator
 
+Indicator = BivariateIndicator = Indicator
```
Zeitsperre: What's happening with the assignments here?

huard: This is a leftover. Should be deleted.

xclim/temperature.py
```diff
@@ -47,6 +47,23 @@ def cfprobe(self, da):
         checks.check_valid(da, 'standard_name', 'air_temperature')
 
 
+class TasminTasmax(BivariateIndicator):
```
Zeitsperre: Build is failing. Looks as though the assignment was performing a necessary operation. Is it a matter of replacing the `BivariateIndicator` here with `Indicator`?

sbiner:
Ça me semble ok

xclim/temperature.py
huard: Correct. Done.

PR #108, opened 2018-11-15
Finch compatibility

huard:
Small changes to make sure finch can load the indices and run them inside WPS processes.


tlogan2000:
@huard build tests are not happy?  Seemed like very minor changes??

PR #112, opened 2018-11-19
Add indicator classes - continued 

tlogan2000:
Added more temperature indicators classes and test

Note also edited - growing season length calculation (take min of the resample instead of max)

### DIFF
--- a/xclim/temperature.py
+++ b/xclim/temperature.py
@@ -75,74 +75,95 @@ heat_wave_index = Tasmax(identifier='hw_index{thresh}',
                          compute=_ind.heat_wave_index,
                          )
 
-tmmean = Tas(identifier='tmmean',
-             units='K',
-             long_name="Mean daily mean temperature",
-             standard_name="air_temperature",
-             cell_methods='time: mean within days time: mean over days',
-             description="{freq} mean of daily mean temperature.",
-             keywords='',
-             compute=_ind.tg_mean, )
+tm_mean = Tas(identifier='tmmean',
+              long_name="Mean daily mean temperature",
+              units='K',
+              standard_name="air_temperature",
+              cell_methods='time: mean within days time: mean over days',
+              description="{freq} mean of daily mean temperature.",
+              keywords='',
+              compute=_ind.tg_mean, )
 
-txmean = Tasmax(identifier='txmean',
-                required_units='K',
-                long_name='Mean daily maximum temperature',
+tx_mean = Tasmax(identifier='txmean',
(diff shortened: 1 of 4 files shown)


xclim/temperature.py
```diff
@@ -75,33 +75,114 @@ def cfprobe(self, dan, dax):
...
-                standard_name='tasmax',
-                description='Maximum daily maximum temperature.',
+tm_mean = Tas(identifier='tmmean',
+              long_name="Mean daily mean temperature",
+              units='K',
+              standard_name="air_temperature",
+              cell_methods='time: mean within days time: mean over days',
+              description="{freq} mean of daily mean temperature.",
+              keywords='',
+              compute=_ind.tg_mean, )
```
huard: I think it's confusing that the instance, the identifier and the compute function all have different names. We probably want to fix that before the next release.

xclim/temperature.py
```diff
@@ -75,33 +75,114 @@ def cfprobe(self, dan, dax):
...
+extreme_temperature_range = TasminTasmax(identifier='etr',
+                                         long_name='Intra-period Extreme Temperature Range',
+                                         units='K',
+                                         standard_name='air_temperature',
+                                         description='{freq} range between the maximum of max temperature (TXx)'
+                                                     ' and the minimum of min temperature (TNn)',
+                                         compute=_ind.extreme_temperature_range,
+                                         )
+
+cold_spell_duration = Tasmin(identifier='csdi{window}',
```
huard: csdi_{window} ?

xclim/temperature.py
tlogan2000: Yes.  I prefer the underscore but was trying at least to match variable 'id' with the cf-metadata spreadsheet from Lars

xclim/temperature.py
tlogan2000: Underscores for both cases of {thresh} and {window} would be my preference but am trying to fit with the metadata sheet as much as possible... Also noticed that we may want to ensure that {thresh} or {window} values are converted to integers before being string formatted (e.g. some functions have 'thresh=5.0' as defaults ,,, gives 'gdd5.0'..  Can we think of examples where thresh would actually have a decimal?

tlogan2000: Also, For the specific case of csdi (as well as some others) metadata doesn't incorporate the window size (e.g. ICCLIM is always 6).

huard:
Good point. Please open a ticket about this.

Le mar. 20 nov. 2018 09:04, Travis Logan <notifications@github.com> a
écrit :

Issue #80, opened 2018-10-25
Update NetCDF History information


Zeitsperre:
One of the useful things that bash utilities like `ncks` and `cdo` do is they append comments to the history field of outputted netcdf files the processes that have been performed on the files. For example:
```
:Conventions = "CF-1.6" ;
		:history = "Sun Jul 29 07:19:00 2018: cdo -O -f nc sellonlatbox,-96,-72,39,52 /<servername>/smith/scripts/eccc_07132018/crcm5_data/bbw/psl/psl_bbw_199701_se.nc /<servername>/smith/scripts/eccc_07132018/cdo_subset/bbw/psl/psl_bbw_199701_se_ECCCsubset.nc\n"
```
It would be good to add a function that on exit or on write success adds a history entry that specifies:
* The date that the file was modified
* The `xclim` function performed
* The filepaths (I/O)

Balinus:
I'd drop the file paths for security reasons.

Zeitsperre:
That seems fair. Maybe just keep the file base names?

Balinus:
Good idea!

Zeitsperre:
@huard This seems to be addressed in #116. Close?

PR #116, opened 2018-11-21
Attribute assignment improvements

### LINKED ISSUE #114: Climate indice class identifiers / dynamix naming
For climate indicator classes:

- [x]  harmonize instance, identifier and the compute function syntax : e.g. for the temperature.tm_mean class, id=tmmean, compute=tg_mean().  Lack of undescore in id is trying to follow the proposed CF-metadata 

- [x] Dynamic variable names : Decide on naming convention when using  cases of {thresh} and {window} in order to separate from the root variable name. My preference would be to use underscores, however this is not the pattern using the CF  metadata sheet (Lars and Claus). E.g. gdd5 (cf spreadsheet) would be gdd_5 (preferred).  Also, for certain cases  (e.g. csdi ) proposed metadata doesn't incorporate the window size (e.g. ICCLIM is always 6).  However xclim allows users to specify the window (if desired)

- [x] For variable naming ensure that {thresh} or {window} values are converted to integers before being string formatted (e.g. some functions have 'thresh=5.0' as defaults ...  gives 'gdd5.0'..

huard:
- converts floats to integers when possible in tagged attributes 
- added history attribute
- concatenate cell_methods from indicator with source data cell_method

Fixes #114 and #80 partially (file names in history signature not implemented)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ pint
 inspect2
 unittest2
 six
+cftime
--- a/tests/test_indices.py
+++ b/tests/test_indices.py
@@ -408,7 +408,6 @@ class TestRainOnFrozenGround:
 
 class TestTg10p:
     def test_simple(self, tas_series):
-
         i = 366
         tas = np.array(range(i))
         tas = tas_series(tas, start='1/1/2000')
@@ -424,7 +423,6 @@ class TestTg10p:
 
 class TestTg90p:
     def test_simple(self, tas_series):
-
         i = 366
         tas = np.array(range(i))
         tas = tas_series(tas, start='1/1/2000')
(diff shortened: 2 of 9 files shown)


Zeitsperre:
I think that timestamp test might be prone to fail since it really depends on the processing speed of the Travis CI instance. Is there a way of increasing the range of acceptable times? Maybe with a datetime.now(), datetime.now()+1s, datetime.now()+2s...
 ```
assert "[2018-11-22 01:37:13] tmin0(da, thresh=0.0, freq='YS')" in 
    ("[2018-11-22 01:37:12] tmin0(da, thresh=0.0, freq='YS')", "[2018-11-22 01:37:13] tmin0(da, thresh=0.0, freq='YS')", "[2018-11-22 01:37:14] tmin0(da, thresh=0.0, freq='YS')")
```

PR #117, opened 2018-11-23
Pinned versions and Python3.8-dev

Zeitsperre:


Zeitsperre:
Python3.8 will be in alpha in next 2 months so I'm probably jumping the gun. In any case, it's looking good.

PR #132, opened 2018-12-11
Add precip temp class

tlogan2000:
Added Rain on frozen ground index.  Cleaned up the class descriptions

### DIFF
--- a/xclim/precip.py
+++ b/xclim/precip.py
@@ -7,6 +7,25 @@ class Pr(Indicator):
     context = 'hydro'
 
 
+class PrTas(Indicator):
+    required_units = ('mm/day', 'K')
+    context = 'hydro'
+
+
+rain_on_frozen_ground_days = PrTas(identifier='rain_frzgr',
+                                   standard_name='number_of_days_with_lwe_thickness_of_'
+                                                 'precipitation_amount_above_threshold',
+                                   long_name='Number of rain on frozen ground days',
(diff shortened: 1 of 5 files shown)


xclim/precip.py
```diff
@@ -7,69 +7,89 @@ class Pr(Indicator):
...
+
+rain_on_frozen_ground_days = PrTas(identifier='rain_frzgr',
+                                   units='days',
+                                   standard_name='number_of_days_with_lwe_thickness_of_'
+                                                 'precipitation_amount_above_threshold',
+                                   long_name='Number of rain on frozen ground days',
+
+                                   description="{freq} number of days with rain above {thresh} "
+                                               "after a series of seven days "
+                                               "with average daily temperature below 0℃. "
```
Zeitsperre: It looks as though the `℃` unicode character isn`t allowed in Python2.7. @tlogan2000 How attached are you to these? There's probably a way to preserve them but we need to specify the unicode encoding. Unofrtunately, this is something that `six` doesn't catch easily.

tlogan2000: The temperature.py has a # -*- coding: utf-8 -*- at the beginning of the file and it seems to be ok ... 
have added to precip.py  we'll see whether it fixes the issue

Zeitsperre: Oh wow. Yeah, that fixes it. Learn a new thing every day.

PR #139, opened 2019-01-08
Ndim runlength funcs

tlogan2000:

Multidimensional run_length function coded directly in xarray 

1.  improved compatibility with time-chunked arrays vs ufunc vesion
2. Calculates run_length values for all grid cells simulataneously

No

* **Other information**:

* When merging, please put `closes #{issue number}` in your comment to auto-close the issue that your PR addresses. Thanks!


tlogan2000:
Simple test on 1 year of data (Canada domain ~10km resolution)
```python
import numpy as np
import time

ds = xr.open_mfdataset('/media/sf_VMshare/XCLIM_ncfiles/NRCAN/*pr*1990*.nc', chunks={'time': 31})
da = ds.pr
# da.values = values.transpose()
bool1 = da * 3600 * 24 >= 1
group = bool1.resample(time='YS')
#lorig = group.apply(rl.longest_run_ufunc)
# New version (n-dims)
start = time.time()
l3d = group.apply(rl.longest_run, dim='time')
l3d.compute()
end = time.time()
print('elapsed time N-dim: ',end - start)

# old version
bool1 = bool1.chunk({'time':-1,'lon':30}) # ufunc reuqires all time-steps in the chunk
group = bool1.resample(time='YS')
start = time.time()
l3d =  group.apply(rl.longest_run_ufunc)
l3d.compute()
end = time.time()
print('elapsed time ufunc : ',end - start)
elapsed time N-dim:  13.250417232513428
elapsed time ufunc :  92.19970297813416

**Comment by tlogan2000** [comment]
@Zeitsperre can you provide insight to why the travis - build is failing for this pr?

**Comment by Zeitsperre** [comment]
Since Bottleneck is installed with xarray already, I deactivated it in xclim requirements. New errors now arise from deprecated employments of pytest `fixtures` for the py27 and py36 builds. Testing on Travis now.

File: tests/test_run_length.py [hunk 1]
```diff
@@ -3,41 +3,60 @@
 import xarray as xr
 import pandas as pd
 import numpy as np
+import os
```
Zeitsperre: I think we're using `Pathlib` in our testing code. We may as well use it consistently. 
```suggestion
from pathlib import Path
```

Zeitsperre:
Figured out the `pytest.fixture` issue: At the top of the testing scripts we call the testing fixtures by name. ie:

```
from xclim.testing.common import tas_series
TAS_SERIES = tas_series
```

This is a workaround in order to prevent PyCharm from removing these imports automatically when we run the code formatting. The error is being raised because when we call them as class instances with `()`, it think we're trying to use them outside of a test (which isn't permitted).

I've also re-added Bottleneck as it looks like we do need to explicitly install this library. Tests are now passing on my machine.

huard:
See https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions
I think we could avoid this problem by declaring the fixtures in a testconf.py file.

Zeitsperre:
Tests are failing again. I think it has to do with the base versions on numpy/scipy for xclim and xarray being different (1.15 for xclim vs 1.12 for xarray). I'll let xarray dictate the installation of them and see if that fixes it.

Zeitsperre:
Build still failing. It seems like the issue is very specifically localized to `bottleneck`. No idea why everything works locally and via `tox` on my machine but not on Travis. Any ideas @huard ?

huard:
Doesn't it work if you remove bottleneck from the requirements.txt file ?

Zeitsperre:
Probably XD

Zeitsperre:
Retriggered a build of master. Confirmed that it's a bug. Looks like something changed with one of the dependencies over the holidays. We need to open a ticket with either the numpy, xarray or bottleneck maintainers.

PR #146, opened 2019-01-23
Fix runlength backfill bug

tlogan2000:
I found some strange behavior from da.bfill() not catching nans in the run_length calculation.  Seems to be only when the data is a dask array with chunk=1 for time.  Changing the chunking seems to fix the issue. Have also added a subsequent .ffill to make sure that all nans are caught.

Also changed 'rolling' calls which now throw errors if we specify the dimension

### DIFF
--- a/xclim/indices.py
+++ b/xclim/indices.py
@@ -121,7 +121,7 @@ def base_flow_index(q, freq='YS'):
       Base flow index.
     """
 
-    m7 = q.rolling(time=7, center=True).mean(dim='time').resample(time=freq)
+    m7 = q.rolling(time=7, center=True).mean().resample(time=freq)
     mq = q.resample(time=freq)
 
     m7m = m7.min(dim='time')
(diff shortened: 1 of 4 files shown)


tlogan2000:
I have posted an issue on xarray re: the bfill methods https://github.com/pydata/xarray/issues/2699
Hopefully they have some recommendation / guidance ... forcing the chunksize to 30 (as in the new code) seems to work in my test cases  but maybe not in all?...

xclim/indices.py
```diff
@@ -991,8 +991,8 @@ def max_n_day_precipitation_amount(pr, window=1, freq='YS'):
     """
 
     # rolling sum of the values
-    arr = pr.rolling(time=window, center=False).sum(dim='time')
-    return arr.resample(time=freq).max(dim='time')
+    arr = pr.rolling(time=window, center=False).sum()
+    return arr.resample(time=freq).max()
```
huard: Just to be clear, the latest version of xarray sets the default aggregation dimension for the rolling and resample operations?

tlogan2000: not resample. Sorry my mistake will adjust and resubmit.

tlogan2000:
> There does not seem to be any test of the `rle` function.
> What happens if a run is longer than the chunk size ?

Tests were added to test_run_length but only ensuring that rl.longest_run, first_run etc gives the same result as rl.longest_run_ufunc (1 dimesional original code).   I think maybe subfunctions of rle and rle_1d do not have their own test? 

For chunk size: Judging by the response to my question https://github.com/pydata/xarray/issues/2699 I it would technically be possible to have very long runs and end up with entire chunks that are nans and we would still have a problem. The simplest way to avoid problems for now would be to chunk time as the entire length :chunks={time: -1} ... and adjust other dims in an automated fashion to keep memory size reasonable.  In the rle code  specifically the time size would always be <= 368 for all case (resample of 'YS' being the largest possibility + start and end 'dummy' time steps)

PR #151, opened 2019-02-05
Tmin_days_below index with class & tests

tlogan2000:

tasmin_days_below index & class

* **Other information**:
makes use of the recent generic threshold utility

### DIFF
--- a/xclim/indices.py
+++ b/xclim/indices.py
@@ -958,6 +958,38 @@ def liquid_precip_ratio(pr, prsn=None, tas=None, freq='QS-DEC'):
     return ratio
 
 
+def tn_days_below(tasmin, thresh=-10.0, freq='YS'):
+    r"""Number of days with tmin below a threshold in
+
+    Number of days where daily maximum temperature exceed a threshold.
+
+    Parameters
+    ----------
+    tasmax : xarray.DataArray
+      Maximum daily temperature [℃] or [K]
+    thresh : float
+      Threshold temperature on which to base evaluation [℃] . Default: -10℃.
+    freq : str, optional
+      Resampling frequency
+
+    Returns
+    -------
+    xarray.DataArray
+      Number of summer days.
+
+    Notes
+    -----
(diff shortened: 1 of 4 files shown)


xclim/indices.py
```diff
@@ -958,6 +958,38 @@ def liquid_precip_ratio(pr, prsn=None, tas=None, freq='QS-DEC'):
     return ratio
 
 
+def tn_days_below(tasmin, thresh=-10.0, freq='YS'):
+    r"""Number of days with tmin below a threshold in
```
huard: Missing end of sentence (units ?)

xclim/indices.py
```diff
@@ -958,6 +958,38 @@ def liquid_precip_ratio(pr, prsn=None, tas=None, freq='QS-DEC'):
...
+
+    Number of days where daily maximum temperature exceed a threshold.
```
huard: not exceed

xclim/temperature.py
```diff
@@ -73,6 +74,15 @@ def compute(*args, **kwds):
         """The function computing the indicator."""
 
 
+tn_days_below = Tasmin(identifier='tnlt_{thresh}',
+                       units='days',
+                       standard_name='number_of_days_with_air_temperature_below_threshold',
+                       long_name='Number of days with Tmin < {thresh}C',
+                       description="{freq} number of days where daily minimum temperature is below {thresh}℃",
+                       cell_methods='time: minimum within days time: sum over days',
```
huard: i don't think it's a case of "sum over days".

tlogan2000: Cell methods on the indicators become very confusing I find but  I took this based on the proposed cf standard excel sheet (from Lars Barring et al)
e.g.  Tropical nights (which is the same but simply greater than) they have:

number_of_days_with_air_temperature_above_threshold | time: minimum   within days time: sum over days | Number of   Tropical Nights (Tmin > 20C)
-- | -- | --

PR #162, opened 2019-02-13
Subset utilities

tlogan2000:
added subsetting utilities

### DIFF
--- a/xclim/utils.py
+++ b/xclim/utils.py
@@ -238,10 +238,10 @@ def subset_gridpoint(da, lon, lat, year_bnds=''):
     if np.all(da.lon > 0) and lon < 0:
         lon += 360
 
-    try:
+    if len(da.lon.shape)==1 & len(da.lat.shape)==1:
         out = da.sel(lon=lon, lat=lat, method='nearest')
 
-    except:
+    else:
 
         dist = np.hypot(da.lat - lat, da.lon - lon)
         iy, ix = np.unravel_index(np.argmin(dist, axis=None), da.lat.shape)
--- a/xclim/utils.py
+++ b/xclim/utils.py
@@ -198,8 +198,9 @@ def subset_gridpoint(da, lon, lat, year_bnds=''):
     """Extract a nearest gridpoint from datarray based on lat lon coordinate.
     Time series can optionally be subsetted by year(s)
 
-            Return a subsetted data array (or dataset) for the grid point falling nearest the input longitude and latitude
-            coordinates. Optionally subset the data array for years falling within provided year bounds
+            Return a subsetted data array (or dataset) for the grid point falling nearest the input
+            longitude and latitudecoordinates. Optionally subset the data array for years falling
+            within provided year bounds
 
             Parameters
             ----------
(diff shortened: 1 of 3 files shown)


xclim/utils.py
```diff
@@ -124,6 +124,143 @@ def percentile_doy(arr, window=5, per=.1):
...
+    if np.all(da.lon > 0) and lon < 0:
+        lon += 360
+
+    try:
+        out = da.sel(lon=lon, lat=lat, method='nearest')
+
+    except:
+
+        dist = np.hypot(da.lat - lat, da.lon - lon)
+        iy, ix = np.unravel_index(np.argmin(dist, axis=None), da.lat.shape)
```
huard: This is an approximation as it does not measure distances in km. We could assume the earth is a sphere to convert degrees in km to keep things simple.

xclim/utils.py
```diff
@@ -124,6 +124,143 @@ def percentile_doy(arr, window=5, per=.1):
...
+        args[xydims[0]] = iy
+        args[xydims[1]] = ix
+        out = da.isel(**args)
+
+    if year_bnds:
+        year_bnds = np.asarray(year_bnds)
+        if len(year_bnds) == 1:
+            time_cond = da.time.dt.year == year_bnds
+        else:
+            time_cond = (da.time.dt.year >= year_bnds.min()) & (da.time.dt.year <= year_bnds.max())
```
huard: xarray supports slicing with strings, eg http://xarray.pydata.org/en/stable/indexing.html#indexing-with-dimension-names, which I feel could make your function more general. 

If for example you had start=None, end=None as parameters, you could do 
```
start = str(start) if start is not None
end = str(end) if end is not None
da.sel(time=slice(start, end))
```
This would let users get all the data from the start to a given year, or from a given year to the end.

tlogan2000: ok will change for both.. bbox and gridpoint ..  
FYI I used slicing for my initial bbox function but discovered that some .nc (nrcan I thin) have inverted latitudes (da,lat values start high and diminish... ). The raw datagrid is upside down if you use a simple imshow with no coords. To use slicing with this you need to invert the start / end values (e.g. slice(max(lat),min(lat)).   Can still work but requires checks to see if dims are ascending or descending ... ds.where work in either case..

huard: Hum, good point. Do you think this is something we could fix upstream ?

xclim/utils.py
Balinus: geopy will provide the functionality as far as I understand. It's mostly plug and play.
https://geopy.readthedocs.io/en/stable/#module-geopy.distance

tlogan2000: used pyproj to achieve this.  seems easier to calculate distances between point of interest as all gripdpoints without looping

xclim/utils.py
tlogan2000: Have added individual start and end year inputs .... have kept the ds.where logic for now vs slicing

tlogan2000: > Hum, good point. Do you think this is something we could fix upstream ?
We could probably apply a xarray.DataArray.sortby before slicing. http://xarray.pydata.org/en/stable/generated/xarray.DataArray.sortby.html
Would be curious about whether this ends up being slower than my current ds.where operation..

tlogan2000:
IS everyone alright if I merge this to master?  I think I have addressed all points

PR #173, opened 2019-03-04
remove K2C modifier in indices.py

tlogan2000:
#128 

Removes the K2C multiplier threshold indices calls.  xclim.indices can now be used on any dataarray / threshold combination having the same units.  Default threshold are in Celsius

Yes - calling xclim.indices directly no longer assumes that temperature data is in K and threshold is in C, but instead assumes that both data and threshold have the same units

There is no difference when calling indicators from the temperature.py classes. As before unit conversion is handled by the convert_units method.   However threshold temperature index classes now have required units of 'degC' to match default threshold values

* **Other information**:

* When merging, please put `closes #{issue number}` in your comment to auto-close the issue that your PR addresses. Thanks!


tlogan2000:
Note as far as I can tell this should replace @Zeitsperre's PR https://github.com/Ouranosinc/xclim/pull/129.  Branch was really old so I basically just re-did the work here

huard:
Doesn't that mean that CMIP temperature in K would be converted to C by default? Won't that add to the compute time ?

tlogan2000:
Only threshold indices have C as required units.  Mean, max, min etc stay in Kelvin

tlogan2000:
The output units are always days or degree days in some cases.  so we stay Cf-compliant

huard:
What I mean is that by setting required_units to C, you force a unit conversion on all K based temperature, although it is not strictly necessary for the computation.

huard:
@tlogan2000  Please take a look at commit.

tlogan2000:
ok will have a look ...

huard:
I think we should probably try to be backward compatible with the current behavior up until the 1.0 release to avoid breaking David Caron's code and others'. We could add a UserWarning saying unitless thresholds will not be supported starting with 1.0.

Balinus:
I think it might be a good practice to do all the planned breaking change in a pre-release before 1.0. In other words say the last release is 0.8. Do a 0.9 release with all the warnings and API breaking changes, then do a 1.0 release which is identical to 0.9, but without warnings, etc...

huard:
@tlogan2000  Will work on this PR this morning.

tlogan2000:
@huard ...just pushed some doc string stuff.  I will let you run with it if you want.  I am busy for the rst of the day

tlogan2000:
I think we need still implement the temporary warning when calling a threshold index with an integer or float : 

e.g.
calling 
```python
xci.tx_days_above(da, thresh=25) 
```
I thin we want to give a warning for now that the thresh should be defined as a string '25 degC' but then still run and assume 'deg C' right?
Right now I get an error

tlogan2000:
Also, small note I can't seem to give the precip threshold values in anything but 'mm/d' or 'mm/s' ...   The data can have cf units but the thresh needs to be mm/day or similar.  

Maybe we just don't want to support other possibilities for simplicity?
e.g. following give an errors ...
```python
xci.wetdays(pr, thresh='5 kg m-2 d-1', freq='MS')
xci.wetdays(pr, thresh='5 mm d-1', freq='MS')

**Comment by tlogan2000** [comment]
> I think we need still implement the temporary warning when calling a threshold index with an integer or float :
> 
> e.g.
> calling
> 
> ```python
> xci.tx_days_above(da, thresh=25) 
> ```
> 
> I think we would want to give a warning (for now) that the thresh should be defined as a string '25 degC' but then still run and assume 'deg C' right?
> Right now I get an error

@huard I think the warning could potentially be implemented in utils.convert_units_to: 
i.e adding an additional elif with check for an int or float ... thoughts?
```python
    # Target units
    if isinstance(target, units.Unit):
        tu = target
    elif isinstance(target, (str, xr.DataArray)):
        tu = cfunits2pint(target)
    else:
        raise NotImplementedError

**Comment by tlogan2000** [comment]
maybe not... utils.convert_units_to seems to be used in too many contexts i think..  maybe a temporary decorator applied on the threshold indices is the easiest solution.

**Comment by huard** [comment]
Yeah, it's more complicated than I thought. Will look into it.

**Comment by tlogan2000** [comment]
Not sure if you're finished but the latest commit still doesn't seem to give the desired warning.. see example below 
```python
nc_file = os.path.join(TESTS_DATA, 'NRCANdaily', 'nrcan_canada_daily_tasmax_1990.nc')
ds = xr.open_dataset(nc_file)

xci.tx_days_above(ds.tasmax,30)
```

We might need call your modified convert_unit_to in the decorator before it applys _check_units (e.g. if val is either a float or int)?

line 262 in utlis.py
```python
for name, val in bound_args.arguments.items():
    _check_units(val, bound_units.arguments.get(name, None))
```

huard:
I was not apparently ; ) 
I just figured that DeprecationWarning is ignored by default, unless it's raised in the __main__. So when I was testing in the console it worked, when I tested in the test suite it worked, but no message really appeared in the real world...

tlogan2000:
@huard I just added the very small changes that we looked at together... I think we are good to go. 

See small commits in utils.py
_check_units 
convert_units_to

Do you have time to give this a final look before merging?

tests/test_utils.py
```diff
@@ -363,13 +363,20 @@ def test_dimensionality(self):
...
             out = utils.convert_units_to(10, units.mm / units.day, context='hydro')
-            assert out.m == 10
+            assert out == 10
+
+        with pytest.warns(FutureWarning):
+            tas = tas_series(np.arange(365), start='1/1/2001')
+            out = ind.tx_days_above(tas, 30)
+            out1 = ind.tx_days_above(tas, '30 degC')
+            out2 = ind.tx_days_above(tas, '303.15 K')
+            np.testing.assert_array_equal(out, out1, out2)
```
huard: This will compare out with out1, and out2 will be the error message if it fails. Replace with 
```
np.testing.assert_array_equal(out, out1)
np.testing.assert_array_equal(out, out2)
```

tlogan2000: Ok .. I think I do this elsewhere will double check

huard:
@davidcaron Just a heads up that this PR introduce a change in the way thresholds are passed to the indicator function. In short, the thresholds need to specify the units: 
`thresh='0C'` instead of just `thresh=0`. The old interface should still work but raise a FutureWarning. Note that the conversion assumes that if context is none, then the value is a temperature in celsius, and if content is hydro, then the threshold is a precipitation in mm/day. We'll remove that pass-through for version 1.0. 
Finch hasn't been modified yet to account for that. Do you want us to hold on to this PR until I modify Finch ?

davidcaron:
Thanks for the heads up. What we could do in finch in the meantime is target a specific commit instead of the head of the master branch.

But if you want you can merge this, it won't break the deployed instance of finch.

PR #182, opened 2019-04-01
Streamflow related indicators

huard:

Adds the following indicators for stream flow values: 
- freq_analysis: compute the n-year return period values
- stats: compute statistical indicators (mean, max, min, std, var, count, sum, argmin, argmax)
- doy_qmax, doy_qmin : day of the year of max or min

This PR makes an effort to reduce code duplication by taking advantage of generic functions. Work to auto-generate more indicators has started but it proved to be difficult to get it right. Another branch called `gen-stats` is created to pursue that work. 

No


tlogan2000:
@huard j'essaie de regarder ça ce pm ou demain matin

tests/test_streamflow.py
```diff
@@ -0,0 +1,27 @@
+from xclim import streamflow
+import numpy as np
+
+
+def test_base_flow_index(ndq_series):
```
tlogan2000: Tests seem somewhat sparse here testing only for metadata additions and not output values

xclim/generic.py
```diff
@@ -0,0 +1,265 @@
...
+    if not indexer:
+        selected = da
+    else:
+        key, val = indexer.popitem()
+        time_att = getattr(da.time.dt, key)
+        selected = da.sel(time=time_att.isin(val)).dropna(dim='time')
+
+    return selected
+
+
```
tlogan2000: This is where I have spent most of my time reviewing.  I think I have my head wrapped around it and everything seems functional but I found the mixing of resample and seasonal subsetting using the **indexer in a single function somewhat confusing.   

I wonder if it would it be better to have users explicitly run 'select_time' to subset the data for a given season or month before using freq_analysis or other in a separate call.  This is how I would do it my and I suppose I still could as the indexer is optional but it took me sometime do understand that it does both ...

tlogan2000: Are there applications where there is a time subset that would combined with any other resample frequency other than YS?

xclim/generic.py
```diff
@@ -0,0 +1,265 @@
...
+    xarray.DataArray
+      An array of values with a 1/t probability of exceedance or non-exceedance when mode is high or low respectively.
+
+    """
+    # Apply rolling average
+    if window > 1:
+        da = da.rolling(time=window, center=False).mean()
+
+    # Assign default resampling frequency if not provided
+    freq = freq or default_freq(**indexer)
```
tlogan2000: There is a default freq='YS' in the function parameters so default_freq() would never be used?

xclim/generic.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # Extract the time series of min or max over the period
+    sel = select_resample_op(da, op=mode, freq=freq, **indexer).dropna(dim='time')
+
```
tlogan2000: I assume we drop nans because the scipy fitting function will not accept them, but do we need to check to make sure there is a large enough sample to properly fit the dist in the subsequent fa() call?

tlogan2000:
I have asked @sbiner to take a look a this as well so I will hold off my approval for now.  My relatively few comments aside everything seems OK for a first version of the streamflow functions.  General comment (maybe because I have little background in hydrology) would be that a good set of tutorial material would be beneficial

tests/test_streamflow.py
huard: There is an output test in test_indices.

xclim/generic.py
huard: Yes, for DJF, you'd want YS-DEC. 

To clarify: 
- `select_time` is the function that only selects a time period. 
- `select_resample_op` applies an operation on a selected period at a given resampling frequency. 
- `frequency_analysis` is the convenience function wrapping these operations together:
  - rolling average
  - min/max operation
  - period selection
  - resampling
  - `fa` on the resulting array. 

So basically, you could do the array preparation on your own, then call `fa` on the result. Since there are a lot of streamflow indices to compute (for all seasons, min/max, different rolling averages) I thought it was best to provide that kitchen sink function.

xclim/generic.py
tlogan2000: i.e Should the function def have freq=None ?

huard: Unless you pass None. You've put your finger on the tender spot of the PR. Not clear to me what to do here. One option is to set the default frequency based on the period. So for example, it's always 'YS' unless seasons is DJF. The issue with this is that you might want to take the full year, but split the year in July. So, I think we need to expose the `freq` option, the question is how to handle the default. Options: 
- Set freq=None, and then call default_freq to get the default. The problem is that it's not apparent in the signature what the default value is going to be;
- Set freq='YS', and only call default_freq if freq is None;
- Set freq as a positional argument, so users need to pass a value.

huard: Now I remember ! But the reason I've chosen this option is that our Indicator class expects the signature to contain freq and its default. If we change that, I need to go back in Indicators and modify the code. Doable, but I was lazy. So let's decide how it should be, and I'll make it work.

tlogan2000: With the code as is I think many users would not think to explicitly set freq=None in order to get the appropriate default for the DJF season (despite the fact that you wrote it in the docstring. e.g. I missed it and was looking fairly closely )

tlogan2000: > * Set freq=None, and then call default_freq to get the default. The problem is that it's not apparent in the signature what the default value is going to be;

If changes to the Indicator class are not a major headache .. this would maybe be the best option in my mind.  Could the docstring include a more explicit mention of the defaults by season generated by default_freq()

xclim/generic.py
huard: That's something that should be implemented in the missing method.

xclim/generic.py
huard: Yes, let's try that.

huard: Done.

xclim/generic.py
huard: Opened an issue to account for period selection in missing function.

PR #197, opened 2019-05-01
Refactoring

### LINKED ISSUE #180: Architecture clean-up
### Description
I think we should take a moment and clean the files and namespace before the next release. 

For example, there is a lot of stuff in utils that I think should go into their own file. I don't like the fact that we have `temperature.py` and `precip.py` files for indicators, it seems arbitrary. 

Another question is then what should be exposed in the `xclim` namespace. 

Suggestion to clean up utils: 
- `base.py`for the Indicator class
- `generic.py` for generic functions used by indices.
- `ens.py` for ensemble related functions
- ... ?

Options for indicator classes: 
- follow the CMIP realm notation: `xclim.atmos` for all related atmospheric indicators, and `land` for hydrological processes, 'ocn' for ocean related stuff, etc. 
- all indicators directly in `xclim`

Zeitsperre:

Refactoring of `utils.py` and `indices.py` and `indicators.py`

Yes. Indices can now be accessed either via the `xclim.indices.etc()` or directly via `xclim.indices._simple.etc()`. Indicators are now to be found within `xclim.atmos`, with other indicators to be located within their specific domains eventually (e.g. `xclim.land`). Docs now subdivide indices according to their true location.

* **Other information**:
fixes #180 


Zeitsperre:
sphinx-autodoc builder for climate indices is broken. Need to figure out how to expose them to the sphinx.

xclim/__init__.py
```diff
@@ -93,7 +94,7 @@ def __build_icclim(mode='warn'):
                'TGx': indices.tg_max,
                'TGn': indices.tg_min,
                'TX90p': indices.tx90p,
-               'TX10p': indices.tx10p,
+               'TX10p': indices .tx10p,
```
huard: blank

tlogan2000: whitespace?

Zeitsperre: Yeah, that's an error on my part.

tlogan2000: fixed now

Zeitsperre:
@huard The underscore is meant to effectively "hide" these files while the `__all__` is used to expose the indices that we want them to access via the `xclim.indices`. Users can still use them if they really want to, but it suggests that these files are in the backend of the package and could change names or locations between versions without us needing to mention them. 

So long as the API is stable (`from xclim import indices`), how we organize the backend should be adaptable to the situation (domains, number of indicators, etc.). One step closer to v1.0

PR #216, opened 2019-05-30
Fix 183

### LINKED ISSUE #183: Solve problem with {} tags in identifier
### Description
In the indicator class, we use the identifier attribute as the unique identifier of a given class. Since this was also used as the output variable name for the created netCDF, we introduced formatting tags in the identifier attribute. The problem is that the same identifier is also used in other contacts (finch) to identify uniquely the process, but before the class is called, we can't format the identifier string. 

We should have two distinct attributes: the unique class identifier, and the output variable name. The identifier should not contain {}, but the output variable name could. 

More generally, we should clarify which attributes need to be clearly defined without arguments tags, and which can include tags.

huard:

Indicator class has a `var_name` attribute used to name the output variable. The identifier identifies the function, so has no {} tags. 

Likely yes. This PR needs more testing before merging. 

Fixes #183


xclim/atmos/_precip.py
```diff
@@ -48,7 +49,8 @@ class PrTas(Indicator2D):
                                     compute=indices.max_n_day_precipitation_amount,
                                     )
 
-wetdays = Pr(identifier='r{thresh}mm',
+wetdays = Pr(identifier='wetdays',
+             var_name='r{thresh}mm',
```
tlogan2000: ```atmos.wetdays(pr, thresh='10 mm/day')``` 
gives 'r10 mm/daymm' as an output variable name

tlogan2000: We would want to somehow extract the value of the pint unit here I think

xclim/atmos/_precip.py
```diff
@@ -48,7 +49,8 @@ class PrTas(Indicator2D):
...
              units='days',
              standard_name='number_of_days_with_lwe_thickness_of_precipitation_amount_at_or_above_threshold',
              long_name='Number of Wet Days (precip >= {thresh} mm)',
```
tlogan2000: same idea here ... this is giving 'Number of Wet Days (precip >= **10 mm/daymm**)

huard:
@tlogan2000 should be good now.

PR #220, opened 2019-06-04
Solve a couple of issues with frequency analysis

### LINKED ISSUE #218: frequency_analysis does not propagate attributes when window > 1
* xclim version: master

### Description
Units are lost and cause mismatch with expected units.

huard:
- Propagate units in frequency analysis
- Matched calendar types in missing_any. If time indexes are different (different calendars), the comparison of the number of days in each period does not work as expected. 
- Raise more informative error when no data is selected in period. 

Bug fix.

Should not. 

Fixes #218


tests/test_checks.py
```diff
@@ -117,3 +124,13 @@ def test_season(self, tasmin_series):
 
         miss = checks.missing_any(ts, freq='YS', season='SON')
         np.testing.assert_equal(miss, [False])
+
+    def test_hydro(self):
+        fn = os.path.join(TESTS_DATA, 'Raven', 'q_sim.nc')
+        ds = xr.open_dataset(fn)
+        miss = checks.missing_any(ds.q_sim, freq='YS')
+        np.testing.assert_array_equal(miss[:-1], False)
```
tlvu: since you did `miss[:-1]`, you extract the last value of the array, so it's not an array anymore, should this be just `assert_equal`?

tests/test_checks.py
```diff
@@ -100,6 +104,9 @@ def test_month(self, tasmin_series):
         miss = checks.missing_any(ts, freq='YS', month=8)
         np.testing.assert_equal(miss, [True])
 
+        with pytest.raises(ValueError):
```
tlvu: if you also want to validate the error message, `pytest.raises(ValueError, match=r"No data for selected period."):`.  Useful if you raise the same Exception but for various different reasons.

huard: fixed

huard:
I found another issue related to missing values. The frequency analysis output is not a time series, so we cannot apply the missing  value flag on it. It should apply to intermediate computations, and we have no mechanisms to support this at the moment. I think I'll just remove the missing checks for now, but we may have to customize the indicator class for this case.

tests/test_checks.py
```diff
@@ -1,19 +1,21 @@
+from pathlib import Path
+
 import numpy as np
 import pandas as pd
 import pytest
-import os
 import xarray as xr
+
+from xclim import checks
```
tlvu: unused import in this commit?

tlvu: oh nevermind, miss read diff

tests/test_checks.py
```diff
@@ -132,7 +132,7 @@ def test_hydro(self):
         ds = xr.open_dataset(fn)
         miss = checks.missing_any(ds.q_sim, freq='YS')
         np.testing.assert_array_equal(miss[:-1], False)
-        np.testing.assert_equal(miss[-1], True)
+        np.testing.assert_array_equal(miss[-1], True)
```
tlvu: oops sorry, gave you bad review !

PR #223, opened 2019-06-05
Modify runlength usage - options for using 1D and Nd versions of code

### LINKED ISSUE #217: run length calculation optimization
Indices using n-Dim runlength #139 functions on small spatial  subsets is inefficient.
see #213 for details

We can potentially attempt to determine when to use 1d/ufunc method vs n-Dim method based on grid size

ideally this would be dealt with directly in run_length.py... 
edit - more detailed discussion here : https://github.com/Ouranosinc/xclim/issues/213#issuecomment-497706760

tlogan2000:
- [ ] bumpversion (minor / major) has been called
- [ ] Tags have been pushed

Modify runlength calculation usage to optimize speed #217

no

* **Other information**:

first attempt to fix #217

### DIFF
--- a/xclim/run_length.py
+++ b/xclim/run_length.py
@@ -71,10 +71,7 @@ def longest_run(da, dim='time', ufunc_1Dim='auto'):
         """
     if ufunc_1Dim == 'auto':
         npts = get_npts(da)
-        if npts <= npts_opt:
-            ufunc_1Dim = True
-        else:
-            ufunc_1Dim = False
+        ufunc_1Dim = bool(npts <= npts_opt)
 
     if ufunc_1Dim:
         rl_long = longest_run_ufunc(da)
@@ -107,10 +104,7 @@ def windowed_run_events(da, window, dim='time', ufunc_1Dim='auto'):
         """
     if ufunc_1Dim == 'auto':
         npts = get_npts(da)
-        if npts <= npts_opt:
-            ufunc_1Dim = True
-        else:
-            ufunc_1Dim = False
+        ufunc_1Dim = bool(npts <= npts_opt)
 
     if ufunc_1Dim:
         out = windowed_run_events_ufunc(da, window)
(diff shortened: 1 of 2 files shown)


tlogan2000:
@huard this is a first attempt based on test results in #217.  Opinions on whether this is a reasonable approach?

tlogan2000:
test are failing ... hold off for a bit

xclim/run_length.py
```diff
@@ -7,6 +7,16 @@
 from warnings import warn
 
 logging.captureWarnings(True)
+npts_opt = 9000
+
+
+def get_npts(da):
```
huard: Docstring

xclim/run_length.py
```diff
@@ -41,7 +51,7 @@ def rle(da, dim='time', max_chunk=1000000):
     return d
 
 
-def longest_run(da, dim='time'):
+def longest_run(da, dim='time', ufunc_1Dim='auto'):
```
huard: I think pep suggest lowercase variable names.

xclim/run_length.py
```diff
@@ -120,22 +159,34 @@ def first_run(da, window, dim='time'):
...
-        da.swap_dims({dim: 'time'})
-    da = da.astype('int')
-    i = xr.DataArray(np.arange(da[dim].size), dims=dim).chunk({'time': 1})
-    ind = xr.broadcast(i, da)[0].chunk(da.chunks)
-    wind_sum = da.rolling(time=window).sum()
-    out = ind.where(wind_sum >= window).min(dim=dim) - (
-        window - 1)  # remove window -1 as rolling result index is last element of the moving window
+    if ufunc_1Dim == 'auto':
+        npts = get_npts(da)
+        ufunc_1Dim = bool(npts <= npts_opt)
```
huard: isn't this a bool anyway ?

xclim/run_length.py
tlogan2000: I agree,  I do not completely understand why this doesn't scale efficiently.  I'm not sure we have any other explicit uses of ufuncs(?) ...

tlogan2000:
> I think it's weird that we need to do this. Do other ufuncs suffer from the same issues ?
ufunc scaling is actually close to perfectly linear (blue line) with the number of points.  I think the extra steps in the n-dim version (index variables, bfills etc etc) are maybe simply a bit too much overhead to be worth it when grid dimensions are small.

image

huard:
rle_1d computes three different things: 
- the values taken by each run
- each run's length
- the starting position

I think we could probably speed that function by assuming we pass a boolean and look only for runs of True.

PR #226, opened 2019-06-12
Implement fire weather index

### LINKED ISSUE #204: Implement "fire weather index"
TODO

huard:
- [ ] bumpversion (minor / major) has been called
- [ ] Tags have been pushed

Adds new indicator.

No

* **Other information**:
Fixes #204 


huard:
@tlogan2000 First go at it. Performance probably sucks and docstrings need some love. Also, should we expose all underlying indicators as xclim indicators ? So far only fwi is exposed.

tlogan2000:
> @tlogan2000 First go at it. Performance probably sucks and docstrings need some love. Also, should we expose all underlying indicators as xclim indicators ? So far only fwi is exposed.

Yes. exposing sub indicators would be good ... I know that some users use at the very least the drought code (dc) individually

tlogan2000:
We might want to take some time to look more deeply into numba for the ufunc calls?  It could possibly help with the runlength performance as well

https://medium.com/capital-one-tech/dask-numba-for-efficient-in-memory-model-scoring-dfc9b68ba6ce
https://towardsdatascience.com/how-i-learned-to-love-parallelized-applies-with-python-pandas-dask-and-numba-f06b0b367138

huard:
Ok. I suggest we keep optimizations for a later time. One issue at the moment is that not all lines of codes are exercised by the test suite. I'd like us to make sure we get full coverage before optimizing. Ideas of where we'd get a larger sample of fwi values with a larger range of inputs ? Another option is to create a synthetic data set from the original fortran code.

tlogan2000:
> Ok. I suggest we keep optimizations for a later time. One issue at the moment is that not all lines of codes are exercised by the test suite. I'd like us to make sure we get full coverage before optimizing. Ideas of where we'd get a larger sample of fwi values with a larger range of inputs ? Another option is to create a synthetic data set from the original fortran code.

I agree re:optimizing was just mentioning it. I have come across numba a few times now (again yesterday) but is probably not a priority...  I will see if I can find another source of test data (input / output) for FWI.   
I think implementing a mechanism of np.cumsum on the entire fire season at once would likely be beneficial.  From memory I think right now the code has to run in a loop using previous days values as t0?

huard:
Correct. It wasn't obvious to me however how we could vectorize the code.

Balinus:
I have a test dataset for FWI. Sending it through email and letting you put it where it needs to be.

tlogan2000:
Even once code is complete we may need to investigate how to determine a season start date if no snow cover data is avaliable ( e.g. CC projections?). Not sure b is appropriate for areas with lots of snow accumulation.  

Taken from 
http://nofc.cfs.nrcan.gc.ca/bookstore_pdfs/29152.pdf

(a) In regions normally covered by snow during
the winter, begin calculations on the third day
after snow has essentially left the area to which
the fire danger rating applies.
(b) In regions where snow cover is not a
significant feature, begin calculations on the
third successive day that noon temperatures of
12 °C or higher have been recorded.
In either case, use the following starting
values:
FFMC 85; DMC 6; DC 15

Balinus:
A (similar) method here:

http://cwfis.cfs.nrcan.gc.ca/background/dsm/fwi

FWI Calculation Start-up

To determine spring start-up dates and starting fuel moisture code values, the procedures described in Turner and Lawson (1978) are used. There are two methods, depending on snow cover.

Method 1
For stations that report significant snow cover during the winter, start-up occurrs when the station has been snow-free for three consecutive days. Significant snow cover is defined as a mean snow depth of more than 10 cm, with snow cover present at least 75% of the days in January and February. Start-up values are as follows:

    Fine Fuel Moisture Code (FFMC) set to 85
    Duff Moisture Code (DMC) set to 6
    Drought Code (DC) set to 15*

Method 2
For stations that do not report significant snow cover during the winter, start-up occurs when the mean daily temperature is 6°C or higher for three consecutive days. This temperature represents the approximate lower limit for plant growth. Start-up values are as follows:

    FFMC set to 85
    DMC set to 2 times the number of days since precipitation
    DC set to 5 times the number of days since precipitation*

*Where overwinter precipitation data are available, the DC is set according to an alternative, more rigorous method, Turner and Lawson (1978), which takes into account the fact that saturation of the deep fuel layers may not occur over the winter. In areas with low overwinter precipitation, the starting value for the DC can be significantly higher than the default.

tlogan2000:
Possibel test data using MERRA2

https://data.giss.nasa.gov/impacts/gfwed/

tlogan2000:
bingo 
https://open.canada.ca/data/en/dataset/3d528848-3d43-41c9-83e3-11eae5c50ee8

huard:
Does not include the FWI values...

tlogan2000:
I'll see if people from the CFS can provide some input output data that we can use to validate with

huard:
Ok, in the meantime I'm trying to compare results with GFWED.

huard:
Managed to run the FWI calculation on MERRA and I'm getting values fairly close to GWFED values northward of the 30th latitude. There are differences south of 30. The code has probably been adapted, but I can't find the code they're using. 
Oh, and it's extremely slow... 
Can't spend a lot more time on it. Has someone some time to investigate the gfwed differences and code optimization ?

tlogan2000:
I can try to dig into this probably next week though... Meetings today and tomorrow

huard:
https://github.com/SantanderMetGroup/fireDanger/blob/devel/R/fwi1D.R

huard:
GFWED Code: https://portal.nccs.nasa.gov/datashare/GlobalFWI/v2.0/20181216.GFWEDCode.tar.gz

tlogan2000:
> GFWED Code: https://portal.nccs.nasa.gov/datashare/GlobalFWI/v2.0/20181216.GFWEDCode.tar.gz

Are we back to Matlab?? :)

huard:
No, but it's useful to see how they computed the FWI so may resolve any differences between their results and ours.

tlogan2000:
Was only kidding.  It seems like a good start.  Personally will be able to follow it better than R....  I'll likely have time to look more at this thurs/fri this week

aulemahal:
Pre-christmas push: We now have a working code for all indices! It packs all the computation into one large temporal iterator
The individual indices are still iterated spatially, but could easily be vectorized, either by replacing math by numpy or using numba's `@vectorize or both. I'll do some benchmarking for that.
I tested with data from Merra2 over one year and compared with the calculations of GFWED. Instead of starting from scratch, I started the calculations with the indices from GFWED. I get a mean error of 20% on all indices, but it's getting there. (Also, I am not using the proper `snow_depth` data yet.

Zeitsperre:
@aulemahal I'm wondering about the timeline on this PR. Think it could be closed by Friday?

huard:
Seems ambitious. I suggest this goes into 0.14.

aulemahal:
Indeed, Friday the 10 is a bit ambitious, but one or two weeks might be feasible.

.github/CONTRIBUTING.rst
```diff
@@ -125,57 +123,58 @@ Before you submit a pull request, please follow these guidelines:
...
+.. warning::
+     Try to keep your contributions within the scope of the issue that you are addressing.
      While it might be tempting to fix other aspects of the library as it comes up, it's better to
      simply to flag the problems in case others are already working on it.
 
      Consider adding a "**# TODO:**" comment if the need arises.
 
 3. Pull requests should raise test coverage for the xclim library. Code coverage is an indicator of how extensively tested the library is.
-   If you are adding a new set of functions, they **must be tested** and **coverage percentage should not decrease.**
+   If you are adding a new set of functions, they **must be tested** and **coverage percentage should not significantly decrease.**
```
Zeitsperre: Yeah, this is a much more reasonable requirement.

aulemahal:
It's getting better! We now have the exact same results as the Matlab code, but there are still differences with the GFWED reference dataset. Here is a plot showing the mean of all subindices around the St-Lawrence valley:

devfwi_all_VSL_tsdiff

In a map view, here for the drought code, we have:
devfwi_DC_QC

All of this is a consistent ~10% error, there's a bit more work to put here. Also, I haven't touched the tests or the Indicators yet.

tlogan2000:
This looks great. I think if we have a close to perfect match vs the matlab code from GFWED (using same input data) then we can't do much more than ask GFWED contacts if there is a pre-processing of the input Merra2 data (or of the output data) that we are unaware of.

tlogan2000:
Moving forward do we know if the GFWED methodology (in terms of matlab code) is updated / tweeaked regularly?  
It would be good to try to monitor any adjustments that are made over time (within reason).  Could maybe simply  have the matlab code on our github and pull the code form their ftp site regularly to spot any small differences

tlogan2000:
Dgging around on GFWED ftp I found a 'corrected' version of MERRA2 input data..

https://portal.nccs.nasa.gov/datashare/GlobalFWI/v2.0/wxInput/MERRA2.CORRECTED/

tlogan2000:
> Moving forward do we know if the GFWED methodology (in terms of matlab code) is updated / tweeaked regularly?
> It would be good to try to monitor any adjustments that are made over time (within reason). Could maybe simply have the matlab code on our github and pull the code form their ftp site regularly to spot any small differences

Answeing myself ... Seems like the data / code is versioned with dates in the .m filename etc so should be easy to monitor

aulemahal:
> Digging around on GFWED ftp I found a 'corrected' version of MERRA2 input data..
That stands for "bias-corrected precipitation" and is what I used. The details on which combinations of data are used is here : https://data.giss.nasa.gov/impacts/gfwed/  I might compare again using another precipitation source.

aulemahal:
Good news everyone! Turns out, the differences were from the preprocessing of the input data. Might be something with the "local time" conversion or unit conversion, I'm not sure. But using directly the input data provided on the GFWED website yields the same values for the basic indices (@huard I had to modify very slightly the DMC code, did you use another source than GFWED at first?
The non-significant differences for the composite indices are likely from difference in math ops between Matlab and Python.
devfwi_all_VSL_tsdiff_good

huard:
My initial version took some inspiration from an R implementation.

Balinus:
This might be of interest: https://www.nature.com/articles/sdata201932

Vitolo, C. et al. A 1980–2018 global fire danger re-analysis dataset for the Canadian Fire Weather Indices. Sci. Data. 6:190032 https://doi.org/10.1038/sdata.2019.32 (2019).

aulemahal:
(Woups @huard, looks like we had a merge battle.)

Anyway, here is a final first version. Meaning, I think this is mergeable, but I will open issues right after for a few things:

 - Compute start up and shut down without `snow_depth`.
 - Expose more indexes through `atmos`, right now only `drought_code` is exposed. There is a `xc.indices.fire_weather_indexes` that returns the 6 things, but our `Indicator` class doesn't support multiple outputs. **Question** : Should we create a `MultiIndicator` class OR expose individual indexes and simply drop everything (really not efficient if you want more than one).
 - Optimize the computation. I haven't vectorized everyting yet, I wanted to try different options before changing the subfunctions.

If any of you have ideas on how to reduce the code complexity, je suis preneur! I did it like this because of the limitations from `xr.apply_ufunc`, not wanting to duplicate code and the need to compute everything in the same loop to stay efficient...

PR #236, opened 2019-06-18
[WIP] simplify subset_gridpoint

davidcaron:
- [ ] Tags have been pushed

We have a dataset that is ordered lon,lat instead of lat,lon. I believe this is a bit uncommon...

In the `subset_gridpoint` function, the current implementation assumes that we have lat,lon. When doing the `isel`, these dimensions were inverted.

No

### DIFF
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -107,30 +107,7 @@ def subset_gridpoint(da, lon, lat, start_yr=None, end_yr=None):
     >>> dsSub = subset.subset_gridpoint(ds, lon=-75,lat=45,start_yr=1990,end_yr=1999)
     """
 
-    g = Geod(ellps='WGS84')  # WGS84 ellipsoid - decent globaly
-    # adjust negative/positive longitudes if necessary
-    if np.all(da.lon > 0) and lon < 0:
-        lon += 360
-    if np.all(da.lon < 0) and lon > 0:
-        lon -= 360
-
-    if len(da.lon.shape) == 1 & len(da.lat.shape) == 1:
-        # create a 2d grid of lon, lat values
-        lon1, lat1 = np.meshgrid(np.asarray(da.lon.values), np.asarray(da.lat.values))
-
-    else:
-        lon1 = da.lon.values
-        lat1 = da.lat.values
-    shp_orig = lon1.shape
-    lon1 = np.reshape(lon1, lon1.size)
-    lat1 = np.reshape(lat1, lat1.size)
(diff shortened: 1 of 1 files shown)


xclim/subset.py
```diff
@@ -129,12 +129,9 @@ def subset_gridpoint(da, lon, lat, start_yr=None, end_yr=None):
     dist = dist.reshape(shp_orig)
 
     iy, ix = np.unravel_index(np.argmin(dist, axis=None), dist.shape)
-    xydims = [x for x in da.dims if 'time' not in x]
 
-    args = dict()
-    args[xydims[0]] = iy
-    args[xydims[1]] = ix
-    out = da.isel(**args)
+    out = da.isel(lat=iy, lon=ix)
```
huard: I think a cleaner solution for later would be to compute new x, y coordinates using a distance preserving projection lon, lat->x, y , then use sel(x=.., y=.., method='nearest')

davidcaron: I agree... please take a look at the next commit. It might look a bit drastic, but I think we don't even need to compute the distances on the ellipsoid here. I think taking the nearest latitude and longitude values leads to the same result. 

I might be missing something, but if we are working with latitudes and longitudes, I think we can do that.

The gridpoint selection is roughly 10 times faster this way.

What do you think?

huard: My guess is that will work for most cases, but there will be instances where the selected point will not be the closest in terms of physical distance. For the BCCAQv2 dataset, I don't think we would have such cases.

tlogan2000: We could possibly introduce a 'fast_select' input flag  which would use the ```.sel()``` method.  If fast_select is false it would use the current implementation using true distances.   I think the true distance method is likely to be useful for data with relatively coarse spatial resolutions

tlogan2000: However the ```.sel()``` method will likely not work on data whose dimensions are 'rotated pole coordinates'  ```rlon``` and ```rlat```

davidcaron: A few lines above, the `da.lon` and `da.lat` attributes are accessed, so these would also need to change in case of `rlon` and `rlat` coordinates?

Because the grid will always be evenly spaced (the current implementation represents this using `np.meshgrid` and looks for the nearest point), even if the distance used to determine the nearest point is not the distance on the ellipsoid, the nearest lat and lon values should still correspond to the same indices... unless we are dealing with a very funky projection, I have a hard time coming up with a case where this won't be true.

PR #239, opened 2019-06-25
Pre-Commit Hooks and Black code formatting

Zeitsperre:
- [ ] Tags have been pushed

Pre-commit is a tool that examines changes prior to `git commit`ting them by running "hooks". This allows us to do things automatically check for inconsistencies with PEP8 as well as ensure that `.yaml` is written correctly and other goodies. For instance, the `import` order and other PyCharm-like code formatting tools can be run at commit time even from command-line, as well as automatic call formatting using the latest python conventions (e.g. changing string formatting from `"%s" % foo` to `"{}".format(foo)"`).

Functionally, no. Black is a set of python code compliances based largely on PEP8 but breaks PEP8 in a few small ways (e.g. line-length max is dynamic instead of static, spaces between `:` are sometimes allowed, etc.). Like other autoformatting tools, this changes the code in small ways for readability/maintainability but has no effect on the function.

* **Other information**:
The goal of this change is to reduce the amount of time spent performing PEP8 compliancy on the code base and to follow industry-developed code formatting conventions at the same time.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -60,38 +60,62 @@ Get Started!
 Ready to contribute? Here's how to set up `xclim` for local development.
 
 1. Fork the `xclim` repo on GitHub.
+
 2. Clone your fork locally::
 
     $ git clone git@github.com:Ouranosinc/xclim.git
 
 3. Install your local copy into a virtualenv. Assuming you have virtualenvwrapper installed, this is how you set up your fork for local development::
 
+    # For virtualenv environments:
     $ mkvirtualenv xclim
+
+    # For Anaconda/Miniconda environments:
+    $ conda create -n xclim python=3.6
+
     $ cd xclim/
-    $ pip install -e .    
+    $ pip install -e .
 
 4. Create a branch for local development::
 
     $ git checkout -b name-of-your-bugfix-or-feature
 
-   Now you can make your changes locally.
+   Now you can make your changes locally!
 
-5. When you're done making changes, check that your changes pass flake8 and the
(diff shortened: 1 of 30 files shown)


tlvu:
Very interesting this pre-commit hook.  Does that run locally only or on TravisCI?

Zeitsperre:
It does both. I have it integrated into the project itself from hooks within .git and it integrates with PyCharm fairly well (some assembly required). Running `git commit` launches the checks and if anything changes, it'll send up a warning and perform the formatting; Once everything passes, the commit will be stored.

Since `black` is effectively a subset of PEP8 guidelines, it works almost exactly like `flake8` or `autopep8`. I have it running via a `tox.ini` configuration but it can be configured to run from `.travis.yml`.

Zeitsperre:
To be more clear, the pre-commit hook is specifically for local development, but all the hooks that are called can be configured as checks within Travis CI. If we wanted to have an autoformatting service within the repo, we'd need to configure a GitHub hook service/bot.

tlvu:
I think local development is already a very good start.  If we see people are bypassing it locally and still pushing bad code, then we can think about enabling it on TravisCI.

Zeitsperre:
@huard I'm planning on adding a few notes on installing/configuring pre-commit to the docs later today and bumping the minor version. I don't want to keep this PR open too long since it contributes fairly large changes to the code base. Wondering what your thoughts are.

Zeitsperre:
That's fair. Old(/bad) habits die hard. I'll perform a squash merge and creates entries for both major changes once I've made the changes to the docs.

PR #240, opened 2019-06-25
Doc cleanup and examples

### LINKED ISSUE #13: Documentation: "Example Workflows"

### LINKED ISSUE #159: Write post on xclim's progress
### Description

Write a couple of paragraphs on xclim, usage and performance.

Zeitsperre:
- [ ] Tags have been pushed

Changed some wording in the docs for grammar and consistency with development practices

No.

* **Other information**:
fixes #159 
fixes #13 

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1,13 +1,95 @@
 Usage
 =====
 
-To use xclim in a project::
+Basic example
+-------------
+
+To use xclim in a project:
+
+.. code-block:: python
 
     import xclim
 
+To open NetCDF (`.nc`) climate data sets, use xarray:
+
+.. code-block:: python
+
+    import xarray as xr
+
+    data_file = "tas-daily-historical-data.nc"
+    ds = xr.open_dataset(data_file)
+
+.. note::
+    Calculations are performed on the variable of interest present within the data set (in this example, **ds.tas**) and not the data set itself, so be sure to specify this when using xclim's indices and other calls.
+
+To perform a simple climate indice calculation:
+
+.. code-block:: python
+
+    gdd = xclim.indices.growing_degree_days(ds.tas, thresh='10.0 degC', freq='YS')
+
+To plot a time slice of the output, we recommend the following:
+
+.. code-block:: python
+
+    import matplotlib.pyplot
+
+    # for a summary statistics histogram
+    gdd.plot()
+
+    # for a specific geographical coordinate
+    gdd.isel(lon=-75, lat=47).plot()
+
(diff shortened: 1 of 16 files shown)


Zeitsperre:
For reference, the issue with integrating Jupyter Notebooks into the docs here is due to a few factors/bugs:
- Pandoc is needed to parse the cells and create sphinx text out of them. If the HOME is not set in the environment (ie: the TOXENV), the build will fail. This is well documented [here](https://github.com/bebraw/pypandoc/pull/104)
- TODO: Ipython

Zeitsperre:
@huard Thanks for noting that. I almost forgot!

I'll ask @tlogan2000 to re-run the cells as the outputs are not 100% correct at the moment. Some of the examples were based on files on his desktop or accessible via mounts he configured. Once the cells are updated, I'll send another review request.

huard:
Ideally the examples would be self-contained, so that anyone can run them and get the same answer.

Zeitsperre:
Good point. 

I'll see if I can simplify the MF-dataset example. It's the only one that would be complicated without access to the database. Might solve more than one problem.

huard:
I think this would fix #13

tlogan2000:
@huard @Zeitsperre  made a few changes adds to the workflow notebook.  I think we should be good to go ...

Zeitsperre:
@huard I still have a few things to add to `usage.py` but feel free to review.

huard:
My feeling is that the notebook is too long, and should start with a simple computation of a climate indicator, highlighting the fact that all the netcdf attributes are there, that units are handled, etc. 

I would create another notebook for subsetting utilities., and another about optimizations. 

That being said, I'm totally OK with this PR as is and we can fine-tune later, including using a web accessible input file.

docs/usage.rst
```diff
@@ -1,13 +1,95 @@
...
+
+Slicing and subsetting with xarray
+----------------------------------
+
+
+Using the xclim.icclim module
+-----------------------------
+`ICCLIM <https://github.com/cerfacs-globc/icclim>`_ is a software platform / python library for performing climate indice calculations. It includes roughly 50 types of indicators as defined by the `European Climate Assessment & Dataset project <https://www.ecad.eu/>`_.
+
+For those familiar with ICCLIM, xclim has created one-to-one mappings of their indices. Effectively, these are simply indice calculations methods that have been renamed to be consistent with ICCLIM's terminology.
```
huard: not all indices, and not a perfect mapping.

PR #247, opened 2019-07-02
Fix 236

tlogan2000:
### Pull Request Checklist:
- [X] Docs have been added / updated (for bug fixes / features)
- [X] HISTORY.rst has been updated

- [ ] Tags have been pushed

Bug fix, #236

no

* **Other information**:
subset_gridpoint() will now use `.sel` method if lon, lat dims are in the dataset. Otherwise it uses old version based on distance 

This PR fixes #236

### DIFF
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -58,19 +58,7 @@ def subset_bbox(da, lon_bnds=None, lat_bnds=None, start_yr=None, end_yr=None):
         )
 
     if start_yr or end_yr:
-        if not start_yr:
-            start_yr = da.time.dt.year.min()
-        if not end_yr:
-            end_yr = da.time.dt.year.max()
-
-        if start_yr > end_yr:
-            raise ValueError("Start date is after end date.")
-
-        year_bnds = np.asarray([start_yr, end_yr])
-        da = da.where(
-            (da.time.dt.year >= year_bnds.min()) & (da.time.dt.year <= year_bnds.max()),
-            drop=True,
-        )
+        da = subset_time(da, start_yr=start_yr, end_yr=end_yr)
 
     return da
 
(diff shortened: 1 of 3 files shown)


xclim/subset.py
```diff
@@ -114,36 +114,48 @@ def subset_gridpoint(da, lon, lat, start_yr=None, end_yr=None):
...
-    dist = dist.reshape(shp_orig)
-
-    iy, ix = np.unravel_index(np.argmin(dist, axis=None), dist.shape)
-    xydims = [x for x in da.dims if "time" not in x]
-
-    args = dict()
-    args[xydims[0]] = iy
-    args[xydims[1]] = ix
-    out = da.isel(**args)
+    # check if trying to subset lon and lat
```
huard: This will fail if lat or lon == 0.

xclim/subset.py
```diff
@@ -114,36 +114,48 @@ def subset_gridpoint(da, lon, lat, start_yr=None, end_yr=None):
...
+        if hasattr(da, "lon") and hasattr(da, "lat"):
+            # adjust negative/positive longitudes if necessary
+            if np.all(da.lon > 0) and lon < 0:
+                lon += 360
+            if np.all(da.lon < 0) and lon > 0:
+                lon -= 360
+
+            dims = list(da.dims)
+
+            if "lat" in dims and "lon" in dims:
```
huard: Add comment to explain the logic.

xclim/subset.py
huard: line 118

tlogan2000:
> Does it also fix the issue raised by David C ?

The issue of different dimension orders? If yes, then it should be solved as I no longer create a 2d meshgrid out of 1d lon lat dim values but instead use the `sel` method.  The 'distance' method is therefore only used when lon, lat dims are not present (i.e. rlat and rlon) and is calculated directly on the irrelgular 2d lon, lat values already present in the dataset. 

Just to make sure I added a test for this in test_utils in `test_irregular` where I transpose the data before running the subset

tlogan2000:
test failing ... @huard wait a minute.

tlogan2000:
Test should be working now...

xclim/subset.py
tlogan2000: now checks to see if lon/lat is `None`

PR #258, opened 2019-08-21
Accept float in days_over_precip_thresh

### LINKED ISSUE #79: Percentile threshold indices : Frequencies?
For indices such as 'R95p' (with the following definition taken from icclim)
-  Days with RR > 95th percentile of daily amounts (very wet days)
(days)

- Let RRwj be the daily precipitation amount at wet day w (RR ≥ 1.0 mm)
of period j and let RRwn95 be the 95th percentile of precipitation at wet
days in the 1961–1990 period. Then counted is the number of days where:

- RRwj > RRwn95

Does the definition of RRwn95 (the percentile threshold value calculated from the reference period) change with the frequency? i.e if the R95p resampling value is seasonal does the RRwn95 value get calculated seasonally? or is the percentile threshold always annual?  ICCLIM docs are unclear on thisI

huard:
### Pull Request Checklist:

- [ ] Tags have been pushed

* Add support for simple annual percentiles. 

No

* **Other information**:
TODO
- [ ] Review code and naming.
- [x] Apply same logic to other relevant functions

This PR fixes #79

### DIFF
--- a/xclim/indices/_multivariate.py
+++ b/xclim/indices/_multivariate.py
@@ -558,7 +558,7 @@ def days_over_precip_thresh(pr, per, thresh="1 mm/day", freq="YS"):
     ----------
     pr : xarray.DataArray
       Mean daily precipitation flux [Kg m-2 s-1] or [mm/day]
-    per : xarray.DataArray or str
+    per : xarray.DataArray
       Daily percentile of wet day precipitation flux [Kg m-2 s-1] or [mm/day].
     thresh : str
        Precipitation value over which a day is considered wet [Kg m-2 s-1] or [mm/day].
@@ -570,10 +570,6 @@ def days_over_precip_thresh(pr, per, thresh="1 mm/day", freq="YS"):
     xarray.DataArray
       Count of days with daily precipitation above the given percentile [days]
 
-    Notes
-    -----
-    The percentile should be computed for a 5 day window centered on each calendar day for a reference period.
-
     Example
     -------
     >>> p75 = percentile_doy(historical_pr, per=0.75)
@@ -583,7 +579,7 @@ def days_over_precip_thresh(pr, per, thresh="1 mm/day", freq="YS"):
     thresh = utils.convert_units_to(thresh, pr)
 
     tp = np.maximum(per, thresh)
-    if isinstance(per, xr.DataArray):
+    if "dayofyear" in per.coords:
         # Create time series out of doy values.
         tp = utils.resample_doy(tp, pr)
 
@@ -606,7 +602,7 @@ def fraction_over_precip_thresh(pr, per, thresh="1 mm/day", freq="YS"):
     ----------
     pr : xarray.DataArray
       Mean daily precipitation flux [Kg m-2 s-1] or [mm/day].
-    per : xarray.DataArray or str
+    per : xarray.DataArray
       Daily percentile of wet day precipitation flux [Kg m-2 s-1] or [mm/day].
     thresh : str
        Precipitation value over which a day is considered wet [Kg m-2 s-1] or [mm/day].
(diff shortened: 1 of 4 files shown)


xclim/indices/_multivariate.py
```diff
@@ -567,8 +567,8 @@ def days_over_precip_thresh(pr, per, thresh="1 mm/day", freq="YS"):
     ----------
     pr : xarray.DataArray
       Mean daily precipitation flux [Kg m-2 s-1] or [mm/day]
-    per : xarray.DataArray
-      Daily percentile of wet day precipitation flux [Kg m-2 s-1] or [mm/day]
+    per : xarray.DataArray or string (scalar with units)
```
Zeitsperre: I don't think this will work with IDEs. From my understanding, this field acts as a call signature hint, the same as what we might find in `def function(a: str = 'foo', b: int or float = 123.321, c: xarray.DataSeries = ds)`.
 
Outside of types joined with `or` or with `[ ]`, the only other thing allowed for this type of docstring is the `, optional` indicator that references the `typing` class of `Optional` (`def func(a: Optional[str] = None`). 

I'd suggest adding that the units are needed in the docstring and just raise a verbose ValueError if there aren't any units. 
```suggestion
    per : xarray.DataArray or str
```

huard:
Is the name `resample_doy` sufficiently informative ? `resample_from_doy` ?

huard:
@Zeitsperre Are the azure failures expected ? Is this something we really want to invest in ?

Zeitsperre:
@huard Yeah, I'm thinking it might be worthwhile to move from Azure as well. The goal was simply to perform the same sort of platform checks as macOS. I learned a few weeks ago that Travis CI recently added a Windows image. I'd much rather use their infra.

I'll open an issue for replacing Azure with Travis CI for Windows builds.

huard:
@tlogan2000 Could you take a look ?

Zeitsperre:
@huard The Azure builds are no longer valid. There was a misconfiguration in their .yaml and I didn't see the value in investing time into making them all work. I feel like we can wait for Travis CI to create some Windows images.

For this PR, I would simply ignore those checks. Everything is passing on Travis.

tlogan2000:
> @tlogan2000 Could you take a look ?

I will take a look at this soon... sorry it slipped through a crack I think

tlogan2000:
@huard with the roll back of float/string input what new features does this PR still include?  Is it the doy_resample()?

huard:
Yes, that's mostly it. I'd call it a clean-up of how doy percentiles are handled. 

Temperature indicators require a doy percentile, while precip percentiles also accept a single "annual" value.

tlogan2000:
> Yes, that's mostly it. I'd call it a clean-up of how doy percentiles are handled.
> 
> Temperature indicators require a doy percentile, while precip percentiles also accept a single "annual" value.

Sounds good

huard:
Made change in docstring. Won't really work as-is however until xarray 0.13 because keep_attrs does not yet work for DataArray.quantiles.

Zeitsperre:
@huard if you want to update the history and `bumpversion patch` master, I can pull the new master and send it to PyPI/conda-forge.

huard:
@Zeitsperre 
So bumpversion does not actually change the tag anymore ? I had to commit manually the change and call git tag myself... weird.

Zeitsperre:
The tagging is re-enabled on another PR since I realize it doesn't cause commits to fail (only `commit = true` seems to do that). I'll inform people when that is in master.

I had it turned off since it was conflicting with pre-commit. It turns out that the `bumpversion` python library has a few bugs with automatically adding lines (causing code formatting to fail). This might be fixed in a fork that Birdhouse uses called `bump2version` but I need to look into it.

Issue #257, opened 2019-08-21
Stabilizing the API


Zeitsperre:
I was watching a really great talk by a MongoDB developer on how to warn of deprecations in APIs. The talk is on [youtube](https://www.youtube.com/watch?v=dqDnB6jKzcE) but his guide is available here: https://emptysqua.re/blog/api-evolution-the-right-way/.

Among the things he mentioned is how best to use libraries like `warnings` with stack-levels, and using the `*` in function call signatures to demand users use keywords, rather than order to launch processes. Is this something that we might want to consider integrating by v1.0?

huard:
Sounds reasonable.

Zeitsperre:
Let's discuss this in the meeting.

Zeitsperre:
The API has been stable (no major breaking changes) for at least 6 months now. I think it's safe to close this issue.

PR #266, opened 2019-09-09
Added a temperature threshold for freezethaw cycle events

### LINKED ISSUE #265: Harmonize freeze-thaw cycle computation with PCC
PCC is using thresholds different than 0.

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

A threshold is now implemented for freezethaw cycles 

Yes. With the exception of Atmos Indices, variables with thresholds will eventually require user-specified key=value pairs. This is now implemented fro daily_freezethaw_cycle calculations via the "UNSET " keyword prepended to the threshold. Calculations should eventually use this approach alongside other changes such as keyword call enforcement, e.g.:
```
def function(*,input=None, key1=word1, key2=word2):
    do stuff
```
* **Other information**:

This PR fixes #265

### DIFF
--- a/xclim/indices/_multivariate.py
+++ b/xclim/indices/_multivariate.py
@@ -154,9 +154,15 @@ def cold_and_dry_days(tas, tgin25, pr, wet25, freq="YS"):
 
 
 @declare_units(
-    "days", tasmax="[temperature]", tasmin="[temperature]", thresh="[temperature]"
+    "days",
+    tasmax="[temperature]",
+    tasmin="[temperature]",
+    thresh_tasmax="[temperature]",
+    thresh_tasmin="[temperature]",
 )
-def daily_freezethaw_cycles(tasmax, tasmin, thresh="0 degC", freq="YS"):
+def daily_freezethaw_cycles(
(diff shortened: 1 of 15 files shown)


tlogan2000:
FYI - PCC has indivudual tmin and tmax thresholds ...

xclim/indices/_multivariate.py
```diff
@@ -153,18 +154,34 @@ def cold_and_dry_days(tas, tgin25, pr, wet25, freq="YS"):
...
+    "days",
+    tasmax="[temperature]",
+    tasmin="[temperature]",
+    thresh_tasmax="[temperature]",
+    thresh_tasmin="[temperature]",
+)
+def daily_freezethaw_cycles(
+    tasmax,
+    tasmin,
+    thresh_tasmax="UNSET 0 degC",
```
huard: What purpose serves this UNSET ?

Zeitsperre: It's a trick I picked up to warn users of a future deprecation. 

If they don't set the thresh_tasmax or thresh_tasmin manually, this creates a very visible warning that in the future they will need to do so. We are eventually going to strictly integrate the `*` into function calls so that call signatures cannot be performed based on order. This relates to the discussion brought up in #257.

Issue #278, opened 2019-10-02
Implement `subset_polygon` function


huard:
### Description

Create a function that takes as input a single polygon and a xarray.Dataset/xarray.DataArray and returns a cropped dataset/dataarray over the polygon, with masked values outside the polygon.

tlogan2000:
I think there has been some miscommunication here I thought that this type of discussion had already taken place and had more or less indicated to @Zeitsperre to try to go forward with the methodology using geopandas seen in #285 

I suggest taking a step back:

Here is a list of potential libraries that can subset polygons:
* rioxarray
* geoxarray
* xgeo
* regionmask
* rasterx

@Zeitsperre before continuing with #285 (beyond the couple tests you had almost finished) I think we could all use some summary knowledge on what the capabilities of each of these packages are in terms of the following use cases (@huard @Zeitsperre @RondeauG  feel free to add to the list that I might of forgot)

* Reprojection capabilities - Handling all positive longitudes 

* Reprojection capabilities - Edge case of above : handling prime meridian crossing. E.g. when polygon crosses 0 deg lon but ds longitude values are all positive.  With #285  we saw that we need to split the polygon in order to subset properly. How/can other libraries deal with this?

* Curvilinear lat/lon grids - Can these libraries handle subsetting of rotated pole data such as CRCM5

* Multipolygon geometries - Can the input shape be a Multipolygon? Note this was my main reason for employing the spatial-join method recycled in #285 

* Grid Centroids not in polygon : The polygon is small enough that not grid centroids are fall within it.  Can it detect gridcells that intersect (e.g. lon/lat bounds) or use a buffer as suggest

* Could #285 pass these other libraries' test suites?

huard:
I'd add to that 
* user interface (how intuitive/user-friendly is the interface to subset a Dataset ?)

Zeitsperre:
Fixed with #285

PR #285, opened 2019-11-01
Polygon subsetting

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

This adds the `subset.subset_shape()` function which allows polygon subsetting using a GeoJSON, Shapefile, single-layer GeoPackage, or any other `fiona` readable formats. The function can be called as follows:
`ds_subset = xclim.subset.subset_shape(ds, shape=geo_file)`

No*. Because of the need to project NetCDF files in order for them to be understoud by tools like rasterio, netCDFs that typically don't have CRS information need to be converted via `rioxarray`. This requires users to have loaded their NetCDF with the rioxarray library installed. Examples can be found in the docstring for the function.

* **Other information**:
`rioxarray` opens up new opportunities for features, such as writing DataArrays directly to GeoTIFF format and other GIS functions. This demands more investigation.


xclim/subset.py
```diff
@@ -125,11 +133,188 @@ def func_checker(*args, **kwargs):
...
+      CRS of the geometries provided. If passing GeometryCollections as shapes, CRS must be explicitly stated.
+    use_all_features : bool
+      Use either the first found feature geometry or the union of all found features. Default: False.
+    start_yr : int
+      Deprecated
+        First year of the subset. Defaults to first year of input data-array.
+    end_yr : int
+      Deprecated
+        Last year of the subset. Defaults to last year of input data-array.
+
```
huard: You're not documenting geometry.

xclim/subset.py
```diff
@@ -125,11 +133,188 @@ def func_checker(*args, **kwargs):
...
+    Subset multiple variables in a single dataset
+    >>> ds = xarray.open_mfdataset(['pr.day.nc','tas.day.nc'])
+    >>> dsSub = subset.subset_bbox(ds, shape="/path/to/polygon.shp", start_yr='1990', end_yr='1999')
+     # Subset with year-month precision - Example subset 1990-03-01 to 1999-08-31 inclusively
+    >>> prSub = subset.subset_time(ds.pr, shape="/path/to/polygon.shp", start_date='1990-03', end_date='1999-08')
+    # Subset with specific start_dates and end_dates
+    >>> prSub = \
+            subset.subset_time(ds.pr, shape="/path/to/polygon.shp", start_date='1990-03-13', end_date='1990-08-17')
+    """
+    try:
```
huard: Add a few comments to explain what's going on.

xclim/subset.py
```diff
@@ -125,11 +133,188 @@ def func_checker(*args, **kwargs):
...
+      Defaults to first day of input data-array.
+    end_date : Optional[str]
+      End date of the subset.
+      Date string format -- can be year ("%Y"), year-month ("%Y-%m") or year-month-day("%Y-%m-%d").
+      Defaults to last day of input data-array.
+    da_crs : Optional[Union[int, dict, str]]
+      CRS of the xarray.DataArray or xarray.Dataset provided. Default: dict(epsg=4326).
+    shape_crs : Optional[Union[int, dict, str]]
+      CRS of the geometries provided. If passing GeometryCollections as shapes, CRS must be explicitly stated.
+    use_all_features : bool
```
huard: That seems counter-intuitive ? Why would you pass a shapefile and not want to use all features ?

huard: Also, this argument is for the benefit only of the decorator. I don't think this is the intended use of decorators.

xclim/subset.py
```diff
@@ -125,11 +133,188 @@ def func_checker(*args, **kwargs):
...
+        raise
+
+    ds_out = xarray.Dataset(data_vars=None, attrs=da.attrs)
+    for v in da.data_vars:
+        if "lon" in da[v].dims and "lat" in da[v].dims:
+            dss = da[v]
+            dss.rio.set_spatial_dims(x_dim="lon", y_dim="lat", inplace=True)
+            dss.rio.write_crs(crs, inplace=True)
+            ds_out[v] = dss.rio.clip(
+                geometry, crs=dss.rio.crs, all_touched=True, drop=True, invert=False
```
huard: Is it possible to get the weights as well ? That is, the fraction of each grid cell's area included in the subset.

huard: If it's not provided out of the box by rioxarray, I suggest we open a ticket for this.

xclim/subset.py
```diff
@@ -125,11 +133,188 @@ def func_checker(*args, **kwargs):
...
+                ds_out[v] = da[v].sel(lat=ds_out.lat)
+            elif "lon" in da[v].dims:
+                ds_out[v] = da[v].sel(lon=ds_out.lon)
+            else:
+                ds_out[v] = da[v]
+
+    if start_date or end_date:
+        ds_out = subset_time(ds_out, start_date=start_date, end_date=end_date)
+
+    return ds_out
```
huard: It seems you're returning a Dataset whether or not the input was a DataArray or a Dataset.

huard:
FYI, this is how ESMValTool does it: https://github.com/ESMValGroup/ESMValCore/blob/92ba9538b51eec92db578c3cb92eb343d9a93e24/esmvalcore/preprocessor/_area.py#L313

How does it compare to your approach ?

Zeitsperre:
@huard No comparison. Much of the work I do is to ensure that the data is well understoud by rioxarray. I'm operating under the assumption that that extension is optimized for cutting data. 

From my interactions with the maintainer of rioxarray, they seem keen to see contributors suggest changes. I can possibly contribute some of this handling to that project to help simplify the logic here. One gain I have in my approach is that it handles multi-dimensional data fairly well while the approach you linked, from what I can tell only handles 2-dimensional data.

xclim/subset.py
Zeitsperre: Zonal statistics is something in rioxarray but it requires a bit of work on both xclim and rioxarray. I can look more into this in another PR if we want to have regional statistics reporting.

xclim/subset.py
Zeitsperre: I added this in order to quickly test for shapefiles and left it in as a "feature". Removing in now.

xclim/subset.py
Zeitsperre: Fixed

xclim/subset.py
Zeitsperre: Done.

tlogan2000:
FYI Failed tests for subset.py seem to be due to the release of cftime>=1.0.4.  released 10/21/2019. https://github.com/Unidata/cftime

XARRAY devs seems to be on it as the build  versus their master seem to pass.

xclim/subset.py
huard: To be clear, we'd be interested in having, for each cell within the subset, the percent area of the total region area. This would account for the both partial intersection of a grid with the polygon, and differences in individual grid cell areas. Ok, let's open a new PR once this one's done.

xclim/subset.py
Zeitsperre: This is now dealt with.

huard:
@Zeitsperre I'm simply wary of

1. leveraging new emerging libraries with few contributors,
2. adding dependencies.  

I'm not against it, but it should be the result of an exercise of pros vs cons, especially if you feel we'll need to invest time in upstream development. 

On the minus side, I see for example that it breaks on Python3.5, which I think we're still running internally, right ? 

On the plus side, we could replace our subset_bbox by the rioxarray implementation, and I see other valuable features that we could exploit. 

I'd be surprised ESMValTool's subset only applied to 2D fields. What makes you think that ?

Zeitsperre:
@huard I'm with you on this.

Fairly disappointed about the Python3.5 non-support. PyPI reported that it was 3.5-compatible so I figured the failures were due to my code. This may be a bust for the time being. Regrettably, I can't port this to miranda either as I'm trying to keep it Python3.5-compatible as well. It might be best to put a pin in this until a later time (or make a few functions exclusive to 3.6+, an idea I really don't like).

I was under that impression ESMValTools doesn't support it based on the logic of some of the checks. I should give it another look. The other method out there is to leverage GeoPandas and Shapely to make a virtual grid that's used to perform basic map algebra. It's currently not clean but I can try that approach next. Too bad.

huard:
xarray is dropping support for 3.5, so we'll have to drop it too. 

Basically, I'd like you to take stock of what our options are, and suggest a solution that sounds solid from a development and maintenance point of view.

Zeitsperre:
@huard As I mentioned briefly in person to @tlogan2000 and yourself, here's a game plan:

- Dropping Python3.5 can happen eventually. Xarray isn't ensuring builds using 3.5 are passing so eventually, this will be the case here. Let's open a ticket and set the next minor version goals to drop 3.5 as well.

- The rioxarray is a neat library and approach to this problem, so rather than erase the work I've done, I will be porting the subsetting code to https://github.com/Ouranosinc/miranda as I have already used it in conjunction with that library for other projects. The code for manually warping the lats and lons of the NetCDF is something that the rioxarray maintainer is interested in that I'll try contributing in my free time. Once that project has matured a little, stabilized its API, or been integrated into xarray more formally, we can integrate it here.

- I'm going to explore the approach for subsetting polygons using GeoPandas. rather than modify the grids of the NetCDFs, reprojecting and manually converting geocoordinates of GeoPandas means we aren't manipulating the NetCDFs as much, which is probably better for a version 1 of the function. We can also leverage GeoPandas for other GIS-like functions and optimize the `subset_bbox` function at the same time. It's heavier on dependencies but we know it works.

I should be able to start on this later today or tomorrow.

tests/test_subset.py
```diff
@@ -517,6 +517,22 @@ def test_warnings(self):
...
+    meridian_geojson = os.path.join(TESTS_DATA, "cmip5", "meridian.json")
+
+    def test_simple(self):
+        ds = xr.open_dataset(self.nc_file)
+
+        with pytest.warns(UserWarning):
+            sub = subset.subset_shape(ds, self.meridian_geojson)
+
+        assert len(sub.tas) == 12
+        np.testing.assert_equal(np.mean(sub.tas.isel(time=0)), 285.057)
```
huard: Add comment explaining where the expected value comes from.

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
...
+         transforming the lons from WGS84 to WGS84 +lon_wrap=180 (longitudes from 0 to 360).
+
+         Returns a GeoDataFrame with the new features in a wrap_lon WGS84 projection if needed.
+        """
+        try:
+            poly = kwargs["poly"]
+            x_dim = kwargs["x_dim"]
+            wrap_lons = kwargs["wrap_lons"]
+        except KeyError:
+            return func(*args, **kwargs)
```
huard: It seems dangerous to me to let a decorator "fail" silently. I think if you use a decorator in your code, you expect it to work, no ?

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
                     kwargs[lon] -= 360
                 else:
                     kwargs[lon][kwargs[lon] < 0] -= 360
+
         return func(*args, **kwargs)
 
     return func_checker
 
 
+def wrap_lons_and_split_at_greenwich(func):
```
huard: Isn't this solving an issue that regionmask has already solved ?

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
...
+                "+proj=longlat +ellps=WGS84 +lon_wrap=180 +datum=WGS84 +no_defs"
+            )
+            kwargs["poly"] = poly
+
+        return func(*args, **kwargs)
+
+    return func_checker
+
+
+@wrap_lons_and_split_at_greenwich
```
huard: Again, how does this compare with regionmask ?

xclim/subset.py
Zeitsperre: I think this is in case the function that gets called doesn't require any of this treatment. I was hoping to show that it could leave early if the kwargs aren't present, but I could raise an error or a warning to suggest that this check isn't being performed?

xclim/subset.py
Zeitsperre: Yes, I see that Regionmask this function as well. There were a few reasons why I didn't consider integrating Regionmask. This is worth a longer discussion, but it looks as though the maintainer of Regionmask is considering some of the approaches shown here. Perhaps we can work on that project and integrate it into our work here?

xclim/subset.py
Zeitsperre: I haven't explored Regionmask too much, but it's essentially the same logic as Regionmask's `utils.wrap_angle`, except deployed as an automatic check with a wrapper.

tests/test_subset.py
Zeitsperre: Done.

xclim/subset.py
huard: Considering that the subsetting functions affect a lot of people, I think it's important we discuss and document high-level decisions before implementation.

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
...
+    >>> prSub = subset.subset_shape(ds.pr, shape="/path/to/polygon.shp", start_date='1990-03', end_date='1999-08')
+    # Subset with specific start_dates and end_dates
+    >>> prSub = \
+            subset.subset_shape(ds.pr, shape="/path/to/polygon.shp", start_date='1990-03-13', end_date='1990-08-17')
+    """
+    if "ts" in ds.coords:
+        ds = ds.drop("ts")
+
+    poly = gpd.GeoDataFrame.from_file(shape)
+
```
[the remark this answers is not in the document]
Zeitsperre: I would also specify in the docstring that the base units are those of the vector, and not the xarray object. Thanks for the suggestion.

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
...
+        )
+        gdf_points = gpd.GeoDataFrame(df, geometry="Coordinates", crs=shifted)
+    else:
+        gdf_points = gpd.GeoDataFrame(
+            df, geometry="Coordinates", crs=CRS().from_epsg(4326)
+        )
+
+    # spatial join geodata points with region polygons
+    point_in_poly = gpd.tools.sjoin(gdf_points, poly, how="left", op="within")
+
```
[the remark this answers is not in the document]
Zeitsperre: Sounds good!

xclim/subset.py
huard: Does your implementation pass regionmask test suite ?

tlogan2000:
https://github.com/Ouranosinc/xclim/pull/285#discussion_r352714780
With respect to regionmask..  activity was pretty sparse from 2016 until just recently
https://github.com/mathause/regionmask/graphs/code-frequency

I think I had come across it but was unsure whether it was still being maintained ...

huard:
Also https://geosynopsis.github.io/xgeo/docs/html/examples.html

```
dsout = ds.geo.subset(vector_file='vector.shp',crop=True)
```

Zeitsperre:
Just to be clear, I'm not opposed to eventually integrating a library that specifically performs subsetting functionality; The issue I find is that there are a lot of projects in development to perform exactly this:

- rioxarray
- geoxarray
- xgeo
- regionmask
- rasterx

There's currently a discussion thread on https://github.com/corteva/rioxarray/issues/4 looking into how best to integrate the work on shapefile subsetting of xarray objects. For the time being, I feel like having a function that can eventually be replaced by something emergent from theses projects would be a good way forward.

xclim/subset.py
Zeitsperre: I've looked into it now and the test approach for what I do here and the regionmask approach aren't exactly compatible. I do need to beef up the tests on this PR, but because the base type needed for the `create_mask` function is a GeoPandas dataframe, it's not all that similar to regionmask in how to define a new test region.

xclim/subset.py
```diff
@@ -125,11 +182,412 @@ def func_checker(*args, **kwargs):
...
+        """
+        try:
+            poly = kwargs["poly"]
+            x_dim = kwargs["x_dim"]
+            wrap_lons = kwargs["wrap_lons"]
+        except KeyError:
+            return func(*args, **kwargs)
+
+        if wrap_lons:
+            if np.min(x_dim) < 0 or np.max(x_dim) > 360:
```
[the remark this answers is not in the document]
Zeitsperre: Are you saying that we need to make a case for when both the NetCDF and Shape are in WGS84? In that scenario, we wouldn't want to perform any wrapping of longitudes I imagine.

huard:
Ok. I'll raise the broader issue with our European colleagues to try to find out what a long-term community solution could look like.

tlogan2000:
I suggest pushing the broader discussion to https://github.com/Ouranosinc/xclim/issues/278#issuecomment-560543182

tlogan2000:
Triggered rebuild... Kmeans failing?

Zeitsperre:
Just spoke with @aulemahal about this. scikit-learn released a new version that corrected a bug that our setup was affected by. The fix will be in another PR.

aulemahal:
Exact, this is issue #317 .

tlogan2000:
Geopandas dependencies on windows are tricky to install I think.  I would update the xclim install documentation and point towards geopandas for dependencies.. 

I would probably have a note saying windows users (potentially Mac?) should  consider using conda to install geopandas to save themselves headaches

Zeitsperre:
@huard Most of it seems to be working. There's a persistent error on the Python3.8 build that makes me think that rasterio and xarray are not getting along. It seems to be an HDF5 file access error. Have you run into this in the past?

tlogan2000:
> @huard Most of it seems to be working. There's a persistent error on the Python3.8 build that makes me think that rasterio and xarray are not getting along. It seems to be an HDF5 file access error. Have you run into this in the past?

Do we still need rasterio for this PR?

RondeauG:
> @huard Most of it seems to be working. There's a persistent error on the Python3.8 build that makes me think that rasterio and xarray are not getting along. It seems to be an HDF5 file access error. Have you run into this in the past?

What's the error, exactly? I haven't tested your new version of the polygon subsetting, but last week I kept having the following error when using it alongside `engine='h5netcdf'` and dask (seemed to work fine without dask, though): `ValueError: cannot set WRITEABLE flag to True of this array`

Googling this error points towards an incompatibility with pandas (or possibly geopandas) and hdf5: https://github.com/pandas-dev/pandas/issues/24839 and https://stackoverflow.com/questions/54210073/pd-read-hdf-throws-cannot-set-writable-flag-to-true-of-this-array

Zeitsperre:
Okay, I want to say it's 98% there. We just need to figure out how to install libspatialindex-dev on the Windows build and I want to say we can push this. @huard What say you?

Zeitsperre:
@huard Will merge this as soon as Travis CI clears. Windows build is still not working, but that's not due to the library, but the trickiness of its dependencies. Will push a new version with that and then we can bump the xarray version and all the dependencies. Work for you?

Issue #286, opened 2019-11-04
subset_bbox adds dimensions and removes chunks on data_vars such as "rotated_pole"


RondeauG:
* xclim version: 0.11.0-beta
* Python version: 3.6
* Operating System: Unix

### Description
Opening multiple netCDF files (each netCDF's dimension = 8760, 130, 155). In this case, the files were NA-CORDEX hourly precipitation : pr_NAM-44_ICHEC-EC-EARTH_historical_r3i1p1_DMI-HIRHAM5_v1_1hr_(years).nc

**ds = xr.open_mfdataset(sub_files, drop_variables=["time_vectors", "ts", "time_bnds"])**
ds:
Dimensions:       (rlat: 130, rlon: 155, time: 245424)
Coordinates:
    lon           (rlat, rlon) float64 dask.array<chunksize=(130, 155), meta=np.ndarray>
    lat           (rlat, rlon) float64 dask.array<chunksize=(130, 155), meta=np.ndarray>
  * rlat          (rlat) float64 -28.4 -27.96 -27.52 ... 27.48 27.92 28.36
  * rlon          (rlon) float64 -33.88 -33.44 -33.0 -32.56 ... 33.0 33.44 33.88
  * time          (time) datetime64[ns] 1981-01-01T00:30:00 ... 2043-12-31T23:30:00
Data variables:
    rotated_pole  (time) |S1 b'R' b'R' b'R' b'R' b'R' ... b'R' b'R' b'R' b'R'
    pr            (time, rlat, rlon) float32 dask.array<chunksize=(8760, 130, 155), meta=np.ndarray>

**ds_subset = subset.subset_bbox(ds, lat_bnds=lat_bnds, lon_bnds=lon_bnds,
                                           start_date=str(np.min(years)), end_date=str(np.max(years)))**
ds_subset:
Dimensions:       (rlat: 49, rlon: 44, time: 245424)
Coordinates:
    lon           (rlat, rlon) float64 -84.98 -84.36 -83.74 ... -43.53 -42.93
    lat           (rlat, rlon) float64 45.73 45.66 45.59 ... 58.39 58.13 57.87
  * rlat          (rlat) float64 -1.12 -0.68 -0.24 0.2 ... 19.12 19.56 20.0
  * rlon          (rlon) float64 8.36 8.8 9.24 9.68 ... 25.96 26.4 26.84 27.28
[...]

PR #287, opened 2019-11-05
fix-286

tlogan2000:
fix bug that adds unwanted dims to variables using curvlinear subset_bbox when applied to a xr.DataSet

### Pull Request Checklist:

- [ ] Tags have been pushed

Bug fix - subset_bbox can add unwanted coordinates/dims to variables when applied to an entire dataset   #286 

No

* **Other information**:

This PR fixes #xyz

### DIFF
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -247,9 +247,7 @@ def subset_bbox(
             # If da is a xr.DataSet Mask only variables that have the
             # same 2d coordinates as da.lat (or da.lon)
             for var in da.data_vars:
-                if [d for d in da[var].dims if d in list(da.lat.dims)] == list(
-                    da.lat.dims
-                ):
+                if set(da.lat.dims).issubset(da[var].dims):
                     da[var] = da[var].where(lon_cond & lat_cond, drop=True)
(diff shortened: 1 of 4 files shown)


xclim/subset.py
```diff
@@ -243,7 +243,17 @@ def subset_bbox(
             lon_cond = in_bounds(lon_b, da.lon)
 
         # Mask coordinates outside the bounding box
-        da = da.where(lon_cond & lat_cond, drop=True)
+        if isinstance(da, xarray.Dataset):
+            # If da is a xr.DataSet Mask only variables that have the
+            # same 2d coordinates as da.lat (or da.lon)
+            for var in da.data_vars:
+                if [d for d in da[var].dims if d in list(da.lat.dims)] == list(
```
huard: Could you use a set here instead ? 
``` if set(lat.dims).issubset(da[var].dims)```

tlogan2000: Was unaware of issubset() method...  will have to remember it

huard:
The `set` object is very handy. Worthwhile to know its methods.

PR #289, opened 2019-11-07
Options for precipitation_accumulation to compute solid and/or liquide phases

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

Moves "precipitation_accumulation" to the `_multivariate.py` file. Adds two optional arguments to pass the mean temperature and phase selection. If `phase` is `solid`, only time steps where T < 0°C are summed. If it is `liquid`, time steps with T >= 0°C are summed. If it is not provided, goes back to the usual behaviour, summing everything.

Nope

* **Other information**:
Tests have been added.
 `xclim.atmos.precip_accumulation` has been modified so the attributes reflect the change, however, I have not yet found a way to modify the given standard name. (There is an ambiguity if `phase` is "liquid" and it should be different for "solid".

### DIFF
--- a/xclim/atmos/_precip.py
+++ b/xclim/atmos/_precip.py
@@ -107,9 +107,9 @@ daily_pr_intensity = Pr(
 precip_accumulation = Pr(
     identifier="prcptot",
     units="mm",
-    standard_name="lwe_thickness_of_precipitation_amount",
-    long_name="Total{phase:>1}precipitation",
-    description="{freq} total{phase:>1}precipitation.",
+    standard_name=lambda **kw: "lwe_thickness_of_snowfall_amount" if kw['phase'] == 'solid' else "lwe_thickness_of_precipitation_amount",
+    long_name=lambda **kw: "Total {} precipitation".format(kw['phase']) if kw['phase'] != 'both' else "Total precipitation",
+    description=lambda **kw: "{freq} " + ("total {} precipitation".format(kw['phase']) if kw['phase'] != 'both' else "total precipitation"),
     cell_methods="time: sum within days time: sum over days",
     compute=indices.precip_accumulation,
 )
--- a/xclim/atmos/_precip.py
+++ b/xclim/atmos/_precip.py
@@ -110,12 +110,14 @@ precip_accumulation = Pr(
     standard_name=lambda **kw: "lwe_thickness_of_snowfall_amount"
     if kw["phase"] == "solid"
     else "lwe_thickness_of_precipitation_amount",
-    long_name=lambda **kw: "Total {} precipitation".format(kw["phase"])
-    if kw["phase"] != "both"
-    else "Total precipitation",
+    long_name=lambda **kw: "Total {} precipitation".format(kw["phase"]).replace(
+        "both ", ""
+    ),
     description=lambda **kw: "{freq} "
     + (
-        "total {} precipitation".format(kw["phase"])
+        "total {} precipitation, estimated as precipitation when T {} 0".format(
+            kw["phase"], "<" if kw["phase"] == "solid" else ">="
+        )
         if kw["phase"] != "both"
         else "total precipitation"
     ),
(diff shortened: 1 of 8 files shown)


Zeitsperre:
_Passes Black and Flake8 compliance on first push to PR._

Yes. This is how it was all meant to be.

aulemahal:
Ok, so here is one solution to the "standard_name" problem.
Now, in `xclim.atmos`, cf attributes can be given as callables. Just before calling `str.format`, the callable is called with the arguments are keyword arguments. 
Solves it with some generality, but makes quite ugly lines in the `atmos` definitions...

aulemahal:
A test was added and the atmos-generated description was modified.

xclim/indices/_multivariate.py
```diff
@@ -538,6 +539,72 @@ def liquid_precip_ratio(
...
+    When the mean temperature is over 0 degC, precipitatio is assumed to be liquid rain and snow otherwise.
+
+    Parameters
+    ----------
+    pr : xr.DataArray
+      Mean daily precipitation flux [Kg m-2 s-1] or [mm].
+    tas : xr.DataArray, optional
+      Mean daily temperature [℃] or [K]
+    phase : str, optional,
+      Which phase to consider, "liquid" or "solid", if None (default), both are considered.
```
huard: default is set to "both", not None

xclim/indices/_multivariate.py
```diff
@@ -538,6 +539,72 @@ def liquid_precip_ratio(
...
+    --------
+    The following would compute for each grid cell of file `pr_day.nc` the total
+    precipitation at the seasonal frequency, ie DJF, MAM, JJA, SON, DJF, etc.:
+
+    >>> pr_day = xr.open_dataset('pr_day.nc').pr
+    >>> prcp_tot_seasonal = precip_accumulation(pr_day, freq="QS-DEC")
+    """
+    freq = freq or "YS"
+
+    if phase != "both":
```
huard: "both" or None ?

xclim/indices/_multivariate.py
```diff
@@ -538,6 +539,72 @@ def liquid_precip_ratio(
...
+        tu = units.parse_units(tas.attrs["units"].replace("-", "**-"))
+        fu = "degC"
+        frz = 0
+        if fu != tu:
+            frz = units.convert(frz, fu, tu)
```
huard: Can be simplified with `utils.convert_units_to`.

huard:
I've proposed changes to the PR. Feel free to dismiss them.

xclim/atmos/_precip.py
```diff
@@ -107,9 +107,20 @@ class PrTas(Indicator2D):
 precip_accumulation = Pr(
     identifier="prcptot",
     units="mm",
-    standard_name="lwe_thickness_of_precipitation_amount",
-    long_name="Total precipitation",
-    description="{freq} total precipitation.",
+    standard_name=lambda **kw: "lwe_thickness_of_snowfall_amount"
```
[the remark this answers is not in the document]
aulemahal: Makes sense!
So I could just write to new function that only call the main one.

xclim/indices/_multivariate.py
aulemahal: Woups. Scusi...

xclim/indices/_multivariate.py
aulemahal: This is a choice that we could make. What do you prefer?
For me "both" was more explicit.

xclim/atmos/_precip.py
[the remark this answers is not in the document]
huard: Yep, done in my commit.

xclim/indices/_multivariate.py
huard: Not a strong opinion on this. I think `None` is more universal.  @tlogan2000 ?

tlogan2000: No strong opiinon either but I would maybe go with ```None``` to be consistent with some other default behavoir in xclim as a whole

aulemahal: 2 vs 1, allons-y.

PR #291, opened 2019-11-07
Support Python3.8

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

Pyproj does not currently have a wheel distribution that Travis CI with Python3.8 supports. This requires that Pyproj rely on system-installed Proj binaries. This PR adds those binaries to the build with the hope that they will be enough to satisfy the pyproj requirements and pass the build.

No.

* **Other information**:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ before_install:
 install:
     - pip install -U tox-travis
 
-script: tox -v -e $TOXENV
+script: tox -vv -e $TOXENV
 
 before_deploy:
     - echo "$TRAVIS_TAG" "$TRAVIS_COMMIT"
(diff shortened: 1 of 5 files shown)


tlvu:
>  Pyproj does not currently have a wheel distribution that Travis CI with Python3.8 supports. This requires that Pyproj rely on system-installed Proj binaries. This PR adds those binaries to the build with the hope that they will be enough to satisfy the pyproj requirements and pass the build.

This is a very clear explanation.

Can you put it in the commit message next time.

If just looking at the commit message, I would not understand Why the commit was needed.  Of course I can go to the PR but it's nice to not have to jump through hoops.

Think later in time when someone annotate the code and come across the commit.

Zeitsperre:
Python3.8 build is now passing. The build requires a few binaries as well as Cython to build pyproj and scipy. This will eventually not be the case but for now we can say we are formally supporting Pythin3.8.

huard:
Ready to add 3.8 in the setup.py metadata ?

PR #293, opened 2019-11-12
Add heatwave total

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

Addition of a new indice to compute the *total* length of heat waves in a period. 

No.

* **Other information**:
Has the same _standard\_name_ as `heat_wave_max_length`, but they both aren't exactly that anyway.

### DIFF
--- a/xclim/indices/_multivariate.py
+++ b/xclim/indices/_multivariate.py
@@ -36,6 +36,7 @@ __all__ = [
     "heat_wave_max_length",
     "heat_wave_total_length",
     "liquid_precip_ratio",
+    "precip_accumulation",
     "rain_on_frozen_ground_days",
     "tg90p",
     "tg10p",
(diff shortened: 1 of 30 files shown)


xclim/indices/_multivariate.py
```diff
@@ -479,6 +481,62 @@ def heat_wave_max_length(
...
+    -----
+    See notes and references of `heat_wave_max_length`
+    """
+    freq = freq or "YS"
+    thresh_tasmax = utils.convert_units_to(thresh_tasmax, tasmax)
+    thresh_tasmin = utils.convert_units_to(thresh_tasmin, tasmin)
+
+    cond = (tasmin > thresh_tasmin) & (tasmax > thresh_tasmax)
+    group = cond.resample(time=freq)
+    total_count = group.apply(lambda d, dim=None: rl.windowed_run_count(d, window, dim=dim), dim="time")
```
huard: Why are you passing `dim=None` ?

I think `group.apply(rl.windowed_run_count, window=window, dim="time")` would work as well.

tlogan2000: This was my same question...  the apply call doesn't seem as 'clean' as heat wave max length etc.  Are there advantages to using a lambda func here?

aulemahal: Yeah... I wasn't sure how to do it, I think my new commit fixes it.

xclim/indices/_multivariate.py
```diff
@@ -533,7 +533,7 @@ def heat_wave_total_length(
 
     cond = (tasmin > thresh_tasmin) & (tasmax > thresh_tasmax)
     group = cond.resample(time=freq)
-    total_count = group.apply(lambda d, dim=None: rl.windowed_run_count(d, window, dim=dim), dim="time")
+    total_count = group.apply(lambda d: rl.windowed_run_count(d, window, dim='time'))
```
huard: You can avoid the lambda altogether. Apply passes keyword arguments to the function.

xclim/indices/_multivariate.py
tlogan2000: I thin in this case we could simply do:

```return group.apply(rl.windowed_run_count, window=window, dim="time")```

xclim/indices/_multivariate.py
tlogan2000: Whoops commented in wrong spot:
https://github.com/Ouranosinc/xclim/pull/293#discussion_r345916778

aulemahal: Ok! Did that. Sorry, I'm still not a complete xarray master.

tlogan2000: No worries, me neither for that matter... there's a lot of stuff in there,

tlogan2000:
@Zeitsperre Travis builds are failing on this PR .. any insights?

Zeitsperre:
Bottleneck pushed a new version last night (1.3.0). Seems that it demands a specific version of numpy that doesn't agree with what is installed by default. I get the impression that this is bad package and I'm thinking we wait a day to see what's happening before stepping in. If this continues, the fix can be implemented here by pinning Bottleneck at 1.2.1.

huard:
Should we bump the version to 0.12.0 with this ?

PR #301, opened 2019-11-19
Fix function signatures and standardize all to xarray

### LINKED ISSUE #300: Indices default freq set to None
The latest xclim version sets a number of `freq` to None instead of "YS". The result is that freq is not set, it fails. The tests don't seem to catch that, which implies they all set freq to something.

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

Standardizes all signatures to set default freq as either "YS" or "QS-DEC" and removes optional typing. 

Yes. Default values for freq are now explicitly set.

* **Other information**:
Setting a default value is not the same as having an optional call signature. Optional implies that keyword is not necessary to run the function. For indices, `freq` is always necessary.

This PR fixes #300

### DIFF
--- a/xclim/indices/_simple.py
+++ b/xclim/indices/_simple.py
@@ -514,7 +514,7 @@ def max_1day_precipitation_amount(pr: xarray.DataArray, freq: str = "YS"):
     at an annual frequency:
     >>> import xarray as xr
     >>> pr = xr.open_dataset('pr.day.nc').pr
-    >>> rx1day = max_1day_precipitation_amount(pr, freq: str = "YS")
+    >>> rx1day = max_1day_precipitation_amount(pr, freq="YS")
     """
 
     out = pr.resample(time=freq).max(dim="time", keep_attrs=True)
@@ -549,7 +549,7 @@ def max_n_day_precipitation_amount(pr, window: int = 1, freq: str = "YS"):
     >>> import xarray as xr
     >>> da = xr.open_dataset('pr.day.nc').pr
     >>> window = 5
-    >>> output = max_n_day_precipitation_amount(da, window, freq: str = "YS")
+    >>> output = max_n_day_precipitation_amount(da, window, freq="YS")
     """
 
     # rolling sum of the values
--- a/xclim/indices/_simple.py
+++ b/xclim/indices/_simple.py
@@ -1,16 +1,11 @@
-import logging
-
-import xarray as xr
+import xarray
 
 from xclim import run_length as rl
 from xclim import utils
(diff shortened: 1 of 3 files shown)


xclim/indices/_simple.py
```diff
@@ -517,30 +512,30 @@ def max_1day_precipitation_amount(pr, freq="YS"):
     --------
     The following would compute for each grid cell of file `pr.day.nc` the highest 1-day total
     at an annual frequency:
-
+    >>> import xarray as xr
     >>> pr = xr.open_dataset('pr.day.nc').pr
-    >>> rx1day = max_1day_precipitation_amount(pr, freq="YS")
+    >>> rx1day = max_1day_precipitation_amount(pr, freq: str = "YS")
```
huard: Find/replace gone wrong.

xclim/indices/_simple.py
```diff
@@ -551,10 +546,10 @@ def max_n_day_precipitation_amount(pr, window=1, freq="YS"):
     --------
     The following would compute for each grid cell of file `pr.day.nc` the highest 5-day total precipitation
     at an annual frequency:
-
+    >>> import xarray as xr
     >>> da = xr.open_dataset('pr.day.nc').pr
     >>> window = 5
-    >>> output = max_n_day_precipitation_amount(da, window, freq="YS")
+    >>> output = max_n_day_precipitation_amount(da, window, freq: str = "YS")
```
huard: Same.

xclim/indices/_threshold.py
```diff
@@ -1,17 +1,12 @@
-import logging
-
 import numpy as np
-import xarray as xr
```
huard: Why is the shorthand xr not ok ?

huard:
Changing default values is a breaking change. I'm still wondering how this went unnoticed for so long.

xclim/indices/_threshold.py
Zeitsperre: Both perfectly fine. Matter of preference, I suppose. I'll avoid making these kinds of changes in the future.

Zeitsperre:
Everything seems to be passing on my end. Once accepted, should I push a new version with these changes?

huard:
Yes, please, since we need a tag for Finch.

Zeitsperre:
New version is on PyPI. conda version will be available in a few hours.

huard:
Thanks for the quick response. Much appreciated.

huard:
Hold on, there were a couple of bad defaults. PR upcoming. Again, this highlights a gap in our test suite.

PR #302, opened 2019-11-19
bugfix: remove space in freq default.

huard:
### Pull Request Checklist:

- [ ] Tags have been pushed

Bug fix.

### DIFF
--- a/xclim/indices/_simple.py
+++ b/xclim/indices/_simple.py
@@ -232,7 +232,7 @@ def tn_min(tasmin: xarray.DataArray, freq: str = "YS"):
 
 
 @declare_units("[temperature]", tasmax="[temperature]")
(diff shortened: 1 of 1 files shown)


huard:
Our test suite has a problem if it does not catch that kind of thing. I'll open a ticket.

Zeitsperre:
Ouch. That's an oversight. I think most all of our test show us explicitly calling function with all signatures set, which isn't how users might use the library. Maybe the jupyter notebook approach would be a way of satisfying both showing the use cases and simultaneously testing for use cases that might be employed.

Zeitsperre:
Issue a hotfix?

huard:
Yep, again, I need a tag for Finch, and this breaks at "import finch"...

PR #310, opened 2019-11-25
Reduce UserWarnings and adapt to future deprecations

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

This PR prepares the test ensemble and some functions for future deprecations that are present in  xarray 0.14 and reduces the instances of FutureWarnings for changes to the Xclim API by catching them with `pytest.warns` and `pytest.raises`. This PR also suppresses UserWarnings and DeprecationWarnings that are raised by design of the xclim testing ensemble that are meant to inform the user of old/inappropriate function calls.

Yes. `xarray.DataArray().drop(variables)` has been changed to `xarray.DataArray().drop_vars(names=set(variables))` and tests using subset now use `str` instead of `int` for time-based slicing unless explicitly testing for warnings/exceptions.

* **Other information**:


tests/test_utils.py
```diff
@@ -688,35 +694,39 @@ def test_simple(self):
...
-            da, lon_bnds=self.lon, lat_bnds=self.lat, start_yr=yr_st, end_yr=yr_ed
+            da, lon_bnds=self.lon, lat_bnds=self.lat, start_date=yr_st, end_date=yr_ed
         )
         assert np.all(out.lon >= np.min(self.lon))
         assert np.all(out.lon <= np.max(self.lon))
         assert np.all(out.lat >= np.min(self.lat))
         assert np.all(out.lat <= np.max(self.lat))
-        np.testing.assert_array_equal(out.time.dt.year.max(), yr_ed)
-        np.testing.assert_array_equal(out.time.dt.year.min(), yr_st)
+        np.testing.assert_array_equal(out.time.dt.year.max(), np.array(int(yr_ed)))
```
huard: Do we need an `np.array` here ?

xclim/ensembles.py
```diff
@@ -103,21 +103,21 @@ def ensemble_mean_std_max_min(ens: xr.Dataset) -> xr.Dataset:
...
+    >>> from pathlib import Path
+    >>> ncfiles = Path().rglob('*tas*.nc')
     Create ensemble dataset
     >>> ens = ensembles.create_ensemble(ncfiles)
     Calculate ensemble statistics
     >>> ens_means_std = ensembles.ensemble_mean_std_max_min(ens)
     >>> print(ens_mean_std['tas_mean'])
     """
-    dsOut = ens.drop(ens.data_vars)
+    dsOut = ens._drop_vars(names=set(ens.data_vars))
```
huard: Question: Am I right in thinking that using drop in user code won't raise a warning, because it's a PendingDeprecationWarning which is disabled by default ? If so, I don't think we need to go this fast in removing `drop` statements. It forces us to use a private method (`_drop_vars`) which is not great practice. 

My suggestion would be to modify this PR to use the new methods `drop_vars`, and keep it on the back burner until the time xarray moves `drop` from PendingDeprecationWarning to DeprecationWarning. Then we make the switch and update the xarray minimum version.

tests/test_utils.py
Zeitsperre: Unfortunately, yes. The format of `out.time.dt.year.max()` is a numpy integer while the `yr_st` and `yr_ed` are strings. Even after casting them to `int`, their datatype is not the same so loading them into `np.array` objects is necessary. The alternative would be to cast the values of `out.time.dt.year.max()` to `str` for comparison.

xclim/ensembles.py
Zeitsperre: Sounds good. The changes here aren't urgent; I had about an hour yesterday and figured it would be good to get on top of some of the Deprecations coming our way. I imagine that once cftime v1.0.4.3 comes around, this will allow us to bump our versions up to xarray 0.14.3 where these deprecations will be relevant.

tests/test_utils.py
huard: But why did it work before ?

Zeitsperre: yr_st and yr_ed were sent in as integers using the `start_year` and `end_year` keywords. In newer versions of xclim, we've moved away from this by demanding `str` objects that can be parsed as "YYYY", "YYYY-MM" or "YYYY-MM-DD" and passed to `start_date` or `end_date`. For backward compatibility, using the old keywords is allowed, but a warning is raised that this will be removed in a future version.

When we implemented these changes, we added a test to assert that a warning is raised, but we didn't modify existing tests using the old/soon-to-be-deprecated calls. This PR does that, hence why the changes here are needed.

Zeitsperre: Ah, I see your question now. It's weird that simply passing them as `int` worked previously. For some reason, the tests weren't working when only casting `str` to `int`. I figured loading them into arrays made the most sense. Mysterious.

huard: Ok, no biggy, was just curious.

Zeitsperre:
@huard Warnings in this PR are now exclusively due to called functions from dependencies. There's a handful of DeprecationWarnings for xarray 0.15 that might need some attention paid to. Outside of that, things are good to merge on this PR.

tests/test_subset.py
```diff
@@ -102,11 +102,12 @@ def test_time_dates_outofbounds(self):
...
 
-        out = subset.subset_gridpoint(
-            da,
-            start_date="{yr_st}-01".format(yr_st=yr_st),
-            end_date="{yr_ed}-01".format(yr_ed=yr_ed),
-        )
+        with pytest.warns(UserWarning):
+            out = subset.subset_gridpoint(
+                da,
+                start_date="{yr_st}-01".format(yr_st=yr_st),
```
huard: Since we dropped 3.5, maybe you could use f-strings here ?

Zeitsperre: True. Will do.

Zeitsperre: Actually, changing this will lead to Black failures. This will need to happen in a new PR, but noted.

Issue #184, opened 2019-04-02
growing season length memory use


tlogan2000:
### Description
Calling growing season length calculation on large multiyear datasets (e.g. mf_datset of nrcan over canada) causes memory errors / accumulation.
Dask writing of chunks to netcdf (xr.to_netcdf) does not seem to work as expected and RAM use quickly accumulates and crashes... 

Running calculation on single years can alleviate the problem but still requires large amounts of memory

Unsure as to why this occurs ... Is it a dask/xarray problem or something in our xclim code

Balinus:
Is it the only indice that behave like that on large datasets?

tlogan2000:
so far yes.

PR #311, opened 2019-11-26
Optimize rolling dask

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

Create a new util function  `rolling` to optimize the rolling sums and means over dask arrays.
Sum and mean are implemented as they both use the same `numpy.cumsum` trick. Other rolling operations (min, max, std) could be implemented later on.

The function does the computation lazily, which xarray's `rolling()` didn't. It defaults to xarray's version is the data is not in a dask array as it uses a special dask function :  `dask.Array.map_overlap()`. 

This function is faster than xarray's even for datasets with only one chunk along the rolled dimension. And it solves MemoryErrors when using huge datasets.

I did some Jedi tricks to get the same behaviour as xarray's version when NaNs are encountered. Without the second  `np.cumsum`, we can reduce the computation time by 30-50%, but NaNs will be treated as zeros.

No

* **Other information**:
I called the function `xc.utils.rolling`, but in comparison to `xr.DataArray.rolling`, it actually does the sum or mean. So same name, different behaviour.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,7 @@ History
 * Enhancement to utils.Indicator to have more dynamic attributes using callables.
 * Added indices `heat_wave_total_length` and `tas` / `tg`.
 * Fixed a bug with typed call signatures that caused downstream failures on library import.
+* Added a `_rolling` util function to fix memory issues on large dask datasets.
 
 0.11.x-beta (2019-10-17)
 ------------------------
--- a/tests/test_precip.py
+++ b/tests/test_precip.py
@@ -22,51 +22,41 @@ class TestRainOnFrozenGround:
         TESTS_DATA, "NRCANdaily", "nrcan_canada_daily_tasmin_1990.nc"
     )
 
-    def test_3d_data_with_nans(self):
+    @pytest.mark.parametrize("prunits,prfac", [("kg m-2 s-1", 1), ("mm/day", 86400)])
+    @pytest.mark.parametrize(
+        "tasunits,tasoffset,chunks", [(None, None, {"time": 73.0}), ("C", K2C, None)]
+    )
+    def test_3d_data_with_nans(self, prunits, prfac, tasunits, tasoffset, chunks):
         pr = xr.open_dataset(self.nc_pr).pr
-        prMM = pr.copy()
-        prMM.values *= 86400.0
-        prMM.attrs["units"] = "mm/day"
+        pr2 = pr.copy()
+        pr2.values *= prfac
+        pr2.attrs["units"] = prunits
 
-        tasmax = xr.open_dataset(self.nc_tasmax).tasmax
-        tasmin = xr.open_dataset(self.nc_tasmin).tasmin
+        tasmax = xr.open_dataset(self.nc_tasmax, chunks=chunks).tasmax
+        tasmin = xr.open_dataset(self.nc_tasmin, chunks=chunks).tasmin
         tas = 0.5 * (tasmax + tasmin)
         tas.attrs = tasmax.attrs
-        tasC = tas.copy()
-        tasC.values -= K2C
-        tasC.attrs["units"] = "C"
+        tas2 = tas.copy()
+        if chunks:
+            tas2 = tas2.chunk(chunks)
+        if tasunits:
+            tas2.values -= tasoffset
+            tas2.attrs["units"] = tasunits
 
-        prMM.values[10, 1, 0] = np.nan
+        pr2.values[10, 1, 0] = np.nan
         pr.values[10, 1, 0] = np.nan
 
-        out1 = atmos.rain_on_frozen_ground_days(pr, tas, freq="MS")
-        out2 = atmos.rain_on_frozen_ground_days(prMM, tas, freq="MS")
-        out3 = atmos.rain_on_frozen_ground_days(prMM, tasC, freq="MS")
-        out4 = atmos.rain_on_frozen_ground_days(pr, tasC, freq="MS")
-        pr.attrs["units"] = "kg m-2 s-1"
-        out5 = atmos.rain_on_frozen_ground_days(pr, tas, freq="MS")
-        out6 = atmos.rain_on_frozen_ground_days(pr, tasC, freq="MS")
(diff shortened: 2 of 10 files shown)


aulemahal:
So, it finally wasn't that hard. I added a function that takes care of rolling operation other than sum and mean. It acts the same way as the conventional rolling : creates a new dimension and copies* data so that the rolling op can be applied along that new dim. But it still does this chunk-wise and is lazy, thus not going into a MemoryError on huge datasets.

* I believe it returns a view of the values, not actually copying them.

Also, this can accept any function, not only the numpy ones.

huard:
Questions: Can this be somehow ported to xarray ?

aulemahal:
Yes and no, as far as I understang xarray's code. The whole "rolling" module is completely exempt to any explicit call to dask.  I don't really see any elegant way to add this code. Also, there is some time in the overhead, that I just skip. What exactly does this overhead do, and is it useful? I do not know.
During this exploration, I did found that you can to `ds.rolling(time=5).mean(allow_lazy=True)`. Which saves up some memory on compute time if you select a specific slice futher along. However, there is still something I don't understand, because it uses 20-40% more memory than my version and is slower.
So, it would be possible, but I don't see a way that would make an acceptable PR.

tlogan2000:
This might indirectly be a solution to #184 where i could never completely understand the snowballing memory problems.  I will try to see if this helps once merged to master

tlogan2000:
> Questions: Can this be somehow ported to xarray ?
I think we could minimally create an issue on the xarray github describing the problems we are experiencing for long time-series and present @aulemahal 's solution / work around as being more efficient.   They might decide to find a way to integrate it themselves?

huard:
I've read similar memory issues happening when bottleneck is not installed. Could you confirm this is not the issue here.

huard:
See https://github.com/pydata/xarray/issues/3332 Might be a place where you could describe what you've done.

tlogan2000:
Messing around with various reduction options I see that using 'std'  results in very small diffs between xr and utils.rolling() not sure why..   

        std_nd_xr = ds_nd.pr.rolling(time=5).std()
        std_nd_xc = utils.rolling(
            ds_nd.pr, window=5, dim="time", mode="std", keep_attrs=False
        )
        std_dask = utils.rolling(ds_dask.pr, window=5, dim="time", mode="std")

        xr.testing.assert_identical(std_nd_xr, std_nd_xc)  # fails : 'allclose' works
        xr.testing.assert_allclose(std_dask, std_nd_xr)

tlogan2000:
It might be prudent to include a chunked version of tests for functions that rely on rolling (our indirectly runlength) in the indicator test suite?  

E.g. tests on .nc files in test_precip.py, test_temperature.py etc currently do not chunk the test datasets before calculating indices

Example from test_precip
```
class TestMaxNDay:
    # testing of wet_day and daily_pr_intensity, both are related

    nc_file = os.path.join(TESTS_DATA, "NRCANdaily", "nrcan_canada_daily_pr_1990.nc")

    def test_3d_data_with_nans(self):
        # test with 3d data
        pr = xr.open_dataset(self.nc_file).pr
        
        ##### potential added line
        pr_dask = xr.opend_dataset(self.nc_file, chunks=dict(time=10)).pr

        prMM = xr.open_dataset(self.nc_file).pr
        prMM.values *= 86400.0
        prMM.attrs["units"] = "mm/day"
        # put a nan somewhere
        prMM.values[10, 1, 0] = np.nan
        pr.values[10, 1, 0] = np.nan
        wind = 3
        out1 = atmos.max_n_day_precipitation_amount(pr, window=wind, freq="MS")
        out2 = atmos.max_n_day_precipitation_amount(prMM, window=wind, freq="MS")
       
        ##### potential added line
        out3 = atmos.max_n_day_precipitation_amount(pr_dask, window=wind, freq="MS")
       
        cont'd assertion tests ....

```
Maybe this overkill as we already test rolling generally in test_utils.py.. opinions?

huard:
I think you're right, but I would like this to be parameterized. We don't want to duplicate all tests, we need to use pytest fictures to parameterize test input so that each test runs with both chunked and un-chunked datasets. This goes back to #52 ... I suggest we raise the priority on this.

aulemahal:
> Messing around with various reduction options I see that using 'std' results in very small diffs between xr and utils.rolling() not sure why..

@tlogan2000 I wasn't able to replicate your problem... Are you sure  `ds_nd` in your example is __not__ a dask array? If it is, then `rolling()` defaults to my method, which gives slightly different results.

@huard I confirm bottleneck is installed in my environment. Also, AFAIK bottleneck is not used by xarray when rolling on a dask array.

I'll modifiy the tests with parametrization in mind.

aulemahal:
So, I realize (tnks @tlogan2000 ) that my solution was float-dependent, which isn't good. Fixed it, but this makes me wonder if other edge cases are not covered in my code, so I decided to put the function private. Like : if ever xarray fixes the problem, or we find a more convenient solution, I don't want this to be a breaking change. 
For now, I believe the current solution is better or similar to doing: 
```python
    r = da.rolling(dim=win)
    a = r.construct('roll').reduce(np.func, dim='roll', allow_lazy=True)
    a = a.where(r._counts() >= win)
```
where the last line is to take care of NaN values and edges. Going through the `construct -> reduce`, saves some overhead and uses less memory than going directly to  rolling().reduce()`.

These changes were really useful for my own script generating indices for the ouranos website, but I would understand if it isn't appropriate for xclim, as it's more a xarray problem.

aulemahal:
Oh! And I added tests following recommendations from #52.

tlogan2000:
I see the py3.5 build is failing but we will be dropping 3.5 shortly anyhow.   Black and flake 8 failed as well but should be an easy fix..

aulemahal:
So, I fixed that black thing.
The failing build was due to a implementation bug in python 3.5 (https://bugs.python.org/issue9232), I fixed that too.

aulemahal:
I suggest we merge this for now.
We should definitely complete #316 before v1.0, but this is our best shot for the moment.

PR #312, opened 2019-11-26
Add tas (tg) indice and tests

### LINKED ISSUE #308: Indicator for average temperature based on tasmin and tasmax
### Description

Create a new indicator computing the daily average temperature from the daily min and max temperature.

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

Adds an indice that simply computes the daily mean temperature from maximum and minimum temperatures.

No

* **Other information**:
Should I also add an  `xclim.atmos` Indicator? Seems to me that it isn't appropriate since it's not *really* an indicator!

This PR fixes #308

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,7 @@ History
 * Added a `tolerance` argument to `subset_gridpoint` raising an error if distance to closest point is larger than tolerance.
 * Created land module for standardized access to streamflow indices.
 * Enhancement to utils.Indicator to have more dynamic attributes using callables.
-* Added indice `heat_wave_total_length`.
+* Added indices `heat_wave_total_length` and `tas` / `tg`.
 * Fixed a bug with typed call signatures that caused downstream failures on library import.
 
 0.11.x-beta (2019-10-17)
--- a/tests/test_indices.py
+++ b/tests/test_indices.py
@@ -748,6 +748,16 @@ class TestTGXN90p:
         assert out[5] == 25
 
 
+class TestTas:
+    def test_tas(self, tasmin_series, tasmax_series, tas_series):
+        tas = tas_series(np.ones(10))
+        tasmin = tasmin_series(np.zeros(10))
+        tasmax = tasmax_series(np.ones(10) * 2)
+
+        tas_xc = xci.tas(tasmin, tasmax)
+        xr.testing.assert_equal(tas, tas_xc)
+
+
 class TestTxMin:
     @staticmethod
     def time_series(values):
--- a/tests/test_temperature.py
+++ b/tests/test_temperature.py
@@ -175,13 +175,16 @@ class TestETR:
 
 
 class TestTmean:
-    nc_file = os.path.join(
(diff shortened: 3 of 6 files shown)


aulemahal:
Added the atmos indicator. However, I add to hack the `Indicator.missing()` method. Right now it only checks for missing days in the data. But that doesn't make sense for not-resampling indicators, like this one. So I return "no missing values" if "freq" is not an argument. This could be modified if we add non-resampling missing value checks later on.

huard:
Another option would be to subclass the missing method with a function that returns False. 
That being said, I think we should add a test to check that when a tasmin is missing but not a tasmax, the result is a NaN.

aulemahal:
1) You mean like:
```python
class TasminTasmaxNoResample(TasminTasmax):
    @staticmethod
    def missing *args, **kwds):
        return False
```
? Seems a very specific solver for a more general problem, no?

Also, xarray already does the check for an addition:
```python
a = xr.DataArray([1, np.nan, 1], dims=('x',), coords={'x': [0, 1, 2]})
b = xr.DataArray([1, 2, 1, 1], dims=('x',), coords={'x': [0, 1, 2, 3]})
a + b
```
gives: `xarray.DataArray (x: 3)> array([ 2., nan,  2.])  Coordinates:   * x        (x) int64 0 1 2`
So, missing values are omitted and NaNs are propagated.

huard:
Yes exactly. To me this communicates better what is going on to readers. Happy to be convinced otherwise. @tlvu Any opinion on this ? 

Perfect ! Just want to make sure that as these inputs go through all the Indicator methods, the same is true. They should in theory, but in practice theory is different from practice.

tlvu:
> Another option would be to subclass the missing method with a function that returns False.
> That being said, I think we should add a test to check that when a tasmin is missing but not a tasmax, the result is a NaN.
> Yes exactly. To me this communicates better what is going on to readers. Happy to be convinced otherwise. @tlvu Any opinion on this ?

Having a static method that always return False is weird to me but if that's how the API/library behave/expect then fine by me.

tlogan2000:
> To be used as a WPS service, we have to expose an indicator in atmos. I think that was what Travis was looking for. Can be another PR.

Yes. The problem for many of our datasets is that we often estimate 'tas' from 'tasmin' and 'tasmax'.  Using ```xclim``` / ```xarray``` we can do this on the fly and then calculate any index requiring tas as input .  However calling the same calculation via the finch wps server requires the resource netcdf to have the input variable already.

We might want to think about creating/writing to disk the a set of 'tas' .nc files for our own datasets on the thredds server.  However a vision of remotely connecting to other nodes will likely result in a similar issue so this will still be of use I think

aulemahal:
I see!
Here I come with another proposition. @huard , it goes to your first comment and as the __add__ of xarray already does this, it's kinda useless for this specific case. But, as long as atmos requires input to be daily, I think giving a default `freq='D'` to the missing check makes sense. And as it flags values missing from the input dataset, still does what we expect it to.
So I am pushing away the issue to the future, when we want to make atmos more general :).

PR #315, opened 2019-11-30
Bump bottleneck version

qwhelan:
While investigating https://github.com/pydata/bottleneck/issues/280, I noticed that `tox` was using `pip==10.0` which is relatively ancient and does not support PEP 517-style builds.

By specifying `requires = pip >= 19.0`, we can force a modern version of `pip` and also resolve the build issues preventing `bottleneck=1.3.1` from being used.


Zeitsperre:
Thanks for looking into this!

PR #318, opened 2019-12-05
Fix calc_perc and add xr kwargs to ensembles

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

(The checklist is empty, but I don't think this PR needs anything? It's so small...)

Fixes a bug in `ensembles._calc_perc` where it would assume that the first dimension of the passed dataset was "realization". This fix removes the tests that were made to assure time in the second dimensions (which is useless) and adds a transpose when needed, to bring "realization" as axis 0.

Also, this adds `**xr_kwargs` to  `create_ensemble` so that when datasets are passed as strings or Paths, some options can be given to the opening call. This can be useful to change the default chunking of `{'time': 10}` for mf datasets or pass a `preprocess` callable (which is why I made this change, for my scripts).

No, only adds options.

### DIFF
--- a/tests/test_ensembles.py
+++ b/tests/test_ensembles.py
@@ -183,37 +183,42 @@ class TestEnsembleReduction:
     nc_file = os.path.join(TESTS_DATA, "EnsembleReduce", "TestEnsReduceCriteria.nc")
 
     def test_kmeans_rsqcutoff(self):
-        ds = xr.open_dataset(self.nc_file)
+        if pytest.importorskip("sklearn", minversion="0.22"):
+            ds = xr.open_dataset(self.nc_file)
 
-        # use random state variable to ensure consistent clustering in tests:
-        [ids, cluster, fig_data] = ensembles.kmeans_reduce_ensemble(
-            data=ds.data, method={"rsq_cutoff": 0.9}, random_state=42, make_graph=False
-        )
+            # use random state variable to ensure consistent clustering in tests:
+            [ids, cluster, fig_data] = ensembles.kmeans_reduce_ensemble(
+                data=ds.data,
+                method={"rsq_cutoff": 0.9},
+                random_state=42,
+                make_graph=False,
+            )
 
-        assert ids == [0, 1, 3, 4, 6, 7, 8, 10, 11, 15, 18, 20, 22]
-        assert len(ids) == 13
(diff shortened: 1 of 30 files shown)


tests/test_ensembles.py
```diff
@@ -112,17 +112,23 @@ def test_create_unequal_times(self):
...
         )
         np.testing.assert_array_equal(
-            np.percentile(ens["tg_mean"][:, 0, 5, 5], 50), out1["tg_mean_p50"][0, 5, 5]
+            np.percentile(ens["tg_mean"].isel(time=0, lon=5, lat=5), 50),
+            out1["tg_mean_p50"].isel(time=0, lon=5, lat=5),
         )
         np.testing.assert_array_equal(
-            np.percentile(ens["tg_mean"][:, 0, 5, 5], 90), out1["tg_mean_p90"][0, 5, 5]
+            np.percentile(ens["tg_mean"].isel(time=0, lon=5, lat=5), 90),
+            out1["tg_mean_p90"].isel(time=0, lon=5, lat=5),
```
Zeitsperre: Commas at the end of signatures will fail in Python3.5. Running black with `--target-version py35` will deal with this problem.

tests/test_ensembles.py
```diff
@@ -357,23 +369,27 @@ def test_kmeans_modelweights(self):
         "matplotlib.pyplot" not in sys.modules, reason="matplotlib.pyplot is required"
     )
     def test_kmeans_rsqcutoff_with_graphs(self):
-        ds = xr.open_dataset(self.nc_file)
-
-        # use random state variable to ensure consistent clustering in tests:
-        [ids, cluster, fig_data] = ensembles.kmeans_reduce_ensemble(
-            data=ds.data, method={"rsq_cutoff": 0.9}, random_state=42, make_graph=True
-        )
+        if pytest.importorskip("sklearn", minversion="0.22"):
```
Zeitsperre: Not familiar with this feature. Could this be written as a decorator to the test function?

tlogan2000: Not sure but we don't need the 'if' apparently @aulemahal tells me

aulemahal: I haven't been able to find examples or ways to do this as decorators, but the line could be simply put at the top of the function (no need for `if`). It raises `pytest.Skip` if the package cannot be found with this minimum version.

Issue #303, opened 2019-11-19
Test suite should exercise default values


huard:
Many tests specify default values, so when defaults are changed or wrong, no test fails. 

TODO: Run all indicators using their default values using a pytest fixture and a small synthetic input array.

PR #319, opened 2019-12-06
Add heatwave options

aulemahal:
### Pull Request Checklist:

- [ ] Tags have been pushed

This adds options to the 3 main heat waves indices  so that the definition used by the INSPQ can be computed. Instead of searching for N *consecutive* days where temperatures are above thresholds, they look for days where the length-N running mean is above thresholds. 

Furthermore, they use a weighted mean, so now, `window` can be a sequence of floats. Checks are made to be sure this option is only used with the "mean" mode and that the weights sum to 1.

Atmos indices were modified with dynamic descriptions and long names to fit what mode was used. and tests were added by transforming the previous tests to parametrized tests (as suggested by #52) that also try to follow #303.  Also, this fixes an overlooked bug from #293: `heat_wave_total_length` wasn't accessible through `atmos` and had no tests.

Nope.

* **Other information**:

I chose the names "consecutive" for the default behaviour and "mean" for the added one. Y'all agree?

### DIFF
--- a/xclim/indices/_multivariate.py
+++ b/xclim/indices/_multivariate.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Union, Sequence
 
 import numpy as np
 import xarray
@@ -338,7 +338,7 @@ def heat_wave_frequency(
     tasmax: xarray.DataArray,
     thresh_tasmin: str = "22.0 degC",
     thresh_tasmax: str = "30 degC",
-    mode: str = 'consecutive',
+    mode: str = "consecutive",
     window: Union[int, Sequence[float]] = 3,
     freq: str = "YS",
 ) -> xarray.DataArray:
@@ -401,7 +401,9 @@ def heat_wave_frequency(
     https://doi.org/10.1175/1520-0450(2001)040<0762:OTDOAH>2.0.CO;2
     """
     if mode == "consecutive" and not np.isscalar(window):
-        raise ValueError("Window as sequence of weights is only valid for the 'mean' mode.")
+        raise ValueError(
+            "Window as sequence of weights is only valid for the 'mean' mode."
+        )
 
     thresh_tasmax = utils.convert_units_to(thresh_tasmax, tasmax)
     thresh_tasmin = utils.convert_units_to(thresh_tasmin, tasmin)
@@ -412,10 +414,27 @@ def heat_wave_frequency(
         if not np.isscalar(window):
             weights = np.array(window)
             window = len(window)
+            if weights.sum() != 1:
+                raise ValueError("Weights used for the window must sum to 1.")
         else:
-            weigths = np.array([1/window] * window)
-        cond = ((_rolling(tasmin, window=window, dim="time", mode=lambda x, axis: np.sum(x * weights, axis=axis)) > thresh_tasmin) &
-                (_rolling(tasmax, window=window, dim="time", mode=lambda x, axis: np.sum(x * weights, axis=axis)) > thresh_tasmax))
+            weights = np.array([1 / window] * window)
+        cond = (
+            _rolling(
+                tasmin,
+                window=window,
+                dim="time",
+                mode=lambda x, axis: np.sum(x * weights, axis=axis),
+            )
+            > thresh_tasmin
+        ) & (
+            _rolling(
+                tasmax,
+                window=window,
+                dim="time",
+                mode=lambda x, axis: np.sum(x * weights, axis=axis),
+            )
+            > thresh_tasmax
+        )
 
     group = cond.resample(time=freq)
     return group.apply(rl.windowed_run_events, window=window, dim="time")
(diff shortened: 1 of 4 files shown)


xclim/indices/_multivariate.py
```diff
@@ -376,21 +383,59 @@ def heat_wave_frequency(
...
+                tasmax,
+                window=window,
+                dim="time",
+                mode=lambda x, axis: np.sum(x * weights, axis=axis),
+            )
+            > thresh_tasmax
+        )
+
     group = cond.resample(time=freq)
     return group.apply(rl.windowed_run_events, window=window, dim="time")
```
tlogan2000: Line 440 - Would using ```mode='mean'``` not result in wanting to set window = 1 in the final ```rl.windowed_run_events``` call?  

I.e. if mode=='mean' and window == 3 we will already have done the rolling weighted average of n=3 so we do not need to find run_events of >= 3 as well in the final return right?

xclim/indices/_multivariate.py
```diff
@@ -447,21 +499,59 @@ def heat_wave_max_length(
...
+                window=window,
+                dim="time",
+                mode=lambda x, axis: np.sum(x * weights, axis=axis),
+            )
+            > thresh_tasmax
+        )
+
     group = cond.resample(time=freq)
     max_l = group.apply(rl.longest_run, dim="time")
     return max_l.where(max_l >= window, 0)
```
tlogan2000: Line 557. Same comment as before ... I think the run_length window condition would already met when mode=='mean'

aulemahal: You are right! That adds the strange case where we have `mode='mean'` and `window=N`, but a max length under N. But hey, not our fault, @inspq.

Zeitsperre:
@aulemahal You gotta run black on the tests as well. Do you use pre-commit?

xclim/indices/_multivariate.py
tlogan2000: > You are right! That adds the strange case where we have `mode='mean'` and `window=N`, but a max length under N. But hey, not our fault, @Inspq.

We could potentially also add N-1 to max_l when mode='mean'?   

Does the INSPQ report talk about  heat wave length definitions in their report?

tlogan2000: Thinking out loud here not sure which option makes more sense... 

On the first day when the the N-day moving window meets heatwave threshold conditions is heat wave length equal to N or equal to 1?

@aulemahal @huard opinions?    I'm thinking more and more that it should be N....

aulemahal: I thought I understood that if the N-day moving window qualified as a heatwave, it still was a 1-day heatwave. But, rereading the INSPQ report, the shortest reported heatwaves all have 3 days... Does that mean 1 = 3? The, adding N-1 to all found lengths would make sense.
However, neither the INSPQ or both its sources for this indicator discuss the "length" of the heat waves.

huard: I think the output from both modes should be comparable. So if for consecutive if we get 3, we should get 3 for the mean mode. 

Still unsure whether this should be an independent indicator. There are probably other heatwave definitions out there. I think it will be easier to maintain and test a set of simple independent indicators than one big indicator with lots of options.

aulemahal: The more I look into the reports, the less sure I am about what they mean. Here is an excerpt from the 2015 report:

> Ainsi, une vague régionale de chaleur extrême est définie ainsi : au minimum 3 jours consécutifs pendant lesquels les moyennes mobiles des températures maximales et minimales,  observées aux stations météorologiques de référence des RSS, atteignent les seuils de chaleur extrême préalablement définis.

Should we understand here that the threshold is taken on the N-window running means, but that we **still need** N consecutive days where the condition is respected? That would mean, that we only need to add a rolling call, but all the rest stays the same.

tlogan2000: > I think the output from both modes should be comparable. So if for consecutive if we get 3, we should get 3 for the mean mode.
> 
> Still unsure whether this should be an independent indicator. There are probably other heatwave definitions out there. I think it will be easier to maintain and test a set of simple independent indicators than one big indicator with lots of options.

Our original thinking was that there are already 3 'types' of heatwave calculations (frequency, max_length, and now total days) repeating each type with a new definition we end up with 6 indicators which starts to be confusing

aulemahal:
So, I discussed with the INSPQ Experts and they do the following:
They take the 3-day running mean and define heatwave as when this exceed the thresholds. But the whole 3-day window is part of the heatwave. 
Also, the weighted mean is only for forecasts.

As this is quite different from what we already have and is quite specific to the INSPQ, @tlogan2000  and I suggest that we drop this PR.

Zeitsperre:
@aulemahal This PR will be the last to go in before we bump the xarray version. I still have a few things to do to fix the doc build on ReadTheDocs and seeing that this PR doesn't rely on newest xarray, I think we can add this in before necessary changes are finalized.

tlogan2000:
I agree that this is getting very specific and we can close for now... We will keep the code and use it for specific project in quesiton but won't add it to the overall xclim library for now

PR #320, opened 2019-12-16
Implement sea ice extent and area indicators [WIP]

### LINKED ISSUE #314: Sea ice extent indicator
### Description

In DACCS we planned a couple of sea-ice indicators. We could have: 
- Arctic/Antarctic sea ice area
- Arctic/Antarctic sea ice extent  
- Sea ice fluxes

huard:
### Pull Request Checklist:

- [ ] Tags have been pushed

New indicators.

* **Other information**:
We'll need a simple way to get the areacello variable from CMIP models. 

This PR fixes #314


xclim/seaIce/_seaice.py
```diff
@@ -0,0 +1,47 @@
...
+    compute=indices.sea_ice_extent,
+)
+
+
+sea_ice_area = SicArea(
+    identifier="sea_ice_area",
+    units="m^2",
+    standard_name="sea_ice_area",
+    long_name="Sea ice area",
+    description="The sum of ice-covered areas where sea ice concentration is at least {thresh}.",
```
tlogan2000: The descriptions between ice_area and ice_extent are really similar.  To the point that I am pretty confused as to the difference between the two.. What is the nuance exactly?

xclim/utils.py
```diff
@@ -54,10 +54,15 @@
...
 units = pint.UnitRegistry(autoconvert_offset_to_baseunit=True)
+
 units.define(
-    pint.unit.UnitDefinition("percent", "%", (), pint.converters.ScaleConverter(0.01))
+    pint.unit.UnitDefinition(
+        "percent", "%", ("pct",), pint.converters.ScaleConverter(0.01)
+    )
 )
 
+units.define("1 = fraction = [] = frac")
```
tlogan2000: What is this doing?  Related to dimensionless case I think?

xclim/indices/_threshold.py
```diff
@@ -717,6 +719,68 @@ def maximum_consecutive_tx_days(
...
+    thresh : str
+      Minimum sea ice concentration for a grid cell to contribute to the sea ice extent.
+
+    Returns
+    -------
+    Sea ice area [m²].
+
+    Notes
+    -----
+    To compute sea ice area over a given region, first mask the sea ice concentration.
```
tlogan2000: line 744 - aren't we doing this for them in line 748?

xclim/indices/_threshold.py
```diff
@@ -717,6 +719,68 @@ def maximum_consecutive_tx_days(
     return group.apply(rl.longest_run, dim="time")
 
 
+@declare_units("[area]", sic="[]", area="[area]", thresh="[]")
+def sea_ice_area(sic, area, thresh="15 pct"):
+    """Return the total sea ice area.
+
+    Sea ice area measures the total sea ice covered area where sea ice concentration is above a threshold,
```
tlogan2000: I think I need a much clearer description between the sea_ice_area vs sea_ice_extent.   Probably super clear for people who know the subject better than myself but still could be helpful

xclim/indices/_threshold.py
```diff
@@ -717,6 +719,68 @@ def maximum_consecutive_tx_days(
...
+    -------
+    Sea ice area [m²].
+
+    Notes
+    -----
+    To compute sea ice area over a given region, first mask the sea ice concentration.
+    """
+    thresh = units.Quantity(thresh)
+    t = utils.convert_units_to(thresh, sic)
+    out = xarray.dot(sic.where(sic >= t, 0), area) * thresh.to("dimensionless") / t
```
tlogan2000: line 748 - Am getting even more confused between extent and area I think.   Why do we need to scale by t?

xclim/atmos/_temperature.py
```diff
@@ -90,6 +90,7 @@ def cfprobe(self, dan, dax):
         for da in (dan, dax):
             checks.check_valid(da, "cell_methods", "time: maximum within days")
```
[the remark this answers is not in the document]
huard: Yes. Fixed.

xclim/indices/_multivariate.py
```diff
@@ -281,9 +282,10 @@ def daily_temperature_range_variability(
 
        vDTR_j = \frac{ \sum_{i=2}^{I} |(TX_{ij}-TN_{ij})-(TX_{i-1,j}-TN_{i-1,j})| }{I}
     """
+    q = 1 * utils.units2pint(tasmax) - 0 * utils.units2pint(tasmin)
```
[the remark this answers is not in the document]
huard: Previously, the declare_units decorator would impose the output units on the output. I thought it would be less error prone if the decorator rather tried to convert the actual output units to the desired output units instead (see line 379 in utils.py). 

In this case, we have a difference in temperature. If the input is in celcius, the difference will be in celsius. If we convert to Kelvin, we'll subtract 273.15, which does not make sense since it's a difference of temperature, and not an absolute temperature.

This line is used to get the actual units (delta_deg or delta_C) from pint. I've added a check to make sure that the output units are "K" and it works on my end. What version of pint have you checked this with ?

tests/conftest.py
```diff
@@ -155,3 +155,28 @@ def _per_doy(values, calendar="standard", units="kg m-2 s-1"):
...
+def areacella():
+    """Return a rectangular grid of grid cell area. """
+    r = 6100000
+    lon_bnds = np.arange(-180, 181, 1)
+    lat_bnds = np.arange(-90, 91, 1)
+    dlon = np.diff(lon_bnds)
+    dlat = np.diff(lat_bnds)
+    lon = np.mean(lon_bnds[:2]) + np.cumsum(dlon)
+    lat = np.mean(lat_bnds[:2]) + np.cumsum(dlat)
+    area = (
```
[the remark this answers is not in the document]
huard: Oups. Fixed.

xclim/indices/_threshold.py
```diff
@@ -690,6 +692,68 @@ def maximum_consecutive_tx_days(tasmax: xarray.DataArray, thresh="25 degC", freq
...
+    Returns
+    -------
+    Sea ice extent [m²].
+
+    Notes
+    -----
+    To compute sea ice extent over a given region, first mask the sea ice concentration.
+    """
+    t = utils.convert_units_to(thresh, sic)
+    extent = xarray.where(sic >= t, 1, 0)
```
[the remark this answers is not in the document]
huard: Hum, I think I wanted the output to be integers, but your suggestion works as well. Changed.

xclim/seaIce/_seaice.py
huard: sea_ice_area sums the area covered by sea ice. 
sea_ice_extent sums the grid cell area of all grid cells with ice.

xclim/utils.py
huard: Correct. Some netcdf files use units=1 for dimensionless quantities. This probably needs to be better tested however.

xclim/seaIce/_seaice.py
tlogan2000: > sea_ice_area sums the area covered by sea ice.
> sea_ice_extent sums the grid cell area of all grid cells with ice.

Ok thanks.  So using the same threshold (15%) sea_ice_extent should be greater than sea_ice_area?

xclim/indices/_threshold.py
huard: Line 748 masks grid cells where sic is below a threshold.

xclim/indices/_threshold.py
huard: http://nsidc.org/arcticseaicenews/faq/#area_extent

xclim/indices/_threshold.py
tlogan2000: > Line 748 masks grid cells where sic is below a threshold.

Ok. We are referring to spatial subsets (more or less) in this note right.  Could maybe change to something like : To compute sea ice area over a subregion, first mask or subset the input sea ice concentration data.

xclim/indices/_threshold.py
tlogan2000: I think I get it now...

tlogan2000: Calculation makes sense to me with the faq link ...

xclim/indices/_threshold.py
tlogan2000: Is this link official / stable enough to add in the notes for these two indicators?

xclim/utils.py
huard: In fact, this doesn't seem to work. Will remove for now.

xclim/indices/_threshold.py
huard: Sea ice concentration is either given as % or a fraction [0,1]. This does the conversion on the output rather than the input. Clarified.

xclim/indices/_multivariate.py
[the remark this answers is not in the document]
tlogan2000: > Previously, the declare_units decorator would impose the output units on the output. I thought it would be less error prone if the decorator rather tried to convert the actual output units to the desired output units instead (see line 379 in utils.py).
> 
> In this case, we have a difference in temperature. If the input is in celcius, the difference will be in celsius. If we convert to Kelvin, we'll subtract 273.15, which does not make sense since it's a difference of temperature, and not an absolute temperature.
> 
> This line is used to get the actual units (delta_deg or delta_C) from pint. I've added a check to make sure that the output units are "K" and it works on my end. What version of pint have you checked this with ?

pint == 0.9 when running ```test_static_daily_temperature_range```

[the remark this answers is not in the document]
tlogan2000: > pint == 0.9 when running `test_static_daily_temperature_range`

I still get 'kelvin' which version do you have?

xclim/indices/_threshold.py
Zeitsperre: If you want to play it safe, you could use the site image from archive.org (the wayback machine):

https://web.archive.org/web/20200103090158/https://nsidc.org/arcticseaicenews/faq/#area_extent

This is an archived image of the page according to the date and time specified (2020-01-03 09:01:58). That's about as stable as it gets.

xclim/seaIce/_seaice.py
huard: Correct, unless sic is 100% everywhere.

xclim/indices/_multivariate.py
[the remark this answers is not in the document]
tlogan2000: Sorry this was happening in the xclim.indices call only... when using xclim.atmos module the output units are corrected to 'K'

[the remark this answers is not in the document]
huard: Yes, I see now, fixed anyway.

xclim/indices/_threshold.py
huard: Done.

tlogan2000:
@huard max1day precip test failing when units are 'mm day-1'
the new  string formatting (utils.py line 276) is unhappy

PR #332, opened 2020-01-06
Update code base to Py36 conventions

### LINKED ISSUE #328: Begin employing Python3.6 features
@Ouranosinc/xclim-core 

Welcome back and a happy new year to everyone!

I quickly wanted to write that with the newest version of xclim that should be pushed to PyPI and Conda-Forge later this week, we will have formally dropped Python3.5. What this means:

New goodies (see https://docs.python.org/3/whatsnew/3.6.html for the full list):
- We can now use f-strings, ie: `f"{foo} has a value of {bar}."` in lieu of `"{0} and {1}".format(foo, bar)`
- In addition to function signature typing we can now use variable typing, ie: `this_dict: Dict[str, int] = {}`
- We can now use underlines when typing out very large numbers for readability, ie: `1_000_000` in lieu of `1000000`.
- Pathlib now supports os.PathLike objects (this is not as exciting but might change some file access static type checks).

This also means that I'll be updating the Black code formatting to be consistent with Python3.6. The changes will be in a PR and will slightly change all of the code base (e.g. removing commas at the end of call signatures).

This has been a small announcement of the coming changes). This issue will be closed with a PR in the coming weeks that integrates these changes.

Zeitsperre:
### Pull Request Checklist:

- [ ] Tags have been pushed

This PR updates the code base to use Python3.6 conventions for Black. Almost all changes are aesthetic except for a few new integrations such as f-strings.

Code will not be compatible with Python3.5 (Compatibility was already broken with xarray 0.14) so, in summary, no.

* **Other information**:

This PR fixes #328


xclim/indices/_multivariate.py
```diff
@@ -5,6 +5,7 @@
 
 from xclim import run_length as rl
 from xclim import utils
+from xclim.utils import _rolling
```
aulemahal: This contradicts #334! I guess a merge of master will erase it.

Zeitsperre: Yep. I need to re-merge master and do some adjustments before this goes live. You'll need to change your Back preferences in all future PRs once this is merged.

Issue #298, opened 2019-11-18
TestMissingAnyFills::TestMonth random failures


Zeitsperre:
### Description

TestMissingAnyFills::test_month tends to fail intermittently, particularly on the macOS Python3.7 build but also on Linux and Windows at a lesser frequency. Caution suggests that if this specific case is failing sometimes, there must either be a bug within xclim or xarray or there is some RNG-dependent call or test setup that we've coded that has some failing edge cases.

I'm not sure what the urgency of this bug is but it's been plaguing TravisCI builds for several months so I'm opening an issue to at least have this on the board. Here's the stack:
```
________________________ TestMissingAnyFills.test_month ________________________
self = <test_checks.TestMissingAnyFills object at 0x11b253150>
tasmin_series = <function tasmin_series.<locals>._tasmin_series at 0x11a409f80>

    def test_month(self, tasmin_series):
        ts = tasmin_series(np.zeros(36))
        miss = checks.missing_any(ts, freq="YS", month=7)
        np.testing.assert_equal(miss, [False])
    
        miss = checks.missing_any(ts, freq="YS", month=8)
        np.testing.assert_equal(miss, [True])
    
        with pytest.raises(ValueError, match=r"No data for selected period."):
            miss = checks.missing_any(ts, freq="YS", month=1)
    
        miss = checks.missing_any(ts, freq="YS", month=[7, 8])
        np.testing.assert_equal(miss, [True])

        ts = tasmin_series(np.zeros(76))
        miss = checks.missing_any(ts, freq="YS", month=[7, 8])
>       np.testing.assert_equal(miss, [False])

E       AssertionError: 
E       Items are not equal:
E        ACTUAL: <xarray.DataArray (time: 1)>
E       array([ True])
E       Coordinates:
E         * time     (time) datetime64[ns] 2000-01-01
E        DESIRED: [False]
```

aulemahal:
Hypothesis
On line 206-8 of `checks.py::missing_any()`, we do:
```
sda = xr.DataArray(data=np.empty(len(t)), coords={"time": t}, dims=("time",))
st = generic.select_time(sda, **indexer)
sn = st.notnull().resample(time=freq).sum(dim="time")
```
AFAIK, `np.empty` creates an empty array, without initializing its values, so you get what was in the memory before the initialization. There is a non-null chance that a NaN was there. Thus, both in `generic.select_time` and `st.notnull()`, this value is dropped and the test fails.

Whatcha think? 

[...]

PR #339, opened 2020-01-08
Improvement of ensemble_percentile | Fix of #298

### LINKED ISSUE #169: Optimize/improve  xclim ensemble stats utilities
There are currently functional versions of the following utilities but they could use improvement / polishing:
```python
xclim.utils.utils.create_ensemble()
xclim.utils.ensemble_percentiles()
xclim.utils.calc_percentiles_simple()
xclim.utils.calc_percentiles_blocks()
xclim.utils.calc_perc()
```
**Ideas to explore:**

For ensemble creation :
-  create dimension using IndexVariable  (instead of just a string) and make sure that it has attribute axis='E'. 

**Try to achieve dask enabled percentile calculation (remove 'time_block' method):**

- try DataArray.reduce(np.nanpercentile, dim='realization') - Check performance (nanpercentile seems slow)

- use DataArray.apply(xclim.utils.calc_perc, dim='realization') - 'calc_perc' only applies nanpercentile where needed ... should technically be faster

aulemahal:
### Pull Request Checklist:
    - This PR fixes #169

Improves the computation of ensemble percentiles according to the benchmarking done in Ouranosinc/xclim-benchmark#2, as suggested in #169. Computation is lazy and faster than the previous implementation, but can consume more memory, depending of the chunksize. An option to try to keep the incoming chunksize after merging the chunks along 'realization' is added as a keyword `keep_chunk_size` (replacing `time_block`). By default, it uses the same size threshold of 2e8 elements as the previous implementation to choose whether to rechunk or not.

Also, adds minor modifications to the `create_ensemble` function, following #169. 'realization' is now an `IndexVariable' with range-like values, and the `axis='E'` attribute.

Moreover, in #318, I forgot to pass `**xr_kwargs` to `_ens_align_datasets()`. Fixed.

Yes, the loss of the `time_block` keyword might break existing scripts!


xclim/ensembles.py
```diff
@@ -338,106 +361,18 @@ def _ens_align_datasets(
...
-        )
-
-    nan_count = np.isnan(arr).sum(axis=0)
-    out = np.percentile(arr, p, axis=0)
-    if np.any((nan_count > 0) & (nan_count < arr.shape[0])):
-        arr1 = arr.values.reshape(arr.shape[0], int(arr.size / arr.shape[0]))
-        # only use nanpercentile where we need it (slow performace compared to standard) :
-        nan_index = np.where((nan_count > 0) & (nan_count < arr.shape[0]))
+def _calc_perc(arr, p=50):
+    nan_count = np.isnan(arr).sum(axis=-1)
```
huard: docstring

huard:
I suggest we release 0.13 once this is merged.

tlogan2000:
This looks cool but will need a bit of time to review.  @aulemahal I would maybe want to get you to benchmark using some of the climatedata.ca data (lots of gridcells (all of canada), 24 realizations 151 timesteps  ) ...  

Quick question for @aulemahal ... between the two new option in  https://github.com/Ouranosinc/xclim-benchmark/pull/2 (apply and apply rechunk) if I understand apply is the 'faster' option but can use more memory... anyway to combine the two with a rule of thumb chunksize? e.g. if chunksizes > n apply_rechunk else apply...

aulemahal:
@tlogan2000  I generated 25 realizations of random data made of 4 chunks of 50x1000x500. I think this resembles  what you were thinking of? 

It gives the following:
ensemble_percentile_memuse_ManyRealizations_FewLargeChunks_SmallTimeDim
Without rechunking, it behaves as expected, but the default rechunking fails (highest red line). I think this is due to the small size of the time dimensions and its chunks here (4x50). Thus, when rechunking, we get chunks of 2x1000x500, which seems not efficient. 

I added a test so that the rechunking is done on the dimensions with the largest chunks, that the lowest red line. It's incredibly better, dask is so mysterious... Coming up in the PRs.

PR #342, opened 2020-01-10
Add command line interface

### LINKED ISSUE #195: Command line interface to compute indicators
### Description

To help users who are not familiar with python, it would be useful to have a command line interface to compute indicators on a designated netCDF file.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #195 

Addition of a command line interface to xclim! As suggested in #195, uses `Click` to generate the cli tool. This is still WIP at the moment but it is usable. Docs and tests still to be added.

To get the right metadata for each indicator, I modified the `Indicator` class to extract the first docstring line for each parameter in the indice.

It registers a `xclim` "console_script" with setuptools so it can be used as :
```
xclim -i input.nc -o output.nc atmos.tg_mean --freq='YS'
```
In this example, it computes `tg_mean` on the variable `tas` in `input.nc` and outputs to `output.nc`.

Custom variable names can be passed:
```
xclim -i input.nc -o output.nc atmos.tg_mean --tas temperature ---freq='YS'
```

When given multiple commands, it writes them all to the file:
```
xclim -i input.nc, -o output.nc atmos.tg_mean atmos.growing_degree_days --thresh "4 degC"
```

A list of all available indicator and information about them can be given with:
```
xclim list   # All indicators in atmos, land and seaIce, name and long_name only.
xclim list -i land  # All indicators in xclim.land with their long_name + docstring abstract
xclim info seaIce.sea_ice_extent     # Detailed metadata info
xclim seaIce.sea_ice_extent --help  # Only info on how to call it
```

Any comments?

No.


xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+        short_help=indicator.long_name,
+    )
+
+
+@click.command(short_help="List indicators.")
+@click.argument("module", required=False, nargs=-1)
+@click.option(
+    "-i", "--info", is_flag=True, help="Prints more details for each indicator."
+)
+def list(module, info):
```
Zeitsperre: Codacity is raising errors over this redeclaration. If this is the proper way of using click, maybe adding a one exemption to the code linter here would be good.

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+import inspect
+
+import click
+import xarray as xr
+from dask.diagnostics import ProgressBar
+
+import xclim as xc
+from xclim import atmos
+from xclim import land
+from xclim import seaIce
```
Zeitsperre: Just an FYI, but it's no longer necessary to import these individual modules, just `ensembles` and `subset` need explicit import.

Zeitsperre:
Great work! Seeing as this is a major development, it would be good to add a page to the documentation (between `utilities` and `contributing` pages) that explains the behaviour and shows some shell output examples. I can give you a hand with this if you'd like.

xclim/cli.py
aulemahal: Or we could choose another name, like "list-indices", or "indices". Or use the `get_command` to return a properly-named function when the uses enters "list".

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+    dsin = _get_input(ctx)
+    dsout = _get_output(ctx)
+
+    for key, val in params.items():
+        if (
+            indicator._sig.parameters[key].default is inspect._empty
+        ):  # a Dataset is expected
+            if key == "tas" and val is None and key not in dsin.data_vars:
+                # Special case for tas.
+                params["tas"] = atmos.tg(dsin.tasmin, dsin.tasmax)
```
huard: Brittle. What guarantees that tasmin and tasmax are part of the inputs ? I would argue against special cases here.

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+        ctx.obj["xr_kwargs"].pop("combine", None)
+        ds = xr.open_dataset(arg, **ctx.obj["xr_kwargs"])
+    ctx.obj["input"] = ds
+    return ds
+
+
+def _get_output(ctx):
+    if "ds_out" not in ctx.obj:
+        dsin = _get_input(ctx)
+        ctx.obj["ds_out"] = dsin.drop_vars(dsin.data_vars)
```
huard: I don't think this is the expected behavior. If you have a file with 100 variables, they're all going to get copied except the inputs, correct ?

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+            with formatter.section(
+                click.style("Indicators in module ", fg="blue")
+                + click.style(f"{modname}", fg="yellow")
+            ):
+                formatter.write_dl(
+                    [
+                        (
+                            click.style(name, fg="yellow")
+                            + (
+                                (" (" + ind.var_name + ")")
```
huard: Use f-strings.

huard: List comprehensions are for one liners. Use a proper for loop here.

Zeitsperre:
```suggestion
                                (f" ({ind.var_name})")
```

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+    ctx.obj["ds_out"] = dsout
+
+
+def _create_command(name):
+    indicator = _get_indicator(name)
+    params = []
+    for name, param in indicator._sig.parameters.items():
+        params.append(
+            click.Option(
+                param_decls=["--" + name],
```
Zeitsperre:
```suggestion
                param_decls=[f"--{name}"],
```

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+                        (
+                            click.style(name, fg="yellow")
+                            + (
+                                (" (" + ind.var_name + ")")
+                                if ind.var_name != name
+                                else ""
+                            ),
+                            ind.long_name
+                            + (
+                                ("\n" + ind.abstract.replace("\n    ", ""))
```
Zeitsperre: (f"\n {', '.join(a.splitlines())}")

xclim/cli.py
```diff
@@ -0,0 +1,239 @@
...
+                )
+
+        command.format_options(ctx, formatter)
+
+        click.echo(formatter.getvalue())
+
+
+class XclimCli(click.MultiCommand):
+    def list_commands(self, ctx):
+        return ["list", "info"]
```
Zeitsperre: Is there a reason why this is wrapped in a list? Wouldn't a tuple work the same?

xclim/cli.py
aulemahal: This copies all attributes and coordinate variables from the input to the output. I guess if you had 100 coordinates that might be too much, but that's a improbable case, no?

huard: Agreed, but even if there is a just of couple of extra unrelated variables in my input data, I'd be surprised to see them pop up in the output. Is suggest explicitly copying over global attributes that are identical for all input files.

xclim/cli.py
aulemahal: From what I've seen in the last months, it's quite common that `tas` is missing. I agree that this is brittle and I was going to add a check and warning before doing it, but I think it should be available.

xclim/cli.py
```diff
@@ -83,7 +85,8 @@ def _process_indicator(indicator, ctx, **params):
                     params[key] = dsin[val or key]
                 except KeyError:
                     raise click.BadArgumentUsage(
-                        f"Variable {val or key} absent from input dataset. You can provide an alternative name with --{key}",
+                        "Variable {val or key} absent from input dataset. "
```
Zeitsperre: Are you missing an `f` here?

aulemahal: I am!

aulemahal:
Note: This PR is staling because of bugs when parsing the signature of wrapped partial indicators, which is kinda related to #364.

Zeitsperre:
@aulemahal Just curious, but do you think this PR is closable this week? Just trying to set milestone goals.

aulemahal:
Hum, we maybe could, but the upcoming refactor of the Indicator class would mean to rework this part again... Unless this is explicitly needed, I would wait for the refactor! 
BTW, it now works for most common cases, however I need to make more robust tests.

tests/test_cli.py
```diff
@@ -61,27 +55,49 @@ def test_indicator_help(indicator, indname):
...
-    tasmin = tasmin_series(np.ones(366,) + 270.15, start="1/1/2000")
-    tasmax = tasmax_series(np.ones(366,) + 272.15, start="1/1/2000")
-    pr = pr_series(np.ones(366,), start="1/1/2000")
-    ds = xr.Dataset(data_vars={"tasmin": tasmin, "tasmax": tasmax, "pr": pr})
+    tasmin = tasmin_series(
+        np.ones(
+            366,
+        )
+        + 270.15,
+        start="1/1/2000",
```
Zeitsperre, line 72: If you remove the trailing `,` after `start`, does it still explode?

aulemahal: It do. Thanks!

tests/test_cli.py
```diff
@@ -0,0 +1,220 @@
...
+    assert out.tg_mean[0] == 1.0
+    assert out.growing_degree_days[0] == 0
+
+
+def test_missing_variable(tas_series, tmp_path):
+    tas = tas_series(
+        np.ones(
+            366,
+        ),
+        start="1/1/2000",
```
Zeitsperre, line 177:
```suggestion
        start="1/1/2000"
```

tests/test_cli.py
```diff
@@ -0,0 +1,220 @@
...
+    results = runner.invoke(
+        cli,
+        [
+            "-i",
+            str(input_file),
+            "-o",
+            str(output_file),
+            "tg_mean",
+            "growing_degree_days",
+        ],
```
Zeitsperre, line 160:
```suggestion
            "growing_degree_days"
        ]
```

tests/test_cli.py
```diff
@@ -0,0 +1,220 @@
...
+        cli,
+        [
+            "-i",
+            str(input_file),
+            "-o",
+            str(output_file),
+            "tn_mean",
+            "--tasmin",
+            "tas",
+        ],
```
Zeitsperre, line 129:
```suggestion
            "tas"
        ]
```

tests/test_cli.py
```diff
@@ -0,0 +1,220 @@
...
+    out = xr.open_dataset(output_file)
+    assert out.tn_mean[0] == 1.0
+
+
+def test_indicator_chain(tas_series, tmp_path):
+    tas = tas_series(
+        np.ones(
+            366,
+        ),
+        start="1/1/2000",
```
Zeitsperre, line 143:
```suggestion
        start="1/1/2000"
```

aulemahal:
@huard  @tlogan2000  @Zeitsperre  This is now ready for review!

I am currently testing all indicators locally, just to be sure.

The think I like the least is the way we get the list of available indicators, to repeat here things I said in the slack: The current registry is useful for creating new indicators but not for listing existing ones as it contains the classes, not the instances. Getting a usable indicator requires calling the class and thus, overwriting it and triggering a warning. I suggest creating a second registry with the instances. This PR would be an appropriate place for this change.   `registry -> class_registry` +  `registry`.

Also there is this naming problem. Using the registry (as opposed to parsing the modules) requires the use of the identifier. Example: `rain_on_frozen_ground_days` is referred to as `rain_frzgr`.

Zeitsperre:
Could the PR that fixes the registry ensure that `xclim indicators` reports available indicators by alphabetical order?

aulemahal:
héhé. I could do this right now also.  That's another thing that the current registry can't do : classify the indicators by their "realm", as this info is not stored in the indicator itself.

Zeitsperre:
Much better!

Zeitsperre:
It's giving me what I want!
```
$ xclim -i tas.sresb1.giss_model_e_r.run1.atm.da.nc -v -o test.nc growing_degree_days --thresh="10 degC"
$ panoply test.nc
```
image

aulemahal:
I had almost forgotten, but this PR improves the doc parsing by parsing parameters and their description (in  `core.formatting.parse_doc`) + their default and annotation (in `Indicator.__new__`).  The last commit adds some tests.

I'm think this new functionality is interesting for wrapping purposes as the parameters doc is now returned in `Indicator.json()`.

Also, we could push it even further by generating Indicator docstrings from it, instead of simply copying the compute function's one. To make sure indices have properly written docstrings, we could compare the generated output to the raw one and fail tests when differences are found.

xclim/indices/generic.py
```diff
@@ -305,7 +305,15 @@ def fa(
...
 
 
-def frequency_analysis(da, mode, t, dist, window=1, freq=None, **indexer):
+def frequency_analysis(
+    da: xr.DataArray,
+    mode: str,
+    t: Union[int, Sequence[int]],
+    dist: str,
+    window=1,
+    freq=None,
```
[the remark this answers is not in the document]
Zeitsperre: Also, `from typing import Optional` !

huard (approved):
Calling `xclim --help` feels a bit slugish. Takes about 2sec to execute. Any way to speed this up ? could be done in another PR. 

I get errors when building the docs locally: 
```
FileNotFoundError: [Errno 2] No such file or directory: b'/home/david/src/xclim/docs/notebooks/out1.nc'
```

xclim/cli.py
```diff
@@ -0,0 +1,304 @@
...
+            )
+    return ctx.obj["ds_out"]
+
+
+def _process_indicator(indicator, ctx, **params):
+    """Add given climate indicator to the output dataset from variables in the input dataset.
+
+    Cmputation is not triggered here if dask is enabled.
+    """
+    click.echo(f"Processing : {indicator.identifier}")
```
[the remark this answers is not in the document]
aulemahal: Could be. So the non-verbose mode would only have the progress bar? Or nothing at all?

xclim/cli.py
```diff
@@ -0,0 +1,304 @@
...
+@click.option("-v", "--verbose", help="Make it more verbose", count=True)
+@click.option("--version", is_flag=True, help="Prints xclim's version number and exits")
+@click.option(
+    "--dask-nthreads",
+    type=int,
+    help="Start a dask.distributed Client with this many threads and 1 worker. If not specified, the local schedular is used. If specified, '--dask-maxmem' must also be given",
+)
+@click.option(
+    "--dask-maxmem",
+    help="Memory limit for the dask.distributed Client as a human readable string (ex: 4GB). If specified, '--dask-nthreads' must also be specified.",
```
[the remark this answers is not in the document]
aulemahal: It felt like this would be extremely machine-dependent and that I wouldn't be able to find a good generic default number?  Like 4GB would be safe for small machines but completely useless on large ones...

xclim/cli.py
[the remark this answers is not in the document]
huard: My own preference would be just the progress bar, but curious to hear others on this.

xclim/cli.py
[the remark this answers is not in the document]
Zeitsperre: Would establishing a baseline from a percentage make sense? For example, report available memory, take half and round to the nearest GB?

I imagine this might be overkill and would probably break if it encounters wacky OSes or server infrastructure. hmmm...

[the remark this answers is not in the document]
huard: Ok, then let's leave it at that and see how it goes. Maybe keep @Zeitsperre suggestion for later.

xclim/cli.py
```diff
@@ -0,0 +1,304 @@
...
+            raise click.BadOptionUsage(
+                "output", "No output file name given.", ctx.parent
+            )
+    return ctx.obj["ds_out"]
+
+
+def _process_indicator(indicator, ctx, **params):
+    """Add given climate indicator to the output dataset from variables in the input dataset.
+
+    Cmputation is not triggered here if dask is enabled.
```
Zeitsperre, line 71:
```suggestion
    Computation is not triggered here if dask is enabled.
```

Zeitsperre:
Would it be possible to squelch `DeprecationWarnings` and `FutureWarnings` from xarray at runtime by default? I feel like this information isn't going to be useful for someone making use of the CLI, assuming they just need to calculate a file or two.

I imagine suppressing warnings of that nature or from xarray/pandas might be enough? https://docs.python.org/3.8/library/warnings.html#temporarily-suppressing-warnings

Zeitsperre (approved):
This looks great to me, if that counts for anything.

aulemahal:
@huard  Sorry, I don't think I can influence the loading time much for `xclim --help`... All happens before it even enters the method... 
As for the doc building error, I guess it was fixed in later commits, since it doesn't happen on RTD...

@Zeitsperre I could filter **all** warnings when the verbose mode is not required?  I can't use a context since `Click` is calling each method, so I have to modify the warning filter permanently in the main function.  But this would also ignore our cf-compliance and data-checks warnings, I'm not sure that's what we want...

Zeitsperre:
> @Zeitsperre I could filter **all** warnings when the verbose mode is not required? I can't use a context since `Click` is calling each method, so I have to modify the warning filter permanently in the main function. But this would also ignore our cf-compliance and data-checks warnings, I'm not sure that's what we want...

No, that definitely isn't what we want. It was just a thought to make it look a bit better. It's too bad that we can't just filter out specific types of warnings, but that's fine. This PR is more than ready IMO.

aulemahal:
@Zeitsperre  I realised that I was more thinking of squelching warnings from specific modules instead of specfic _types_ of warnings, the latter being quite easy to do. So, the last commit removes all FutureWarnings and DeprecationWarnings from the output, only if the verbose mode is **not** enabled.

Otherwise, I'll merge once the build finishes.

PR #347, opened 2020-01-13
Multi-region masks

### LINKED ISSUE #344: Subset with individual features
* xclim version: 0.13.0

It would be great to have an option for us to explode the features of a shapefile and subset spatially according to their boundaries. This has been implemented in a few libraries but it wouldn't take much work to modify our current `subset_shape` and `create_mask` to have this behaviour.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #344

Modifies `xclim.subset.wraps_lons_and_split_at_greenwich` so that each feature of multi-feature `GeoDataFrame` is treated individually. This fixes any possible issues of passing a list of geometries to `xclim.subset.create_mask`, thus the final values in the mask are the expected index of the associated feature.

We tried to have a way to transfer names given in the `GeoDataFrame` to the returned mask, but didn't find any satisfactory method, so the recommended way would be:
```python
from xclim import subset
import xarray as xr
import geopandas as gpd
ds = xr.open_dataset('example.nc')
polys = gpd.read_file('regions.json')
mask = subset.create_mask(x_dim=ds.lon, y_dim=ds.lat, poly=polys)
ds = ds.assign_coords(regions=mask)
region_names = xr.DataArray(polys.id, dims=('regions',)))
ds = ds.groupby('regions').mean().assign_coords(region_names=region_names)
```

This is also in the improved docs.

No. Unless someone was expecting a multi-valued mask with shapes crossing 0° longitude.


Zeitsperre:
I think you may have forgotten to add the multi-feature json to the PR. Once that's in, I'll take a look. Thanks!

PR #348, opened 2020-01-17
Reformat lon and lat dimnames for atypically named data dimensions

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

This PR includes another checking decorator (`check_latlon_dimnames`; not married to the function name) that renames the latitude and longitude coordinates so that they are synonymous with the rest of our subsetting tools. There are plenty of edge cases where this can occur ("x" and "y"; "longitude" and "latitude"; "lats" and "lons") and this is indeed a case for data extracted via the CDSAPI (ERA5, ERA5-Land and other Copernicus data sets).

As it stands, any xarray.Dataset or DataArray that is read by `subset_bbox`, `subset_gripoint`, or `susbset_shape` that has non `lat` and `lon` coordinates will have those coordinates renamed. One way to prevent that from happening would be to style the decorator to rename these dimensions to their previous names after the subsetting function has been passed on them. This might require more work, I'm not sure.

* **Other information**:

### DIFF
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -183,8 +183,8 @@ def check_latlon_dimnames(func):
                     dims = argument.dims
                 else:
                     raise TypeError
-            except (IndexError, TypeError):
-                logging.error(f"No file or no dimensions found in arg {argument}.")
+            except TypeError:
+                logging.error(f"No file or no dimensions found in arg `{argument}`.")
                 formatted_args.append(argument)
                 continue
             try:
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -178,13 +178,10 @@ def check_latlon_dimnames(func):
         formatted_args = list()
         conv = dict()
         for argument in args:
-            try:
-                if isinstance(argument, (xarray.DataArray, xarray.Dataset)):
-                    dims = argument.dims
-                else:
-                    raise TypeError
-            except TypeError:
-                logging.error(f"No file or no dimensions found in arg `{argument}`.")
+            if isinstance(argument, (xarray.DataArray, xarray.Dataset)):
+                dims = argument.dims
+            else:
+                logging.info(f"No file or no dimensions found in arg `{argument}`.")
                 formatted_args.append(argument)
                 continue
 
--- a/xclim/subset.py
+++ b/xclim/subset.py
@@ -187,26 +187,21 @@ def check_latlon_dimnames(func):
                 logging.error(f"No file or no dimensions found in arg `{argument}`.")
                 formatted_args.append(argument)
                 continue
-            try:
-                if not {"lon", "lat"}.issubset(dims) or {"rlon", "rlat"}.issubset(dims):
 
-                    if {"long"}.issubset(dims):
-                        conv["long"] = "lon"
-                    elif {"latitude", "longitude"}.issubset(dims):
-                        conv["latitude"] = "lat"
-                        conv["longitude"] = "lon"
-                    elif {"lats", "lons"}.issubset(dims):
-                        conv["lats"] = "lat"
-                        conv["lons"] = "lon"
-                    else:
-                        pass
-                    argument = argument.rename(conv)
-
-            except ValueError:
-                logging.error(
-                    f"Lat and Lon dimensions are not found among dimensions: {list(dims)}."
-                )
-                raise
+            if not {"lon", "lat"}.issubset(dims) or not {"rlon", "rlat"}.issubset(dims):
+                if {"long"}.issubset(dims):
+                    conv["long"] = "lon"
+                elif {"latitude", "longitude"}.issubset(dims):
+                    conv["latitude"] = "lat"
+                    conv["longitude"] = "lon"
+                elif {"lats", "lons"}.issubset(dims):
+                    conv["lats"] = "lat"
+                    conv["lons"] = "lon"
+                if not conv:
+                    warnings.warn(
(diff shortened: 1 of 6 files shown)


xclim/subset.py
```diff
@@ -164,6 +165,37 @@ def func_checker(*args, **kwargs):
...
+            except (IndexError, TypeError):
+                logging.error(f"No file or no dimensions found in arg {argument}.")
+                formatted_args.append(argument)
+                continue
+
+            if not {"lon", "lat"}.issubset(dims):
+                if {"latitude", "longitude"}.issubset(dims):
+                    argument = argument.rename({"latitude": "lat", "longitude": "lon"})
+            formatted_args.append(argument)
+        return func(*formatted_args, **kwargs)
```
huard: Question: should we convert the output dimension names back to their original values ? 
Also, what if the name is "long" instead of "lon" ?

Zeitsperre: I feel like it would be good to be consistent with inputs and outputs, even if they aren't *exactly* using the standard. I was thinking that a way of implementing it could be using a dictionary to keep track of name changes and, if that dictionary isn't empty, call the function within the decorator and perform a dimension rename on exit.

huard: Just thinking here, but another way to go about this is to write a function `to_cf` that would try to convert a dataset to the CF convention. 
What I'm concerned here is that with this decorator, the subset function has an undocumented and unexpected side-effect (dimension name change).

Zeitsperre: `long` is absolutely possible and an edge case that could go in here.

Zeitsperre: The alternative is to figure out what the longitude and latitude dimensions are and tell all subsetting functions to use those instead of `lon` and `lat`. That might be a lot more work.

Zeitsperre: This is why I think it would be good to not change the output dimension names from what they were at the start. I do like the `to_cf` method you're thinking of though. Could we push some of this code to a function like that?

huard: I think so. And you know what would be nice ? If the `to_cf` method could also output a NcML file embedding this conversion from original file to cf-compliant.

tlvu: Just an idea here, I know nothing about CF conventions so it might sound noob or impossible.

Are we allowed to create alias/alternate name/symlink to the old name that is not good?  So there are no "unexpected side-effect (dimension name change)", we are just adding an alias to make it compliant.

Zeitsperre: Honestly, the idea behind this is simply to make it easier to perform operations on those coordinates. An alias could certainly work for that.

@huard The idea of getting things closer to CF-conformance is definitely something to bring into xclim. I think we've hinted at it in a few issues. Might be good to open a new one with the laundry list of what we would want (similar to the data quality assurance checks discussion).

huard: The alias idea is not crazy, but this is something that would go into xarray. Given that renames are simple operations, I'm not convinced this would fly.

xclim/subset.py
```diff
@@ -164,6 +165,37 @@ def func_checker(*args, **kwargs):
...
+            return func(*args, **kwargs)
+
+        formatted_args = list()
+        for argument in args:
+            try:
+                if isinstance(argument, (xarray.DataArray, xarray.Dataset)):
+                    dims = argument.dims
+                else:
+                    raise TypeError
+            except (IndexError, TypeError):
```
aulemahal: I might be missing something, but what could raise a `IndexError` here? Are there cases of DataArrays or Datasets without `dims`?

Zeitsperre: For that, I can't say for sure. I imagine that an empty Dataset is possible. I figured that's a reasonable case that could happen.

huard: I suggest explicit type checking and error handling is done only in cases where they provide added-value: 
- the standard error message is unclear
- there is a potential for silent errors

Zeitsperre: Type-checking here does serve a purpose: In the case of shape_subset, a shape can currently be passed as an arg. This prevents us from trying to grab dims from an unopened vector file so that much of the code is needed. The IndexError can definitely be removed.

xclim/subset.py
```diff
@@ -164,6 +165,56 @@ def func_checker(*args, **kwargs):
...
+
+        formatted_args = list()
+        conv = dict()
+        for argument in args:
+            try:
+                if isinstance(argument, (xarray.DataArray, xarray.Dataset)):
+                    dims = argument.dims
+                else:
+                    raise TypeError
+            except TypeError:
```
aulemahal: I'm guessing  we could get rid of the try-except here too?

xclim/subset.py
```diff
@@ -164,6 +165,56 @@ def func_checker(*args, **kwargs):
...
+                    conv["lons"] = "lon"
+                if not conv:
+                    warnings.warn(
+                        f"lat and lon-like dimensions are not found among arg `{argument}` dimensions: {list(dims)}."
+                    )
+                argument = argument.rename(conv)
+
+            formatted_args.append(argument)
+
+        final = func(*formatted_args, **kwargs)
```
aulemahal: Are there cases where the output is more than one dataset? Do we want to cover them ?

aulemahal: Right now I'm thinking of the FWI where latitude is needed and the core indicator function returns 6 datasets. But this is out of scope, I agree.

xclim/subset.py
Zeitsperre: see #350 for the follow-up discussion on this. We can eventually port some of this code to a better method once we figure out how best to proceed.

PR #359, opened 2020-02-07
Metadata internationalization

### LINKED ISSUE #352: Internationalization of metadata
A need for internal Ouranosinc projects, but potentially for external users, is to have available metadata in another language then english. As far as I know, there is not CF-standardized way of adding multilingual metadata in a netCDF, so we're our own for this.
Also, while python has the GNU `gettext` tool built-in, it might not be what we want. But, I have found no examples of other python packages similar to xclim implementing such a feature. So on our own here to.

In this thread, I'd like to have people's input on the way to go forward. I have narrowed it down to 3 main options + ancillary questions :

Internationalization support in `atmos`, `seaIce` and `land`

1) None. This is out-of-scope for `xclim`
2) Gettext-based translation. We could ship pre-compiled gettext files with xclim and use python's gettext to generate the metadata. We could use tools like [`babel`](http://babel.pocoo.org/en/latest/) to help.
3) Json-based translation. As we have a limited number and explicit list of text fields to translate, we could json files to xclim, for each locale, mapping indice names and fields to translated strings. Would need are own tools to generate the fields.

How to (for 2 and 3):

A) Monkey-patching. If loading xclim and calling `xclim.add_locale('fr')`, all indices would be monkey-patched so that they contain the translated fields too.
B) `set_locale` + context. The `Indicator` class would be modified so it adds the translations from the globally- or context-set `locale` option.  Similar to xarray's `set_options`.

Metadata:
I suggest adding "_fr" and similar to the localized attributes.

My favorite way is 3B.  What do you think?

aulemahal:
### Pull Request Checklist:
    - This PR fixes #352 

This PR adds the possibility to get metadata information in other language than the hard-coded english. It does so by adding a `locale` submodule that handles dictionaries (json) providing translated fields for each Indicator in `atmos`, `land` and `seaIce`. This PR also provides the french translations (or at least my version of them).

The additional locales are activated either globally with:
```
xclim.locale.set_locales('fr-CA', 'de')
```
or within a context with:
```python
with xclim.locale.metadata_locale('fr-BE', 'de'):
    # create indicators with french and german meta
# not anymore
```
The available translated fields can accessed at any time with
```
xclim.atmos.tg_mean.translate_attrs('de-AT')
```
In the examples above, it is assumed that json files for french and german exist. However, the system is flexible. Let's say that `fr-CA.json` and `de.json` are stored in `xclim/locale/`, they would match in all cases. The priority goes as: 1) exact locale name match, 2) locale without territory match (`de-AT` matches `de`), 3) locale of another territory match (`fr-BE` matches `fr-CA`).

If a full json file is not provided by xclim, but exists elsewhere, it can be given as a tuple:
```
xclim.locale.set_locales( ('eo', '/path/to/esperanto.json') )
```
Finally, a dictionary can be passed in the tuple:
```
xclim.locale.set_locales( ('tlh', klingon_translation_dict) )
```

The translation dictionaries are structured like this:
```json
{
    "attrs_mapping" : {
        "modifiers": ["", "f", "mpl", "fpl"],
        "YS" : ["annuel", "annuelle", "annuels", "annuelles"],
    },
    "atmos.dtrvar" : {
        "long_name": "Variabilité de l'intervalle de température moyen",
        "description": "Variabilité {freq:f} de l'intervalle de température moyen."
    }
}
```
The `attrs_mapping` subdict and its `modifiers` list are mandatory. `modifiers` should be as long as the longest sequence of translations in `attrs_mapping`. `modifiers` defines formatting strings that can be used with the formatting keyword in the translated strings. The keys of `attrs_mapping` should be the same as in `xclim.utils.default_formatter.mapping`.  In our case here, we know that the values passed to `freq` are translated into words by `xclim.indicator.Indicator.format`, so we redefine its possible values in the translation dict, and add modifiers so xclim knows which declension to use.

This is done by a new object: `xclim.utils.AttrFormatter`. It is initiated by a list of modifiers and a list of potential keyword translation (the keys of `attrs_mapping`).  The Indicator retrieves the appropriate formatter and translated attrs during the output metadata processing.

The attributes that xclim will try to translate are given by `xclim.locale.TRANSLATABLE_ATTRS`. For now I put `long_name`, `description`, `comment` (but it is completely unused) and `title`. We could put more. As for the suggestion of translating the description of an indicator's parameters, it is not yet parsed, so kinda inaccessible. But I have plans for that, see an upcoming issue/PR.

Yes. To simplify our job, I removed one level of hierarchy in `xclim.utils.Indicator._attrs_mapping`. Before, translation of keywords was specific for each field, but this feature was not used anywhere. The only field with translations defined differently was `cell_methods` but there are no indicators with format strings for their `cell_methods` field. Also, I removed the `xclim.utils.format_keywords` function as it was unused in xclim, it was a weaker version of `Indicator.format()`.

* **Other information**:
I'm missing tests and a full documentation page, but I wanted to have your opinions.


xclim/locales/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+            json.load(pkg_resources.resource_stream(__package__, f"{locale}.json")),
+        )
+    if isinstance(locale[1], dict):
+        return locale
+    with open(locale[1]) as locf:
+        return locale[0], json.load(locf)
+
+
+class TranslatableStr(UserString):
+    """A string that can to translate certain arguments passed to format()`.
```
huard: typo

xclim/locales/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+        removed from the arguments and replaces by all versions of `keyword` + `modifier`.
+
+        Example
+        -------
+        Let's say the string "The dog is {adj1}, the goose is {adj2}" is to be translated
+        to french and that we know that possible values of `adj` are `nice` and `evil`.
+        In french, the genre of the noun changes the adjective (cat = chat is masculine,
+        and goose = oie is feminine) so we initialize the string as:
+
+        >>> s = TranslatableStr("Le chien est {adj1_m}, l'oie est {adj2_f}",
```
huard: You could possibly use custom formatters instead: https://tobywf.com/2015/12/custom-formatters/

aulemahal: This indeed looks better. And maybe will make it less restrictive. Thanks!

tlogan2000:
@tomLandry, @davidcaron 

FYI on work translating metadata

cothel:
Désolée pour mon intervention tardive. Je trouve que la traduction des noms de variables comme les **long name** et même les **standards name** est un jeu dangereux qui devrait pas être laissé aux mains d'une fonction de traduction. Cela nous a causé des problèmes dans le passé (les flux radiatifs sont un bon exemple de cauchemar). C'est pourquoi nous avons choisi de ne plus traduire nos listes de variables pour le MRCC. 
À mon avis si vous tenez à la traduction, il faut construire un dictionnaire hyper-vérifié et conforme à l'OMM.

aulemahal:
@cothel C'est justement ce qui m'a fait éviter l'utilisation de quelque chose à la gettext. Ici, c'est un dictionnaire statique qui fait la traduction et pour les langues principales (i.e. le français), ces traductions sont inclues au processus de versionnage et de vérification de git(hub), ça rend ça plus safe je pense. Et je ne veux pas traduire `standard_name` c'est certain.  Pour l'instant seulement  `long_name` et `description` sont traduits et inclus aux fichiers netCDF sortant.

Zeitsperre:
Be sure to `bumpversion patch` on this branch at some point.

aulemahal:
@davidcaron  Is this feature an urgent need for the external uses of xclim? 
I.e.  Should we push on this and release a 0.15 version before going into the whole xclim refactoring process that could introduce breaking changes or do we do both in the same release.

davidcaron:
Sorry, I didn't see your comment yesterday. I would say ideally for the portal it would be released somewhere before or around March 16th.

davidcaron:
Just started reviewing :)

davidcaron:
I'm starting to test the finch integration and I noticed something: is it possible the "abstract" properties are missing? For example:

```
>>> xci.json()  
{
'identifier':'rain_frzgr',
'abstract':'Number of days with rain above a threshold after a series of seven days below freezing temperature.\n    Precipitation is assumed to be rain when the temperature is above 0℃.',
'description':'{freq} number of days with rain above {thresh} after a series of seven days with average daily temperature below 0℃. Precipitation is assumed to be rain when thedaily average temperature is above 0℃.',
...
}
```

In `fr.json`
``` json
"atmos.rain_frzgr": {
        "long_name": "Nombre de jours de pluie sur sol gelé",
        "description": "Nombre {freq:m} de jours avec plus de {thresh} de pluie suivant sept jours consécutifs où la température était sous 0℃. Les précipitations sont considérés comme de la pluie lorsque la température moyenne est au-dessus de 0℃.",
        "title": "Nombre de jours de pluie sur sol gelé"
    },
```

So the "abstract" property holds a generic text without the placeholders for specific inputs.

aulemahal:
There are in fact missing, I set the "xclim.locales.TRANSLATABLE_ATTRS` to `["long_name", "description", "comment", "title"]`. I will add `'abstract'` and translate them.

aulemahal:
Le Grand Tour de Magie (LGTM)

PR #360, opened 2020-02-11
Fix boolean checking with np.all

davidcaron:
Fix bug when casting numpy array to boolean.

The test case (`test_run_length.TestFirstRun.test_simple`) doesn't test for this case, because the index is None.

### DIFF
--- a/xclim/run_length.py
+++ b/xclim/run_length.py
@@ -460,7 +460,7 @@ def first_run_ufunc(
         kwargs={"window": window},
     )
 
-    if index is not None and ~(np.isnan(ind)):
+    if index is not None and np.all(~(np.isnan(ind))):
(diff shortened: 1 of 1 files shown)


davidcaron:
I ran into this when computing the `freshet_start` indicator.

xclim/run_length.py
```diff
@@ -460,7 +460,7 @@ def first_run_ufunc(
         kwargs={"window": window},
     )
 
-    if index is not None and ~(np.isnan(ind)):
+    if index is not None and np.all(~(np.isnan(ind))):
```
huard: Would `ind.notnull().all()` be clearer ?

davidcaron: Yes, definitely!

PR #367, opened 2020-02-13
Lazy subset shape, MultiPolygon crossing meridian, DataArray in subset_shape.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

I  was using data through open_dap and realized that `subset_shape` wasn't lazy. The last part was doing a `dropna()` which triggers a download of all the data, not cool.

Then while playing with my fix, I realized that passing a `DataArray` to the method made it fail. Because of an iteration over `ds.data_vars`.

Furthermore, I realized that my fix that allowed passing a `GeoDataFrame` directly to the method was also broken 'cause `wrap_lons_and_split_at_greenwich` is modifying the thing *in place*. 

At last, this *in place* thing was discovered because the decorator doesn't accept `MultiPolygon` geometries. Fixed that too.

* **Other information**:

In order for the tests to pass, I had to modifiy a value in `test_subset.py::TestSubsetShape.test_wraps`. Why? Mystère et boule de gomme.


xclim/subset.py
```diff
@@ -256,7 +257,10 @@ def func_checker(*args, **kwargs):
                     )
 
                     # Create a meridian line at Greenwich, split polygons at this line and erase a buffer line
-                    union = Polygon(cascaded_union(feature.geometry))
+                    if feature.geometry.type == "MultiPolygon":
```
Zeitsperre: What about `if isinstance(feature.geometry, shapely.geo.Multipolygon):`?

HISTORY.rst
```diff
@@ -4,6 +4,7 @@ History
 
 0.14.x
 ------
+* Fix in `subset.subset_shape` to make it lazy.
```
Zeitsperre:
```suggestion
* Rewrote the `subset.subset_shape` function to allow for dask.delayed (lazy) computation.
```

Zeitsperre:
I'm considering bringing the coverage up on these utilities (and dealing with the crufty CRS handling). Would it be possible to test a few of your new case conditions?

PR #379, opened 2020-02-27
Date-based indices, docstring and test cleanup and mappings notebook

Zeitsperre:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #103 
    - This PR now also closes #391 
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] HISTORY.rst has been updated (with summary of main changes)
- [x] `bumpversion (minor / major / patch)` has been called on this branch
- [ ] Tags have been pushed (`git push --tags`)

* **What kind of change does this PR introduce?** 

This adds a mock module (`xclim.pcc`) alongside the `icclim` module and adds several new indices (`growing_season_end`, `date_of_last_spring_frost`, `longest_hot_spell`, `hot_spell_frequency`, `dry_days`,  `maximum_consecutive_frost_free_days`). The naming of the mappings follows from the variable names available on the PCC Climate Atlas information page.

For some indices, the simpler definitions for similar events (heat waves, for example) required a new indice calculation with fewer inputs. These new indices are distinctly named to not be confused with existing, more complex indices (`spells` in the case of heat waves).

The docstrings and some of the notes are not accurate and require some more time to examine. Tests have not yet been written.

* **Does this PR introduce a breaking change?** 
No, core functionality hasn't been touched. A handful of typos have been corrected in docstrings outside the scope of this PR.

* **Other information**:

PR #380, opened 2020-02-27
Fix #378 : Retrieve coordinate values in first_run.

### LINKED ISSUE #378: Make run_length methods return the coordinate values
Right now, functions like `rl.first_run` return indexes (integers) along the given dimension. Retrieving the coordinate values (like the datetime objects) is not an easy task and usually gets done by unelegant hacky lines. There should be an option on the run_length methods to do this.

Moreover, it should be lazy (which might not be simple) and we could add options to return sub-indexers (like "time.dt.dayofyear"). And, there could be a way to manage NaNs in the input : right now functions like `first_run` return NaN if there are any.

And I think there is a potential need for a "last_run" function, which is also currently done by a hacky workaround.

Ça pourrait intéresser @sbiner et @Zeitsperre !

aulemahal:
Add feat to first_run to retrieve coord values instead of indexes.
Make first_run completely lazy, add last_run
Fix freshet to work with the new first_run

### Pull Request Checklist:
    - This PR fixes #378

This adds functionality so that the ouput of `run_length.first_run` can be the coordinate value, instead of the index. Does so lazily and NaN-aware with a new `run_length.lazy_indexing` function.
Also, adds a `run_length.last_run` function.

Thus, given a boolean DataArray `da`, one can do:
```python
from xclim.run_length import first_run
first_index = first_run(da, dim='time', window=1)
first_datetime = first_run(da, dim='time', window=1, coord=True)
first_doy = first_run(da, dim='time', window=1, coord='dayofyear')
first_month = first_run(da, dim='time', window=1, coord='month')
```
Passing a string to `coord` only works on datetime coordinates. I don't know of other interesting cases? The function directly accesses the `da.time.dt` object.

Yes. I removed the option to return the index in `run_length.first_run_1d`, not to duplicate functionalities. This was used directly by `xclim.indices.freshet_start`, which I modified.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.15.x
 ------
+* Add feature to retrieve coordinate values instead of index in `run_length.first_run`. Add `run_length.last_run`.
 * Fix bug in subset_gridpoint to work on lat/lon coords of any dimension when they are not a dimension of the data.
 
 0.14.x (2020-02-21)
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.14.0
+current_version = 0.14.1-beta
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+).(?P<patch>\d+)(\-(?P<release>[a-z]+))?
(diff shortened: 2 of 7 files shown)


Zeitsperre:
Feel free to `bumpversion patch` on this and merge. I'll integrate this new functionality into #379 afterwards.

Issue #351, opened 2020-01-23
Improve Fire Weather Indexes


aulemahal:
As of #226, we now have code to compute the Fire Weather Indexes, but it is still quite slow and restrictive.

We need to: 

- [x] Vectorize the three code (DC, DMC, FFMC) functions, either directly or with [numba](https://numba.pydata.org/)
- [x] Add ways of computing the start up and shut down of the indexes. Right now, only the GFWED way, using the snow depth from the previous 60 days is implemented.
- [x] Find a way to get test data without bloating up xclim. Right now, the exposed indices are not tested.
- [x] Find a nice way to expose the individual indices OR expose a `MultiIndicators` object in `atmos`. Right now, only the `drought_code` is accessible through `atmos` and the others are computable in one go through `xclim.indices.fire_weather_indexes`.

As for the second point: we could try to use the "dry start up" method everywhere. The [CFS](https://cwfis.cfs.nrcan.gc.ca/background/dsm/fwi) has more info about a DC-specific method. Also, in the paper below, they didn't use any start up/shut down for the three codes and had an arbitrary threshold for masking the compound indexes (20% snow-covered cell area). 

Vitolo, C. et al. A 1980–2018 global fire danger re-analysis dataset for the Canadian Fire Weather Indices. Sci. Data. 6:190032 https://doi.org/10.1038/sdata.2019.32 (2019).

aulemahal:
I am currently working on the vectorization of the code functions and here are the computation times for different version of the FFMC:

1. Current loop on spatial dimensions : 35 s
2. Same function but vectorized with `np.vectorize` : 40 s
3. Same function but vectorized with numba : 450 ms
4. Rewritten ND numpy function  : 1.1 s

(data is 125x100, iterated on a full year)
The problem with #4 is that it uses a lot of compound `np.where()` and I found it difficult to maintain readability while optimizing the computation. On the other hand, it allows to boost computation without adding a new dependency.
We could have both options implemented to make numba optional, but that's a bit ugly.  It's easier to do this with 2 and 3 as we only need to redefine the "vectorize" decorator.

So. Are we ok with adding numba as dependency, or do we prefer stick to numpy and still accerated by a lot anyway?

@huard @tlogan2000 @Zeitsperre

tlogan2000:
[...]

PR #388, opened 2020-03-06
Improve fire weather

aulemahal:
### Pull Request Checklist:

Improvements of the FWI indices as planned in #351. This PR vectorizes the codes using numba, which shoud accelerate computation a lot (~ 60x), but adds a new dependency.
Also, I promoted some parts of the big iterator to their own functions. This allowed to write a cleaner start_up / shut_down method that could be improved as we implement other such methods.

Only if you can't install numba.


xclim/indices/fwi.py
```diff
@@ -607,6 +649,64 @@ def _fire_weather_calc(*args, **params):
     return tuple(ind_data.values())
 
 
+def _fwi_arg_compressor(
```
huard: This compressor / decompressor approach does not feel pythonic. I understand this is what allows users to request different combination of indices, correct ? 

What are our other options ?

xclim/indices/fwi.py
```diff
@@ -8,11 +8,36 @@
...
+shut down and the subsequent spring start up. The principal method uses the snow depth, a variable which
+is not always available. xclim offers less restrictive options usable with model climate data. In the list
+below, parameters between "" refer to those listed below.
+
+Shut down methods:
+
+    `'temperature'`
+        Pixels where the average temperature of the last "startShutDays" is below "tempThresh" are shut down.
+    `'snow_depth'`
+        In addition to the `'temperature'` condition, pixels where the average snow depth of the last "startShutDays"
```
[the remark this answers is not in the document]
Zeitsperre: Grid points?

xclim/indices/fwi.py
```diff
@@ -8,11 +8,36 @@
...
+        Same as above, but the *wet* start is only used on pixels where: 1) the average snow depth of the last "snowCoverDaysCalc" is above "minWinterSnoD" and
+        2) at least "minSnowDayFrac" % of the last "snowCoverDaysCalc" had a snow depth above "snoDThresh". For all other pixels, the *dry* start is used,
+        where DC and DMC are started with their "DryStartFactor" multiplied by the smallest number between "snowCoverDaysCalc" and the number of days since
+        the last rain event of at least "precThresh" mm.
 
-Default values for the following parameters are stored in the DEFAULT_PARAMS dict. The current implementation doesn't use all those parameters, so it might be useless to modify them.
 
 Parameters
 ----------
+Default values for the following parameters are stored in the DEFAULT_PARAMS dict. The current implementation doesn't use all those parameters, so it might be useless to modify them.
```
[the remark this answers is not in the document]
aulemahal: We could.

xclim/indices/fwi.py
aulemahal: Yes, the goal was to only pass the arrays that are needed. `xr.apply_ufunc` can only accept arrays that are defined at least over the `input_core_dims` dimensions. Another option is to set unnecessary arrays to empty arrays and empty tuples in `input_core_dims`. 
It might need less complexity in functions. I'll look into that.

aulemahal: Turns out I was wrong. Passing "None" to `apply_ufunc` is not a problem. I removed the `compressor` and `decompressor` functions and added simpler verification in `fire_weather_ufunc`.

Zeitsperre:
So the code coverage loss is expected due to `numba` creating optimized C code. We're looking into ways of maybe parametrizing running the code at least once with numba enabled and at least once without it enabled. 

For more information: https://stackoverflow.com/questions/26875191/analyzing-coverage-of-numba-wrapped-functions

setup.py
```diff
@@ -25,6 +25,7 @@
     "numpy>=1.16",
     "xarray>=0.14.1",
     "scipy>=1.2",
+    "numba",
```
Zeitsperre: Should we peg a version?

xclim/indices/fwi.py
```diff
@@ -108,79 +138,103 @@
...
 
 
-def day_length(lat):
-    """Return the average day length by month within latitudinal bounds."""
-    lat_bnds = (-90, -30, -15, 15, 33, 90)
-    i = np.digitize(lat, lat_bnds) - 1
-    return DAY_LENGTHS[i]
-
+@jit
+def day_length(lat, mth):  # pragma: no cover
```
Zeitsperre: Nice trick. Let's do our best not to abuse this ;)

xclim/indices/fwi.py
```diff
@@ -108,79 +138,103 @@
...
+        dlf = DAY_LENGTH_FACTORS[0, :]
+    elif 15 > lat >= -15:
+        return 1.39
+    elif 90 >= lat >= 15:
+        dlf = DAY_LENGTH_FACTORS[2, :]
+    return dlf[mth - 1]
+
+
+@vectorize
+def fine_fuel_moisture_code(t, p, w, h, ffmc0):  # pragma: no cover
```
Zeitsperre: Would it break functionality if these were kwargs?

xclim/indices/fwi.py
```diff
@@ -667,75 +671,76 @@ def fire_weather_ufunc(
...
+        (rh, "rh", ["DMC", "FFMC"], True),
+        (ws, "ws", ["FFMC"], True),
+        (snd, "snd", ["snow_depth"], True),
+        (tas.time.dt.month, "month", ["DC", "DMC"], True),
+        (lat, "lat", ["DC", "DMC"], False),
+        (dc0, "dc0", ["DC"], False),
+        (dmc0, "dmc0", ["DMC"], False),
+        (ffmc0, "ffmc0", ["FFMC"], False),
+    )
+    args = []
```
Zeitsperre: Matter of personal preference, but I find verbose/literals to be more readable.

setup.py
aulemahal: I thought of it, but AFAIK,  the minimum version with `jit` and `vectorize` (with the unctionalities we need) has numpy 1.16 as a dependency. So, pinning numpy implicitly pins numba. If you think explicit pinning is  better, let's do it, I don't these things...

xclim/indices/fwi.py
aulemahal: No, but I believe the compiler wants positional arguments, so we wouldn't gain much.

setup.py
Zeitsperre: That should be fine then. It might be an issue for the conda package but I'll deal with it if it comes up.

xclim/indices/fwi.py
Zeitsperre: Got it.

PR #393, opened 2020-03-10
Add relative humidity utility

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Adds a utility function to convert temperature and dew point temperature to relative humidity, useful for example in the fire weather indexes computation when RH is not directly available.


aulemahal:
Je pense que @Paattrriicckk ou @EmilieBresson connaissent mieux que moi ce calcul-là? Est-ce que ma version a de l'allure?

```
RH = 100 * exp( -L * ( T - Td) / (Rw * T * Td ) )
```

Paattrriicckk:
Salut Pascal, il y a plusieurs formulations possibles pour l'humidité relative; on pourra en jaser demain.

aulemahal:
I think this is good enough for now.

@RondeauG  Est-ce que ça répond à ton besoin futur?
@Paattrriicckk Si tu as le teps, pourrais-tu jeter un coup d'oeil pour nous assurer que je fait pas n'importe quoi? J'ai pris quelques méthodes un peu au hasard, je pense qu'on en propose un nombre raisonnable pour l'instant mais si tu penses qu'une autre est plus commune et mériterait de se trouver ici, c'est pas difficile à ajouter.

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+    invalid_values : {"clip", "fill", None}
+        What to do with values outside the 0-100 range.
+        If "clip" (default), clips everything to 0 - 100,
+        if "fill", replaces values outside the range by np.nan
+        if None, does nothing.
+
+    Notes
+    -----
+
+    In the followig, let :math:`T`, :math:`T_d`, :math:`q` and :math:`p` be the temperature,
```
huard:
```suggestion
    In the following, let :math:`T`, :math:`T_d`, :math:`q` and :math:`p` be the temperature,
```

RondeauG:
De ce que je vois, ton code fait seulement la conversion de humidité spécifique vers humidité relative ? J'aurai besoin d'aller dans les deux directions : 

`huss --> hurs --> post-process hurs --> huss`

aulemahal:
Ah ouais! Ça fait ça.

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+        Air Pressure
+    ice_thresh : str
+        Threshold temperature under which to switch to equations in reference to ice instead of water.
+        If None (default) everything is computed with reference to water.
+    method : {"dewpoint", "goffgratch46", "sonntag90", "tetens30", "wmo08"}
+        Which method to use, see notes.
+    invalid_values : {"clip", "fill", None}
+        What to do with values outside the 0-100 range.
+        If "clip" (default), clips everything to 0 - 100,
+        if "fill", replaces values outside the range by np.nan
```
huard: `fill`/`fillna` is usually used to fill nan with other values. Maybe choose another name ? `mask` ? `nan` ?

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+
+        ... math::
+
+            RH = 100\\frac{w}{w_{sat}}
+            w = \\frac{q}{1-q}
+            w_{sat} = 0.622\\frac{e_{sat}}{P - e_{sat}}
+
+    The methods differ by how :math:`e_{sat}` is computed. In all cases, :math:`log(e_{sat})`
+    is an empirically fitted function (usually a polynomial) where coefficients can be different
+    when ice is taken as refernce instead of water. Available methods are:
```
huard:
```suggestion
    when ice is taken as reference instead of water. Available methods are:
```

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+        dtas = convert_units_to(dtas, "degK")
+        tas = convert_units_to(tas, "degK")
+        L = 2.501e6
+        Rw = (461.5,)
+        rh = 100 * np.exp(-L * (tas - dtas) / (Rw * tas * dtas))
+        desc = (
+            "Relative humidity computed from temperature and "
+            f"dew point temperature with L = {L} and Rw = {Rw}"
+        )
+        method = "dewpoint"
```
huard: Wait, why is setting `dtas` changing the value of `method` ? `method` is an input, no?

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+            source = "Goff and Gratch (1946)"
+        elif method in ["wmo08", "WMO08"]:
+            e_sat = xr.where(
+                ref_is_water,
+                611.2 * np.exp(17.62 * (tas - 273.16) / (tas - 30.04)),
+                611.2 * np.exp(22.46 * (tas - 273.16) / (tas - 0.54)),
+            )
+            source = "WMO (2008)"
+        else:
+            raise NotImplementedError(f"Method {method} is not (yet) implemented.")
```
huard: I think you want a ValueError here. Method {method} is not in {method_options}.

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+
+        w = huss / (1 - huss)
+        w_sat = 0.62198 * e_sat / (ps - e_sat)
+        rh = 100 * w / w_sat
+
+        desc = f"Relative humidity computed from temperature, specific humidity and pressure. The saturation vapor pressure was calculated according to {source}."
+        if ice_thresh is not None:
+            desc += f" The computation was done in reference to ice for temperatures below {ice_thresh}."
+
+    from xclim import __version__
```
huard: Should probably be a top import.

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
     wind.attrs["units"] = "m s-1"
 
     return uas, vas
+
+
+@declare_units(
```
huard: This should go into indices, no ?

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+
+    from xclim import __version__
+
+    if invalid_values == "clip":
+        rh = rh.clip(0, 100)
+    elif invalid_values == "fill":
+        rh = rh.where((rh <= 100) & (rh >= 0))
+
+    rh.name = "rh"
+    rh.attrs["standard_name"] = "relative_humidity"
```
huard: Shouldn't this stuff be in an Indicator instance ?

xclim/core/utils.py
```diff
@@ -162,3 +164,208 @@ def sfcwind_2_uas_vas(wind: xr.DataArray = None, windfromdir: xr.DataArray = Non
...
+    rh.attrs["long_name"] = "Relative Humidity"
+    rh.attrs["description"] = desc
+    rh.attrs["units"] = "%"
+    rh.attrs[
```
huard: History handling should probably not be done by individual indices, it should be a generic operation in the Indicator class. If there is custom information to pass from the indice, I think we want to clarify how it should be done in general. Rule so far was that indices did not mess with attributes, except for units.

xclim/core/utils.py
aulemahal: Oh yeah "mask"! I didn't like "fill" but couldn't find better ideas, let's go with mask.

xclim/core/utils.py
aulemahal: It's because "method" is irrelevant when computing RH from the dew point temp. Thus, I override the method so it can be formatted correctly further down. It's explained in the doc, but we could also raise a ValueError on wrong argument inputs.

xclim/core/utils.py
aulemahal: Well. Maybe that's a discussion we could have. I feel this is the same kind of function as the surface wind ones or `xc.indices.tas`. They do not perform any indice-like stats and are completely dimension-independent, thus I think they don't really belong to the indices library.  Here I used the "declare_units" simply not to have to manage that part as it is important here, while not really in `tas()` and the surface wind utils.

aulemahal: We could make a `xclim.indices.utils` file and put them there, but I would vote against throwing them in the same places as real indices.

xclim/core/utils.py
aulemahal: Same as above: I think not cause data quality and all the checks the Indicator class does are irrelevant. But I'm open to discuss it!

aulemahal:
Suggestion after discussion with @Zeitsperre : 
Let's move this thing, the surface wind funcs, and the tas helper to a new `xclim.indices.variables` submodule (maybe not the best name?).  Also, I could move the history updating to a function/decorator and use it there. The aim would be to have an new class of methods that are not indices (cause no stats) or indicators (cause no health checks nor need for the whole formatting machinery), but still have updated metadata (and history) and unit checks when needed. The checks could all be optional to have less restrictive methods. 
Whatcha think @huard  and @tlogan2000 ?

huard:
I remain unconvinced ; ) I suggest a telco.

aulemahal:
@huard  So I moved everything to indices and atmos. I called the files "conversion" cause that's how I think they are going to be used. Also made the tests simpler without the huge dataset that I took from somewhere on the net.

aulemahal:
> Any relation with https://unidata.github.io/MetPy/latest/api/generated/metpy.calc.html ? Is this something we could leverage ?

Sorry I never answered to that. MetPy does way more than what this PR. It could be used side-by-side with xclim for small scale projects. However, it returns `pint.Quantity` objects, not `DataArray`s and is *not* lazy. For use in a big-data pipeline, that's not cool. 


Also, the current state of this PR fails because I used `wrapped_partial` in the way that #394 makes it work. Thus, I am now in a git mess. I'll wait for answers on mahmoud/boltons#242 to see the best way forward.

Zeitsperre:
@aulemahal Looks like things are good on the boltons front. Anything I can do to help?

aulemahal:
Once we merge #394 here, I believe very little will be left to do!

xclim/indices/_conversion.py
```diff
@@ -0,0 +1,435 @@
...
+    vas = convert_units_to(vas, "m/s")
+
+    # Wind speed is the hypotenuse of "uas" and "vas"
+    wind = np.hypot(uas, vas)
+
+    if not return_direction:
+        wind.attrs["units"] = "m s-1"
+        return wind
+
+    # TODO Attributes should be set by the indicator, but there are no multi-output indicators, so we set them anyway if return_direction is True,
```
huard: Raise this issue in the course of the refactoring.

Zeitsperre:
@aulemahal don't forget to `bump2version patch`!

Zeitsperre:
Looks like one of the libraries hasn't yet built a wheel for Python3.8. This should resolve soon.

xclim/core/utils.py
aulemahal: It does, which is why its now the case ;)

PR #394, opened 2020-03-11
Better partial wrapping 

### LINKED ISSUE #392: Investigate using boltons.funcutils to create modules
### Description

functools.partials generate objects that are interpreted as data by the `help` function. Maybe boltons could fix this. 

```
help(xclim.icclim)
```

aulemahal:
### Pull Request Checklist:
    - This PR fixes #392 

Rewrites  `xclim.core.utils.wrapped_partial` to get more meaningful signature and help messages. Also, allows hard and soft argument overriding. Example:
```python
from inspect import signature

def func(a, b=1, c=1):
    print(a, b, c)

newf = wrapped_partial(func, b=2)
signature(newf)  # > (a, *, c=1)
newf(1)  # >  1, 2, 1

newf = wrapped_partial(func, suggested=dict(c=2), b=2)
signature(newf)  # >  (a, *, c=2)
newf(1)  # >  1, 2, 2
```

The  `boltons.funcutils.wraps` almost did the same thing, but not quite, so I grosso-modo copied and arranged its code to get the behaviour I think we want. This now creates a function that acts as a brand new one by removing the reference to the previous that is usually stored in `func.__wrapped__`.  As such, it removes the need for a different treatment in the `Indicator` class.

Not really.

* **Other information**:

Because of the new test file "test_utils.py", I branched this PR out of the RH one, so that one needs to be merged before this one.


Zeitsperre:
@aulemahal Just curious, but presently the ICCLIM mock module now found in `xclim/indictors/__init__.py` does not make use of any `wrapped_partial` calls. Should these be changed to make use of this new functionality?

aulemahal:
Now that mahmoud/boltons#242 is closed, I simplified the thing greatly. It has to be merged into the RH branch since this one needs those changes. 
boltons 20.1 is not yet available on conda, we'll merge #393 when it is.

PR #398, opened 2020-03-16
Made units.convert_units_to lazy

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz
- [] Documentation has been added / updated (for bug fixes / features)

Simply modifies `xc.core.units.convert_units_to` so that it is lazy when receiving a DataArray with dask. 

No

* **Other information**:

When accessing a DataArray's values we have 2 choices : `da.values` returns the values, loading the array if needed, `da.data` returns the underlying container, either a dask array or a numpy array. With our current config, Pint won't force-cast data into a numpy array if the thing acts like an array.

### DIFF
--- a/tests/test_units.py
+++ b/tests/test_units.py
@@ -1,3 +1,4 @@
+import dask.array as dsk
 import numpy as np
 import pytest
 import xarray as xr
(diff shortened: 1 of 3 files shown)


tests/test_units.py
```diff
@@ -70,6 +71,11 @@ def test_fraction(self):
         out = convert_units_to(xr.DataArray([10], attrs={"units": "%"}), "")
         assert out == 0.1
 
+    def test_lazy(self, pr_series):
```
huard: I thought this was covered in test_indicators.test_delayed.

aulemahal: There is no units conversion in that test! In fact, most calls to `convert_units_to` in `xclim.indices` only act on thresholds and stuff like that, which does not impact the laziness. And the indices where the dataarray was converted were not tested for laziness.

huard: Ok thanks for catching this !

PR #400, opened 2020-03-18
`Scaling` downscaling method

huard:
Very basic downscaling method that applies a mean correction. 

I think the objective here is to name things correctly, document and test the utility functions, restructure the `train` and `predict` functions to fold into the structure proposed by @aulemahal.


aulemahal:
I inadvertently pushed code directly to the xsd branch, sorry. @huard , you might be interested in looking at "downscaling/methods.py" where I have a version of Detrended Quantile Matching (directly copied from R code).

huard:
@aulemahal I left a few comments in methods.py and just added some code for eqm which probably doesn't work. 

I think we want to be able to train and then predict. This allows us to save the correction factors for inspection or later use. 

I suggest we try to get these three in good shape (scaling, dqm, eqm) before going further (tests, some documentation, architecture, naming).

xclim/downscaling/utils.py
```diff
@@ -118,7 +118,7 @@ def adapt_freq(
     group: str = "time",
     window: int = 1,
 ):
-    """
+    r"""
```
Zeitsperre, line 121: If you're using raw strings, doesn't that affect the double backslashes in your latex formulas?

huard: sphinx chokes on formulas otherwise. Suggestion ?

Zeitsperre: Nothing. I just wasn't sure if that was meant to be there. Good to know!

Zeitsperre:
@huard Everything seems to be in working order here but I just realized why this isn't building on Travis CI: We currently only have build triggered against `master`. If we want to have a development-like branch (like `xsd`) we need to add it as a branch that Travis CI should build. I'll add it now and hopefully that will fix the problem.

aulemahal:
@huard Just pushed a `Grouper` object in addition to the `QuantileMapping`. So, what do you think of a structure as now available in `base.py`? The gain is not enormous, but it does remove duplicate code. Also, I think it can help have a more uniform output style for all steps.

huard:
I like the `Grouper` class, but I think `interp` should be an argument of `add_index`. Also, it does not suppport grouping with "time" only. 

I'm less convinced by the `QuantileMapping` class. The reason is that we want developers to be able to add new methods on their own. It's not obvious to me how you would subclass `QuantileMapping` to support other methods. I think it raises the difficulty threshold to contribute a new method, at least for now. I haven't lost hope that we'll be able to write a clear, extensible class, but we're not there yet. 

So my suggestion for now to advance in this direction would be : 
1. Add support for "time" group in Grouper class
1. Use Grouper in all current methods (qdm, dqm, scaling, eqp, loci). 
1. Move `extrapolate_qm` into `predict`. 
1. Hide `add_cyclic_bounds` into methods that interpolate (broadcast, interp_on_quantiles)
1. Make sure `train` outputs are uniform in terms of naming and attributes.

aulemahal:
Idée:
Pull the normalization and detrending out of the QuantileMapping class. So the roles are :
`QM.train` : Get the quantiles and the correction factors.
`QM.predict` : Extrapolate the quantiles, interp the corrections factors to fut, apply the correction.

QDM is then a subclass of QM where the interpolation is done from the ranks of fut.
DQM and EQM are the same as QM, but one should normalize sim and fut and detrend+retrend fut in the process (but that action is independent of the QM itself).
It gets the idea of Patrick of a modular construction. 

In this scheme, we could have a `pipeline.py` file that gives sample uses of the tools in downscaling as function taking in `obs, sim, fut, **kwargs` and outputting `qm, fut_corr`. Given more as _examples_ than methods to be used as-is.

PR #413, opened 2020-04-06
Multi gridpoints support in subset_gridpoint

### LINKED ISSUE #412: subset multipoint (scattered locations)
For a couple of recent projects I have had to extract multiple scattered site locations (weather stations, sample sites etc etc) from a netcdf and thought that the steps could form the basis of a `subset.subset_scattered()` or similar named function
 
For a large number of stations it was significantly faster than looping on `subset_gridpoint()`
One known limitation would be that (for now at least) it selects nearest neighbor points based on lon, lat values so does not account for geodistance calculations as in `subset_gridpoint()`

```
import xarray as xr
import numpy as np
import scipy.spatial.ckdtree as ckdtree

dsTest = xr.open_dataset(ncfile)
sites = dict(id=['site1','site2','site3'], lat = [46.4, 47.5, 48.6], lon = [-70.45, -75.65, -67.756]) 

# Stack lon, lat values together 
ds_stack = dsTest.stack(point=('lat', 'lon'))

# create 1d array of netcdf lon lats 
combined_x_y_arrays = np.array(list(ds_stack.point.values))

#site lon lats
cents = np.array(list(zip(sites['lat'], sites['lon'] )))

# Create ckdtree with model 1d array
mytree = ckdtree.cKDTree(combined_x_y_arrays)

# query ckdtree for indexes of site locations
dist, indexes = mytree.query(cents)

# extract from netcdf 
ds_pts = ds_stack.isel(point=indexes)

# save original stacked netcdf point values
points = ds_pts.point

# reassign dataster 'point' coordinates as integers (nb* do this because we cannot export Mutli-index 'point' using 'to_netcdf')
id_array = xr.DataArray(np.arange(0, len(cents)), dims='point',
                        coords={'point': ds_pts.point})
ds_pts = ds_pts.assign_coords({'point': id_array})

# reassign point lat/lon values as a 1d array on tuples
tup1 = tuple(zip(points.lat.values, points.lon.values))
arr = np.empty(len(tup1), dtype='O')
arr[:] = points
points_array = xr.DataArray(arr, dims='point',coords={'point': ds_pts.point})
ds_pts = ds_pts.assign_coords({'point_latlon':points_array})

# add site id or name (optional)
site_array = xr.DataArray(sites['id'], dims='point',coords={'point': ds_pts.point})
ds_pts = ds_pts.assign_coords({'site':site_array})
ds_pts

aulemahal:
### Pull Request Checklist:
    - This PR fixes #412 

Modifies `subset.subset_gridpoint` and `subset.distance` to also accept list or DataArrays of coordinates. In fact, the function now works as if lat and lon are 1D DataArrays with dimensions "site". A new decorator makes the conversion.

The "site" dimension is squeezed if only one point was requested, keeping the old behaviour.

Also added an `add_distance` flag. When True, the output has a `distance` coordinate giving the distance in m along "site".

Yes. I did not know what to do with the case where one of the points is too far, but not the others. So, I decided to simply mask all points where distance > tolerance in the output, **instead of raising an error**. So in the previous case where one requests 1 point but there is none close enough, instead of raising an error, the method returns a DataArray subsetted on the closest point, but with only NaNs.
Is this ok? We could also raise an error only if a single point was requested, that seemed incoherent to me, but is easy to do.


aulemahal:
You are right that with 1D array, `sel()` would give a 2D output. But the magic happens because lat and lon are DataArrays with a common "site" dim. So `.sel()` is able to align them. It happens in the `convert_lat_lon_to_da` decorator.

Zeitsperre:
@aulemahal I'm working on something to get rid of the `check_date_signature` and other small changes, in case the reality of four decorators seems a bit excessive at the moment.

tlogan2000:
> Cool! In fact, do we really need to provide time subsetting in `subset_gridpoint`? I don't really understand why we are offering this there. It would shorten the method, lower its complexity **and** remove one decorator!

This is a leftover when subset_time was not exposed as it's own method...  we left it there as a convenience for people who wanted to potentially do all their subsetting in a single call. ..  I think it potentially makes wps calls more flexible as well.

tlogan2000:
> You are right that with 1D array, `sel()` would give a 2D output. But the magic happens because lat and lon are DataArrays with a common "site" dim. So `.sel()` is able to align them. It happens in the `convert_lat_lon_to_da` decorator.

Okay I will take you rword for it... will look again tomorrow morning to make sure i understand

tlogan2000:
> @aulemahal I'm working on something to get rid of the `check_date_signature` and other small changes, in case the reality of four decorators seems a bit excessive at the moment.

for the date signature decorator is it even necessary to have this decorator explicityl on subset_gridpoint, subset_bbox etc? ....  We are calling subset_time within the various methods and subset_time uses the same decorator already ...

tlogan2000:
> > @aulemahal I'm working on something to get rid of the `check_date_signature` and other small changes, in case the reality of four decorators seems a bit excessive at the moment.
> 
> for the date signature decorator is it even necessary to have this decorator explicityl on subset_gridpoint, subset_bbox etc? .... We are calling subset_time within the various methods and subset_time uses the same decorator already ...

Scratch that ... subset_time doesn't use the signature check right now .... but could we not move it there?

aulemahal:
Couldn't we just remove `check_date_signature` ? It mainly throws a deprecation warning, and we had it for a few versions now!

Zeitsperre:
@aulemahal I've already done that in my branch. Will push later today.

aulemahal:
Oh! Then I'm merging.

PR #414, opened 2020-04-07
Fix 322

### LINKED ISSUE #322: some comments on polygon subsetting
Sorry to chip in - I just saw #285 and couldn't resist. For `regionmask` I found that the `shapely.vectorized` functions are faster than even converting lat/ lon to `Points` and it is also faster than `[poly.intersects(point) for point in points]` (which is basically what `gpd` does internally, although it also applies some tricks to make it faster).

To get the same as `intersects` you have to combine `touches` and `contains`:

``` python
from shapely import geometry
from shapely import vectorized
import numpy as np

lon = np.arange(-180, 181)
lat = np.arange(89.5, -90, -1)

outline = np.array([[-100.0, 50.0], [-100.0, 28.0], [-80.0, 28.0], [-80.0, 50.0]])
poly = geometry.Polygon(outline)

LON, LAT = np.meshgrid(lon, lat)
points = [geometry.Point(pt) for pt in zip(LON.flatten(), LAT.flatten())]

a = np.array([poly.intersects(point) for point in points]).reshape(LAT.shape)

b = vectorized.touches(poly, LON.flatten(), LAT.flatten()).reshape(LAT.shape)
c = vectorized.contains(poly, LON.flatten(), LAT.flatten()).reshape(LAT.shape)

(a == (b | c)).all()
```
And some timings:
``` python

%timeit points = [geometry.Point(pt) for pt in zip(LON.flatten(), LAT.flatten())]

%timeit a = np.array([poly.intersects(point) for point in points]).reshape(LAT.shape)

%timeit b = vectorized.touches(poly, LON.flatten(), LAT.flatten()).reshape(LAT.shape)
%timeit c = vectorized.contains(poly, LON.flatten(), LAT.flatten()).reshape(LAT.shape)

```
I get
```
~ 370 ms
~ 185 ms
~ 7.64 ms
~ 4.96 ms
```

Note: `a = ` is likely faster in `gpd.tools.sjoin` as he uses an `rtree` and `np.vectorize` internally, but still... (I could not get `gpd.tools.sjoin` to work, so I wasn't able to time it).

In `regionmask` I trade a tiny bit of precision for a bit of time; but also get a consistent point-on-border behavior (https://github.com/mathause/regionmask/blob/93bcc1a4a9656c13e1d6ecb0656b12980d471d4e/regionmask/core/mask.py#L259-L264).

I wasn't sure - do you allow to subset more than one Polygon at a time? Then your results may depend on the order of the polygons (if the polygons touch and you have points on the border).

Anyway, feel free to close...

---
edit: fix example

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #322 

This PR provides a new engine for creating masked grids using `shapely.vectorize` as an option by default (over the `geopandas.sjoin` method). This provides a significant speedup as suggested by @mathause in #322. The original back-end is provided in the case of GeoDataframes containing numerous small features (where `geopandas.sjoin` performs better).

This also cleans up a lot of the `subset_shape` code, extracting warnings to a separate function and supporting the CF-compliant `crs_wkt` field that is written to Datasets when they are subsetted by `subset_shape`.

Yes:
- This PR removes the deprecated `start_yr` and `end_yr` call signatures for time subsetting functions. Any scripts that still used them (despite the UserWarnings will now fail). 
- Behaviour of `subset_gridpoint` is also changes such that when called only with time dimensions (`start_date` and/or `end_date`) the function will now fail. Time subsetting is still allowed so long as a grid points location is provided.
- Any Datasets that are run through `subset_shape` will now has a `crs` variable, whether they are lon_wrapped or not. Essentially, this makes data that has been established to be in WGS84 more CF-compliant as this information is now appended by default.

* **Other information**:

The new changes to `subset_gridpoint` necessitated updates to the existing test ensemble for `subset.py`. There is now a `TestSubsetClass` to support these changes.


xclim/subset.py
```diff
@@ -377,18 +342,83 @@ def create_mask(
     >>> region_names = xr.DataArray(polys.id, dims=('regions',)))
     >>> ds = ds.assign_coords(regions_names=region_names)
     """
-    wgs84 = CRS("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")
-    # poly.crs = wgs84
+    _check_has_overlaps(polygons=poly)
```
tlogan2000, line 345: I would make this option I think... It can be super long with a shape that has lots of polyogns

xclim/subset.py
```diff
@@ -377,18 +342,83 @@ def create_mask(
...
+    >>> mask = subset.create_mask(x_dim=ds.lon, y_dim=ds.lat, poly=polys)
+    >>> ds = ds.assign_coords(regions=mask)
+    Operations can be applied to each regions with  `groupby`. Ex:
+    >>> ds = ds.groupby('regions').mean()
+    Extra step to retrieve the names of those polygons stored in the "id" column
+    >>> region_names = xr.DataArray(polys.id, dims=('regions',)))
+    >>> ds = ds.assign_coords(regions_names=region_names)
+    """
+    wgs84 = CRS(4326)
+    _check_has_overlaps(polygons=poly)
```
tlogan2000, line 421: same comment as before - optional.

xclim/subset.py
```diff
@@ -434,16 +464,16 @@ def create_mask(
 def subset_shape(
     ds: Union[xarray.DataArray, xarray.Dataset],
     shape: Union[str, Path, gpd.GeoDataFrame],
+    vectorize: bool = False,
```
tlogan2000, line 467: After time tests I think we would want to make vectorize the default.   In fact is there any reason to keep the sjoin() version at all?   We might want to test performance a bit more I suppose

tlogan2000: Anyways my small number of tests are pretty conclusive that vectorize is faster so I would put it as the dafault

xclim/subset.py
```diff
@@ -848,7 +903,7 @@ def subset_gridpoint(
     # Subset with specific start_dates and end_dates
     >>> prSub = subset.subset_time(ds.pr,lon=-75,lat=45, start_date='1990-03-13',end_date='1990-08-17')
     """
-
+    dist = None
```
tlogan2000, line 906: Why are we needing to initialize dist?

tlogan2000:
Looks pretty good made a few comments.  Is @aulemahal tagged on the review as well? We might want his input before finalizing...  Right now I would say a potential slow down of the vectorize method is the 'looping' on individual polygons.. Pascal might have a suggestion there?

For the other CRS issues I am not sure if I completely understand what the problems are ... The code seems to do what we want it to do.

xclim/subset.py
Zeitsperre: Code style. I'll see if I can rearrange some things so as to not have to do this.

tlogan2000: Not a big issue i was merely curious

Zeitsperre:
One thing I'm curious about is that the vectorize version no longer performs lon_wrapping. If everything is still working fine without it, should we drop it?

tlogan2000:
> One thing I'm curious about is that the vectorize version no longer performs lon_wrapping. If everything is still working fine without it, should we drop it?

Is there a wrap_lons test using 'vectorize' ?

Zeitsperre:
Checking the tests now.

xclim/subset.py
```diff
@@ -781,16 +796,57 @@ def in_bounds(bounds: Tuple[float, float], coord: xarray.Coordinate) -> bool:
...
 
 
+def _check_has_overlaps(polygons: gpd.GeoDataFrame):
+    non_overlapping = []
+    for n, p in enumerate(polygons["geometry"][:-1], 1):
+        if not any(p.overlaps(g) for g in polygons["geometry"][n:]):
+            non_overlapping.append(p)
+    if len(polygons) == len(non_overlapping):
+        return
+    return False
```
aulemahal, line 812: Shouldn't we issue a warning here? The output of the this is not checked above, so I'm not sure what it really does?

Zeitsperre: On it right now.

tlogan2000:
> Checking the tests now.
I think the wrapping lon is ok w/ vectorize.

The original `create_mask` method seems to explicitly assign as CRS to the flattened netcdf gridpoints where as the vectorized method does not .... however the decision to split polygon along the meridian is based on whether those same .nc lons are all positive.  We then assume the polygon should be assigned the `"+proj=longlat +datum=WGS84 +no_defs +type=crs +lon_wrap=180"`  crs. 

As such even though we do not officially assign a crs to the .nc points in the vectorize() call we can safely assume that the data treatment will be the same.  Right?

Zeitsperre:
@aulemahal perhaps you can help me with understanding something. From the test output:
```
test_subset.py::TestSubsetShape::test_wraps[True] 
test_subset.py::TestSubsetShape::test_no_wraps[False] 
test_subset.py::TestSubsetShape::test_all_neglons 
test_subset.py::TestSubsetShape::test_rotated_pole_with_time 
test_subset.py::TestSubsetShape::test_small_poly_buffer 
test_subset.py::TestSubsetShape::test_mask_multiregions 
```
Why is it that the `test_wraps` and `test_no_wraps` only report one configuration from the `pytest.mark.paramterize`. Is this typical behaviour?

aulemahal:
Hum, I don't know, it never happened to me. 
I checked in the Travic CI output and both tests were flagged as passed...

Zeitsperre:
OK, if it's just weirdness on my end, I;m fine with that. Good to merge?

aulemahal:
On my last PR we started to discuss the usefulness of having time subsetting on the spatial subsetting methods. I'm not sure why we offer it there? And also, unlike the other 2, `subset_gridpoint` accepts _only_ time subsetting, which seems strange to me. I would suggest we remove at least this option there.

xclim/subset.py
```diff
@@ -924,7 +924,9 @@ def subset_gridpoint(
     # Subset with specific start_dates and end_dates
     >>> prSub = subset.subset_time(ds.pr,lon=-75,lat=45, start_date='1990-03-13',end_date='1990-08-17')
     """
-    # check if trying to subset lon and lat
+    if lat is None or lon is None:
+        raise ValueError("Insufficient coordinates provided to locate grid point(s).")
+
     if lat is not None and lon is not None:
```
aulemahal, line 930: I think if we remove this level of condition, complexity will go down.

Zeitsperre: Right.

tlogan2000:
I think we could remove the 'only_time' option from gridpoint.  However I think we should keep the combined (space, time) option.   In a WPS I think it is advantageous to be able to do both in a single call.  As each individual WPS call creates an output netcdf if these are separated finch users would end up creating a lot of intermediate data files....

aulemahal:
Right, didn't think of that!

Zeitsperre:
How's it looking now?

tlogan2000:
> How's it looking now?

will try to look... was in a meeting with @aulemahal

aulemahal:
CodeFactor didn't update I think. But it **L**ooks **G**ood **T**o **M**e!

PR #433, opened 2020-05-05
Add support for `freq=None` in missing_* functions

### LINKED ISSUE #432: Support `freq=None` in missing_* functions
### Description

This will simplify the Indicator class logic and facilitate switching the missing method.

huard:
### Pull Request Checklist:
    - This PR fixes #432

missing functions can now handle freq=None, which is useful when computing climatologies. 

No

* **Other information**:

### DIFF
--- a/xclim/indicators/atmos/_conversion.py
+++ b/xclim/indicators/atmos/_conversion.py
@@ -20,13 +20,7 @@ class Converter(Indicator):
 
     def missing(self, *args, **kwds):
         """Return whether an output is considered missing or not."""
-        # Converters should propagate null values themselves. Otherwise, the code below would do it.
-        """
-        from functools import reduce
-        import numpy as np
-        miss = (da.isnull() for da in args)
-        return reduce(np.logical_or, miss)
-        """
+        # Converters should propagate null values themselves.
         return False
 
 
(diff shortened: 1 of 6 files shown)


xclim/indicators/atmos/_conversion.py
```diff
@@ -19,6 +19,10 @@ class Converter(Indicator):
     def validate(self, da):
         """Input validation."""
 
+    def missing(self, *args, **kwds):
+        """Return whether an output is considered missing or not."""
+        return False
```
aulemahal, line 24: Not sure why we would skip checks on "_conversion" indicators? Usually we can trust xarray to do it anyway, but what if?  An invalid "tasmin" should imply an invalid "tas", and so on.

huard: Totally agree, but `missing` is intended for data reductions along time. In this case, we have a scalar operation (n -> n), so I assume there is no need for the missing function. This is a check I could add.

aulemahal: Euh, but isn't that what this PR adds? Missing methods for non-reducing indicators?

huard: No, it's missing methods for indicators reducing on entire series (no resampling): n -> 1

PR #434, opened 2020-05-05
Add function `at_least_n_valid` to check for missing values

### LINKED ISSUE #221: Enable missing checks for frequency analysis
### Description
Missing value checks applies to time series, but freq analysis does not return a time series. Rather, the time series is an intermediate result. 

We should create a special Indicator subclass able to support missing value checks within the freq analysis computation.

huard:
### Pull Request Checklist:
    - This PR fixes #221

Add new function to check for missing values.

No.

* **Other information**:


xclim/core/checks.py
```diff
@@ -383,3 +390,27 @@ def missing_pct(da, freq, tolerance, **indexer):
       A boolean array set to True if period has missing values.
     """
     return MissingPct(da, freq, **indexer)(tolerance=tolerance)
+
+
+def at_least_n_valid(da, freq, n, **indexer):
```
aulemahal, line 395: I think it would be better to have a matching name like `missing_????`, but I can't think of one so ok.

huard: @tlogan2000 Ideas ?

Issue #364, opened 2020-02-12
Make indices and indicators easier to maintain and parse


aulemahal:
# Description

I think the structure of xclim needs a major refactoring to better organise the indices and indicators and make it easier for automated tools to parse the available methods.

Right now we define indices that take care of the computation and of the units. Then we create an indicator that adds checks and metadata. A few of those indicators called wrapped versions of the indices, but mostly the relationships are 1-to-1.

The problem:
 - There is a lot of duplicated code for very similar algorithms.
 - Both `xclim.indices` and `xclim.atmos` are crowded and hard to maintain.

I suggest the following structure:
Indices: Base algorithms, with some options and general names and doc
Indicator: Wrapped partial of indices, setting some options and defaults (+ metadata)
Aliases : For creating indicator that  have different metadata and defaults for the same computation.

`Aliases` would simply update an existing Indicator with new metadata and new default values for parameters.

Ex:
Indice : `degree_days(temp, ref_temp, max_temp, min_temp, sign)`
Indicator : `heating_degree_days -> degree_days(tas, 17 degC, 17 degC, None, -1)`
Aliases : `freezing_index -> heating_degree_days (ref_temp = 0 degC)`

Caveats:
The parsed docstring would be less relevent to the actual indicator than right now, but there could be an easy way to modify this as we modify the metadata.
This will take time.

huard:
One thing we probably want to keep are the custom docstrings. I agree that computationally, heating and cooling degree days are similar, but ideally the docstrings would be different to explain what each function is doing, provide examples and references. 

So totally agree we could replace similar code with generic functions, but I would preserve the individual indices to keep those docstrings.

aulemahal:
Makes sense!

## Decorator idea
We could implement this idea by having a Indicator decorator?

Example:
```python
@create_indicator(
    identifier="blabla",
   **other metadata)
def heating_degree_days(tas, thresh='17 degC'):
    """Specific doc-string"""
    return degree_days(tas, ref_temp=thresh, max_temp=thresh, min_temp=None, sign=-1)
```

Pros:
- Easy to implement
- Signature is correctly seen by interactive editors (ipython, pycharm)
- Doc-string easily overrided

Cons:
- Contact with the compute function is lost
- Unit handling has to be done again

## Third level idea
[...]

PR #435, opened 2020-05-05
Cleaning up of Indicator class

huard:
### Pull Request Checklist:
    - See #364 

- The `cfprobe` method has been renamed to `cfcheck` and the `validate`  method has been renamed to `datacheck`. 
- The algorithm for missing values is identified by its registered name, e.g. "any", "pct", etc, along with its `missing_options`.
- More importantly, instantiating `Indicator` creates a new subclass on   the fly and stores it in a registry, allowing users to subclass existing indicators easily. 

At the moment, instantiating Indicator creates an instance:
`Indicator(...) -> Indicator object`

What this PR does is created an Indicator subclass and register it, then instantiate the subclass:
`Indicator(identifier='tmax', ...) -> TMAX subclass -> TMAX object`

`TMAX` is stored in `xclim.core.indicator.registry, so it's possible to do:

```
class TMAXPlus(xclim.core.indicator.registry["TMAX"]):
  bla bla
```
This also opens the door to passing `missing` and `validate` functions when instantiating a new Indicator, ie something like:
`new_ind = registry["TMAX"](identifier="new_ind", missing="pct")`

While this is a change for users maintaining custom indicators, normal users of xclim should see no change in behavior. 

Only for users defining their own indicators. 

* **Other information**:


xclim/core/base.py
```diff
@@ -0,0 +1,407 @@
...
+        return out
+
+    @staticmethod
+    def cfprobe(self, *das):
+        """Check input data compliance to expectations.
+        Warn of potential issues."""
+        return True
+
+    @staticmethod
+    def compute(self, *args, **kwds):
```
aulemahal, line 382: I'm having a hard time understanding how this works: you are declaring a `staticmethod` but still having a `self` argument? Isn't the point of static methods to be independent of the object instantiation?

xclim/core/base.py
```diff
@@ -0,0 +1,407 @@
...
+                key: {
+                    "default": p.default if p.default != p.empty else None,
+                    "desc": "",
+                }
+                for (key, p) in self._sig.parameters.items()
+            }
+        )
+
+        # if six.PY2:
+        #     out = walk_map(out, lambda x: x.decode('utf8') if isinstance(x, six.string_types) else x)
```
aulemahal, line 325: I believe this can be removed now!

xclim/core/base.py
```diff
@@ -0,0 +1,407 @@
...
+    context = "none"
+    title = ""
+    abstract = ""
+    keywords = ""
+    references = ""
+    comment = ""
+    notes = ""
+
+    # Subclass registry
+    registry = {}
```
aulemahal, line 119: Is there a reason for having the registry as an attribute of the Indicator class, instead of the module variable?

xclim/core/base.py
```diff
@@ -0,0 +1,407 @@
...
+
+    @staticmethod
+    def check_identifier(identifier):
+        """Verify that the identifier is a proper slug."""
+        if not re.match(r"^[-\w]+$", identifier):
+            warnings.warn(
+                "The identifier contains non-alphanumeric characters. It could make life "
+                "difficult for downstream software reusing this class.",
+                UserWarning,
+            )
```
aulemahal, line 272: I think I would be in favor of being stricter about this. What does @Zeitsperre think?

xclim/core/base.py
huard: Well, that's a bug.

xclim/core/base.py
huard: To differentiate descendants. So Indicator2D has its own registry. Open to changing this, or registering subclasses at the module level and class level.

aulemahal: Is the Indicator vs Indicator2D the only division? Honestly, I don't really understand the need of having a subclass of Indicator just to change `_nvar`, why not simply change it in the subclass itself? Or even leave it to the object instantiation. That's what I did in  `_conversion.py`, it didn't seem useful to create many subclasses only to change `_nvar`.

Also, if we restrict input arguments to DataArrays and parameters to numbers or strings, then we could easily infer `nvar` from the annotations of `compute`?

huard: This may have been done to facilitate wrapping in the WPS server. We can probably work something out without this distinction.

xclim/core/base.py
Zeitsperre: Yes, this could cause bigger problems down-the-line. Until we explicitly support things like emojis, let's not let them be passed along.

huard:
This is going well, but I'm concerned it could introduce subtle changes and new bugs. I suggest we push back the merge to the next release.

huard:
@aulemahal This should be reviewed and merged only after #485, #488 and #489 are merged. 

@Zeitsperre I suggest we release 0.18 before merging this to give us some time to track issues related to this change.

xclim/core/indicator.py
```diff
@@ -79,56 +161,184 @@ class Indicator:
...
+        if self.missing == "from_context" and self.missing_options is not None:
+            raise ValueError(
+                "Cannot set `missing_options` with `missing` method being from context."
             )
 
-        # Default value for `var_name` is the `identifier`.
-        if self.var_name == "":
+        # Validate hard-coded missing options
+        kls = MISSING_METHODS[self.missing]
+        self._missing = kls.execute
```
aulemahal, line 239: Same comment as in #485: if I understand correctly, this line will break if self.missing is 'from_context', no? It will invoke the empty `FromContext` class.

huard: This deserves to be tested, but in principle it should execute the missing algorithm set by options.

huard:
@MatProv Here is a PR based on our refactoring discussions.

aulemahal:
Otherwise, many thanks @huard, this is spotless and will help a lot!

xclim/indicators/atmos/_temperature.py
```diff
@@ -56,54 +54,39 @@
 # E.g. http://vocab.nerc.ac.uk/collection/P07/current/BHMHISG2/
 
 
-class Tas(Indicator):
+class Tas(Daily):
     """Class for univariate indices using mean daily temperature as the input."""
 
-    def cfprobe(self, tas):
-        cfchecks.check_valid(tas, "cell_methods", "**time: mean within days*")
+    @staticmethod
```
MatProv, line 60: Is it normal that Tas.cfcheck() is static while, for example, Tasmin.cfcheck() Tasmax.cfcheck() are not static?

xclim/indicators/land/_streamflow.py
```diff
@@ -12,44 +8,33 @@
 __all__ = ["base_flow_index", "freq_analysis", "stats", "fit", "doy_qmax", "doy_qmin"]
 
 
-class Streamflow(Indicator):
+class Streamflow(Daily):
```
MatProv, line 11: It can be a good idea to add the indicators inheritance changes to the changelog, if any regression later on. Also, any regression anticipated, specifically on this change? (applies to other places)

xclim/indicators/land/_streamflow.py
```diff
@@ -12,44 +8,33 @@
...
-    def missing(self, *args, **kwds):
-        """Return whether an output is considered missing or not."""
-        from functools import reduce
-
-        indexer = kwds["indexer"]
-        freq = kwds["freq"] or generic.default_freq(**indexer)
-
-        miss = (missing.missing_any(da, freq, **indexer) for da in args)
-        return reduce(np.logical_or, miss)
+    missing = "any"
```
MatProv, line 22: Are the `missing = "skip|any"` in indicators documented, to avoid interpretation?

xclim/indicators/land/_streamflow.py
```diff
@@ -12,44 +8,33 @@
...
 
 
 # Disable the missing value check because the output here is not a time series.
```
MatProv, line 25: Why this is not in docstring?

MatProv:
@huard I left some comments, simply ignore if not relevant. 
Other than that, it looks good to me, good work guys.

xclim/indicators/atmos/_temperature.py
huard: No. Fixed.

xclim/indicators/land/_streamflow.py
huard: Done.

xclim/indicators/land/_streamflow.py
huard: Added skip to indicator docstring. Rewrote the missing.py module docstring and included it in the docs.

xclim/indicators/land/_streamflow.py
huard: Replaced "any" by "at_least_n" and set `missing_options`. Fixed bug related to hard-coded missing options.

xclim/core/indicator.py
huard: `from_context` is the default value for the missing attribute. Added a test for this.

PR #437, opened 2020-05-06
Control xclim's behavior with xc.set_options

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Introduces a new `set_options` object that mirrors xarray's `set_options` system.
This allows for user control on xclim's behavior either globally or within a context. For now, I implement the following : 

1) additional metadata locales (simply moved the already existing mechanism)

2) Missing method control. A generic `checks.missing_default(da, freq, **indexer)` calls whatever method is registered in "check_missing". For methods with options, I created specific options so that this can be controlled. User can implement new methods and register them with the `xc.core.options.register_missing_method` decorator.

3) Validation checks strictness."validate_inputs" can be one of "log", "warn" or "raise". The idea is the following: check methods should be declared with the `xc.core.checks.check` decorator and are to throw a `xc.core.utils.ValidationError` whenever the check fails.  The decorator catches the error and either rethrows it, or writes the message to a warning, or to the log. 

One would set options globally with  
```python
xc.set_options(check_missing='wmo')
```
or within a context:
```python
with xc.set_options(missing_wmo={'nm':4, 'nc':4}):
    ... do stuff ...
```
The `metdata_locale` object is now merged into the `set_options`. To change the locales, one should now do:
```python
xc.set_options(metadata_locales=['fr'])
```


xclim/core/checks.py
```diff
@@ -25,28 +31,45 @@
 # I suggest we use `check` for weak checking, and `assert` for strong checking.
 # Weak checking would log problems in a log, while strong checking would raise an error.
 #
+# ANSWER to the above:
+#   The use of set_options with a switch for raising vs warning vs logging would render these terms useless.
+#
```
huard, line 36: Sounds good.

xclim/core/checks.py
```diff
@@ -383,3 +316,16 @@ def missing_pct(da, freq, tolerance, **indexer):
       A boolean array set to True if period has missing values.
     """
     return MissingPct(da, freq, **indexer)(tolerance=tolerance)
+
+
+def missing_default(da, freq, **indexer):
```
huard, line 321: Maybe rename to "missing_from_context" ?

huard: Also, is there a way to avoid special casing the arguments ?

xclim/core/checks.py
```diff
@@ -56,108 +79,15 @@ def assert_daily(var):
...
     if not var.time.to_pandas().is_monotonic_increasing:
-        raise ValueError("time index is not monotonically increasing.")
-
-
-def check_valid_temperature(var, units):
-    r"""Check that variable is air temperature."""
-
-    check_valid(var, "standard_name", "air_temperature")
-    check_valid(var, "units", units)
-    assert_daily(var)
```
huard, line 75: Why remove this ?

aulemahal: Hum. No solid reasons, before getting to this solution I began to write another that was not coherent with these "batch" checkers, which is why I removed them. But I can put them back, were they used elsewhere?  Sorry!

xclim/core/checks.py
aulemahal: You mean for the options like in "wmo" and "pct"? I didn't find any that fitted the `set_options` paradigm...

xclim/core/checks.py
huard: I mean, this could also do with some refactoring and automating, possibly using cf-checker, but yes, let's keep them around until we replace them.

xclim/core/checks.py
aulemahal: There could be a dict-like option "missing_options"  that takes dict-like arguments and update itself with them?
Ex: `xc.set_options(check_missing='pct', missing_options={'pct':{'tolerance': 0.04}})`
It's a lot of nesting, but it's more generic. 
Accordingly, the `register_missing_method` would take an `options` arg to register the available options and default values.

huard: I was thinking about the last option, and in fact it can inspect the function automatically.

aulemahal:
Ok, so in order to make everything pass with the new paradigm, I had to make xclim stricter. Now the checks we put in the `cf_probe` of Indicators now actually raise errors, which required some tweaking in a lot of tests and modifying the NRCAN testdata that was missing a `cell_methods` field. 

Major changes that we might want to discuss:

- Metadata checks, with `xclim.core.checks.check_valid` now accept posix-style wildcards (* , ?, [abc], [!abc]).  This is particularly useful for checking the `cell_methods` that could contain many things before the actual part we want to check. Am I right @Zeitsperre ?   In our case, it was needed for `tas` indicator where `tas` was created from `tasmin` and  `tasmax` with `atmos.tg`.

- `cf_probe` has two ways of being written : either the variables are passed by keywords, with the same name defined in the `compute` func which are hopefully standardized,  OR passed by position, when the signature-binding fails. Before it was positional only and this change was triggered because we had cf_probes with arguments in the wrong order.  @huard ?

- New way of registering missing method options:
```python
@register_missing_method("max_valid", validator=lambda n: n > 0)
def missing_max_valid(da, freq, n=1, **indexer):
    ... the code ...

# And change the options / method with:
with xc.set_options(check_missing="max_valid", missing_options={'max_valid': {'n': 54}}):
    ... some code ...
```

huard:
Question: Why do we need a validator inside `register_missing_method` ? Shouldn't such validation be done by the function itself ? Or you want to catch possible errors before the computation runs ?

huard:
You could have a `validate` method in `MissingBase` subclasses, and register the class instead of the function.

tests/conftest.py
```diff
@@ -120,7 +120,7 @@ def _q_series(values, start="1/1/2000"):
             coords=[coords],
             dims="time",
             name="q",
-            attrs={"standard_name": "dis", "units": "m3 s-1"},
+            attrs={"standard_name": "streamflow", "units": "m3 s-1"},
```
huard, line 123: I'm not sure this is the standard. I think it's `water_volume_transport_in_river_channel`. To be confirmed.

xclim/core/indicator.py
```diff
@@ -310,15 +315,15 @@ def missing(*args, **kwds):
         freq = kwds.get("freq")
         indexer = kwds.get("indexer") or {}
 
-        # We flag any period with missing data
-        miss = (checks.missing_any(da, freq, **indexer) for da in args)
+        # We flag periods according to the currently set missing data method
+        miss = (checks.missing_from_context(da, freq, **indexer) for da in args)
```
huard, line 319: I think this can be the default, but I suggest Indicators support the possibility of imposing a given missing function that could not be modified by the context. We can probably do this in the Indicator refactoring however.

aulemahal: That's a good idea. Then we could move what is currently in the `Indicator.missing` method to the `__call__` and set `Indicator.missing` to  `checks.missing_from_context`.

aulemahal:
The validator is there so that an option cannot be set if it is invalid, so that the error is raised at setting time, not at runtime, when it could be harder to trace the source of the error. But I like the idea of registering the classes instead, the current class/function duality is not necessary.

xclim/core/indicator.py
aulemahal: Or even, scratch `missing_from_context` and put this mechanism in the base `Indicator.missing`. Is there really a need for  `mìssing_from_context` outside of the Indicators? I guess a user would make the explicit choice of method and options...

huard: I don't see one, so yes, `missing_from_context` could be a method of `Indicator`. If `missing` is not explicitly passed to Indicator, then it's using the default. I suggest we make this change in the refactoring branch however.

aulemahal:
Changed the missing method registration to use the classes. Also added a complete example in the notebook. This same notebook should be extended to cover custom Indicator creation in that refactor.

docs/notebooks/options.ipynb
```diff
@@ -79,15 +81,56 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This method checks that there is less than `nm=5` invalid values in a month and that there are no consecutive runs of `nc>=4` invalid values. Thus, every month is now valid."
+    "This method checks that there is less than `nm=5` invalid values in a month and that there are no consecutive runs of `nc>=4` invalid values. Thus, every month is now valid.\n",
+    "\n",
+    "Finally, it is possible for advanced users to register their own method. Xclim's missing methods are in fact based on class instances. Thus, to create a custom missing class, one should implement a subclass based on `xclim.core.checks.MissingBase` and overriding at least the `is_missing` method. The method should take a `null` argument ad  a `count` argument.\n",
```
huard, line 86:
```suggestion
    "Finally, it is possible for advanced users to register their own method. Xclim's missing methods are in fact based on class instances. Thus, to create a custom missing class, one should implement a subclass based on `xclim.core.checks.MissingBase` and overriding at least the `is_missing` method. The method should take a `null` argument and  a `count` argument.\n",
```

docs/notebooks/options.ipynb
```diff
@@ -79,15 +81,56 @@
...
+    "\n",
+    "- `null` is a `DataArrayResample` instance of the resampled mask of invalid values in the input dataarray.\n",
+    "- `count` is the number of days in each resampled periods and any number of other keyword arguments. \n",
+    "\n",
+    "The `is_missing` method should return a boolean mask, at the same frequency as the indicator output (same as `count`), where True values are for elements that are considered missing and masked on the output.\n",
+    "\n",
+    "When registering the class with the `xclim.core.checks.register_missing_method` decorator, the keyword arguments will be registered a options for the missing method. One can also implement a `validate` static method that receives only those options and returns whether they should be considered valid or not."
```
huard, line 93:
```suggestion
    "When registering the class with the `xclim.core.checks.register_missing_method` decorator, the keyword arguments will be registered as options for the missing method. One can also implement a `validate` static method that receives only those options and returns whether they should be considered valid or not."
```

xclim/core/options.py
```diff
@@ -0,0 +1,169 @@
...
+
+def register_missing_method(name, validator=None):
+    def _register_missing_method(cls):
+        sig = signature(cls.is_missing)
+        opts = {
+            name: param.default
+            for name, param in sig.parameters.items()
+            if name not in ["self", "null", "count"]
+        }
+        MISSING_METHODS[name] = {"_cls": cls, **opts}
```
huard, line 91: I suggest `register` fills two dict: 
- {name: class}
- {name: dict(**missing arguments)}

Having a `_cls` item forces you to do all kinds of special casing afterward.

aulemahal:
I think this is not complete.
Maybe one thing to be noted : by splitting the options for the check "loudness" into "data_validation" and "cf_compliance", we added a level of control but also a possibility for confusion:
Checks using the option in "data_validatation" use the decorator `@datacheck` and those for CF standards : `@cfcheck`. The indicator class also has the  `cf_probe` vs `validate` distinction, but the two distinctions are not explicitly connected. 
I think this is ok, but maybe others think differently? @tlogan2000  @huard

huard:
Maybe this is something we can look at in #435 ? Ie clarify the intent of each Indicator method, assign a registry for each kind of check and make sure this is clearly documented.

PR #440, opened 2020-05-08
sdba - Fix parameters serialization - Fix DQM - Update black

### LINKED ISSUE #441: DetrendedQuantileMapping only applies the last correction factor
* xclim version: '0.16.1-beta' (master, as of 2020-05-08)
* Python version: 3.8.2
* Operating System: Linux

### Description
The figure below is the daily average : `plt.plot(data.groupby("time.dayofyear").mean("time").values)`. The black line is my OBS data, the blue line is my RAW data, the green line is the result of EmpiricalQQMapping and the orange line is the result of DetrendedQuantileMapping.

image

These are the options I used to create the orange line:
```
group_month_nowindow = xclim.sdba.utils.Grouper('time.month')
QQ_m_nw_ni = xclim.sdba.adjustment.DetrendedQuantileMapping(nquantiles=50, kind='+', group=group_month_nowindow, extrapolation='constant')
QQ_m_nw_ni.train(obs, hist)
qq_m_nw_ni = QQ_m_nw_ni.adjust(hist, degree=3)
```

And this is `QQ_m_nw_ni.ds.af.plot.contourf()`
image

The huge jump in May leads me to believe that the correction factor associated to the last quantile is what is being applied for the whole distribution.

### Additional Information
In `xclim.sdba.adjustment` and `_adjust` (line 230), I added the line with `+269` (which is roughly the annual average) :

```
sim = apply_correction(sim,
            broadcast(self.ds.scaling, sim, group=self.group, interp=self.interp),
            self.kind,
        )
        sim_fit = PolyDetrend(degree=degree, kind=self.kind).fit(sim)
        sim_detrended = sim_fit.detrend(sim)
        sim_detrended.values = sim_detrended.values + 269
        scen_detrended = super()._adjust(sim_detrended)
        scen = sim_fit.retrend(scen_detrended)
        scen.values = scen.values - 269  # I need to remove that +269
```
With that additional line, I don't have the bug. This leads me to believe that this is a discrepancy between `sim`, which is centered around 269°K, and `sim_detrended`, which is normalized and centered around 0.

aulemahal:
### Pull Request Checklist:
    - This PR fixes issues inserted by a previous PR and fixes #441 

- Changes the way parameters of Adjustment classes are added to the adjustment dataset. Everything is now flattened out and added as individual global attributes. Previously was a dict attribute, but that's not serializable when writing a new netCDF.

- Adds a `detrending` arg to `DetrendedQuantileMapping` to that a user has this flexibility. Also changes the default to linear, instead of the previously wrong constant (degree = 0).
The detrending can still be overridden in `adjust`, as is currently needed in some tests.

No

* **Other information**:

### DIFF
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -9,6 +9,7 @@ from xarray.core.dataarray import DataArray
 from .base import Grouper
 from .base import ParametrizableClass
 from .base import parse_group
+from .detrending import BaseDetrend
 from .detrending import PolyDetrend
 from .processing import normalize
 from .utils import ADDITIVE
@@ -98,10 +99,11 @@ class BaseAdjustment(ParametrizableClass):
         """Set the trained dataset from the passed variables.
 
         The trained dataset should at least have a `af` variable storing the adjustment factors.
-        Adds the adjustment parameters as the "adj_params" dictionary attribute.
+        Adds the adjustment parameters as the "adj_*" attributes.
         """
         self.ds = xr.Dataset(data_vars=kwargs)
-        self.ds.attrs["adj_params"] = self.parameters_to_json()
+        self.ds.attrs["adj_base_method"] = self.__class__.__name__
+        self.ds.attrs.update(self.flatten_parameters(prefix="adj"))
 
     def _train(self):
         raise NotImplementedError
@@ -197,7 +199,7 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     1. A scaling factor that would make the mean of `hist` match the mean of `ref` is computed.
     2. `ref` and `hist` are normalized by removing the group-wise mean.
     3. Adjustment factors are computed between the quantiles of the normalized `ref` and `hist`.
-    4. `sim` is corrected by the scaling factor then detrended using a linear fit.
+    4. `sim` is corrected by the scaling factor then detrended using some fit (linear by default).
     5. Values of detrended `sim` are matched to the corresponding quantiles of normalized `hist` and corrected accordingly.
     6. The trend is put back on the result.
 
@@ -213,6 +215,29 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     .. [Cannon2015] Cannon, A. J., Sobie, S. R., & Murdock, T. Q. (2015). Bias correction of GCM precipitation by quantile mapping: How well do methods preserve changes in quantiles and extremes? Journal of Climate, 28(17), 6938–6959. https://doi.org/10.1175/JCLI-D-14-00754.1
     """
 
+    @parse_group
+    def __init__(
+        self,
+        *,
+        nquantiles: int = 20,
+        kind: str = ADDITIVE,
+        interp: str = "nearest",
+        detrending: Union[int, BaseDetrend] = 1,
+        extrapolation: str = "constant",
+        group: Union[str, Grouper] = "time",
+    ):
+
+        super().__init__(
+            nquantiles=nquantiles,
+            kind=kind,
+            interp=interp,
+            extrapolation=extrapolation,
+            group=group,
+        )
+        if isinstance(detrending, int):
+            detrending = PolyDetrend(degree=detrending, kind=self.kind)
+        self._add_parameter("detrending", detrending)
+
     def _train(self, ref, hist):
         mu_ref = self.group.apply("mean", ref)
         mu_hist = self.group.apply("mean", hist)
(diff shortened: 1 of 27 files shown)


huard:
Not sure I understand. What was the problem with json.dumps ?

aulemahal:
The initial problem, actually, was that I forget to rename a call to "paramters_to_json" in adjustment.py, json.dumps was doing ok, but it created a string, which is not optimal for saving attributes on a dataarray, since everything is frozen in the same one. 
We can't simply put a dictionary as netCDF don't support it. Same thing for adding each attribute individually : some of them have unserializable types : e.g. other `ParametrizableClass` like the `group` arg.

Thus, not the simplest, but I felt like the most useful solution was to flatten the dict, recurse into ParametrizableClasses and cast the rest to strings.

huard:
My concern is about "round-tripping" the parameters. If you wanted to used the saved params to instantiate the class, it would be cumbersome to extract the right parameters and strip the prefix. Could you subclass the `__repr__` method of `ParameterizableClass` to support serialization ?

aulemahal:
Hum, not wrong... But in all cases, in order to save to a netCDF we would need to serialize things in way that would need a not yet existing function to read those back in from the file.

So, we could solve the issue you're raising with a new "from_attributes" class method of ParametrizableClass, that would also recurse and strip the prefixes?

Actually, I'm not sure I understand your `__repr__` idea. Are you suggesting we dump à json string as attribute and then load objects from it? This would also need some tweaking (maybe only a subclass of JSONDecoder to construct ParametrizableClasses).

In the first case, you get a more human-readable set of attributes, in the second a less clunky and more elegant serialization/parsing process.

huard:
I think you could do something like 
```python
class ParametrizableClass(dict):
    """Helper base class that sets as attribute every kwarg it receives in __init__.

    Only parameters passed in the init are considered as such and returned in the
    :py:meth:`ParametrizableClass.parameters` dictionary and the :py:meth:`ParametrizableCalss.parameters_to_json` method.
    """

    @property
    def parameters(self):
        """All parameters as a dictionary."""
        return self.copy()

    def __str__(self):
        return f"<{self.__class__.__name__}: {self.parameters}>"

    def __repr__(self):
        return f"{self.__class__.__name__}({self.parameters})"
```
and then
```python
>>> in_params = dict(anint=4, abool=True, astring="a string", adict={"key": "val"})
>>> obj = ParametrizableClass(**in_params)
>>> json.dumps(obj)
'{"anint": 4, "abool": true, "astring": "a string", "adict": {"key": "val"}}'
```

aulemahal:
Sorry for the many commits I had to test something, will squash this PR.
@Zeitsperre , so you know : I edited the setup.cfg file to flake8 ignores the "import not at top-level" errors, which are induced by the `importorskip` calls in all `sdba` tests.

@huard @RondeauG  This now fixes #441 by normalizing per group before detrending.  Still not the same as a grouped detrending though! Coming up.

xclim/sdba/adjustment.py
```diff
@@ -226,16 +247,40 @@ def _train(self, ref, hist):
...
-    def _adjust(self, sim, degree=0):
+    def _adjust(self, sim, degree=None):
         sim = apply_correction(
             sim,
             broadcast(self.ds.scaling, sim, group=self.group, interp=self.interp),
             self.kind,
         )
-        sim_fit = PolyDetrend(degree=degree, kind=self.kind).fit(sim)
-        sim_detrended = sim_fit.detrend(sim)
+        mu_sim = self.group.apply("mean", sim)
```
huard, line 256: Add comments about what is going on.

xclim/sdba/base.py
```diff
@@ -29,12 +29,21 @@ def __init__(self, **kwargs):
...
         """All parameters as a dictionary."""
         return {key: getattr(self, key) for key in self._parameter_names}
 
     def __str__(self):
+        return (
+            f"<{self.__class__.__name__}({json.dumps(self.parameters, default=str)})>"
+        )
+
+    def __repr__(self):
         return (
```
huard, line 47: Why not use repr to return a string that could be "eval"led  to return the actual object ?

xclim/sdba/base.py
```diff
@@ -29,12 +29,21 @@ def __init__(self, **kwargs):
             setattr(self, key, val)
             self._parameter_names.append(key)
 
+    def _add_parameter(self, name, value):
+        setattr(self, name, value)
+        self._parameter_names.append(name)
+
```
huard, line 35: Why not just subclass dict ? I have the feeling this is over-engineered.

aulemahal: That could be done... The idea was to allow subclasses to declare attributes that wouldn't be considered as "parameters". But I got into this problems with DQM where the sub-subclass adds a parameter to the subclass (and so `super().__init__()` is not enough...).

xclim/sdba/base.py
aulemahal: Well, that's what `__str__` does? Also, I kept the old output, but there is no need really. We could have only one of both.

aulemahal:
I'm not sure I understand what you're suggesting? 
I can close this PR and make 3 separate ones for : parameters serialization, DQM fix and pre-commit hooks update.

huard:
Sorry for the delay, but no, I don't want to waste your time. I suggest to merge this pretty much as is and then fine-tune the other stuff later.

aulemahal:
haha too late. Wasn't that long to do, and I'm waiting on other compilation elsewhere anyway.

aulemahal:
PR replaced by #443, #445 and #446.

PR #447, opened 2020-05-12
sdba - Fix DQM and QDM, refactor adjustment objects 

### LINKED ISSUE #441: DetrendedQuantileMapping only applies the last correction factor

aulemahal:
### Pull Request Checklist:
    - This PR fixes #441 and partially #448,

Fixes #441 by adding a grouped normalization before the detrending in DQM.

No, but the behavior is not exactly what a user could expect from reading Cannon et al. 2015, a note is added in the docs to clarify this.

* **Other information**:

### DIFF
--- a/xclim/sdba/__init__.py
+++ b/xclim/sdba/__init__.py
@@ -9,9 +9,9 @@ Adjustment algorithms all conform to the `train` - `adjust` scheme, formalized w
 Given a reference time series (ref), historical simulations (hist) and simulations to be adjusted (sim),
 any bias-adjustment method would be applied by first estimating the adjustment factors between the historical simulation and the observations series, and then applying these factors to `sim`, which could be a future simulation::
 
-  Adj = Adjustment(group="time.month", interp="linear")
+  Adj = Adjustment(group="time.month")
   Adj.train(ref, hist)
-  scen = Adj.adjust(sim)
+  scen = Adj.adjust(sim, interp="linear")
   Adj.ds.af  # adjustment factors.
 
 The `group` argument allows adjustment factors to be estimated independently for different periods: the full
@@ -37,12 +37,16 @@ A generic bias adjustment process is to layed out like:
 - preprocessing on `ref`, `hist` and `sim` (using methods in `xclim.sdba.processing` or `xclim.sdba.detrending`)
 - creating the adjustment object `Adj = Adjustment(**kwargs)` (from `xclim.sdba.adjustment`)
 - training `Adj.train(obs, sim)`
-- adjustment `scen = Adj.adjust(sim)`
+- adjustment with corresponding arguments `scen = Adj.adjust(sim, **kwargs)`
 - post-processing on `scen` (for example: re-trending)
 
-The train-adjust approach allows to inspect the trained adjustment object. The adjustment information is stored in
+The train-adjust approach allows to inspect the trained adjustment object. The trained information is stored in
 the underlying `Adj.ds` dataset and always has a `af` variable with the adjustment factors. Its layout and the
 other available variables vary between the different algorithm, refer to :ref:`bias-adjustment-algos`.
+
+Parameters needed by the training and the adjustment are saved to the `Adj.ds` dataset as a  `adj_params` attribute.
+Other parameters, those only needed by the adjustment are passed in the `adjust` call and written to the history attribute
+in the output scenario dataarray.
 
 Grouping
 ========
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -201,9 +201,12 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     1. A scaling factor that would make the mean of `hist` match the mean of `ref` is computed.
     2. `ref` and `hist` are normalized by removing the group-wise mean.
     3. Adjustment factors are computed between the quantiles of the normalized `ref` and `hist`.
-    4. `sim` is corrected by the scaling factor then detrended using a linear fit.
+    4. `sim` is corrected by the scaling factor, normalized by the group-wise mean and then detrended using a linear fit.*
     5. Values of detrended `sim` are matched to the corresponding quantiles of normalized `hist` and corrected accordingly.
-    6. The trend is put back on the result.
+    6. The group-wise mean and trend are put back on the result.*
+
+    * Steps 4 and 6 include a group-wise normalization to overcome the fact that detrending is not made group-wise.
+    A future release of xclim will have grouped detrending and not require this extra step any more.
 
     .. math::
 
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -9,7 +9,6 @@ from xarray.core.dataarray import DataArray
 from .base import Grouper
 from .base import Parametrizable
 from .base import parse_group
-from .detrending import BaseDetrend
 from .detrending import PolyDetrend
 from .processing import normalize
 from .utils import ADDITIVE
@@ -22,6 +21,7 @@ from .utils import interp_on_quantiles
 from .utils import invert
 from .utils import map_cdf
 from .utils import MULTIPLICATIVE
+from .utils import rank
 from xclim.core.calendar import get_calendar
 from xclim.core.formatting import update_history
 
@@ -75,6 +75,8 @@ class BaseAdjustment(Parametrizable):
         ----------
         sim : DataArray
           Time series to be bias-adjusted, usually a model output.
+        kwargs :
+          Algorithm-specific keyword arguments, see class doc.
         """
         if not self.__trained:
             raise ValueError("train() must be called before adjusting.")
@@ -93,8 +95,9 @@ class BaseAdjustment(Parametrizable):
                 stacklevel=4,
             )
         scen = self._adjust(sim, **kwargs)
+        params = ", ".join([f"{k}={repr(v)}" for k, v in kwargs.items()])
         scen.attrs["history"] = update_history(
-            f"Bias-adjusted with method {str(self)}", sim
+            f"Bias-adjusted with {str(self)}.adjust(sim, {params})", sim
         )
         return scen
 
@@ -129,16 +132,21 @@ class EmpiricalQuantileMapping(BaseAdjustment):
 
     Parameters
     ----------
+    At init:
+
     nquantiles : int
       The number of quantiles to use. Two endpoints at 1e-6 and 1 - 1e-6 will be added.
     kind : {'+', '*'}
       The adjustment kind, either additive or multiplicative.
-    interp : {'nearest', 'linear', 'cubic'}
-      The interpolation method to use then interpolating the adjustment factors.
-    extrapolation : {'constant', 'nan'}
-      The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details.
     group : Union[str, Grouper]
       The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
+
+    In adjust:
+
+    interp : {'nearest', 'linear', 'cubic'}
+      The interpolation method to use when interpolating the adjustment factors. Defaults to "nearset".
+    extrapolation : {'constant', 'nan'}
+      The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details. Defaults to "constant".
 
     References
     ----------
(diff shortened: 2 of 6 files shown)


xclim/sdba/adjustment.py
```diff
@@ -215,6 +217,29 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     .. [Cannon2015] Cannon, A. J., Sobie, S. R., & Murdock, T. Q. (2015). Bias correction of GCM precipitation by quantile mapping: How well do methods preserve changes in quantiles and extremes? Journal of Climate, 28(17), 6938–6959. https://doi.org/10.1175/JCLI-D-14-00754.1
     """
 
+    @parse_group
+    def __init__(
```
huard, line 221: This class is missing a Parameters section in the docstring.

xclim/sdba/adjustment.py
```diff
@@ -215,6 +217,29 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
...
+        super().__init__(
+            nquantiles=nquantiles,
+            kind=kind,
+            interp=interp,
+            extrapolation=extrapolation,
+            group=group,
+        )
+        if isinstance(detrending, int):
+            detrending = PolyDetrend(degree=detrending, kind=self.kind)
+        self["detrending"] = detrending
```
huard, line 241: A bit weird. Would use `self.detrending` instead.

huard: And why do we need this in the init ?

xclim/sdba/adjustment.py
```diff
@@ -228,16 +253,49 @@ def _train(self, ref, hist):
...
+                interp=self.interp,
+            ),
+            kind=self.kind,
+        )
+
+        # Find trend on sim (for debugging purpose, the detrending is overridable)
+        if degree is not None:
+            detrending = PolyDetrend(degree=degree, kind=self.kind)
+        else:
+            detrending = self.detrending
```
huard, line 283: It is very confusing that we can set detrending in the init, and degree in adjust. One API would be to accept any BaseDetrend subclass in adjust, and default to PolyDetrend when an `int` is given.

xclim/sdba/adjustment.py
```diff
@@ -228,16 +253,49 @@ def _train(self, ref, hist):
...
+
+        sim_fit = detrending.fit(sim_norm)
+        sim_detrended = sim_fit.detrend(sim_norm)
+
+        # Adjust
```
huard, line 288:
```suggestion
        # Adjust using `EmpiricalQuantileMapping.adjust`
```

xclim/sdba/adjustment.py
aulemahal: If I use `self.detrending` it won't be "registered" as a parameter, that was my point with flushing the  `__setattr__ = dict.__setitem__` part.
It is in the init so we can use  `super().__init__` without having to define a meaningless `**kwargs` in `EmpiricalQuantileMapping`.

xclim/sdba/adjustment.py
aulemahal: Hum I actually don't agree with having a `degree` parameter  in the `adjust` at all. I think this is only there as a debug workaround for one of the tests.

xclim/sdba/adjustment.py
aulemahal: True. It didn't need one before, but with `detrending` it now does.

xclim/sdba/adjustment.py
huard: Why ? Let's say you want to compare an adjustment with degree 1 and degree 2, at the moment, you'd have to do something like
```
adj['detrending'] = 2
adj.adjust(sim)
```

xclim/sdba/adjustment.py
huard: But why does it need to be initialized at all if it is an argument of adjust ?

xclim/sdba/adjustment.py
aulemahal: I was seeing the Adjustment objects as "frozen". You can only adjust with a trained object with one set of parameters. This ensures that, for example, the `adj_params` attribute gives a complete description of the adjustment. For changing the detrending degree I would suggest recreating a whole object and train it on the same ref and hist.

huard: What about storing the history with the call from adjust ?
```scen.attrs["history"] = update_history(
            f"Bias-adjusted with method {str(self)}.adjust(sim, **f{kwargs})", sim
)
```
I guess my point is that training is compute intensive, so we should make sure the training object can be reused whenever possible. Otherwise, it doesn't really make sense to split `train` and `adjust`.

aulemahal: We could do that. But to stay coherent, all adjust-only parameters should be passable through adjust, no? (namely: extrapolation and interp).
The train-adjust still makes sense as the adjust could be called on many realizations or models,  but you are right that it's less clear.

huard: Yes, now that these are done in the adjust stage (thanks to the quantile interpolation), these could be `adjust` parameters.

tlogan2000: Can you provide a clearer example of what analysis steps would be like under the different scenarios... i.e. frozen vs flexible adjust objects?

tlogan2000: > I guess my point is that training is compute intensive, so we should make sure the training object can be reused whenever possible. Otherwise, it doesn't really make sense to split `train` and `adjust`.

Aren't training results dependent on the detrend method chosen?  I.e. I don't see how you could change the poly degree without having to retrain your correction factors..

aulemahal:
Merged fixes for QDM in this PR to save time. Dask-arrays needed a new version of `rank` so I replicated the xarray one but with an `apply_ufunc` that allows for dask parallelization.

aulemahal:
My bad, the fix to `xclim.sdba.utils.broadcast` did not fix  #449 for all cases, so I am reverting it for now.
Also, the new rank util partially fixes #448. However, it is not a great use as long as 449 is not fixed...

xclim/sdba/__init__.py
```diff
@@ -37,13 +37,17 @@
 - preprocessing on `ref`, `hist` and `sim` (using methods in `xclim.sdba.processing` or `xclim.sdba.detrending`)
 - creating the adjustment object `Adj = Adjustment(**kwargs)` (from `xclim.sdba.adjustment`)
 - training `Adj.train(obs, sim)`
-- adjustment `scen = Adj.adjust(sim)`
+- adjustment with corresponding arguments `scen = Adj.adjust(sim, **kwargs)`
```
huard, line 40:
```suggestion
- adjustment `scen = Adj.adjust(sim, **kwargs)`
```

xclim/sdba/__init__.py
```diff
@@ -37,13 +37,17 @@
...
 - post-processing on `scen` (for example: re-trending)
 
-The train-adjust approach allows to inspect the trained adjustment object. The adjustment information is stored in
+The train-adjust approach allows to inspect the trained adjustment object. The trained information is stored in
```
huard, line 43:
```suggestion
The train-adjust approach allows to inspect the trained adjustment object. The training information is stored in
```

xclim/sdba/adjustment.py
```diff
@@ -127,17 +132,22 @@ class EmpiricalQuantileMapping(BaseAdjustment):
 
     Parameters
     ----------
+    At init:
```
huard, line 135:
```suggestion
    At instantiation:
```

xclim/sdba/adjustment.py
```diff
@@ -127,17 +132,22 @@ class EmpiricalQuantileMapping(BaseAdjustment):
...
     kind : {'+', '*'}
       The adjustment kind, either additive or multiplicative.
-    interp : {'nearest', 'linear', 'cubic'}
-      The interpolation method to use then interpolating the adjustment factors.
-    extrapolation : {'constant', 'nan'}
-      The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details.
     group : Union[str, Grouper]
       The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
 
+    In adjust:
```
huard, line 144:
```suggestion
    In adjustment:
```

xclim/sdba/adjustment.py
```diff
@@ -228,16 +253,46 @@ def _train(self, ref, hist):
...
+            broadcast(
+                invert(mu_sim, kind=self.kind), sim, group=self.group, interp=interp,
+            ),
+            kind=self.kind,
+        )
+
+        # Find trend on sim (for debugging purpose, the detrending is overridable)
+        if isinstance(detrend, int):
+            detrend = PolyDetrend(degree=detrend, kind=self.kind)
+
```
huard, line 279: Catch case where detrend is not an int.

xclim/sdba/adjustment.py
```diff
@@ -254,17 +309,35 @@ class QuantileDeltaMapping(EmpiricalQuantileMapping):
...
+      The adjustment kind, either additive or multiplicative.
+    group : Union[str, Grouper]
+      The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
+
+    In adjust:
+
+    interp : {'nearest', 'linear', 'cubic'}
+      The interpolation method to use when interpolating the adjustment factors. Defaults to "nearest".
+    extrapolation : {'constant', 'nan'}
+      The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details. Defaults to "constant".
```
huard, line 328: Maybe in a later PR think about a mechanism to have this docstring in `adjust` itself. Fine as it is at the moment.

xclim/sdba/utils.py
```diff
@@ -462,3 +464,47 @@ def _interp_quantiles_2D(newx, newg, oldx, oldy, oldg):
         dask="parallelized",
         output_dtypes=[np.float],
     )
+
+
+def rank(da, dim="time", pct=False):
```
huard, line 469: Can this be submitted as a PR to xarray ?

aulemahal: Maybe. I don't think I covered all xarray cases, but that shouldn't be difficult.

xclim/sdba/adjustment.py
aulemahal: I thought of it, we could simply patch the class doc to `adjust.__doc__` or in a more fancy way.

xclim/sdba/adjustment.py
aulemahal: Hum, but isn't that case covered by the "AttributeError" that would be raised for any objects other than BaseDetrend? I mean it you pass a string, it will raise AttributeError Object of type string has not attribute `fit` which is explicit enough for me!

huard: You're right.

PR #451, opened 2020-05-13
Anuclim and Updated Tox build configuration

### LINKED ISSUE #411: Implement ANUCLIM module &  indicators
Many users in forestry / biology / agriculture use ANUCLIM bioclimatic variables. I suggest creating a  submodule xclim.anuclim for these users

Many indices are already presnt but some will require coding .. Also most of the derived indices (seasonality, etc) here  work of monthly or seasonal datasets so might require expansion of checks etc

Chapter 6 describes indices.
https://fennerschool.anu.edu.au/files/anuclim61.pdf

---

The descriptions below assume that BIOCLIM is using the weekly time step (the default). If
the monthly time step is selected, monthly values are used when calculating these parameters.
The quarterly parameters are not aligned to any calendar quarters. BIOCLIM's definition of a
quarter is any 13 consecutive weeks, (or any consecutive 3 months if running with a monthly
time step). For example, the driest quarter will be the 13 consecutive weeks that are drier than
any other set of 13 consecutive weeks.

- [x] 1) Annual Mean Temperature
The mean of all the weekly mean temperatures. Each weekly mean temperature is the
mean of the weekly maximum and minimum temperatures over the whole year.
- [ ] 2) Mean Diurnal Range(Mean(period max-min))
The mean over the whole year of the weekly diurnal temperature ranges. Each weekly
diurnal range is the difference between that week's maximum and minimum temperature.
- [ ] 3) Isothermality 2/7
The mean diurnal range (parameter 2) divided by the Annual Temperature Range
(parameter 7).
- [x] 4) Temperature Seasonality (C of V)
The temperature Coefficient of Variation (C of V) is the standard deviation of the weekly
mean temperatures expressed as a percentage of the mean of those temperatures. For this
calculation, the mean in degrees Kelvin is used. This avoids the possibility o f having to
divide by zero, but it does mean that the values are usually quite small.
- [x] 5) Max Temperature of Warmest Period
The highest maximum temperature in all weeks of the year.
- [x] 6) Min Temperature of Coldest Period
The lowest minimum temperature in all weeks of the year.
- [ ] 7) Temperature Annual Range (5-6)
The difference between the Max Temperature of Warmest Period and the Min
Temperature of Coldest Period.
- [x] 8) Mean Temperature of Wettest Quarter
The wettest quarter of the year is determined (to the nearest week), and the mean
temperature of this period is calculated.
- [x] 9) Mean Temperature of Driest Quarter
The driest quarter of the year is determined (to the nearest week), and the mean
temperature of this period is calculated.
- [ ] 10) Mean Temperature of Warmest Quarter
The warmest quarter of the year is determined (to the nearest week), and the mean
temperature of this period is calculated. 
- [ ] 11) Mean Temperature of Coldest Quarter
The coldest quarter of the year is determined (to the nearest week), and the mean
temperature of this period is calculated.

...

tlogan2000:
### Pull Request Checklist:
    - This PR fixes #411

Adds functions for a subset anuclim bioclimatic indicators  (those based on temp and precipitation

No

* **Other information**:
Docs strings have been added but we could possibly add an automodule .rst similar to ICCLIM?


xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+    Parameters
+    ----------
+    tasmin : xarray.DataArray
+      average daily minimum temperature [℃] or [K] at daily, weekly, or monthly frequency
+
+    tasmax : xarray.DataArray
+      average daily maximum temperature [℃] or [K] at daily, weekly, or monthly frequency
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 46:
```suggestion
    Parameters
    ----------
    tasmin : xarray.DataArray
      average daily minimum temperature [℃] or [K] at daily, weekly, or monthly frequency
    tasmax : xarray.DataArray
      average daily maximum temperature [℃] or [K] at daily, weekly, or monthly frequency
    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+    seas.attrs["units"] = "%"
+    return seas
+
+
+@declare_units("percent", pr="[precipitation]")
+def precip_seasonality(pr: xarray.DataArray,):
+    r""" ANUCLIM Precipitation Seasonality (C of V)
+    The annual precipitation Coefficient of Variation (C of V) expressed in percent. Calculated as the standard deviation
+    of precipitation values for a given year expressed as a percentage of the mean of those values.
+    See : https://fennerschool.anu.edu.au/files/anuclim61.pdf ch. 6
```
Zeitsperre, line 129: Wouldn't it be better to put outside references under `Notes`?

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+      Mean temperature [℃] or [K] at daily, weekly, or monthly frequency
+
+    op : str
+        Operation to perform :  'warmest' calculate warmest quarter ; 'coldest' calculate coldest quarter
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 199:
```suggestion
    tas : xarray.DataArray
      Mean temperature [℃] or [K] at daily, weekly, or monthly frequency.
    op : str
      Operation to perform :  'warmest' calculate warmest quarter ; 'coldest' calculate coldest quarter.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.
    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+    Notes
+    -----
+    According to the ANUCLIM user-guide https://fennerschool.anu.edu.au/files/anuclim61.pdf (ch. 6), input
+    values should be at a weekly (or monthly) frequency.  However, the xclim.indices implementation here will calculate
+    the result with input data with daily frequency as well. As such weekly or monthly input values, if desired,
+    should be calculated prior to calling the function.
+
+    """
+    # determine input data frequency
+    # determine input data frequency
```
Zeitsperre, line 226:
```suggestion
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+        return out
+
+
+@declare_units("[temperature]", tas="[temperature]", pr="[precipitation]")
+def tg_mean_wetdry_quarter(
+    tas: xarray.DataArray,
+    pr: xarray.DataArray,
+    op: str = None,
+    input_freq: str = None,
+    freq="YS",
```
Zeitsperre, line 264:
```suggestion
    freq: str = "YS",
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+      Total precipitation rate at daily, weekly, or monthly frequency
+
+    op : str
+        Operation to perform :  'wettest' calculate for the wettest quarter ; 'dryest' calculate for the dryest quarter
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 286:
```suggestion
    tas : xarray.DataArray
      Mean temperature [℃] or [K] at daily, weekly, or monthly frequency.
    pr : xarray.DataArray
      Total precipitation rate at daily, weekly, or monthly frequency.
    op : str
      Operation to perform :  'wettest' calculate for the wettest quarter ; 'dryest' calculate for the dryest quarter.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.

    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+      Total precipitation rate at daily, weekly, or monthly frequency
+
+    op : str
+        Operation to perform :  'wettest' calculate wettest quarter ; 'dryest' calculate dryest quarter
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 360:
```suggestion
    pr : xarray.DataArray
      Total precipitation rate at daily, weekly, or monthly frequency.
    op : str
      Operation to perform :  'wettest' calculate wettest quarter ; 'dryest' calculate dryest quarter.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.
    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+
+    Returns
+    -------
+    xarray.DataArray
+       Total precipitation values of the wettest/dryest quarter of each year.
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 368:
```suggestion
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+      Mean temperature [℃] or [K] at daily, weekly, or monthly frequency
+
+    op : str
+        Operation to perform: 'warmest' calculate for the warmest quarter ; 'coldest' calculate for the coldest quarter
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 448:
```suggestion
    pr : xarray.DataArray
      Total precipitation rate at daily, weekly, or monthly frequency.
    tas : xarray.DataArray
      Mean temperature [℃] or [K] at daily, weekly, or monthly frequency.
    op : str
      Operation to perform: 'warmest' calculate for the warmest quarter ; 'coldest' calculate for the coldest quarter.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.

    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+    Parameters
+    ----------
+    pr : xarray.DataArray
+        Total precipitation flux [mm d-1], [mm week-1], [mm month-1] or similar
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 515:
```suggestion
    Parameters
    ----------
    pr : xarray.DataArray
      Total precipitation flux [mm d-1], [mm week-1], [mm month-1] or similar.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.
    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+        Total precipitation flux [mm d-1], [mm week-1], [mm month-1] or similar
+
+    op : str
+        Operation to perform :  'wettest' calculate wettest period ; 'dryest' calculate dryest period
+
+    input_freq : str
+        Input data time frequency - One of 'daily', 'weekly' or 'monthly'
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
```
Zeitsperre, line 562:
```suggestion
    pr : xarray.DataArray
      Total precipitation flux [mm d-1], [mm week-1], [mm month-1] or similar.
    op : str
      Operation to perform :  'wettest' calculate wettest period ; 'dryest' calculate dryest period.
    input_freq : str
      Input data time frequency - One of 'daily', 'weekly' or 'monthly'.
    freq : str
      Resampling frequency; Defaults to "YS".
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,624 @@
...
+    tasmax : xarray.DataArray
+      average daily maximum temperature [℃] or [K] at daily, weekly, or monthly frequency
+
+    freq : str
+      Resampling frequency; Defaults to "YS".
+
+    Returns
+    -------
+    xarray.DataArray
+      The isothermailty value expressed as a percent.
```
huard, line 51:
```suggestion
      The isothermality value expressed as a percent.
```

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,599 @@
...
+            .resample(time=freq)
+            .map(_get_from_other_extreme, args=("pr", "tas", np_op))
+        )
+        out.attrs = pr.attrs
+        out.attrs["units"] = "mm"
+        return out
+
+
+@declare_units("mm", pr="[precipitation]")
+def prcptot(pr: xarray.DataArray, input_freq: str = None, freq: str = "YS"):
```
aulemahal, line 483: So this does the same as `precip_accumulation` but with the `input_freq` flag added? I'm not sure it's a good idea to add  it to the `xclim.indices` namespace...  We could maybe add the `input_freq` management to the existing  precip_accumulation`? Or rename this to remove confusion between the two? Or forbid `input_freq="daily"`?

aulemahal:
1) There was a bug in `_get_from_other_extreme` that I fixed. Also, added tests using dask to `TestTempWetDryPrecipWarmColdQuarter` . However, we still have problems : when `input_freq='daily'` and the data is stored in a dask array, the preliminary resampling operation will create small chunks along 'time' and the `resample` will fail as the chunks are smaller than half the window size... I skipped those tests for now, as the user can easily do the resampling and rechunk before.

2) Shouldn't we use 'D', 'W' and 'M' for `input_freq`? One day, we might use a `xr.infer_freq` to fill this field automatically and that function will return one of those. And we could test with : `input_freq.startswith('D')`.

xclim/indices/_anuclim.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+
+    if op == "wettest":
+        return pr.resample(time=freq).max(dim="time", keep_attrs=True)
+    elif op == "dryest":
+        return pr.resample(time=freq).min(dim="time", keep_attrs=True)
+    else:
+        raise NotImplementedError(
+            f'Unknown operation "{op}" ; op parameter but be one of "wettest" or "dryest"'
+        )
```
Zeitsperre, line 571:
```suggestion
    raise NotImplementedError(
        f'Unknown operation "{op}" ; op parameter but be one of "wettest" or "dryest"'
    )
```

Zeitsperre:
>     2. Shouldn't we use 'D', 'W' and 'M' for `input_freq`? One day, we might use a `xr.infer_freq` to fill this field automatically and that function will return one of those. And we could test with : `input_freq.startswith('D')`.

It would be good if we followed xarray/pandas-like time_freq conventions as closely as possible. I wonder if there are any helper functions from those libraries that we can lean on...?

tlogan2000:
> > 1. Shouldn't we use 'D', 'W' and 'M' for `input_freq`? One day, we might use a `xr.infer_freq` to fill this field automatically and that function will return one of those. And we could test with : `input_freq.startswith('D')`.
> 
> It would be good if we followed xarray/pandas-like time_freq conventions as closely as possible. I wonder if there are any helper functions from those libraries that we can lean on...?

ok. good call.  I think eventually we would like to see a version where input_freq isn't event there and we get the frequency via xr.infer_freq.  However, it currently doesn't support non-standard calendars so this was a work-around

xclim/indices/_anuclim.py
tlogan2000: I agree not too sure what to do here really.... I originally wanted to just do a wrapped partial of precip_accum in the anuclim module but the return applies a ```pint_multiply(pr, 1 * units.day, "mm")```  so if input data is weekly for example it ends up dividing by 7  ...   I thought it was weird to add support for other frequencies to a single index of our 'standard' library (i.e. non anuplin indices)..  `

I can remove daily support but would there a way for me to wrap two separate calls into the "P12_AnnualPrecip" function in the mapped module..  right now it does this ``"P12_AnnualPrecip": wrapped_partial(indices.prcptot, freq="YS")``   would need a lambda if else kind of thing??

Zeitsperre:
@tlogan2000 Would you be upset if I did a find and replace for `dryest` to `driest` ?

tlogan2000:
> @tlogan2000 Would you be upset if I did a find and replace for `dryest` to `driest` ?

haha. ok with me.  I had actually coded driest everywhere while @aulemahal had 'dryest'.  I went with 'dryest' as I was not sure what non-native english speakers would be expecting and personally it was  either/or

Zeitsperre:
I had the impression that `dryest` was an out-west cowboy colloquial spelling, haha.

tlogan2000:
> I had the impression that `dryest` was an out-west cowboy colloquial spelling, haha.

Websters doesn't seem to have 'dryest' so I guess driest would be the most correct

tlogan2000:
> 1. There was a bug in `_get_from_other_extreme` that I fixed, my bad. Also, added tests using dask to `TestTempWetDryPrecipWarmColdQuarter` . However, we still have problems : when `input_freq='daily'` and the data is stored in a dask array, the preliminary resampling operation will create small chunks along 'time' and the `resample` will fail as the chunks are smaller than half the window size... I skipped those tests for now, as the user can easily do the resampling and rechunk before.


Would this be a problem for all the 'wetdry'/'warmcold' quarter indices?  i.e. anything with a rolling or is it specific to those using the sub function?

xclim/indices/_anuclim.py
aulemahal: Maybe you could move the input_freq-dependent part of the code to the `__build_anuclim` function, so its available to the Anuclim module but never made public?

aulemahal:
> Would this be a problem for all the 'wetdry'/'warmcold' quarter indices? i.e. anything with a rolling or is it specific to those using the sub function?

It's specific to the indices resampling daily to weekly and then calling a `xr.apply_ufunc`, so right now only the cross-indices (tas of driest, pr of warmest and so on). 

> out-west cowboy colloquial spelling
Sadly, it was only keb engrish.

xclim/indices/_anuclim.py
tlogan2000: > Maybe you could move the input_freq-dependent part of the code to the `__build_anuclim` function, so its available to the Anuclim module but never made public?

Euhh, Might need some help there ...  :/   Sounds good but not sure how to implement... 

Would this mean the _anucplin.prcptot() would be removed? (Great if possible) and we could somehow shoehorn precip_accum to do what we want ?

huard:
In `tg_mean_warmcold_quarter`, if `input_freq` is daily, why not set `wind` to 13*7 ?

huard:
Also, the dosctring should probably mention that the `freq` should include an offset to avoid chopping the warmest quarter in 2 (e.g. AS-JUL) if you're in the SH.

tlogan2000:
> Also, the dosctring should probably mention that the `freq` should include an offset to avoid chopping the warmest quarter in 2 (e.g. AS-JUL) if you're in the SH.

Right.  This will affect my mapping...  I mapped most as partial with freq='YS'.  ..  Might therfore need to make sure that input freq is annual for some indices (seasonality at the very least)

tlogan2000:
> In `tg_mean_warmcold_quarter`, if `input_freq` is daily, why not set `wind` to 13*7 ?

This would be fine as well.  Initial thought was to try reproduce anuclim output as closely as possible but maybe we don't want to force ourselves to accept their 'weekly' input limitation

huard:
So I wrote a `_quarter` function that could be reused in a couple of indices. I tested it on two of them, I'll let you decide if you want to use that or not. If we decide to change the behavior, then there is only one place it would change.

huard:
I think you could simplify the logic by using "min", "max" as the `op`.

aulemahal:
Woups! We implemented different methods at the same time. But they were easily merged. I moved the "from_other" lazy indexing machinery to `xclim.indices.run_length.lazy_indexing` that can now accept ND inputs. (and to do so, changed `_np_argops` to `_xr_argops`.  For reference, the new way was 1.5x faster on my machine. 
Also, propagated your fix (@huard ) to `prcptot_warmcold_quarter`.

huard:
@aulemahal Cyclical import issue possibly due to: 
```
xclim/sdba/__init__.py:1:0: R0401: Cyclic import (xclim -> xclim.core.options -> xclim.locales -> xclim.core.formatting) (cyclic-import)
```

PR #460, opened 2020-05-25
Add indice, run length helper and tests

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Adds `first_day_below` indice and indicator. Also `run_length.first_run_after_date`.

Nein

* **Other information**:
It will be mainly used to compute the first frost day. I did not name it `first_frost_day` so its more general. However, that makes it in contradiction with `last_frost_day`...

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.17.x (2020-05-15)
 -------------------
+* Added indicator `first_day_below` and run length helper `first_run_after_date`
 * Added support for operations on dimensionless variables (`units = '1'`)
 * Moved `xclim.locales` to `xclim.core.locales` in a batch of internal changes aimed to removed most potential cyclic imports cases.
 * Missing checks and input validation refactored with addition of custom missing class registration (`xclim.core.checks.register_missing_method`) and simple validation method decorator (`xclim.core.checks.check`).
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,10 +1,12 @@
 =======
 History
 =======
-
-0.17.x (2020-05-15)
--------------------
+0.18.x
+------
 * Added indicator `first_day_below` and run length helper `first_run_after_date`
+
+0.17.0 (2020-05-15)
+-------------------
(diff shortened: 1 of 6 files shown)


aulemahal:
The difference is that it doesn't check for a preexisting run. Temperature doesn't have to be above thresh before "date". It's literally a `first_run` with data before the given date masked. I tried to use `run_end_after_date` with the opposite condition, as you suggest), but in addition to my first point, there was also the problem that `run_end_after_date` (or `growing_season_end` will return the last time if the run doesn't end within the timeseries.

HISTORY.rst
```diff
@@ -4,6 +4,7 @@ History
 
 0.17.x (2020-05-15)
 -------------------
+* Added indicator `first_day_below` and run length helper `first_run_after_date`
```
Zeitsperre, line 7:
```suggestion
0.18.x
------
* Added indicator `first_day_below` and run length helper `first_run_after_date`

0.17.0 (2020-05-15)
-------------------
```

xclim/indices/run_length.py
```diff
@@ -388,6 +388,46 @@ def run_end_after_date(
     return end.where(beg.notnull())
 
 
+def first_run_after_date(
+    da: xr.DataArray,
+    window: int,
+    date: str = "07-01",
+    dim: str = "time",
+    coord: str = "dayofyear",
```
Zeitsperre, line 396:
```suggestion
    coord: Opitonal[str] = "dayofyear",
```

HISTORY.rst
aulemahal: woups, thanks!

xclim/indices/run_length.py
aulemahal: haha you sure?

Zeitsperre: It's marked optional in the docstring. Both can't be right?

aulemahal: oui oui you're right, I meant for the "opitonal" ;)

Zeitsperre: Oof. That's embarrassing.

PR #461, opened 2020-05-25
Use convert_calendar in ensemble alignment

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Instead of implicitly doing the calendar conversion, `ensembles.create_ensemble` now has an explicit option to control the output calendar. Control is minimal: only the calendar name can be modified and the default options of  `convert_calendar` are used.

In fact, I broke `create_ensemble` in the last PR that modified the alignment process, but didn't realize until now. '360_day' calendar were broken because of the 30th of Feb. This fix has the same default behaviour than the old code and adds some control.

No, the default for the `calendar` argument is `default`. 

* **Other information**:

### DIFF
--- a/xclim/ensembles.py
+++ b/xclim/ensembles.py
@@ -264,6 +264,9 @@ def _ens_align_datasets(
     resample_freq : Optional[str]
       If the members of the ensemble have the same frequency but not the same offset, they cannot be properly aligned.
       If resample_freq is set, the time coordinate of each members will be modified to fit this frequency.
+    calendar : str
+      The calendar of the time coordinate of the ensemble. For conversions involving '360_day', the align_on='date' option is used.
+      See `xclim.core.calendar.convert_calendar`. 'default' is the standard calendar using np.datetime64 objects.
     xr_kwargs :
       Any keyword arguments to be given to xarray when opening the files.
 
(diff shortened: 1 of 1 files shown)


tlogan2000:
@aulemahal If you are working on ensembles.py I recently had the idea of implementing a `add_perc_as_dim`  flag.   Default would be `False` and would run as currently implemented creating a new variable for each percentile.  If ``True`` the output would be a single dataarray with added 'percentile' dim ( coords == input percntiles).  It should be somewhat 'cleaner' if users want a larger number of percentiles.   I think we could calculate all percentiles at once even instead of looping over values as we do currently?

We could make this a separate issue I suppose

@huard opinion on a 'add_perc_as_dim' option?

aulemahal:
I think it's a good idea! I would also suggest to have this in a separate PR/issue. Shouldn't be difficult to implement. Also, while doing this, we should make ensemble_percentiles work on DataArrays too, the current code is unneedingly dataset-only.

tlogan2000:
> I think it's a good idea! I would also suggest to have this in a separate PR/issue. Shouldn't be difficult to implement. Also, while doing this, we should make ensemble_percentiles work on DataArrays too, the current code is unneedingly dataset-only.

Sounds good to me .. can you open an new issue?  I will try to take a look at this PR in the meantime.  Seems fairly straightforward

huard:
I would suggest reviewing the API for this. It seems to me the *natural* xarray approach would be to always have this new percentile dimension. Then, when writing to file, you could have a `split` method to create individual files for each percentile value.

PR #463, opened 2020-05-25
Ensemble percentiles as dim

### LINKED ISSUE #462: Ensemble percentiles as dimension instead of variables
Currently, `xclim.ensembles.ensemble_percentiles` takes in a dataset and returns a new one with one variable per percentile and data variable. A more "xarray-like" behaviour would be to replace the `realization` dimension by the `percentile` dimension. In fact, `ensemble_percentiles` is more or less a xr-wrapper and workaround because `xr.quantiles` doesn't work on dask arrays. 

Moreover, it would be nice if `ensemble_percentiles` accepted DataArrays directly.

Discussion points:
- Should we keep the old behaviour (creating new variables)?
- If yes, should it be the default?

aulemahal:
### Pull Request Checklist:
    - This PR fixes #462 

Changes the behaviour of `ensembles.ensemble_percentiles` to concatenate the percentiles along a `percentiles` dimension by default. The old behaviour can be obtained by specifying `split=True`. 

Also, the method now accepts DataArrays. (In fact when passing a dataset, it simply loops on the variables with `realization` as a dim.)

Yes!

* **Other information**:

### DIFF
--- a/xclim/ensembles.py
+++ b/xclim/ensembles.py
@@ -206,6 +206,7 @@ def ensemble_percentiles(
                 if "realization" in da.dims
             ]
         )
+        out.attrs.update(ens.attrs)
         out.attrs["history"] = update_history(
             f"Computation of the percentiles on {ens.realization.size} ensemble members.",
             ens,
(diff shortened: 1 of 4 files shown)


xclim/ensembles.py
```diff
@@ -190,58 +196,78 @@ def ensemble_percentiles(
...
-                input_core_dims=[["realization"]],
-                output_core_dims=[[]],
-                keep_attrs=True,
-                kwargs=dict(p=p),
-                dask="parallelized",
-                output_dtypes=[ens[v].dtype],
+            ens = ens.chunk({"realization": -1})
+
+    percs = []
+    for p in values:
```
tlogan2000, line 239: line 239 : Would simply calling _calc_perc on all input values at once work?  i.e. instead of looping.  numpy.percentile / nanpercentile will accept tuples or list I believe.  
It might be faster as percentile calculation will do some sorting function multiple time instead of once??
We can split if desired but would also avoid the concat steps etc etc

e..g
replace kwargs with all input vals??
```
kwargs=dict(p=values),

```

PR #467, opened 2020-05-28
sdba - Grouped detrending

### LINKED ISSUE #442: sdba - Grouped Detrending
We need to implement grouped detrending into sdba.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #442 

All detrending is now made using the `Grouper` class of sdba. The BaseDetrend exposed methods call `group.apply()` with the `_fit` and `_get_trend ` methods of the subclasses.

Performance could be lower for simple detrendings. To code it as elegantly as I could, I didn't code shortcuts. When calling `fit`, the `_fit(da)` method returns a grouped dataset. This dataset is broadcasted back along the 'time' coordinate of the dataarray to detrend and then grouped again for the trend computation (which is a non-grouping grouped operation). For something as simple as `MeanDetrend` it's a lot of overhead...

Yes, in order to achieve this as elegantly as I could, I removed the resampling possibility of  `PolyDetrend`. It could be put back, but it's slightly more complicated. Watcha think @huard?

Results of DQM might be different between this new method and the previous one.

EDIT: Comment on the performance
* **Other information**:
Might not work with the current master of xarray : calling `polyfit` on an element from `groupby` could throw some errors that should be fixed with pydata/xarray#4099.


xclim/sdba/utils.py
```diff
@@ -195,7 +195,7 @@ def broadcast(
     """
     if sel is None:
         sel = {}
-
+    print(group)
```
huard, line 198: debugging leftovers

aulemahal: darn it

xclim/sdba/detrending.py
```diff
@@ -90,47 +126,33 @@ class PolyDetrend(BaseDetrend):
...
     def _fit(self, da, dim="time"):
-        if self.freq is not None:
-            da = da.resample(
-                time=self.freq, label="left", loffset=loffsets[self.freq]
-            ).mean()
-        self._fitds = da.polyfit(dim=dim, deg=self.degree, full=True)
+        return da.polyfit(dim=dim, deg=self.degree)
 
-    def _get_trend(self, da):
+    def _get_trend(self, grpd, dim="time"):
```
huard, line 149: I remember hard-coding `time` in a couple of places in the code. Probably worth writing a test with a `mytime` dimension to check everything work as expected.

aulemahal: Good idée

PR #472, opened 2020-06-02
Re-enable nbsphinx execute except for example.ipynb

### LINKED ISSUE #471: Example notebooks are neither executed nor tested
We took the decision a while ago to move part of the doc to notebooks and have them executed during the doc build, as to add them to the testing suite. Doing so, we added a git hook that strips the notebook outputs, easing the versioning. 

However, since "example.ipynb" is quite heavy and slow, it sometimes times out making it difficult to work on upgrading the docs and some previous PR disabled the notebook executing in sphinx's options. Currently, notebooks are empty on readthedocs.

I suggest we find a solution to have the workflow example be lighter and re-enable the notebook execution in sphinx. We could also add exceptions where needed to exclude this notebook from the nbstripout hook and sphinx execution.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #471 

This adds the following metdata flags to the `example.ipynb` notebook, preventing nbstripout from removing the information if it is present.
```
   "metadata": {
    "keep_output": true
   },
```
I also re-enabled `"automatic"` execution of the `nbsphinx_execute` option. This means it will only execute the notebook if there is not already output in the cells, therefore ignoring `example.ipynb`.

No.

* **Other information**:
This is primarily a bugfix, no need for documentation changes, IMO.

### DIFF
--- a/docs/notebooks/usage.ipynb
+++ b/docs/notebooks/usage.ipynb
@@ -132,7 +132,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "with xclim.set_options(missing_options={'pct': dict(tolerance=0.1)}):\n",
+    "with xclim.set_options(check_missing='pct', missing_options={'pct': dict(tolerance=0.1)}):\n",
     "    # Change the missing method to \"percent\", instead of the default \"any\"\n",
     "    # Set the tolerance to 10%, periods with more than 10% of missing data\n",
     "    #     in the input will be masked in the ouput.\n",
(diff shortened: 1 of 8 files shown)


docs/notebooks/usage.ipynb
```diff
@@ -132,7 +132,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "with xclim.set_options(check_missing='pct', missing_pct=0.1):\n",
+    "with xclim.set_options(missing_options={'pct': dict(tolerance=0.1)}):\n",
```
aulemahal, line 135:
```suggestion
    "with xclim.set_options(check_missing='pct', missing_options={'pct': dict(tolerance=0.1)}):\n",
```
Ach I forgot that earlier, sorry. Setting the option name is still needed.

PR #473, opened 2020-06-04
sdba - Fix rechunking in apply and make sdba easier to import

aulemahal:
### Pull Request Checklist:
    - This PR fixes a slack-raised issue

1) A previous PR added rechunking at the end of `Grouper.apply` calls. However, the default was to rechunk with `{dim: -1}` only if the input had single chunks along `dim`. This changes to use the same chunking in the output  as in the input. Obviously, this only applies to non-grouping operation.

2) Makes it easier to import xclim.sdba by making `adjustment`, `processing`, `detrending`, `utils` and `Grouper` directly in  `sdba`. I suggest using `from xclim import sdba`.

No

### DIFF
--- a/xclim/sdba/__init__.py
+++ b/xclim/sdba/__init__.py
@@ -63,7 +63,7 @@ except ImportError as err:
 else:
     del polyval
 
-from . import adjustment
+from .adjustment import *
 from .base import Grouper
 from . import detrending
 from . import processing
--- a/xclim/sdba/__init__.py
+++ b/xclim/sdba/__init__.py
@@ -63,6 +63,11 @@ except ImportError as err:
 else:
     del polyval
 
+from . import adjustment
+from .base import Grouper
+from . import detrending
+from . import processing
+from . import utils
 
 # TODO: ISIMIP ? Used for precip freq adjustment in biasCorrection.R
 # Hempel, S., Frieler, K., Warszawski, L., Schewe, J., & Piontek, F. (2013). A trend-preserving bias correction &ndash;
--- a/xclim/sdba/base.py
+++ b/xclim/sdba/base.py
@@ -245,7 +245,7 @@ class Grouper(Parametrizable):
                     if d.chunks and self.dim in d.dims
                 ]
                 or [[]],  # pass [[]] if no dataarrays have chunks so min doesnt fail
-                key=lambda d: len(d),
+                key=len,
             )
         else:
             grpd = self.group(da)
--- a/xclim/sdba/base.py
+++ b/xclim/sdba/base.py
@@ -19,7 +19,12 @@ class Parametrizable(dict):
     :py:meth:`Parametrizable.parameters` dictionary, the copy method and the class representation.
     """
 
-    __getattr__ = dict.__getitem__
+    def __getattr__(self, attr):
+        try:
+            return self.__getitem__(attr)
+        except KeyError as err:
+            # Raise the proper error type for getattr
+            raise AttributeError(*err.args)
 
     @property
     def parameters(self):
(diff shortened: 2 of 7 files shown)


xclim/sdba/base.py
```diff
@@ -227,9 +227,9 @@ def apply(
         -------
         DataArray or Dataset
           Attributes "group", "group_window" and "group_compute_dims" are added.
-          If the function did not reduce the array, its is sorted along the main dimension.
-          If the function did reduce the array and there is only one group, it is squeezed out of the output.
-
+          If the function did not reduce the array, its is sorted along the main dimension and chunking along it is conserved (apply performs a rechunking)
```
huard, line 230:
```suggestion
          If the function did not reduce the array, it is sorted along the main dimension and chunking along it is conserved (apply performs a rechunking)
```

xclim/sdba/base.py
```diff
@@ -227,9 +227,9 @@ def apply(
...
+          If the function reduces the array and there is only one group, it is squeezed out of the output.
+          If the function reduces the array, it is rechunked to have only 1 chunk along the new dimension.
```
huard, line 232: Clarify, a bit confused.

xclim/sdba/base.py
huard: Review syntax. `apply`

xclim/sdba/base.py
```diff
@@ -238,20 +238,19 @@ def apply(
...
+                    if d.chunks and self.dim in d.dims
+                ]
+                or [[]],  # pass [[]] if no dataarrays have chunks so min doesnt fail
+                key=len,
             )
         else:
             grpd = self.group(da)
-            dim_is_chunked = (
-                da.chunks is not None and len(da.chunks[da.get_axis_num(self.dim)]) > 1
+            dim_chunks = (
```
huard, line 252: Add explanation comment.

xclim/sdba/base.py
```diff
@@ -282,20 +281,20 @@ def apply(
         out.attrs["group_compute_dims"] = dims
         out.attrs["group_window"] = self.window
 
+        # On non reducing ops, drop the constructed window
+        if self.window > 1 and "window" in out.dims:
+            out = out.isel(window=self.window // 2, drop=True)
```
huard, line 286: Note that drop does not remove the dimension from the array shape. If this is what you want to do, you need to use `squeeze`.

xclim/sdba/__init__.py
```diff
@@ -63,6 +63,11 @@
 else:
     del polyval
 
+from . import adjustment
```
huard, line 66: What about putting the adjustment methods right into the sdba namespace ?

xclim/sdba/base.py
aulemahal: Hum... are you sure? I just tested and it doesn't sem to leave it there? I added a test for this.

xclim/sdba/__init__.py
aulemahal: right, makes more sense.

xclim/sdba/base.py
huard: On my end it drops the coordinates, but not the dimension.

aulemahal:
Kinda breaking change : Any BaseAdjustment object will raise a ValueError if ref, hist or sim contains multiple chunks along the main dimension (group.dim, usually "time"). This simply throws the error earlier and more explicitly. It would have failed anyway.

Technical change : `hasattr(obj, 'name')` returns false by catching an `AttributeError`, which made it useless on `Parametrizable` objects since we patched `__getattr__` by `__getitem__` which raises a `KeyError`... I had to made it a bit more complete than simple patching so `hasattr` works as expected.

PR #477, opened 2020-06-15
sdba - Optimizations for DQM and others

### LINKED ISSUE #455: sdba - improve the speed of DetrendedQuantileMapping
### Description
Doing a few preliminary tests **on a single pixel and without dask**, I obtained the following times : 

DetrendedQuantileMapping, groups='month', 150 years of daily data : ~15 seconds
DetrendedQuantileMapping, groups='day', window=31, 150 years of daily data : ~24 seconds

EmpiricalQuantileMapping, groups='month', 150 years of daily data : ~0.35 seconds
EmpiricalQuantileMapping, groups='day', window=31, 150 years of daily data : ~2.5 seconds

There seems to be a bottleneck in the detrending version. Discussions with @aulemahal point towards the normalization process and `sdba.utils.interp_on_quantiles`. I will continue investigating and post my findings here.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #455

- Added `norm_group` to `DetrendedQuantileMapping` so that a different grouping can be used in the normalization steps. `DQM.adjust` also has a new  `normalize_sim` arg. When True, `sim` is normalized along `norm_group` before being detrended along `group`. When False (default), `sim` is only detrended along `norm_group`. This is quite useful as the detrending process has poor performance compared to simple normalization, but only detrending by month creates artefacts at the month boundaries.  Optimal performance is obtained with: `norm_group='time.dayofyear', group='time.group', normalize_sim=True`.

- Added `BaseAdjustment.save_training()` and  `BaseDetrending.save_fit()` to save the training/fit datasets. This can help dask as adjustment processes create an enormous amount of small operations, especially with "time.dayofyear" grouping. Default is to save to a temporary file that is deleted when the adjustment object is garbage-collected. A persistent file can be created by passing `filename` or the temporary file's directory can be given with `tempdir`.

No

### DIFF
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -175,7 +175,7 @@ class BaseAdjustment(Parametrizable):
 
     def __del__(self):
         # Delete the training data file if it was saved to a temporary file.
-        if self.__ds_is_tempfile and hasattr(self, "_ds_filename"):
+        if self.__ds_is_tempfile and hasattr(self, "_ds_file"):
             self._ds_file.unlink()
 
     def _train(self):
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -303,6 +303,7 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
       The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details. Defaults to "constant".
     normalize_sim : bool
       If True, scaled sim is normalized using `norm_group` and then detrended using `group`.
+        The norm is broadcasted and added back on scen using `interp='nearest'`, ignoring the passed `interp`.
       If False, scaled sim is detrended using `norm_group`.
       This is useful on large datasets using dask, when `norm_group` is a very small division (e.g. 'time.dayofyear')
         because normalisation is a more efficient operation than detrending for similarly sized groups.
@@ -384,7 +385,7 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
         if normalize_sim:
             return apply_correction(
                 scen_anom,
-                broadcast(ds.norm, scen_anom, group=self.norm_group, interp=interp),
+                broadcast(ds.norm, scen_anom, group=self.norm_group, interp="nearest"),
                 self.kind,
             )
         return scen_anom
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -6,6 +6,7 @@ from typing import Optional
 from typing import Union
 from warnings import warn
 
+import h5py
 import numpy as np
 import xarray as xr
 from xarray.core.dataarray import DataArray
@@ -140,8 +141,11 @@ class BaseAdjustment(Parametrizable):
         self.ds = xr.Dataset(data_vars=kwargs)
         self.ds.attrs["adj_params"] = str(self)
 
-    def save_training(
-        self, filename: Optional[Union[Path, str]] = None, tempdir: Optional[str] = None
+    def _save_training(
+        self,
+        filename: Optional[Union[Path, str]] = None,
+        tempdir: Optional[str] = None,
+        use_hdf5_store: bool = False,
     ):
         """Save training data to a (temporary) file.
 
@@ -166,16 +170,34 @@ class BaseAdjustment(Parametrizable):
             os.close(fid)  # Passing file-like objects is too restrictive with xarray.
             self.__ds_is_tempfile = True  # So that the file is deleted when this instance is garbage collected
 
-        self._ds_file = Path(filename)
-        previous_chunking = self.ds.chunks  # Expected behavior is to conserve chunking
-        self.ds.to_netcdf(self._ds_file)
-        # chunks: non-dask data will return an empty set on ds.chunks, but that means 1 chunk for open_dataset
-        # `previous_chunking or None` returns None is ds.chunks was an empty set
-        self.ds = xr.open_dataset(self._ds_file, chunks=previous_chunking or None)
+        if not use_hdf5_store:
+            self._ds_file = Path(filename)
+            previous_chunking = (
+                self.ds.chunks
+            )  # Expected behavior is to conserve chunking
+            self.ds.to_netcdf(self._ds_file)
+            # chunks: non-dask data will return an empty set on ds.chunks, but that means 1 chunk for open_dataset
+            # `previous_chunking or None` returns None is ds.chunks was an empty set
+            self.ds = xr.open_dataset(self._ds_file, chunks=previous_chunking or None)
+        else:
+            self._ds_hdf5 = h5py.File(filename, mode="w")
+            new_ds = {}
+            for name, data in self.ds.data_vars.items():
+                d = self._ds_hdf5.require_dataset(
+                    name, shape=data.shape, dtype=data.dtype
+                )
+                new_ds[name] = data.copy(
+                    data=data.data.store(d, return_stored=True, compute=False)
+                )
+            self.ds = xr.Dataset(data_vars=new_ds)
 
     def __del__(self):
         # Delete the training data file if it was saved to a temporary file.
         if self.__ds_is_tempfile and hasattr(self, "_ds_file"):
+            if hasattr(self, "_ds_hdf5"):
+                self._ds_hdf5.close()
+            else:
+                self.ds.close()
             self._ds_file.unlink()
 
     def _train(self):
@@ -267,9 +289,10 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     The algorithm follows these steps, 1-3 being the 'train' and 4-6, the 'adjust' steps.
 
     1. A scaling factor that would make the mean of `hist` match the mean of `ref` is computed.
-    2. `ref` and `hist` are normalized by removing the group-wise mean.
+    2. `ref` and `hist` are normalized by removing the "dayofyear" mean.
     3. Adjustment factors are computed between the quantiles of the normalized `ref` and `hist`.
-    4. `sim` is corrected by the scaling factor, and detrended group-wise using a linear fit.
+    4. `sim` is corrected by the scaling factor, and either normalized by "dayofyear" and  detrended group-wise
+       or directly detrended per "dayofyear", using a linear fit (modifiable).
     5. Values of detrended `sim` are matched to the corresponding quantiles of normalized `hist` and corrected accordingly.
     6. The trend is put back on the result.
 
@@ -289,9 +312,10 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     kind : {'+', '*'}
       The adjustment kind, either additive or multiplicative.
     group : Union[str, Grouper]
-      The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
-    norm_group : Union[str, Grouper]
-      If given, the normalization steps are done using this group. Otherwise, they use the main `group`.
+      The grouping information use in the quantile mappgin process. See :py:class:`xclim.sdba.base.Grouper` for details.
+      the normalization step is always performed on each day of the year.
+    norm_window : 1
+      The window size used in the normalization grouping. Defaults to 1.
 
     In adjustment:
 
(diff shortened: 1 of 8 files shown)


xclim/sdba/adjustment.py
```diff
@@ -135,6 +140,44 @@ def _make_dataset(self, **kwargs):
         self.ds = xr.Dataset(data_vars=kwargs)
         self.ds.attrs["adj_params"] = str(self)
 
+    def save_training(
```
huard, line 143: Should this be a private method ? 
It's mysterious to me that writing to disk provides speed advantages. So I hesitate to make this a public method, and wonder if it should remain hidden (and called automatically behind the scene).

xclim/sdba/adjustment.py
```diff
@@ -135,6 +140,44 @@ def _make_dataset(self, **kwargs):
...
+        workload when needed.
+
+        Parameters
+        ----------
+        filename : Optional[Union[Path, str]]
+          Filename of the saved file. When given, the file is not considered "temporary"
+          and is not deleted when the Adjustment object is deleted by Python.
+        tempdir : Optional[str]
+          The path to a directory where to save the temporary file. Ignored if `filename`
+          is given.
```
huard, line 161: I would simplify that to one single argument (e.g. path). What are the benefits of splitting the directory and the filename ?

xclim/sdba/adjustment.py
```diff
@@ -247,6 +290,8 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
       The adjustment kind, either additive or multiplicative.
     group : Union[str, Grouper]
       The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
+    norm_group : Union[str, Grouper]
+      If given, the normalization steps are done using this group. Otherwise, they use the main `group`.
```
huard, line 294:
```suggestion
      Grouping for the normalization steps. If None, default to the main `group`.
```

xclim/sdba/adjustment.py
```diff
@@ -256,36 +301,76 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
       The method to use when detrending. If an int is passed, it is understood as a PolyDetrend (polynomial detrending) degree. Defaults to 1 (linear detrending)
     extrapolation : {'constant', 'nan'}
       The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details. Defaults to "constant".
+    normalize_sim : bool
+      If True, scaled sim is normalized using `norm_group` and then detrended using `group`.
+        The norm is broadcasted and added back on scen using `interp='nearest'`, ignoring the passed `interp`.
+      If False, scaled sim is detrended using `norm_group`.
+      This is useful on large datasets using dask, when `norm_group` is a very small division (e.g. 'time.dayofyear')
+        because normalisation is a more efficient operation than detrending for similarly sized groups.
```
huard, line 309: I think it is confusing that the detrending group is controlled by a parameter called `normalize_sim`.

xclim/sdba/base.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         window: int = 1,
         add_dims: Optional[Sequence[str]] = None,
         interp: Union[bool, str] = False,
+        manage_chunking: bool = True,
```
huard, line 53: This publicly exposes something that feels like an implementation detail. Does this need to be part of the public API ? 

There is a tradeoff between flexibility and accessibility. Each time we add another argument, we implicitly ask users  to understand what it does, and then decide which values they should use. At the end, it makes the software less accessible.

xclim/sdba/detrending.py
```diff
@@ -77,8 +89,43 @@ def retrend(self, da: xr.DataArray):
         trend = self.get_trend(da)
         return self._retrend(da, trend)
 
-    def _set_fitds(self, ds):
-        self.fit_ds = ds
+    def save_fit(
```
huard, line 92: Same comment as above. If the tmp file is meant to be deleted, should it really be in the public API ?

xclim/sdba/detrending.py
```diff
@@ -94,6 +141,11 @@ def _get_trend(self, grpd, dim="time"):
     def _fit(self, da):
         raise NotImplementedError
 
+    def __del__(self):
+        # Delete the training data file if it was saved to a temporary file.
+        if self.__ds_is_tempfile and hasattr(self, "_ds_file"):
+            self._ds_file.unlink()
```
huard, line 147: Should the file be `closed` ?

xclim/sdba/processing.py
```diff
@@ -173,9 +173,13 @@ def jitter_under_thresh(x: xr.DataArray, thresh: float):
 
 @parse_group
 def normalize(
-    x: xr.DataArray, *, group: Union[str, Grouper] = "time", kind: str = ADDITIVE
+    x: xr.DataArray,
+    *,
+    group: Union[str, Grouper] = "time",
+    kind: str = ADDITIVE,
+    return_norm: bool = False,
```
huard, line 185: Would this be cleaner to split this into two functions: `mean` and `normalize(x, group, kind, mean=None)`

xclim/sdba/adjustment.py
aulemahal: I'm not sure on how to automate the calling of this, but if possible that would be great!

aulemahal: Reason to why this is useful : dask struggles to produce a task graph when too many operations are needed. When using "time.dayofyear" grouping, there is one "getitem" op per day, so usually 365 doy * 30 years * spatial dims * 2 (ref + hist). This is a humongous amount of operations. My tests showed that dividing the work like this is actually faster than trying to do both adjust and train in one go. The split process was often faster than the task-graph-building of the single process.
Moreover, when trying different adjust for one train, not using  `save_training` makes the training computation re-execute itself each time. (2 / 3)

aulemahal: Reasons to why it is public: Because the alternative for a user that wants this option would be:
1) Call a private function `_save_training`
2) Do it by hand with : 
```
DQM.ds.to_netcdf('...')
DQM.__dict__['ds'] = xr.open_dataset('...', **)
```
Which is even worse quant à moi. (3 / 3)

xclim/sdba/adjustment.py
aulemahal: Simply for the option of making it a temporary file or not.
Example: on our server at ouranos it often makes more sense to save temporary files to `/exec/me/` than to `/tmp`.  But I still wanted the option to save to a persistent file for later inspection/re-use. And, maybe one day, a `Adjustment.from_training_ds('...')` class method.

xclim/sdba/adjustment.py
aulemahal: True. I will try to think of another more explicit name.

xclim/sdba/base.py
aulemahal: You are right, this one could go. I used it to test different scenarios, but it turns out that managing chunking myself only save a few seconds, nothing significative enough to merit a public option. I think there still could be use cases where optimal performance is obtained when `group.apply` does not rechunk its output, but am willing to sacrifice it if I don't find a better option.

xclim/sdba/detrending.py
aulemahal: Same answer as above. For medium and small size cases, this is not beneficial, thus why giving the option as a public method.

xclim/sdba/detrending.py
aulemahal: I believe datasets are closed when garbage-collected? Will check.

xclim/sdba/processing.py
aulemahal: I agree. More lines to write on the user sides, but more consistent api.

xclim/sdba/adjustment.py
huard: Got it. 
Can't you use `.compute()` instead of writing to disk ? 
Could you have a rule deciding whether to force computation or not ? If you have a default value that works well in most cases, it makes this improvement useful for all users, but no one has to care about this until they actually notice a performance issue. 
e.g. something like 

```
force_compute : None, int
  Size of dask operations above which a `compute` on the `train` method will be executed automatically. This is done to avoid the cost of building huge tasks graphs in dask. If None, do not force compute. 
```

aulemahal:
I agree with the spirit of your comments. I believe one strength of xclim.sdba is the use of xarray/dask and the tremendous performance improvement it can provide when compared to other options. For large workloads, dask does have its limits and the `save_training/fit` function is an example of optimization we can add to help dask while staying in kinda-high-level programming.

For the `norm_group/normalize_sim` duo, I am ambivalent. As @RondeauG  and I came to realize is that QM with monthly grouping is not quite as robust as we thought. However, this 'dayofyear' pre-normalization really helps. We could suggest to the user to build its own process, that is in fact what I do in the notebook of #478. 

D'ailleurs, I think that notebook adds explanations where the docstrings could be still unclear.

xclim/sdba/adjustment.py
aulemahal: Hum, I like this `force_compute` idea, will explore.
However, doesn't `compute` put everything in the ram? Our "test" case was tas ad mare usque ad mare, so maybe that's too much for normal ram? In fact, I am following here tip 2 of xarray's http://xarray.pydata.org/en/stable/dask.html#optimization-tips.

RondeauG:
Just to give my opinion on this matter:

As @aulemahal mentioned, we discovered that grouping per month is not reliable enough, especially for months that exhibit large variations (i.e. several degrees of warming between 1st and 30th of June). Therefore, for an adequate adjustment, it is essential to normalize the data using a smaller window (time.dayofyear does the job adequately, especially if using a window).

The issue is that for large datasets, dask is overloaded by those smaller groups (as @aulemahal mentioned in another comment).

These are the conclusions of my tests:

1) `norm_group=Grouper(group='time.dayofyear')` is essential. Any other option (except maybe weekofyear?) will simply create bad results whenever there are strong variations within the groupings decided by the user. Therefore, I don't think that `norm_group/normalize_sim` are good arguments to provide the user, as anything other than `norm_group=Grouper(group='time.dayofyear')` is wrong. A better parameter, if we want one, could be `norm_window`. The difference in results between `window=1` vs. `window=31` seemed rather small, while window=31 was much more costly to run.

2) In any case, using `norm_group=Grouper(group='time.dayofyear')` **is** costly. There's no way around that. We should warn users that if this is too much for `dask`, the domain should be cut in smaller pieces, or the user should use the more optimized (but complex) code provided in @aulemahal 's Notebook. Using his Notebook, I have roughly the same performance as before (~25 minutes to bias correct 15 000 grid cells).

huard:
I'm :-1: on asking users to cut the domain. I think it's our job to solve this problem. 
Is this relevant: https://docs.dask.org/en/latest/array-creation.html#intermediate-storage ?

RondeauG:
If we can solve the problem, I'm all for it. The link you provided mentions "It can be useful as a checkpoint for long running or error-prone computations.", so... maybe?

The main problem that we have is that with the `time.dayofyear` groups, `dask` gets overloaded while building its task tree ("arborescence de tâches") : the process never gets parallelised, the RAM usage just grows and grows until you either stop the code or it reaches the limit of the computer. This is why our solution was to write more often on disk, as this reduces the number of tasks that `dask` has to perform at once.

xclim/sdba/adjustment.py
huard: Yep. I don't know. If it only keeps the training result in memory, then it should not be too bad, right? In other words, if we have `input (DISK) -> chunk streams (RAM) -> train (RAM)`  it should be ok, right ? I mean, we can cap the RAM used by dask, no ? Which will limit the number of chunks it handles at once.

aulemahal: Well, with "time­.dayofyear" and, let's say, nquantiles > 30, the training result is actually *larger* than the input (given we use 30 years). I know that this example doesn't make much sense scientifically, but it is a probable use case, no? 
I am currently exploring a way to use `store()` and a hdf5 dataset (on disk) to keep the train lazy while improving reusability of the training results.

huard: Indeed. OK !

xclim/sdba/adjustment.py
aulemahal: With the removal of `norm_group` I decided to keep `normalize_sim` as the argument name.

xclim/sdba/adjustment.py
aulemahal: `save_training` and `save_fit` removed.

aulemahal:
After discussion with the whole user base, I followed the comments here and removed `save_training` and `save_fit`. Loading the datasets in memory is now the recommended way of dividing the workload in two, the notebook in the other PR has been update as well.

xclim/sdba/adjustment.py
```diff
@@ -289,9 +247,10 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     kind : {'+', '*'}
       The adjustment kind, either additive or multiplicative.
     group : Union[str, Grouper]
-      The grouping information. See :py:class:`xclim.sdba.base.Grouper` for details.
-    norm_group : Union[str, Grouper]
-      If given, the normalization steps are done using this group. Otherwise, they use the main `group`.
+      The grouping information use in the quantile mappgin process. See :py:class:`xclim.sdba.base.Grouper` for details.
```
huard, line 250:
```suggestion
      The grouping information used in the quantile mapping process. See :py:class:`xclim.sdba.base.Grouper` for details.
```

PR #485, opened 2020-06-24
Subclass MissingBase to create FromContext class

huard:
### Pull Request Checklist:
    - This PR fixes #xyz

This PR refactors the `missing.py` file in order to have a uniform mechanism to register missing functions. At the moment, the "missing_from_context" has no associated class. It will be easier for the refactor to have a class for each function. 

No, the missing functions are still available. 

* **Other information**:


xclim/core/missing.py
```diff
@@ -299,41 +275,112 @@ def missing_pct(da, freq, tolerance, **indexer):
...
+
+    def is_missing(self, null, count):
+        """Return whether or not the values within each period should be considered missing or not."""
+        return False
+
+    def __call__(self):
+        return False
+
+
+@register_missing_method("from_context")
```
aulemahal, line 337: I'm not sure I understand how registering the "from_context" option works? Right now, this allows to do: `xc.set_options(check_missing='from_context')`, which will fail with an `NotImplementedError` as this class doesn't implement anything?

huard: You would tell an indicator to use the "from_context" missing method. Then, when calculating the mask, it would use the missing method from context. In your example, you're setting the context to "from_context", which is an infinite loop.

aulemahal: Oh! Sorry I misread the code in #435! Then I suggest we modify the "check_missing" validator in `xclim.core.options` to forbid setting `from_context`, just in case.

PR #489, opened 2020-06-24
Changed signature of four indices from tasmax, tasmin to tasmin, tasmax

### LINKED ISSUE #486: Some indices are defined with (tasmin, tasmax) while others use (tasmax, tasmin)
* xclim version: master

### Description

I think we should stick to one ordering. It creates problems when writing cf compliance checks and will potentially trip users.

huard:
### Pull Request Checklist:
    - This PR fixes #486 

Change the signature of indices and indicators to have a uniform API for multivariate indices. Now all indices use tasmin, tasmax. 

Yes. This changes the signature of 4 indices and indicators. 

* **Other information**:
This does not cause any issue at the moment, because we are binding the arguments to methods using keyword arguments. But in my Indicator class refactor, I was not binding arguments and bugs appeared. I'll try to bind them also to make the Indicator as robust to such variations, but I thought it was worth changing anyway.


huard:
To be clear, making this change is not strictly necessary, so if it breaks too much downstream code, I'm ok with dropping it.

aulemahal:
Not at all! The website's script uses user made xclim calls for now and will use keywords only calls in a next version.

PR #502, opened 2020-07-13
KKZ implementation doc and tests

### LINKED ISSUE #499: Implement KKZ selection algorithm (ensembles.py)
### Description
Implement the KKZ selection technique described in https://journals.ametsoc.org/jcli/article/28/3/1260/106748/Selecting-GCM-Scenarios-that-Span-the-Range-of

Similar to cluster selection but gives nested selection results

nb* requested by P. Grenier

aulemahal:
### Pull Request Checklist:
    - This PR fixes #499

Adds the `ensembles.kkz_reduce_ensemble` method, based on [Cannon (2015)](https://journals.ametsoc.org/jcli/article/28/3/1260/106748/Selecting-GCM-Scenarios-that-Span-the-Range-of). It has a similar signature, input syntax and output as,  `ensembles.kmeans_reduce_ensemble`. It selects points in a phase-space, trying to cover a region a large as possible. 

No.

### DIFF
--- a/tests/test_ensembles.py
+++ b/tests/test_ensembles.py
@@ -402,3 +402,41 @@ class TestEnsembleReduction:
         )
         assert ids == [0, 1, 3, 4, 6, 7, 10, 11, 18, 20]
         assert len(ids) == 10
+
+    @pytest.mark.parametrize(
+        "crit,num_select,expected",
+        [
+            ([0, 1], 5, [16, 19, 20, 15, 9]),
+            ([0, 1], 4, [16, 19, 20, 15]),
+            (np.arange(6), 8, [23, 19, 10, 14, 11, 8, 20, 3]),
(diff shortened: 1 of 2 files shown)


aulemahal:
I'm pretty sure it would be fine to use a standardized metric on already standardized data...
I added the standardization because the article recommends it. A user could have already standardized data or not want to use the standardization, maybe? I can't think of a use case where we wouldn't want standardization... If it seems useless to provide a choice, we could simply force it...

tlogan2000:
@aulemahal added a small test testing the 'weird' case of seuclidean with standardize=True etc... assuming it passes travis buils feel free to merge.   I think we can leave the option to standardize or not.  Maybe there would be an application. Default set to True so people would have to actively change it to False...

Issue #506, opened 2020-07-20
Most doctests are not written to be tested


Zeitsperre:
I'm currently integrating some better doctest handling into the tox builds and have run up against a glaring issue: Many of the documentation examples are reliant on fictional files, for instance in `subset.py`:
```
    Examples
    --------
    >>> import geopandas as gpd
    >>> ds = xarray.open_dataset(path_to_tasmin_file)  # This file doesn't exist --> Failure
    >>> polys = gpd.read_file(path_to_multi_shape_file) # This file doesn't exist --> Failure

    Get a mask from all polygons in the shape file
    >>> mask = create_mask_vectorize(x_dim=ds.lon, y_dim=ds.lat, poly=polys)
    >>> ds = ds.assign_coords(regions=mask)
```
and for the documented examples in `ensembles.py`, they effectively all must be skipped:
```
    Examples
    --------
    Create ensemble dataset
    >>> ens = create_ensemble(temperature_datasets)  # doctest: +SKIP  # This file doesn't exist --> Failure

    Calculate default ensemble percentiles
    >>> ens_percs = ensemble_percentiles(ens)  # doctest: +SKIP # The previous line was skipped so all subsequent lines must be as well
    ...
```
I can skip all the tests that can't be tested for the moment, but unless we add some very small bunk data to the main package or run doctests only when the source files are present (which isn't so much issue for Travis CI or when building from the release source files in `.tar.gz`), most of these tests will need to be skipped.

Wondering if we have any ways of envisioning how we structure doctests. I will link to the PR here once it's been opened, so we can see exactly how many tests are passing/failing.

huard:
@Zeitsperre Status ? If there is no time to do this before the release date, I suggest reverting #507 to reinstate the doctests, even if some were not exercised.

huard:
The issue description mentions "fictional files", but for the record, these are real test files defined in `conftest.py` and injected in the doctest namespace using the pytest fixture `doctest_namespace`. The open question is why are they not recognized by `xdoctest`.

Zeitsperre:
Been working on this today Managed to figure out how best to inject to the `xdoctest_namespace`. Doing my best to catch errors in the doctests, but I may need yours and @aulemahal's help in dealing with them.

Should be able to push a PR this afternoon.

PR #507, opened 2020-07-20
Add xdoctest

### LINKED ISSUE #450: Not all doctests are picked by pytest
### Description

- Not all doctests in docstrings are picked up by pytest.
- I don't think code in the docs is exercised by pytest yet.

For example, in `indices`, `max_n_day_precipitation_amount` runs, but not `max_1day_precipitation_amount`.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #450 

This PR adds the [`xdoctest`](https://github.com/Erotemic/xdoctest) library to the dev build and implements it in the tox builds that check the documentation. This library uses a much better parser and is able to find all the documentation examples within the code and properly run evaluation tests on them.

I've also reformatted many docstrings that simply weren't being formatted properly on ReadTheDocs. Consequently, locally-building the docs via `make docs` at the top-level now raises MANY warnings that should eventually be addressed.

Another problem relates to discussion in #506. I have elected to set all tests that only marginally pass to be skipped entirely. If we can figure out how best to deal with the example test data problem, we can re-enable these checks.

No, though docstring standards are getting generally stricter.

* **Other information**:

I've made the decision to import global and locally-imported libraries within the examples to show users how to gain access to the functions being presented. This is more of a stylistic decision, and we can discuss whether we like examples to be short and sweet or longer and more elaborate, albeit more redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,13 @@ jobs:
         apt:
           packages:
             - libspatialindex-dev
+    - env: TOXENV=doctests
+      name: "Linting suggestions and Doctests"
+      python: 3.6
+      addons:
+        apt:
+          packages:
+            - libspatialindex-dev
     - env: TOXENV=py38
       name: "Python3.8 (Linux)"
       python: 3.8
@@ -55,13 +62,6 @@ jobs:
         - /usr/local/opt/python@3.8/bin/pip3 install -U tox-travis
     - env: TOXENV=py37
       name: "Python3.7 (Linux)"
-      python: 3.7
-      addons:
-        apt:
-          packages:
-            - libspatialindex-dev
-    - env: TOXENV=py37-doctests
-      name: "Python3.7 (Linux + Linting and Doctests)"
       python: 3.7
       addons:
         apt:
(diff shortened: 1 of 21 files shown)


Zeitsperre:
@tlogan2000 Would this check be something that we want as a build failure? Badly-written or error-filled examples could prevent new merges? What say you?

tlogan2000:
> @tlogan2000 Would this check be something that we want as a build failure? Badly-written or error-filled examples could prevent new merges? What say you?

i.e any new examples moving forward ?  I would say yes but we would need to create a standard dataset to run to docs right?

Zeitsperre:
Precisely. That's more or less what #506 is about. 

I'll add the "doctests and linting" build as another check entirely (that runs `pylint` in no-failures mode so that we can see what should be addressed without blocking merges due to "errors"). 

Linting errors are more like very strong code suggestions that will take time to deal with and will need some serious digging and debate, and potentially some refactoring as well. Best to deal with in another issue/PR.

tlogan2000:
ok for me

PR #508, opened 2020-07-23
subset module is replaced by clisops library

cehbrecht:
### Pull Request Checklist:
    - This PR fixes #xyz

This PR removes the subset module and replaces it by the clisops library:
https://pypi.org/project/clisops/

Yes. Presently, all subset.py-based functions will be offloaded to the roocs/clisops library, necessitating a new dependency and an updated API call for users relying on these functions (`from clisops.core import subset`).

Something that we can do is to offer the subset functions without breaking functionality is to offer the subset functions via a `subset.py` module that loads clisops if the library is available.

* **Other information**:

https://github.com/roocs/clisops/issues/31


docs/notebooks/index.rst
```diff
@@ -6,7 +6,6 @@ Examples
    :maxdepth: 1
 
    example
-   subset
```
Zeitsperre, line 9: I wouldn't remove the subset example yet until clisops documentation is online accessible and shows how this is performed.

environment.yml
```diff
@@ -11,13 +11,9 @@ dependencies:
  - cftime>=1.0.4
  - netCDF4>=1.4
  - poppler>=0.67
- - fiona>=1.8
- - geojson>=2.5.0
- - shapely>=1.6
- - geopandas>=0.7
+ - rtree>=0.9.4
```
Zeitsperre, line 14: If `geopandas` is not installed, does xclim still need the rtree library?

Zeitsperre: Also, is the newest clisops available as a conda-forge package?

docs/notebooks/index.rst
cehbrecht: the clisops docs has a copy of the subset example notebook:
https://clisops.readthedocs.io/en/latest/notebooks/subset.html

Should it be linked to the xclim docs?

environment.yml
cehbrecht: clisops conda-forge package is available (currently building). I don't know about the `rtree` ... I remove it and run the tests.

cehbrecht: removed `rtree` ... cleaned requirements in `setup.py`. tests are still working (locally).

tlogan2000:
Quick question:  Looking at the new notebook examples I believe this completely removes xclim.subset correct? 

Alternatively we could simply recreate the xclim.subset module by simply importing clisops?

docs/notebooks/index.rst
Zeitsperre: That's good to see. Seeing as we want users to start using clisops in the near future, it might be good to place a note somewhere that links to the clisops docs. The one thing that we might want to do for the subset example is to run it and leave the output intact so people can see the figures. I can look into porting that to theclisops docs if you want.

Zeitsperre:
> Quick question: Looking at the new notebook examples I believe this completely removes xclim.subset correct?
> 
> Alternatively we could simply recreate the xclim.subset module by simply importing clisops?

@tlogan2000 Exactly that. The new library gets rid of all the GIS dependencies and moves the functions out to more data slicing-and-dicing utility. The ideal usage is to simply import the subset module from there and run with it

environment.yml
Zeitsperre: We might also be able to remove the `spatialindex` and `libspatialindex-dev` apt/homebrew libraries from the travis-ci builds at the same time. Those are only there to support rtree and GeoPandas.

tlogan2000:
> > Quick question: Looking at the new notebook examples I believe this completely removes xclim.subset correct?
> > Alternatively we could simply recreate the xclim.subset module by simply importing clisops?
> 
> @tlogan2000 Exactly that. The new library gets rid of all the GIS dependencies and moves the functions out to more data slicing-and-dicing utility. The ideal usage is to simply import the subset module from there and run with it

I guess my question was whether we should keep an integrated xclim.subset module (probably less confusing for users) even if the code is transfered to clisops ....  e.g xclim's subset.py would basically have one line ... `from clisops.core import subset`

Zeitsperre:
@tlogan2000 for backwards compatibility, we could set that as a FutureWarning ("The subset module is now available under `clisops.core.subset` and will not be installed by default in a future version" or something along those lines).

At the same time, I'm also wondering about the need for `spatialindex`; If we need  `clisops` in order to run our doc examples, then we can't remove it as it's a necessary dependency, even though we don't offer this library. I think linking the docs to clisops makes more sense.

We need to make a decision:
- Do we keep `spatialindex` and list clisops as a dependency in order to run our subset examples, adding a FutureWarning for the API change ?
- Do we remove all subsetting and create an install flavour (e.g. `pip install xclim[gis]`) and have one travis-ci build check for it ?
- Another option ?

Zeitsperre:
@cehbrecht Think you could give me push access to your fork? I think I have some changes in mind to help this along.

cehbrecht:
> @cehbrecht Think you could give me push access to your fork? I think I have some changes in mind to help this along.

@Zeitsperre I have send you an invitation. Thanks.

Zeitsperre:
Thanks Carsten. Will work on this when I have some time today or tomorrow.

docs/api.rst
```diff
@@ -89,8 +89,8 @@ ensembles module
...
-
-.. automodule:: xclim.subset
-   :members:
-   :undoc-members:
-   :show-inheritance:
+.. warning::
+    Subsetting is now offered via `clisops`. The functions offered by clisops
+    will be described here once the subsetting functions API is made available.
+    For now, refer to their documentation here:
+    :doc:`clisops subset examples <clisops:notebooks/subset>`
```
aulemahal, line 96: We could, but I feel a link to the clisops doc is already sufficient? Why should we present the doc of another package?

setup.py
```diff
@@ -61,6 +51,8 @@
     "distributed>=2.0",
 ]
 
+gis_requirements = ["rtree>=0.9", "clisops>=0.3"]
```
aulemahal, line 54: I might have missed something, why are we needing rtree?

Zeitsperre: Okay, so this is an issue that I think needs to be addressed in clisops, but essentially, depending on how GeoPandas is installed (via pip or via conda), `rtree` is not guaranteed to be installed alongside it. This might explain some errors in `clisops` (I'm opening a PR to verify this right now) and, assuming this is the case, `rtree` should 100% be listed as a build dependency in that package, not here. Once I've verified this, I can remove this from here. `clisops` should be managing all its dependencies. Nice catch!

The apt package `libspatialindex[-dev]` is configured as a C-lib dependency in order to satisfy this package, hence why I've removed it from the `py36-nosubset` build.

docs/api.rst
Zeitsperre: This is true. I thought it might be interesting to link directly to their command API as xclim now virtually loads `clisops` to expose their API as ours.

aulemahal: Oh! So this would be a 0.19-temporary thing? Then I agree.

Zeitsperre: Yeah, that's what I envisioned. Who knows? Maybe it would be neat to have an API exposed only if `xclim[gis]` is installed?

Zeitsperre:
This PR is nearly done. Just waiting on https://github.com/roocs/clisops/pull/36 to be merged and a bug fix release to be pushed before I can add the final touches.

Zeitsperre:
@cehbrecht Just to keep you in the loop; This PR is ready, but we're going to integrate this into the September release of xclim. There's already quite a bit going into 0.19 already. Expect this to be merged in the 3-4 weeks.

Thanks for the PR!

HISTORY.rst
```diff
@@ -2,8 +2,30 @@
...
 
-0.19.0 (2020-08-18
-------------------
+0.20.x
+------
+
+Breaking changes
+~~~~~~~~~~~~~~~~
+* `xclim.subset` has been deprecated and now relies on `clisops` to perform specialized spatiotemporal subsetting.
+  In order to maintain functionality, users must install xclim with `xclim[gis]` in order to retain the same functionality.
```
huard, line 11:
```suggestion
* `xclim.subset` has been deprecated and now relies on `clisops` to perform specialized spatiotemporal subsetting. Install with `pip install xclim[gis]` in order to retain the same functionality.
```

HISTORY.rst
```diff
@@ -2,8 +2,30 @@
...
+
+New features and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* xclim now depends on clisops for subsetting, offloading several GIS dependencies that are typically installed by default,
+  whether subsetting utilities are desired. This improves maintainability and reduces the size of a "vanilla" xclim
+  installation considerably.
+
```
huard, line 18:
```suggestion
* xclim now depends on clisops for subsetting, offloading several heavy GIS dependencies. This improves maintainability and reduces the size of a "vanilla" xclim installation considerably.

```

HISTORY.rst
```diff
@@ -2,8 +2,30 @@
...
+Bug fixes
+~~~~~~~~~
+
+Internal changes
+~~~~~~~~~~~~~~~~
+* `xclim.subset` now attempts to load and expose the functions of `clisops.core.subset`. This is an API workaround for backwards compatibility.
```
huard, line 24:
```suggestion
* `xclim.subset` now attempts to load and expose the functions of `clisops.core.subset`. This is an API workaround preserving backwards compatibility.
```

tests/test_subset.py
```diff
@@ -133,6 +164,9 @@ def test_time_incomplete_years(self):
         np.testing.assert_array_equal(out.time.max().dt.day, 30)
 
 
+@pytest.mark.skipif(
```
huard, line 167: Why not use `pytest.importorskip` instead ?

Zeitsperre: For some reason, `pytest.importorskip` was not registering the tests as skipped in the log. Using this method clearly shows in the pytest evaluation the specific tests that were skipped. I want to know which files and tests are being skipped and this seemed like the better approach to get that done.

setup.py
Zeitsperre: This is fixed in `clisops@v0.3.1`

Zeitsperre:
All good for me as well. I've opened a PR on the `clisops` side to configure their conda-feedstock to provide Windows binaries(https://github.com/conda-forge/clisops-feedstock/pull/5), but this shouldn't prevent us from moving forward here. Merging once the build checks clear.

tests/test_subset.py
```diff
@@ -1,16 +1,47 @@
...
+# class TestSubsetImport:
+#     subset = pytest.importorskip('clisops.core.subset', reason="`clisops` subset utilities are not installed.")
+#     pytest.importorskip("rtree", reason="rtree spatial indexing utilities are not installed.")
+
+try:
+    import clisops.core.subset as subset
+    import geopandas as gpd
+except ImportError:
+    subset = False
+    gpd = False
```
huard, line 16:
```suggestion
try:
    import clisops.core.subset as subset
    import geopandas as gpd
except ImportError as e:
    pytestmarkskip_gis = pytest.mark.skip(reason=str(e))
```

tests/test_subset.py
```diff
@@ -1,16 +1,47 @@
...
 
 TESTS_HOME = os.path.abspath(os.path.dirname(__file__))
 TESTS_DATA = os.path.join(TESTS_HOME, "testdata")
 
 
+class TestSubsetRaises:
+    @pytest.mark.skipif(
+        subset is False, reason="`clisops` subset utilities are not installed."
+    )
```
huard, line 25:
```suggestion
    @pytestmarkskip_gis
```

tests/test_subset.py
huard: Good point. See https://github.com/pytest-dev/pytest/issues/5551#issuecomment-508453960 for a work around.

tests/test_subset.py
Zeitsperre: This is raising errors as it changes a lot of the logic behind skipping tests peppered throughout the file. This also causes a problem when clisops is installed as `pytestmarkskip_gis` is undefined. I imagine an `else` statement would work but I'm not sure what to define that decorator as in the case that clisops is found.

Zeitsperre:
I opted to revert the different pytest approach as the current one seemed to be working fine. Ideally, I would like to see our subset tests get streamlined seeing as `clisops` has nearly copied all of our tests already. That would merit another PR, I think.

huard:
Ok, sorry about the distraction.

Zeitsperre:
All good. Thanks for suggesting the different approach. I'm unresolving it so that we remember to try that out the next time a case like this comes up. Speaking of that, do you know how the decorator should be defined in the case I mentioned above?

huard:
`pytest.mark.skipif(False)` ?

PR #510, opened 2020-07-31
Multi indicator subclass

aulemahal:
### Pull Request Checklist:
    - This PR fulfills a goal of #351

~Introduces a new Indicator subclass for MultiIndicators, i.e.: for cases when the compute function outputs multiple dataarrays. This is especially useful for the fire weather indexes, which are calculated together.~
Modifies the Indicator class in depth to change paradigm : Now Indicator have a variable number of output variables.

- Instead of being passed directly, cf attributes are given through "var_attrs", a list of dictionaries.
- Cf attributes are divided into `_cf_var`, the attributes to be given to the variables and `_cf_global`, those to be given to the dataset.  Var: (var_name), standard_name, long_name, description, units, cell_methods, comment. Global: description, references.
- Some var attributes may still be given directly in the class : global settings of units, cell_methods and description will be used in case a var_attrs element doesn't declare them.
- Users have a possibility to switch the output of Indicators with `xc.set_options(output_dataset=True)`. With True, it always return a dataset. With False, it returns a DataArray if there is only one output (the case for almost all Indicators). The switch is to False right now, with a FutureWarning and I suggest we move it to True in 0.20.

The implementation needed many small tweaks elsewhere to work:

- Indicator's method `update_attrs` and `format` are now _classmethods_, so one must pass the set of attributes to update when using them.
- `xclim.core.units.declare_units` now works with multiple outputs : if the output of the indice is a tuple, it iterates over a zip of (out_units, output), so the units must be declared as a sequence.
- `xclim.core.locales.get_local_attrs` now uses the indicator's class name instead of the indicator object itself. Also, the `fill_missing` keyword has been removed. The "fill_missing" behavior is now part of the `Indicator.translate_attrs` method.
- Moved some arguments of `xclim.indices.fire_weather_indexes` from the starred kwargs to the explicit kwargs.
- Added the `"[speed]" dimensionnality.

Oh que oui. The new declaration of Indicators and new scheme could break a lot of downstream code. Most importantly, `Indicator.cf_attrs` and  `Indicator.json` now put the variable attributes in a subkey.

Also, the `update_attrs`  and `format` change are internal but could be breaking? Same with changes to `locales.get_local_attrs`

EDIT: After discussions, we decided to make the change more general the whole PR is update.

### Changes example for `xc.atmos.tg_mean`
<details>

`.to_json()` before:
```
{'identifier': 'tg_mean',
 'var_name': 'tg_mean',
 'abstract': '',
 'keywords': '',
 'standard_name': 'air_temperature',
 'long_name': 'Mean daily mean temperature',
 'units': 'K',
 'cell_methods': 'time: mean within days time: mean over days',
 'description': '{freq} mean of daily mean temperature.',
 'notes': '',
 'parameters': "{'tas': {'default': None, 'desc': ''}, 'freq': {'default': 'YS', 'desc': ''}}"}
```
`to_json()` in this PR:
```
{'identifier': 'tg_mean',
 'title': '',
 'abstract': '',
 'keywords': '',
 'vars': [{'var_name': 'tg_mean',
   'standard_name': 'air_temperature',
   'long_name': 'Mean daily mean temperature',
   'units': 'K',
   'cell_methods': 'time: mean within days time: mean over days',
   'description': '{freq} mean of daily mean temperature.'}],
 'notes': '',
 'parameters': "{'tas': {'default': None, 'desc': ''}, 'freq': {'default': 'YS', 'desc': ''}}"}
```
</details>


HISTORY.rst
```diff
@@ -13,11 +13,14 @@ Breaking changes
   missing values is identified by its registered name, e.g. "any", "pct", etc, along with its `missing_options`.
 * xclim now requires xarray >= 0.16, ensuring that xclim.sdba is fully functional.
 * The dev requirements now include `xdoctest` -- a rewrite of the standard library module, `doctest`.
+* `xclim.core.locales.get_local_attrs` now uses the indicator's class name instead of the indicator itself, also it no longer accept the `fill_missing` keyword. Behavior is now the same as passing `False`.
```
Zeitsperre, line 16:
```suggestion
* `xclim.core.locales.get_local_attrs` now uses the indicator's class name instead of the indicator itself and no longer accepts the `fill_missing` keyword. Behaviour is now the same as passing `False`.
```

aulemahal:
Sorry for the multiple commits... It seems that `pydocstyle` isn't run by pre-commit? I believe that should be the case, no? @Zeitsperre

Zeitsperre:
Ah, it is possible I didn't enable it since there's a bugfix release coming soon. I can look into this. Sorry!

xclim/core/indicator.py
```diff
@@ -536,6 +551,168 @@ class Indicator2D(Indicator):
...
+    r"""Climate multi-indicator base class.
+
+    Climate indicator object that, when called, computes multiple indicator and assigns its outputs a number of
+    CF-compliant attributes. Some of these attributes can be *templated*, allowing metadata to reflect
+    the value of call arguments.
+
+    Instantiating a new multi-indicator returns an instance but also creates and registers a custom subclass.
+
+    The usage of this class is similar to the base `Indicator` except that its output is a _Dataset_ containing
+    multiple arrays istead of a single _DataArray_. Attributes of the children arrays are given through the `children`
```
huard, line 564:
```suggestion
    multiple arrays instead of a single _DataArray_. Attributes of the children arrays are given through the `children`
```

Zeitsperre:
@aulemahal For the pydocstyle check, the PR is #514

xclim/core/indicator.py
```diff
@@ -325,33 +329,35 @@ def update_attrs(self, ba, das):
           Input arrays.
         ba: bound argument object
           Keyword arguments of the `compute` call.
+        attrs : Mapping[str, str]
+          The attributes to format and update.
+        identifier : str
```
huard, line 334: I think it's confusing to use `identifier` to describe something that is not exactly the class identifier.

xclim/core/indicator.py
```diff
@@ -369,8 +375,8 @@ def update_attrs(self, ba, das):
             attrs["cell_methods"] += " " + out.pop("cell_methods")
 
         attrs["history"] = update_history(
-            f"{self.identifier}{ba.signature.replace(parameters=cp.values())}",
-            new_name=vname,
+            f"{identifier or cls.identifier}{ba.signature.replace(parameters=cp.values())}",
```
huard, line 378: Not sure I understand why identifier would be different from `cls.identifier`.

xclim/core/indicator.py
huard, line 564:
```suggestion
    r"""Multi-indicator base class extending `Indicator` to allow the computation of multiple outputs. 

    The usage of this class is similar to the base `Indicator` except that, upon execution, it returns a `Dataset` containing multiple `DataArray` objects. The attributes of individual output `DataArray` are specified using the `children` variable, a sequence of dictionaries mapped to the output variables. Class instance attributes refer to *global* dataset attributes. 
```

xclim/locales/fr.json
```diff
@@ -517,5 +517,46 @@
     "comment": "",
     "title": "Fin de la saison de végétation",
     "abstract": "Jour de l'année de la fin de la saison de végétation. La saison de végétation est définie comme l'intervalle entre la première série de N jours où la température journalière est au-dessus d'un seuil et la première série (après une certaine date) de N jours où elle est sous ce même seuil"
+  },
+ "FWI":{
+  "description": "L'IFM est formé de six composantes qui tiennent compte des effets de la teneur en eau des combustibles et des conditions météorologiques sur le comportement du feu.",
+  "title": "Indices Forêt-Météo",
+  "abstract": "Calcule les 6 indices forêt-météo tel que défini par le Service Canadien des Forêts : l'indice de sécheresse, l'indice d'humidité de l'humus, l'indice du combustible léger, l'indice de propagation initiale, l'indice du combustible disponible et l'indice forêt météo."
```
huard, line 524:
```suggestion
  "abstract": "Calcule les 6 indices forêt-météo tels que définis par le Service Canadien des Forêts : l'indice de sécheresse, l'indice d'humidité de l'humus, l'indice du combustible léger, l'indice de propagation initiale, l'indice du combustible disponible et l'indice forêt météo."
```

xclim/locales/fr.json
```diff
@@ -517,5 +517,46 @@
...
+    "long_name": "Indice du combustible léger (FFMC)",
+    "description": "Évaluation numérique de la teneur en eau de la litière et d'autres combustibles légers. Calculée avec la méthode de mise en route '{start_up_mode}'",
+    "title": "Indice du combustible léger quotidien (FFMC)",
+    "abstract": "L'indice du combustible léger fait partie du système canadien des Indices Forêt-Météo. C'est une évaluation numérique de la teneur en eau de la litière et d'autres combustibles légers."
+  },
+ "FWI.ISI": {
+    "long_name": "Indice de propagation initiale",
+    "description": "Évaluation numérique du taux prévu de propagation du feu.",
+    "title": "Indice de propagation initiale quotidien (ISI)",
+    "abstract": "L'indice dde propagation intiale fait partie du système canadien des Indices Forêt-Météo. C'est une évaluation numérique de le taux prévu de propagation du feu."
```
huard, line 548:
```suggestion
    "abstract": "L'indice de propagation intiale fait partie du système canadien des Indices Forêt-Météo. C'est une évaluation numérique de le taux prévu de propagation du feu."
```

xclim/core/indicator.py
aulemahal: In MultiIndicator the "identifier" passed to `update_attrs` would be something like : "FWI.DC", whereas the class identifier is "FWI".

xclim/core/indicator.py
aulemahal: Right. Could be "sub_id", or "child_id" ?

aulemahal:
@huard @Zeitsperre The last commit is kinda big. I updated the PR text, see above.

xclim/core/options.py
```diff
@@ -22,6 +23,7 @@
     CF_COMPLIANCE: "warn",
     CHECK_MISSING: "any",
     MISSING_OPTIONS: {},
+    OUTPUT_DATASET: False,  # TODO : Change to True in 0.20
```
huard, line 26: You think the default should be to return a dataset ?

aulemahal: Because we chose that multi-output Indicator return Datasets, I thought that _always_ returning a dataset was more consistent. But, your point on backward-compatibility is more important, I will remove the warning and this comment.

xclim/core/indicator.py
```diff
@@ -536,6 +551,168 @@ class Indicator2D(Indicator):
...
+
+        # Assume the first arguments are always the DataArrays.
+        das = OrderedDict()
+        for i in range(self._nvar):
+            das[self._parameters[i]] = ba.arguments.pop(self._parameters[i])
+
+        # Metadata attributes from templates
+        attrs = [
+            self.update_attrs(
+                ba, das, child, identifier=f"{self.identifier}.{child['var_name']}"
```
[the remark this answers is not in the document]
aulemahal: I changed it to "var_id", you might be reading an old version?

aulemahal:
@huard Made several changes to make this backward-compatible.

1) The indicator's __new__ method takes care of putting variable attributes in the `var_attrs` dictionaries. This allows to pass them directly as keywords or even as class attributes. However, this made me realize that using `var_attrs` kinda breaks the inheritance pattern we wanted.... I mean, when a class inherits from another, the `var_attrs` is not merged, but replaced. With the fix, inheritance works, but in a hacky way, since the base Indicator class does not declare the variable attributes...

<details>

Example:
```
xc.atmos.tg_mean.units   # Does not exist, AttributeError

class TG_MEAN2(registry['TG_MEAN']):
    units = 'C'

TG_MEAN2().units  # Exists and is now considered a "global" attribute is some part of the code (but not cf).
```

Indicators now have 3 lists of names : 

- `ind._var_names` : Valid variable attributes
- `ind._cf_var` : Variable attributes to write to the output
- `ind._cf_global` : Global attributes to write to the output

So, instead of having a list of "valid global attributes" I am using : "hasattr(class, name)" to check if an attribute is global. Theoritically, all class attributes  should be global attributes, but allowing variable metadata inheritance breaks that.

</details>

2) Done

Instead of using `ind.__class__.__name__` I swtiched to `ind.identifier.upper()`, since that is what it is.

xclim/core/indicator.py
```diff
@@ -156,20 +187,29 @@ class Indicator:
     # Number of input DataArray variables. Should be updated by subclasses if needed.
     _nvar = 1
 
-    # Allowed metadata attributes on the output
-    _cf_var = [
+    # Allowed metadata attributes for the variables
+    _var_names = [
```
huard, line 191: I think `_var_names` is confusing. I thought it referred to the names of the output DataArrays.

aulemahal:
With the upcoming set changes that revert many lines, this PR would have too much messy commits. I will reopen a new one from a branch branch.

Issue #516, opened 2020-08-05
Document Indicator creation and customization


aulemahal:
With the merge of Indicator class refactoring, need for examples on how to create and customize new Indicator classes arises.
We should explain how a user can create those on-the-fly or in an external package, it could be added to the "customizing and controlling xclim" notebook.

Moreover, the Table of Indicator could be reintegrated to the documentation index and be improved by removing duplicates and adding a bit more info on each indicator + giving a link to the docstring.

huard:
There is already some material here: https://xclim.readthedocs.io/en/stable/api.html#defining-new-indicators

However, there is no mention of multi-output indicators, or the expected frequency of inputs.

PR #519, opened 2020-08-10
Multi outputs indicators

aulemahal:
### Pull Request Checklist:
    - This PR fulfills a goal of #351 and goes in the direction of #342 and #516.

Copy of #510  at midpoint, I created a new PR to limit the number of revert commits. 

So, modifies the class Indicator so that it accepts any number of outputs. Variable metadata can still be passed as strings, but also as lists of strings. The number of outputs is assumed to be `len(indicator.var_name)` . When a string is passed to a multi-output indicator, it is copied on all outputs.

In opposition to the other PR, this version allows to still use classic class inheritance for indicators. Internally, the indicator class uses `Ind.cf_attrs` to manage the number of outputs and their attributes. It is created in the `__new__` method.  Note that this means that variable metadata is _read-only_, which it wasn't before.

The implementation needed many small tweaks elsewhere to work:

- Indicator's method `update_attrs` and `format` are now _classmethods_, so one must pass the set of attributes to update when using them.
- `xclim.core.units.declare_units` now works with multiple outputs : if the output of the indice is a tuple, it iterates over a zip of (out_units, output), so the units must be declared as a sequence.
- `xclim.core.locales.get_local_attrs` now uses the indicator's class name instead of the indicator object itself. Also, the `fill_missing` keyword has been removed. The "fill_missing" behavior is now part of the `Indicator.translate_attrs` method.
- Moved some arguments of `xclim.indices.fire_weather_indexes` from the starred kwargs to the explicit kwargs.
- Added the `"[speed]" dimensionnality.

The declaration, use and output of current single-output indicator hasn't changed. But:

- Accessing variable attribute on the indicator object might return a single string or a list of strings. The attributes concerned are the ones in `Ind._cf_names`.
- `Ind.cf_attrs` is now a list of dictionaries, very often a single dict.
- `Ind.json()` now as an entry "outputs" which is a list of dicts (in fact, a (formatted) copy of `cf_attrs`).

## Other information
The `long_name` json attribute is replaced by `title`. Currently, titles are not a reliable and coherent source of information for indicators, but #516 will take care of that.

### Changes example for `xc.atmos.tg_mean.json()`
<details>

`.to_json()` before:
```
{'identifier': 'tg_mean',
 'var_name': 'tg_mean',
 'abstract': '',
 'keywords': '',
 'standard_name': 'air_temperature',
 'long_name': 'Mean daily mean temperature',
 'units': 'K',
 'cell_methods': 'time: mean within days time: mean over days',
 'description': '{freq} mean of daily mean temperature.',
 'notes': '',
 'parameters': "{'tas': {'default': None, 'desc': ''}, 'freq': {'default': 'YS', 'desc': ''}}"}
```
`to_json()` in this PR:
```
{'identifier': 'tg_mean',
 'title': '',
 'abstract': '',
 'keywords': '',
 'outputs': [{'var_name': 'tg_mean',
   'standard_name': 'air_temperature',
   'long_name': 'Mean daily mean temperature',
   'units': 'K',
   'cell_methods': 'time: mean within days time: mean over days',
   'description': '{freq} mean of daily mean temperature.'}],
 'notes': '',
 'parameters': "{'tas': {'default': None, 'desc': ''}, 'freq': {'default': 'YS', 'desc': ''}}"}
```
</details>


HISTORY.rst
```diff
@@ -31,7 +35,9 @@ Internal changes
 * Indicator subclasses `TasminTasmax` and `PrTas` now inherit from `Daily2D`.
 * Docstring style now enforced using the `pydocstyle` with `numpy` doctsring conventions.
 * Doctests are now performed for all docstring `Examples` using `xdoctest`. Failing examples must be explicitly skipped otherwise build will now fail.
+* Indicator methods `update_attrs` and `format` are now classmethods, attrs to update must as passed.
```
huard, line 38: typo

xclim/core/indicator.py
```diff
@@ -95,112 +98,144 @@ class Indicator:
...
+            values = kwds.get(name, getattr(cls, name))
+            if not isinstance(values, (list, tuple)):
+                values = [values] * n_outs
+            elif len(values) != n_outs:
+                raise ValueError(
+                    f"Attribute {name} has {len(values)} elements but should have {n_outs} according to passed var_name."
+                )
+            for attrs, value in zip(kwds["cf_attrs"], values):
+                if value:
+                    attrs[name] = value
```
huard, line 235: Does it make sense to put this logic into a method ?

xclim/core/indicator.py
```diff
@@ -261,26 +294,49 @@ def __call__(self, *args, **kwds):
...
 
         # Compute the indicator values, ignoring NaNs and missing values.
-        out = self.compute(**das, **ba.kwargs)
+        outs = self.compute(**das, **ba.kwargs)
+        if isinstance(outs, tuple) and len(outs) != n_outs:
+            raise ValueError(
+                f"Indicator {self.identifier} was wrongly defined. Expected {n_outs} outputs, got {len(outs)}."
+            )
+        if n_outs == 1:
+            outs = [outs]
```
huard, line 317:
```suggestion
        if n_outs == 1:
            outs = [outs]
        if len(outs) != n_outs:
            raise ValueError(
                f"Indicator {self.identifier} was wrongly defined. Expected {n_outs} outputs, got {len(outs)}."
            )

```

huard: or maybe better would be 
```
if isinstance(outs, xr.DataArray):
    outs = (outs, )
```

xclim/core/indicator.py
```diff
@@ -261,26 +294,49 @@ def __call__(self, *args, **kwds):
...
         # Mask results that do not meet criteria defined by the `missing` method.
+        # This means all variables must have the same dimensions...
         mask = self.mask(*das.values(), **ba.arguments)
+        outs = [out.where(~mask) for out in outs]
 
-        return out.where(~mask).rename(vname)
+        # Return a single DataArray in case of single output, otherwise a tuple
+        if n_outs == 1:
+            return outs[0]
+        return tuple(outs)
```
huard, line 339: Shouldn't `outs` already be a tuple ?

xclim/core/indicator.py
```diff
@@ -95,112 +98,144 @@ class Indicator:
...
     abstract = ""
     keywords = ""
     references = ""
-    comment = ""
     notes = ""
 
     def __new__(cls, **kwds):
-        """Create subclass from arguments."""
-        identifier = kwds.get("identifier", getattr(cls, "identifier"))
+        """Create subclass from arguments. Keywods of kwds that are class attributes are attributed here."""
```
aulemahal, line 197:
```suggestion
        """Create subclass from arguments."""
```

xclim/core/indicator.py
aulemahal: Could. And it would reduce the complexity of the method...

xclim/core/indicator.py
aulemahal: Because I use list-comprehension, it's a list. But multiple outputs in python is conventionally a tuple.

PR #520, opened 2020-08-11
Enable doctests using xdoctest conventions

### LINKED ISSUE #506: Most doctests are not written to be tested

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #506 

This PR re-enables the doctests that were previously skipped with #507. Minor updates to API calls have been made in order for examples to be synonymous with functions (_the reason for doctests_, one could say). With these changes, all `Examples` in the library are called and evaluated. 

No. `xdoctests` was already a required testing dependency and functional codes have not been touched.

* **Other information**:

Two tests are presently being troublesome:
- `tests/indices/_multivariate.py::cold_spell_duration_index`, as written, is hanging (this could be due to the need for more bunk tasmin data).
- `tests/ensembles.py::kmeans_reduce_ensemble` is giving some esoteric TypeError traceback that I can't make sense of. 
- `tests/subset.py` are all being skipped, as this module is being deprecated very soon.

### DIFF
--- a/xclim/ensembles.py
+++ b/xclim/ensembles.py
@@ -2,7 +2,7 @@
 import logging
 import warnings
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import scipy.stats
@@ -75,15 +75,14 @@ def create_ensemble(
 
     Examples
     --------
-    >>> import glob  # doctest: +SKIP
-    >>> from xclim.ensembles import create_ensemble  # doctest: +SKIP
-    >>> ens = create_ensemble(temperature_datasets) #doctest: +SKIP
-    ...
-    # Using multifile datasets:
-    # Simulation 1 is a list of .nc files (e.g. separated by time)
+    >>> from xclim.ensembles import create_ensemble
+    >>> ens = create_ensemble(temperature_datasets)
+
+    Using multifile datasets:
+    Simulation 1 is a list of .nc files (e.g. separated by time)
     >>> datasets = glob.glob('/dir/*.nc')  # doctest: +SKIP
-    ...
-    # Simulation 2 is also a list of .nc files
+
+    Simulation 2 is also a list of .nc files
     >>> datasets.append(glob.glob('/dir2/*.nc'))  # doctest: +SKIP
     >>> ens = create_ensemble(datasets, mf_flag=True)  # doctest: +SKIP
     """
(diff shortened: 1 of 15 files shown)


Zeitsperre:
Alright, I'll set the failing ones to `+SKIP` for the time being in order for the builds to pass properly (we'll still know which ones are being skipped). Will perform the rest of the PR merge checks now.

xclim/ensembles.py
```diff
@@ -718,11 +718,11 @@ def plot_rsqprofile(fig_data):
...
-    >>> from xclim.ensembles import kmeans_reduce_ensemble, plot_rsqprofile  # doctest: +SKIP
-    >>> crit = xr.open_dataset(path_to_ensemble_file).data  # doctest: +SKIP
-    >>> ids, cluster, fig_data = kmeans_reduce_ensemble(data=crit, method={'rsq_cutoff':0.9}, random_state=42)  # doctest: +SKIP
-    >>> plot_rsqprofile(fig_data)  # doctest: +SKIP
+    >>> from xclim.ensembles import kmeans_reduce_ensemble, plot_rsqprofile
+    >>> is_matplotlib_installed()
+    >>> crit = xr.open_dataset(path_to_ensemble_file).data
+    >>> ids, cluster, fig_data = kmeans_reduce_ensemble(data=crit, method={'rsq_cutoff':0.9}, random_state=42)
+    >>> plot_rsqprofile(fig_data)
     """
```
Zeitsperre, line 726:
```suggestion
    """
        if not MPL_INSTALLED:
        logging.error("Matplotlib is not installed. Unable to execute plotting functions.")
        raise ImportError("matplotlib")

```
Is this a terrible idea?

aulemahal: bah... Not a bad idea, but isn't the name "plt_rsqprofile" already quite explicit on the fact that it does plot? (And thus needs matplotlib)?

Zeitsperre: I suppose so. Nothing prevents this from being called as is. If MPL isn't there, I suppose the errors from python ("plt is not defined" or something like that) would be enough?

huard: Yes, seems overkill. Also, maybe we should avoid creating graphics in the doctests. If it's important to show these graphics, I would place them in the notebook tutorials instead.

aulemahal: True! We could have a notebook on all "ensembles" things, including kmeans and kkz selection.

Zeitsperre: See #522

PR #526, opened 2020-08-17
Add partial support for L-Moments

### LINKED ISSUE #33: Fit distributions using L-moments in addition to ML
We should be able to fit these six distributions using both maximum likelihood and the method of L-Moments:

- [ ] Gumbel (right)
- [ ] GEV
- [ ] LogNormal2 (loc=0)
- [ ] LogNormal3
- [ ] Pearson type III
- [ ] Log-Pearson type III 

This new functionality would be exposed in `indices.generic.fit`, `fa` and `frequency_analysis` as a new parameter `estimator` taking the default value `ml` for the default scipy ML estimator, and `lmom` for L-moments estimator.

huard:
### Pull Request Checklist:
    - This PR partially fixes #33 

Adds a `pwm_fit` method, estimating statistical distribution parameters from a sample using probability weighted moments. 
Only a few distributions are supported. 

No. 

* **Other information**:
Requires the `lmoments3` library, but it is unmaintained anymore. It's not clear if we should add it as a dependency or work on a more robust solution. Needs to be installed from master because the last release is not compatible with recent scipy releases.


xclim/indices/generic.py
```diff
@@ -147,97 +162,205 @@ def fitfunc(arr):
...
+
+        # Fill with NaNs if one of the parameters is NaN
+        params = list(lm3dc.lmom_fit(x).values())
+        if np.isnan(params).any():
+            params[:] = np.nan
+
+        return params
+
+    # xarray.apply_ufunc does not yet support multiple outputs with dask parallelism.
+    data = dask.array.apply_along_axis(fitfunc, da.get_axis_num("time"), da)
```
aulemahal, line 227: I'd suggest making this a duck ops : dask methods will return dask arrays, which is not always wanted!
A simple way, for this specific case is :
`duck = dask.array if isintance(da.data, dask.array.Array) else numpy`
since `np.apply_along_axis` has the same signature as the dask version.

huard:
@Zeitsperre How do you suggest we deal with the lmoments3 dependency? Since it is unmaintained, I'm not sure it's a good idea to include this "officially" in the requirements, but we probably still want to run it in the travis test suite.

Zeitsperre:
Fully agree that adding unmaintained libraries to the requirements is bad form. Plenty of ways forward though.

There's a couple of ways we could deal with it from the CI side:
- Add a new CI build for it (expensive on resources)
- Retrofit an existing CI build for it (like the py36-nosubset build)

In terms of how to proceed requirements-wise, we can treat functions reliant on it the same way we treat plotting functions, (IE creating a global variable like LM3_INSTALLED in the module that uses it, aborting functions that need it when it isn't there).

If we want users to be able to run tests locally with lmoments3 installed, we would put the dependency in the `tox.ini`, and if we only want to run it on Travis CI, we can add it to the `.travis.yml`.

Zeitsperre:
Whichever way forward you want to try, let me know and I can push some changes here.

huard:
@Zeitsperre I suggest we go with `tox.ini` to allow local testing. Ok to bundle the dependency with another suite. We could have a test environment dedicated to all optional external dependencies like roocs. Note that the lmoments3 library has to be installed from master.
`pip install git+https://github.com/OpenHydrology/lmoments3.git`

huard:
@aulemahal Combined fit and fit_pwm into one single function.

xclim/indices/generic.py
```diff
@@ -348,6 +376,21 @@ def get_dist(dist):
     return dc
 
 
+def get_lm3_dist(dist):
+    """Return a distribution object from `lmoments3.distr`."""
+    import lmoments3.distr
+
```
aulemahal, line 382:
```suggestion
```
Without this space, the comment about the manual installation will appear in conventional tracebacks, if the function is called without an installation of lmoments3.
We could also catch the error and raise an ImportError with the same message, but I don't think it is necessary here.

aulemahal:
Woupelaye. The build failures are due to OpenHydrology/lmoments3#6, which is open and inactive since june **2019**...

huard:
Ah right, it's the develop branch.

xclim/indices/generic.py
```diff
@@ -378,10 +378,11 @@ def get_dist(dist):
 
 def get_lm3_dist(dist):
     """Return a distribution object from `lmoments3.distr`."""
-    import lmoments3.distr
+    # fmt: off
+    import lmoments3.distr  # isort: skip
     # The lmoments3 library has to be installed from master.
     # pip install git+https://github.com/OpenHydrology/lmoments3.git
```
huard:
```suggestion
    # The lmoments3 library has to be installed from the `develop` branch.
    # pip install git+https://github.com/OpenHydrology/lmoments3.git@develop#egg=lmoments3
```

Zeitsperre:
Added in the formatting exceptions. Having the import in a function with no space afterwards breaks two hooks, haha.

PR #527, opened 2020-08-17
Adds a `parameter_quantile` method

### LINKED ISSUE #515: Generic quantile method operating on fitted distribution parameters
### Description

At the moment, `generic.fa` does both the fit and the quantile estimation. In some cases, we want to split this in two independent operations, and do 
```
p = fit(...)
q = quantile(p, ...)
```
For example, in a project I compute GEV block maxima over 5 years. This requires a parameter transformation in between then fit and the quantile estimation.

huard:
### Pull Request Checklist:
    - This PR fixes #515

Implements `parametric_quantile` function computing quantiles from pre-computed distribution parameters returned by `fit`. 

No. The `fa` function was modified to use this function internally. 

* **Other information**:
This is useful when parameters have been estimated using a non-standard estimation method.


HISTORY.rst
```diff
@@ -45,7 +49,7 @@ New features and enhancements
 * Create new Indicator `Daily`, `Daily2D` subclasses for indicators using daily input data.
 * The `Indicator` class now supports outputing multiple indices for the same inputs.
 * `xclim.core.units.declare_units` now works with indices outputting multiple DataArrays.
-* Doctests now make use of the `xdoctest_namespace` in order to more easily access mdoules and tesdata.
+* Doctests now make use of the `xdoctest_namespace` in order to more easily access mdoules and testdata.
```
aulemahal, line 52:
```suggestion
* Doctests now make use of the `xdoctest_namespace` in order to more easily access modules and testdata.
```

xclim/indices/generic.py
```diff
@@ -147,97 +147,128 @@ def fitfunc(arr):
...
-    ] = f"{dist} return period values for {getattr(da, 'standard_name', '')}"
+    ] = f"{dist} return period values for {p.attrs.get('standard_name', '')}"
     out.attrs["cell_methods"] = (
         out.attrs.get("cell_methods", "") + " dparams: ppf"
     ).strip()
-    out.attrs["units"] = da.attrs.get("units", "")
-    out.attrs["mode"] = mode
+    out.attrs["units"] = p.attrs["original_units"]
+
     out.attrs["history"] = (
```
aulemahal, line 219: You might want to use `xclim.core.formatting.update_history`? It does the same but also adds a timestamp and xclim's version.

xclim/indices/generic.py
```diff
@@ -147,97 +147,128 @@ def fitfunc(arr):
...
+    else:
 
         def func(x):
-            return dc.ppf(1.0 / t, *x)
-
-    else:
-        raise ValueError(f"Mode `{mode}` should be either 'max' or 'min'.")
+            return dc.ppf(q, *x)
 
     data = dask.array.apply_along_axis(func, p.get_axis_num("dparams"), p)
```
aulemahal, line 199: Not from your changes, but we could use the duck typing approach here to I think. So that the output is not always a dask array.

huard: Done in `fix-33`.

huard: Oups, no, I did it for fit and fit_pwm, not for parametric_quantile.... will do.

xclim/indices/generic.py
huard: done

PR #531, opened 2020-08-19
Add support for indicators based on hourly data

### LINKED ISSUE #419: Extend support to hourly data
Raven will generate hourly streamflow time series, driven by hourly meteorological data. Ideally we would be able to define indicators operating on hourly data.

huard:
### Pull Request Checklist:
    - This PR fixes #419

Adds support for hourly data through a `Daily` subclass of `Indicator`. 
Support hourly data in `missing` functions through a new `input_freq` parameter explicitly stating the expected frequency of input data. 
Add `max_pr_intensity` indices/indicator computing the maximum precipitation intensity (mm/h) over a duration (# hours). This would be used for example to compute IDF. 

Backward compatible API change to missing functions. 

* **Other information**:
**Q**: What kind of documentation should we add to this PR ? One idea would be a notebook on IDF curves, but this will probably require more time than I have at the moment.


tlogan2000:
@huard not sure if I will get to this review before thursday or friday ...  ok with you?  

If you and @aulemahal are certain of the changes feel free to merge I suppose.  If not I will take a look near the end of the week

huard:
@tlogan2000 Previously we assumed that input data was daily. Because now it can either be hourly or daily, we need to communicate this information to a couple of functions, such as the missing value functions. I've chosen "input_freq" but it sounds lame. Also, the `Indicator` class calls this `freq`, which is probably going to get confused with the output freq. So I suggest we use consistent naming before merging. Note that in the missing functions, both the output freq and the input freq are arguments. Suggestions ? 
- resolution
- time_step

tlogan2000:
> @tlogan2000 Previously we assumed that input data was daily. Because now it can either be hourly or daily, we need to communicate this information to a couple of functions, such as the missing value functions. I've chosen "input_freq" but it sounds lame. Also, the `Indicator` class calls this `freq`, which is probably going to get confused with the output freq. So I suggest we use consistent naming before merging. Note that in the missing functions, both the output freq and the input freq are arguments. Suggestions ?
> 
> * resolution
> * time_step

Q- Is this input something that will be exposed to user calls? e.g. in the atmos module?  Or is it handled more less internally by the checks and the added `xr.infer_freq` method?  

If it is seen by user calls then I kind of like the idea of an 'input' distinction ... makes it clearer in my mind.  Or perhaps use 'source'
`source_timestep` or `src_timestep` ?

I would personally avoid `resolution` as I think mnay people associate that only with the spatial aspect.

Again sorry I have not looked at the code at all

tlogan2000:
To follow up : 
FYI - some of the `_anuclim.py` indices were created using an `input_freq` parameter (anuclim typically works on weekly or even monthly data). So changes should be made there to be consistent I think.  

@aulemahal is there potential to remove `input_freq` from anuclim via xr.infer_freq to determine data frequency and a simple check to ensure that input is either  daily, weekly or monthly?

aulemahal:
We could default to `input_freq = None` and compute it with `xr.infer_freq`. The problem is that `infer_freq` is super strict : as soon as you have a missing timeslice it will return `None`. I don't think we can rely on it completely.  Also, the weekly frequency is not official to cftime objects and, thus, not implemented in  `xr.infer_freq`. It could be done, really shouldn't be hard.

Even if it is lame, I like the explicit-ness of `input_frequency`.  `src_timestep` sounds good too.

Zeitsperre:
Sorry if I OOTL on this, but if the `time_frequency` data being supplied can _somewhat_ easily be inferred, why are asking users to supply it to the functions?

aulemahal:
Hum, well it's because of both the sensitivity of  `xr.infer_freq` to small errors and of its inability to infer weekly frequencies. There could be a case where the data is not _exactly_ daily (missing feb 29th, for example), bu the user still wants to pass it through the indicator?

xclim/core/missing.py
```diff
@@ -56,13 +58,15 @@ class MissingBase:
...
-    def __init__(self, da, freq, **indexer):
-        self.null, self.count = self.prepare(da, freq, **indexer)
+    def __init__(self, da, freq, input_freq, **indexer):
+        if input_freq is None:
+            raise ValueError("`input_freq` must be either 'D' or 'H'.")
+        self.null, self.count = self.prepare(da, freq, input_freq, **indexer)
 
     @classmethod
-    def execute(cls, da, freq, options, indexer):
+    def execute(cls, da, freq, input_freq, options, indexer):
```
tlogan2000, line 67: I think I prefer `src_timestep` in the end

tlogan2000 (approved):
This seems to good. Like I mentioned I think I would maybe use 'src_timestep' in order to avoid confusion for users

xclim/core/missing.py
huard: Ok, will make the change and merge.

huard:
For consistency, I'll rename your options for input_freq: 
daily -> D
weekly -> W
monthly -> M

PR #532, opened 2020-08-20
Integrate mypy into the build tests

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

This PR introduces `mypy` to the build testing ensemble which is a very comprehensive type-checking library, and can help in examining whether call signatures and functions outputs are valid across functions, modules, and libraries.

Potentially. Many call signatures show some problems and fixing this will likely change expected types seen for functions in userland. Addressing some errors may require comprehensive changes, but hopefully not.

* **Other information**:

This is currently a draft as mypy report more than 120 errors currently. [Some suggestions on how to best work with mypy to address these problems](https://breadcrumbscollector.tech/mypy-how-to-use-it-in-my-project/) involve progressively upping the strictness of checks and ignoring errors where needed. This is most definitely the case of operator errors for xarray objects, as they are sometimes recognized as `Any` types (and `Any * int` or `Any / float` are correctly recognized as errors). I need to read further from the xarray side whether this can be easily addressed.


aulemahal:
Good idea! This will certainly help the coherence of the package.
I understand the `Any` problematic you raised is the reason behind the "#type: ignore" comments in `_conversion.py`?  If we have too many of those, it could become cumbersome to manage...

Zeitsperre:
Yes, that is a problem. The impression I had was to deal with the errors progressively; If we can deal with the problems that are because of us, we can eventually move towards dealing with the problems impacting us from outside modules.

It's possible that we may need to push some changes towards xarray, but thankfully, they've already made some headway there: https://github.com/pydata/xarray/issues?q=is%3Aissue+mypy

Zeitsperre:
One way to proceed is to explicitly run mypy over a specific file. It will raise errors in multiple files, but it's a lot easier than wading through 100s of errors at a time. 

I'm thinking that the effort that we've put towards added static type hinting checks is the reason why we don't have 1000s or more errors.

xclim/indices/_conversion.py
```diff
@@ -74,7 +78,7 @@ def uas_vas_2_sfcwind(
 
     if not return_direction:
         wind.attrs["units"] = "m s-1"
-        return wind
+        return wind, None
```
Zeitsperre, line 81: @aulemahal This is a change that is potentially breaking, so I'm curious as to what you think here: Depending on inputs to the function, the number of outputs will change. I've come to understand that this isn't the best practice to follow. As such, I figured it would make sense to always return two outputs (`None` or `xr.DataArray`), but this requires a fix/tests to the Indicator class `atmos.wind_speed_from_vector`.

Is there a better way to organize this? Would it make more sense to split these outputs into two separate functions? I need development/design advice.

aulemahal: Oh! That's one change I overlooked when pushing the PR on multiindicators! The `return_direction` is not necessary anymore. I can push a fix today (in a separate PR, its cleaner), so that the function always return 2 DataArrays.

Zeitsperre: Awesome. Thanks!

Zeitsperre:
Pinning CLISOPS below 0.4.0 since the new package has several dependencies that should be addressed outside of this PR

xclim/indices/_anuclim.py
```diff
@@ -194,8 +196,9 @@ def precip_seasonality(
 @declare_units("[temperature]", tas="[temperature]")
 def tg_mean_warmcold_quarter(
     tas: xarray.DataArray,
-    op: str = None,
-    src_timestep: str = None,
+    *,
```
huard, line 199: Is there a consensus on not allowing positional arguments ?
It seems overkill for a function that takes only one dataarray argument.

xclim/indices/fwi.py
```diff
@@ -140,18 +140,26 @@ def day_length(lat, mth):  # pragma: no cover
...
+    elif lat > 90 or lat < -90:
+        raise ValueError("Invalid lat specified.")
+    else:
+        raise ValueError
     return dl[mth - 1]
 
 
 @jit
-def day_length_factor(lat, mth):  # pragma: no cover
+def day_length_factor(lat: Union[int, float], mth: int):  # pragma: no cover
```
huard, line 151: `int` is implied in `float`, no ? So I don't think you need this union.

xclim/core/indicator.py
```diff
@@ -253,7 +253,9 @@ def _parse_docstring(cls, kwds):
         return kwds
 
     @classmethod
-    def _parse_cf_attrs(cls, kwds):
+    def _parse_cf_attrs(
+        cls, kwds: Dict[str, Union[str, Callable, int, bool, List[str]]]
+    ) -> Union[List[Dict[str, str]], List[Dict[str, Union[str, Callable]]]]:
```
aulemahal, line 258: phew... haha That one is complex.
But I don't really understand the difference between : "List[Dict[str, str]]" and "List[Dict[str, Union[str, Callable]]]". Isn't the second only a generalization of the first?

huard (approved):
I suggest we defer the positional argument change. Ideally this would be done in bulk, not for some functions but not others, with a clear logic behind. Otherwise looks good.

xclim/indices/fwi.py
Zeitsperre: Yes, will get on that.

xclim/indices/_anuclim.py
Zeitsperre: Woops, yes, we've discussed this and I think we deferred discussion on this change to much later (maybe v1). Will fix.

xclim/indices/fwi.py
```diff
@@ -236,7 +244,9 @@ def fine_fuel_moisture_code(t, p, w, h, ffmc0):  # pragma: no cover
 
 
 @vectorize
-def duff_moisture_code(t, p, h, mth, lat, dmc0):  # pragma: no cover
+def duff_moisture_code(
+    t, p, h, mth: int, lat: Union[int, float], dmc0: float
```
aulemahal, line 248: That feels a bit weird, no? Also, is there really an integer lat case?

xclim/indices/run_length.py
```diff
@@ -618,7 +618,7 @@ def windowed_run_events_1d(arr: Sequence[bool], window: int):
     return (v * rl >= window).sum()
 
 
-def windowed_run_count_ufunc(x: Sequence[bool], window: int) -> xr.apply_ufunc:
+def windowed_run_count_ufunc(x: Sequence[bool], window: int) -> Callable:
```
aulemahal, line 621: Euh, isn't the result a DataArray? I think we mistakenly wrote that those ufunc return "function" instead of datarrays.

aulemahal (approved):
Made some comments.
Thanks for doing this, looks like a long and meticulous job!

xclim/core/indicator.py
Zeitsperre: Yeah, this needs to be simplified. The errors being raised by mypy might be due to the fact that things being fed to this function COULD be any of these things. As we tighten the types up a bit, I imagine that this will change.

aulemahal: A problem with typing these class methods is that `kwds` can contain _anything_ an indicator can be initialized with. Sure we might be able to come up with a typing compound that covers all types xclim currently uses in its indicators. But, there's not guarantee that a future (or external) subclass might use something fancy outside from those! 
Can we type with "Any"? Or ignore typing?

Zeitsperre: You were right about this, it was nearly too much for `typing` to parse. `kwds: Dict[str, Any]` should work fine though.

PR #536, opened 2020-08-28
Update black in pre-commit hooks, reformat code base, remove python-pandoc

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

This PR makes the xclim code sympatico with recent changes to `black` (v0.19.10 --> v0.20.8). I haven't read the patch notes in-depth (I know there are some interesting experimental features coming, but, in the interest of not messing up everyone's day, they aren't active here yet).

No. Changes are minor. The convention now appears to be to split `args` and `kwargs` in call signatures vertically by default. You can still write them horizontally, but pre-commit hooks will modify to fit the standard.

* **Other information**:

You may need to perform an update to `pre-commit` if you haven't updated it in a long time. `conda update pre-commit` will do the work for you while the commit hooks will update and manage themselves.

### DIFF
--- a/tests/test_run_length.py
+++ b/tests/test_run_length.py
@@ -79,7 +79,12 @@ class TestLongestRun:
 
         # n-dim version versus ufunc
         da3d = xr.open_dataset(self.nc_pr).pr[:, 40:50, 50:68] * 0
-        da3d[0:10,] = da3d[0:10,] + 1
+        da3d[0:10,] = (
+            da3d[
+                0:10,
+            ]
+            + 1
+        )
         da3d = da3d == 1
         lt_orig = da3d.resample(time="M").map(rl.longest_run_ufunc)
         # override 'auto' usage of ufunc for small number of gridpoints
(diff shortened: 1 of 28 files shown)


tests/test_run_length.py
```diff
@@ -79,7 +79,12 @@ def test_start_at_0(self):
 
         # n-dim version versus ufunc
         da3d = xr.open_dataset(self.nc_pr).pr[:, 40:50, 50:68] * 0
-        da3d[0:10,] = da3d[0:10,] + 1
+        da3d[0:10,] = (
+            da3d[
+                0:10,
+            ]
+            + 1
+        )
```
aulemahal, line 87: Yo, que pasa? I'm not sure I understand which pep8 rule the line was breaking? It's not the only one-liner that is now spread out for no apparent reason...

aulemahal:
Oh, synchronized commenting. There seems to be something wrong with line lengths?

Zeitsperre:
Ok, yes, some of this is nasty. I'll have time to look deeper into this this afternoon.

Zeitsperre:
Alright, so it's gotten a bit more strict in some cases. If you are slicing lists, you need to remove your trailing commas or else it will explode the statement. Therefore, this:
` da3d[0:10,] = da3d[0:10,] + 1`
Should now be written as this:
`da3d[0:10] = da3d[0:10] + 1`

This appears to be the worst offender. Other changes are slightly jarring, but this is now the standard. Merges should be relatively painless from here.

aulemahal:
Oh! ok so the trailing comma explodes the lists. Is it the same for calls and lists in the other changes?

Zeitsperre:
@aulemahal From what I can tell, if calls signatures have more than two arg/kwargs, they are now exploded/vertical. As I mentioned above, feel free to write it the way you feel most comfortable. If pre-commit is installed correctly, it'll make changes as needed.

Zeitsperre:
Ok, once this clears, feel free to merge!

PR #539, opened 2020-09-02
Indicator instance registration and realm attribute

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Moves all indicator registering actions to a new `IndicatorRegistrar` class (from which `Indicator` now inherits). Instances are registered by storing weakrefs in a dictionary. Classes stored in the indicator registry have a `get_instance()` method that returns the first existing instance, or raises a ValueError if none are found (either none were created, or they were all destroyed).

Also, to keep the realm info when parsing indicators through the registry, Indicators now have a `realm` attribute. It must be one of atmos, land, seaIce or ocean. Indicators declared in `xclim.indicators` need not declared, it is parsed from the module name. All other Indicators must explicitly set it (either through the parent class or a keyword, as other attributes).

No.

* **Other information**:
I believe packages like `finch` could be improved by parsing indicators through the registry instead of the module. It will ensure that external indicators (even if they don't exist for now) will be included.

### DIFF
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -64,6 +64,7 @@ a `tg_mean` indicator returning values in Celsius instead of Kelvins, you could 
 """
 import re
 import warnings
+import weakref
 from collections import OrderedDict, defaultdict
 from inspect import signature
 from typing import Mapping, Sequence, Union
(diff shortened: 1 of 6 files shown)


xclim/core/indicator.py
```diff
@@ -87,10 +88,40 @@ def datacheck(tas):
...
+        return super().__new__(cls)
+
+    def __init__(self):
+        _indicators_registry[self.__class__].append(weakref.ref(self))
+
+    @classmethod
+    def get_instance(cls):
+        """Return first found instance.
+
+        Raises ValueError is no instance exist (anymore).
```
huard, line 113:
```suggestion
        Raises `ValueError` if no instance exists.
```

xclim/indicators/atmos/_precip.py
```diff
@@ -166,6 +166,7 @@ class PrTas(Daily2D):
 fire_weather_indexes = Daily(
     _nvar=4,
     identifier="FWI",
+    realm="atmos",
```
huard, line 169: Why does this need to be defined ?

huard (approved):

xclim/indicators/atmos/_precip.py
aulemahal: Because the module of a class is the file in which the class is defined, not where it is instantiated. Meaning that as we are using "xclim.core.indicator.Daily" and not one of the classes defined under `xclim.indicators.atmos", the class's `__new__` doesn't "detect" the realm from the module... 
Maybe a more appropriate technique would be to create a "fire weather indexes" class that would have only one instance?

Issue #558, opened 2020-09-17
Accept datasets and variable names in Indicator calls


aulemahal:
To make the indicator calls even easier for users and automated processes, calling an Indicator should be possible with a dataset directly. We already do this in the command line interface, so transferring to the Indicator class should be easy.

Suggested implementation:
Add a new argument `ds_in` to `Indicator.__class__`, it should accept a Dataset and be None by default.
If `ds_in` is passed, variables arguments can be strings, corresponding to variable names. If they are not given, the name of the argument is used as the variable name.

Example:
```python3
ds  # Dataset with variable "tas", the mean daily temperature.

# All following calls would be equivalent:
tg_mean(tas=ds.tas)
tg_mean(ds_in=ds, tas='tas')
tg_mean(ds_in=ds)
```

Potential issues:
- Currently, we copy the signature of the `compute` function to `__call__`. With this, we will need to modify the defaults and type annotations.
- How do we determine that an argument is to be sought as a "variable"? In the CLI we use the annotations, if it is `xr.DataArray`, then it is a variable. This logic can be rediscussed and needs to be tested across all indicators.

Issue #368, opened 2020-02-14
Clean-up of confusing meta data values


huard:
See https://github.com/bird-house/finch/issues/80

### Description

Indicators carry over metadata from the original file, but some of them may not be valid anymore (e.g. frequency) or make the output confusing. We should decide what to do with them (keep, change, update, remove).

For example, should we store the original attributes in a string ? 

## CMIP5 attributes
- [ ] branch_time
- [ ] contact
- [ ] Conventions
- [ ] creation_date
- [ ] experiment
- [ ] experiment_id
- [ ] forcing
- [ ] frequency
- [ ] initialization_method
- [ ] institute_id
- [ ] institution
- [ ] model_id
- [ ] modeling_realm
- [ ] parent_experiment_id
- [ ] parent_experiment_rip
- [ ] physics_version
- [ ] product
- [ ] project_id
- [ ] realization
- [ ] source
- [ ] table_id
- [ ] tracking_id

Zeitsperre:
@Ouranosinc/xclim-core We need to have a meeting about this specifically.

huard:
@Zeitsperre Please propose a date, time and agenda.

Zeitsperre:
Proposed talking points (feel free to modify this message as needed):

- What metadata do we presently affix into `Indicators`/indices?
- How do we currently handle external metadata?
- Which metadata standards and fields should we provide explicit support for?
    - How should we handle exceptions to these standards/accepted fields?
- What fields should we be adding?
    - How can these fields support current project?

Zeitsperre:
Some thoughts:

1. CF-standartd global attributes should not be found in variable attributes. This is an error/misleading. I would remove these, if/when they are found. **This complicates how we write our `history`**. From the summary write-up of CF-1.9 (proposed):

    > 2.6.1 Identification of Conventions
      Requirements:
           > The Conventions attribute must be a single text string containing a list of convention names separated by blank space or commas, one of which shall be the full CF string as described below.
           > Files that conform to the CF version 1.8 conventions must indicate this by setting the global Conventions attribute to contain the CF string value "CF-1.8".

    > 2.6.2 Description of File Contents
      Requirements:
           > The title, history, institution, source, references, and comment attributes are all type string.
      Recommendations:
           > The title and history attributes are only defined as global or groups attributes. If they are used as per variable attributes a CF compliant application should treat them exactly as it would treat any other unrecognized attribute.

[...]

PR #559, opened 2020-09-18
Rename history to xclim_history

aulemahal:
### Pull Request Checklist:
    - This PR is step forward for #368

Changes the `history` attribute that Indicators assign to their output to `xclim_history`.  Also removed the `<No available history>` placeholder strings, an empty string is used instead.

Yes.

* **Other information**:
Right now we are constructing this history string from all the `history` attributes of the input variables. Which doesn't make sense in regards to CF conventions and. from now on, is not coherent with our own attribute name. As long as #558 is not implemented, I don't see any solutions...


huard (approved):

huard (commented):
I suggest setting the `missing_str` to "" instead of "<No available history>". It's not very informative anyway.

aulemahal:
And then I learned about a new github markdown possibility, interesting. "<very interesting indeed>"

Issue #525, opened 2020-08-17
Clean-up test and doctests datasets


huard:
I suggest we review and update our test datasets.  The NRCan datasets that we currently use will be unfamiliar to many users, and have a lot of grid points, which slows down testing. 

I suggest we discuss what datasets should be included in the test, doctest and documentation. 
For example, we could use recent CMIP6 simulations.

huard:
@sbiner I suggest we include some of the observational datasets that caused you issues. 

@Zeitsperre We could create a `xclim-data` repository playing the same role as `xarray-data`, then add an `xclim.tutorial` wrapping `xarray.tutorial`. 

@aulemahal Comments ?

aulemahal:
We had the same discussion in #351, where we evaluated the `intake` option. While it did take care of the machinery part, it did not solve the storage.
`xclim-data` seems a good idea indeed. Also the code of  `xarray.tutorial` is super simple, it would be easy to adapt and that solution adds no dependency.
Are there any challenges to creating a data-only repo?  I think I remember from earlier discussions, that we didn't want to do that, but I have no memory why...

Zeitsperre:
> Are there any challenges to creating a data-only repo? I think I remember from earlier discussions, that we didn't want to do that, but I have no memory why...

Interestingly enough, the folks at roocs have done precisely that with our existing data via sub-modules: https://github.com/roocs/xclim-testdata/tree/master. When the source code for one of their projects is cloned, they also pull this testdata repo using `git clone --recursive <link>`. If anything I suggest we use this repo (and maybe add in some data license texts and Apache 2 License stating that this was forked from xclim).

Integrating this into xclim would be extremely simple:
```
cd tests/
git submodule add https://github.com/roocs/xclim-testdata/
git submodule init
```

huard:
+1 for using xclim-testdata, but -1 for the submodule approach. I would use the `xarray.tutorial` access mechanism to simplify access for users.

aulemahal:
After today's discussions, we discussed the following partial fixes:

1 - Create a `xclim-testdata` github repo and use `xr.tutorial.open_dataset` to read data from it.

2 - Using the implementation suggested in #558, create a large dataset with a sample of all kinds of variables used in xclim. Cycle through all indicators (with the registry) and call them with the dataset (all other parameters default).  

Simple as this, it would at least test that the computation works for most indices. Moreover, we will be able to test if attributes are carried on or not.

[...]

PR #571, opened 2020-10-06
New testing.open_dataset method for downloading test data

aulemahal:
### Pull Request Checklist:
    - This PR goes in the direction of fixing #525.

Adds `open_dataset` to  xclim.testing. Both are copied and adapted from xarray, with the following changes:
    - Use of pathlib instead of os
    - Allow sub-folders in the distant repo.
    - Change default parameters to point to xclim-tesdata.

No

Tests are coming when the structure of xclim-testdata is updated.


Zeitsperre:
@aulemahal Problematic testing case: `test_ensembles.py`

```
class TestEnsembleStats:
    nc_files_simple = glob.glob(
        os.path.join(TESTS_DATA, "EnsembleStats", "*1950-2100*.nc")
    )
    nc_files = glob.glob(os.path.join(TESTS_DATA, "EnsembleStats", "*.nc"))
```

We may need to hardcode these files and then open them before sending them to `create_ensemble()`. What do you think?

aulemahal:
Yuck... We could also create the ensemble beforehand and have it as a netcdf in the repo?

Zeitsperre:
There's a few instances of `xr.open_mfdataset()` in our testing ensemble. Is this possible or would it be better to `concat` opened `Datasets` ?

aulemahal:
Unless the thing tested is the multiplicity of the files, I believe the next test clean up step will remove those calls.  For now, we could indeed concat opened datasets?
As for tests where multiple files is important (like for `xclim.ensembles` I presume), I'm not sure of the way forward...

Zeitsperre:
@aulemahal There seems to be an issue with the FWI Indices tests. It could be that the assertion is not an exact match. Barring the `open_mfdataset` and `create_ensemble` tests, everything is coming together. Could likely finish this today.

aulemahal:
I'll look into the FWI problem.
Also, in regard to your last commit, would it be possible to reactivate the windows test build on this PR? I don't think windows users should have to use `os.path.join` to use `testing.open_dataset`. I'll try to think of a workaround, but would be interested to see the CI build!

Zeitsperre:
> Also, in regard to your last commit, would it be possible to reactivate the windows test build on this PR? I don't think windows users should have to use `os.path.join` to use `testing.open_dataset`. I'll try to think of a workaround, but would be interested to see the CI build!

Yeah, I can re-enable it, but it'll need to be deactivated again before pushing to master. The other option is to use `Path` objects; they are able to determine OS-specific path names as well.

aulemahal:
Easier than I thought!

Zeitsperre:
We just need to revert https://github.com/Ouranosinc/xclim/commit/f23f2f32f1ae221c1e57d3effec7b8339b7a1980 and we're good, I think!

huard (approved):

Issue #575, opened 2020-10-21
`missing` functions do not support cftime objects


huard:
* xclim version: master

### Description

Applying missing function on DataArray whose time coordinate has a cftime calendar. 

### What I Did

```
import numpy as np
import xarray as xr
import xclim
time = xr.cftime_range(start="1950-01-01", periods=1000)
da = xr.DataArray(np.random.rand(1000), dims=("time", ), coords={"time":time})
xclim.core.missing.missing_pct(da, freq="Y", season="DJF", tolerance=.1)
```

### What I Received

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-11-e368fc7adfc3> in <module>
----> 1 xclim.core.missing.missing_pct(da, freq="Y", season="DJF", tolerance=.1)

~/src/xclim/xclim/core/missing.py in missing_pct(da, freq, tolerance, src_timestep, **indexer)
    422 def missing_pct(da, freq, tolerance, src_timestep=None, **indexer):  # noqa: D103
    423     src_timestep = src_timestep or xr.infer_freq(da.time)
--> 424     return MissingPct(da, freq, src_timestep, **indexer)(tolerance=tolerance)
    425 
    426 

~/src/xclim/xclim/core/missing.py in __init__(self, da, freq, src_timestep, **indexer)
     62         if src_timestep is None:
     63             raise ValueError("`src_timestep` must be either 'D' or 'H'.")
---> 64         self.null, self.count = self.prepare(da, freq, src_timestep, **indexer)
     65 
     66     @classmethod

~/src/xclim/xclim/core/missing.py in prepare(self, da, freq, src_timestep, **indexer)
    141         if indexer:
    142             # Create a full synthetic time series and compare the number of days with the original series.
--> 143             t0 = str(start_time[0].date())
    144             t1 = str(end_time[-1].date())
    145             if isinstance(da.indexes["time"], xr.CFTimeIndex):

AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'date
```

tlogan2000:
Strange,  I can treat your random data as input to an `atmos` calculation and use a set_options context and it seems to work?
```
import numpy as np
import xarray as xr
import xclim
from xclim import atmos
time = xr.cftime_range(start="1950-01-01", periods=1000)
da = xr.DataArray(np.random.rand(1000), dims=("time", ), coords={"time":time})
da.resample(time='YS').mean(dim='time')
[...]

PR #583, opened 2020-10-31
Multiple fixes on calendar tools

### LINKED ISSUE #568: Support `dt.datetime` objects in `get_calendar`
* xclim version: 0.20

### Description

Would it be possible to support normal datetime objects in `get_calendar` ?

### LINKED ISSUE #578: Add `365_day` to `datetime_classes`
* xclim version: 0.21

### Description

`365_day` is also used for `noleap`

### LINKED ISSUE #580: Remove user warning for convert_calendar
* xclim version: 0.21

### Description

Getting this message when using align_on convert_calendar. This is unnecessarily annoying. 

```
UserWarning: Argument `align_on` was specified, but none of the source or target calendars is '360_day'. `align_on` will be ignored.
```

### LINKED ISSUE #582: Optionally add values for missing days on calendar conversion
Converting calendars from one that has more days e.g. `all_leap` or  `366_day` towards `default` will drop the appropriate number of days in order to remain consistent throughout the year, but the same is not true currently when performing the inverse.

In the course of some analyses, I realize that having the option to pad the year with an odd `np.NaN` when going from `noleap` to `proleptic_gregorian`/`default` calendar would be a useful option to have to ensure that days don't decouple when comparing multiple years between data of different data sources.

In other words: being able to ensure that 30 years (1981 to 2010) of `noleap` days goes from `len(days) = 11680` to `len(days) == 11688` in a default calendar.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #568, fixes, #575, fixes #578, fixes #580 and fixes #582.

Several small changes:

- `missing` argument added to `convert_calendar`
- No more warning for passing `align_on` uselessly to `convert_calendar`
- Non-failing use of `Missing` objects with  `indexer` and cftime coordinates.
- All objects I could think of can be passed to `get_calendar`. (python datetimes, pandas Timestamps, cftime datetimes and arrays or sequences of them).
- Added `365_day` and `366_day` to the understood CF datetime classes (respectively 'noleap' and 'all_leap').

No.

* **Other information**:
5 issues in 1 PR, my new high-score.

### DIFF
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -6,8 +6,6 @@ Calendar handling utilities
 
 Helper function to handle dates, times and different calendars with xarray.
 """
-
-import datetime
 import datetime as pydt
 from typing import Any, Optional, Sequence, Union
 from warnings import warn
@@ -79,7 +77,7 @@ def get_calendar(obj: Any, dim: str = "time") -> str:
     obj = np.take(
         obj, 0
     )  # Take zeroth element, overcome cases when arrays or lists are passed.
-    if isinstance(obj, datetime.datetime):  # Also covers pandas Timestamp
+    if isinstance(obj, pydt.datetime):  # Also covers pandas Timestamp
         return "default"
     if isinstance(obj, cftime.datetime):
         return obj.calendar
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -8,7 +8,7 @@ Helper function to handle dates, times and different calendars with xarray.
 """
 
 import datetime as pydt
-from typing import Optional, Sequence, Union
+from typing import Any, Optional, Sequence, Union
 from warnings import warn
 
 import cftime
@@ -35,10 +35,11 @@ datetime_classes = {
     "proleptic_gregorian": cftime.DatetimeProlepticGregorian,
     "julian": cftime.DatetimeJulian,
     "noleap": cftime.DatetimeNoLeap,
+    "365_day": cftime.DatetimeNoLeap,
     "all_leap": cftime.DatetimeAllLeap,
+    "366_day": cftime.DatetimeAllLeap,
     "360_day": cftime.Datetime360Day,
 }
-
 
 # Maximum day of year in each calendar.
 max_doy = {
@@ -48,7 +49,9 @@ max_doy = {
     "proleptic_gregorian": 366,
     "julian": 366,
     "noleap": 365,
+    "365_day": 365,
     "all_leap": 366,
+    "366_day": 366,
     "360_day": 360,
 }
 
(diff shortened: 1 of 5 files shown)


xclim/core/calendar.py
```diff
@@ -35,11 +36,12 @@
     "proleptic_gregorian": cftime.DatetimeProlepticGregorian,
     "julian": cftime.DatetimeJulian,
     "noleap": cftime.DatetimeNoLeap,
+    "365_day": cftime.DatetimeNoLeap,
     "all_leap": cftime.DatetimeAllLeap,
+    "366_day": cftime.DatetimeAllLeap,
```
huard, line 41: Note that you'll find this in `cftime._cftime.DATE_TYPES`

xclim/core/calendar.py
```diff
@@ -113,15 +127,20 @@ def convert_calendar(
...
     dim : str
       Name of the time coordinate.
 
     Returns
     -------
     Union[xr.DataArray, xr.Dataset]
       Copy of source with the time coordinate converted to the target calendar.
-      The length of the array is the same as `target` if an array was given, otherwise it stays the same as `source`.
-      Except if source is a `360_day` calendar and `align_on='date'`: then a daily source will be output with 358 dates
+      If `target` is given as an array, the output is reindex to it.
```
huard, line 140:
```suggestion
      If `target` is given as an array, the output is reindexed to it.
```

huard (approved):
Just did a superficial review. I suggest to review the convert_calendar docstring for the english (punctuation missing in some places for clarity).

Zeitsperre:
This is looking good. I'm running some processing scripts to see if it addresses the problems that I ran into. Assuming all goes well, I'll give it a proper review later today. Thanks for the speedy fixes!

xclim/core/calendar.py
```diff
@@ -256,6 +265,72 @@ def interp_calendar(
     return out
 
 
+def date_range(*args, calendar="default", **kwargs):
+    """Wrapper of pd.date_range (if calendar == 'default') and xr.cftime_range (otherwise)."""
```
Zeitsperre, line 269:
```suggestion
    """Wrap pd.date_range (if calendar == 'default') or xr.cftime_range (otherwise)."""
```

xclim/core/calendar.py
```diff
@@ -106,23 +108,27 @@ def convert_calendar(
...
     target : Union[xr.DataArray, str]
       Either a calendar name or the 1D time coordinate to convert to.
       If an array is provided, the output will be reindexed using it and in that case, days in `target`
-      that are missing in the converted `source` are filled by NaNs.
+      that are missing in the converted `source` are filled by `missing` (which defaults to NaN).
     align_on : {None, 'date', 'year'}
       Must be specified when either source or target is a `360_day` calendar, ignored otherwise. See Notes.
+    missing : Optional[any]
+      A value to use for filling in dates in the target that were missing in the source.
+      If `target` is a string, default (None) is not to fill values. If it is an array, default is to fill with NaN.
```
Zeitsperre, line 120:
```suggestion
      A value to use for filling in dates in the target that are missing in the source.
      If `target` is a string, default (None) is to not fill any values. 
      If `target` is an array, default is to fill with numpy.NaN/pandas.na.
```
Not all `NaN`s are made the same. Numpy's `NaN` is float-based, but in the event of an integer variable, does numpy's `np,nan` know to keep the nantype (integer) synonymous with the rest of the values in the array?

Zeitsperre (approved):

xclim/core/calendar.py
aulemahal: True, numpy and xarray do not know how to have a nan type similar to the values. However, using np.nan is xarray's default already. It will recast an entire int array to float. But using `pandas.NA` would anyway recast the int array to an `object` one, which is not better!
Here, passing `target : DataArray, missing = np.nan` has the exact same behavior as not passing `missing` (the previous behavior).
Until xarray/numpy do something about their non-float NaNs, I suggest we stick to what they do.

Zeitsperre: Okay. It's definitely something to take note of. I can see this being a bit jarring in the case of users converting their calendar and seeing the size of the files jump when introducing NaNs. Perhaps an `Note` is a good idea in the docstring ?

PR #584, opened 2020-11-02
Add spatial analog algos from flyingpigeon

### LINKED ISSUE #496: Migrate analog algorithms from flyingpigeon
Flyingpigeon has algorithms for analog computations. We want to migrate them to xclim and use xarray and dask.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #496 

Adapts (copies) the spatial analogues algorithms from flyingpigeon into xclim. For now, I haven't done much more then writing a `spatial_analog` function that wraps the metrics in an `apply_ufunc` call. I tried numba, but all metrics use scipy stuff that numba doesn't understand, so the current solution is to have `vectorize=True` in `apply_ufunc`.

No

### DIFF
--- a/xclim/analog.py
+++ b/xclim/analog.py
@@ -1,0 +1,545 @@
+# -*- encoding: utf8 -*-
+"""
+===============
+Spatial analogs
+===============
+
+Spatial analogues are maps showing which areas have a present-day climate that is analogous
+to the future climate of a given place. This type of map can be useful for climate adaptation
+to see how well regions are coping today under specific climate conditions. For example,
+officials from a city located in a temperate region that may be expecting more heatwaves in
+the future can learn from the experience of another city where heatwaves are a common occurrence,
+leading to more proactive intervention plans to better deal with new climate conditions.
+
+Spatial analogues are estimated by comparing the distribution of climate indices computed at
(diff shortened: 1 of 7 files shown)


aulemahal:
This might be missing a notebook? That can also be done later.

xclim/analog.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+# ---------------------------------------------------------------------------- #
+# -------------------------- Utility functions ------------------------------- #
+# ---------------------------------------------------------------------------- #
+
+
+def standardize(x, y):
+    """
+    deviation.
```
huard, line 163: Missing description.

huard (approved):
Nice job. Agree about the notebook, let's create another issue for it, as well as another issue for the missing algorithms (low-priority).

aulemahal:
@huard  I have reunited both decorators in one. I thought that "drop when NaN" logic should be the same for all metrics. I.e. : if any value of either distributions (x, y) is NaN, return NaN, don't even bother to try the computation.  For some metric functions, doing otherwise was failing, for other it was returning np.Inf. I think it is now more consistent.

huard:
Good to go from my PoV. Let's defer further improvements with the notebook drafting. 
Thanks !

PR #606, opened 2020-12-02
Degree-day depassment date

### LINKED ISSUE #569: Implement degree-day depassment date indicator
#334 # Description

User interest for a calculation of date when the accumulated sum of degree-days reaches a user provided total.

E.g. the date when one reaches a sum of 100 degree-days (5 degC) starting from jan-01

I would suggest a 'generic' calculation allowing a single algorithm for all degree day types (Heating, cooling, growing etc etc). i.e. User provides parameter on the `op`  (<=, >=) as well as a `sum_threshold` (or similar), as well as all other params typically found in the degree day calculations already present.    

It would be interesting to see if there could be flexibility on whether the calculation is performed on daily tmin, tavg or tmax as well.

edit : Adding note that interested parties indicate that a `after_date` param would be ideal

aulemahal:
### Pull Request Checklist:
    - This PR fixes #569

Adds `xclim.indices.degree_day_depassment_date`, which returns the day-of-year when the cumulative degree-day sum exceeds `sum_thresh`. The degree-days are computed using `thresh` and in the `op` direction. i.e.: if  `op = '>'`, `dd = (tas - thresh).clip(0)`
and inversely.  

This definition makes the cumulative degree-day sum a monotonically increasing function. Which raises the question, do we really need an `after_date`? It could be replaced by a `clip` on the result. But, it wasn't much work anyway.

This PR also extracts `xc.indices.generic.compare` from  `xc.indices.generic.threshold_count`. I thought I would need it, but then I didn't.

Finally, now all run_lengths with dates accept `date=None`, which result not performing that check. All hail polymorphism.

No.

* **Other information**:

Indicators, doc and tests to come.

### DIFF
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -19,7 +19,7 @@ __all__ = [
     "cold_spell_days",
     "cold_spell_frequency",
     "daily_pr_intensity",
-    "degree_day_depassment_date",
+    "degree_days_depassment_date",
     "cooling_degree_days",
     "freshet_start",
     "growing_degree_days",
@@ -1369,7 +1369,7 @@ def tropical_nights(
 
 
 @declare_units("", tas="[temperature]", thresh="[temperature]", sum_thresh="K days")
-def degree_day_depassment_date(
+def degree_days_depassment_date(
     tas: xarray.DataArray,
     thresh: str,
     sum_thresh: str,
@@ -1380,7 +1380,7 @@ def degree_day_depassment_date(
     thresh = convert_units_to(thresh, "K")
     tas = convert_units_to(tas, "K")
     sum_thresh = convert_units_to(sum_thresh, "K days")
-    print(sum_thresh, thresh)
+
     if op in ["<", "<=", "lt", "le"]:
         c = thresh - tas
     elif op in [">", ">=", "gt", "ge"]:
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -19,6 +19,7 @@ __all__ = [
     "cold_spell_days",
     "cold_spell_frequency",
     "daily_pr_intensity",
+    "degree_day_depassment_date",
     "cooling_degree_days",
     "freshet_start",
     "growing_degree_days",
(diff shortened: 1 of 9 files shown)


aulemahal:
@tlogan2000  It this what we want?

tlogan2000:
> @tlogan2000 It this what we want?

Will try to review today or tommorrow

xclim/indices/_threshold.py
```diff
@@ -1365,3 +1366,31 @@ def tropical_nights(
...
+    thresh: str,
+    sum_thresh: str,
+    op: str = ">",
+    after_date: str = None,
+    freq: str = "YS",
+):
+    thresh = convert_units_to(thresh, "K")
+    tas = convert_units_to(tas, "K")
+    sum_thresh = convert_units_to(sum_thresh, "K days")
+    print(sum_thresh, thresh)
```
Zeitsperre, line 1383: This can be removed.

Zeitsperre (approved):
I _really_ like this. Very much what I had in mind. Well done, Pascal!
```
dd = degree_day_depassment_date(ds,thresh="5 degC", sum_thresh="500 degC days",freq="YS")
dd[0].plot()
plt.show()
```
Gives me exactly what I expect:
image

aulemahal:
Woups didn't meant to do that.

Zeitsperre (approved):

Zeitsperre:
@aulemahal Ignore the failing build. It'll be fixed in the next week or so.

xclim/indices/_threshold.py
```diff
@@ -1365,3 +1366,74 @@ def tropical_nights(
...
+    if op in ["<", "<=", "lt", "le"]:
+        c = thresh - tas
+    elif op in [">", ">=", "gt", "ge"]:
+        c = tas - thresh
+
+    def _depassment_date(grp):
+        return rl.first_run_after_date(
+            grp.cumsum("time") > sum_thresh,
+            window=1,
+            date=after_date,
```
tlogan2000, line 1436: I don't think this is doing what is intended with the `after_date`.  My understanding from discussion with interested parties would be that value indicates the beginning of when the `cumsum()` should start accumulating values.  

So I think we would want to assign 'nan' or zeros to `grp`  before `after_date` value then let `rl.first_run_after_date()` run without the `date` input

tlogan2000 (changes requested):
Looks good but I believe the `after_date` parameter is not doing exactly what we want. see comment in the code

xclim/indices/_threshold.py
```diff
@@ -1365,3 +1366,74 @@ def tropical_nights(
...
+    tas : xarray.DataArray
+      Mean daily temperature.
+    thresh : str
+      Threshold temperature on which to base degree days evaluation.
+    sum_thresh : str
+      Threshold of the degree days sum, in "K days".
+    op : {">", "gt", "<", "lt", ">=", "ge", "<=", "le"}
+      If equivalent to '>', degree days are computed as `tas - thresh` and if
+      equivalent to '<', they are computed as `thresh - tas`.
+    after_date: str, optional
```
tlogan2000, line 1396: Modify after_date definition to 'start date of degree day accumulation - default is the beginning of resample period' or similar

xclim/indices/_threshold.py
aulemahal: Oh I see! In my approach I thought a user would choose the starting date by passing the appropriate freq, like "AS-JUL" for 1st of july, etc. But I can see how that is limiting.

tlogan2000: That could work but I heard examples of some applications (Building codes or ??) that would  have super specific dates... "04-15" ... basically not always first of the month

tlogan2000 (approved):
Nice job.  FWIW seeing as TravisCI builds are not working I ran the full test suite on my VM

Zeitsperre:
Builds should work now. Will remove this bypass when the new clisops is released.

PR #608, opened 2020-12-03
Better quantity string parsing

aulemahal:
### Pull Request Checklist:
    - This PR fixes #606

Adds a `xc.core.units.str2pint` function that parses string into `pint.Quantity`, but instead of parsing them as _expressions_, it splits the magnitude and the units, casting the first to a float and parsing the second with pint's `parse_units`. This matches more closely what a user expects when writing things like "3 degC days". AFAIU, we meant quantity 3 of unit "degree­­·days" and not "3 times 1 degreeC times 1 day", which is what `parse_expression` does.

`units2pint` now _always_ returns a `pint.Unit` object and doesn't accept string with magnitudes (`units2pint('2 m')` doesn't work anymore). For the sake of polymorphism, `str2pint` also accepts strings without magnitudes (it catches them when the float can't be cast), but then returns a `pint.Quantity` of magnitude 1.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -36,6 +36,7 @@ Internal changes
 * With the help of the `mypy` library, added several typing fixes to better identify inputs/outputs, and reduce object type mutations.
 * Fixed some doctests in `ensembles` and `set_options`.
 * `clisops` v0.4.0+ is now an optional requirements for non-Windows builds.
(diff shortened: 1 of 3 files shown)


huard (approved):
Looks all good. Nice, clean solution.

PR #621, opened 2021-01-19
Ordered docstring parameters as in signature

huard:
Minor fix to docstring to match signature parameter order.


Zeitsperre (approved):
LGTM

PR #630, opened 2021-01-25
Remove explicit netcdf4 dependency

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

This removes the `netCDF4` library from `setup.py` and `environment.yml`. Seeing as it's not needed by `xarray` and we never use any of its methods explicitly, it shouldn't be there.

The NetCDF4 library is only used in one example on the docs, so it's been added and explicitly imported for that example. 

No. 

* **Other information**:


aulemahal (approved):
I held netCDF4 close to my heart and it didn't spark joy anymore. We can let it go.

Zeitsperre:
Almost done. Just need to fix these docs...

marie_kondo

aulemahal:
Oh!  The issue from RTD is actually related to the changes! 
In that complex sdba example, we are saving a dataset and it seems that one of the attributes is in unique and that `scipy.io.netcdf`, which uses netcdf 3, doesn't understand that.

Zeitsperre:
Alright, looks like we need to change some things around. I'll see what I can do.

Zeitsperre:
I've looked into this more, and it appears that all the `python-netCDF4` library does is direct changes to NetCDF files on disk. Seeing as we don't have any functions that rely on this functionality, we shouldn't be bundling it in Xclim.

huard:
I'm not sure I understand. xclim is meant to be used with netCDF files. If I install xclim, I expect to be able to read a netCDF file. no?

Zeitsperre:
@huard Sorry, I misread their docs (https://xarray.pydata.org/en/stable/faq.html?highlight=python-netcdf4#what-other-netcdf-related-python-libraries-should-i-know-about). It seems as though `python-netCDF4` is a redundant dependency nonetheless.

Zeitsperre:
@tlogan2000 @huard This is good to be merged, IMO.

tlogan2000 (approved):
ok fo rme

PR #634, opened 2021-01-27
Improvements for external parsers

### LINKED ISSUE #619: Various improvements for automated parsing of xclim
I am putting here a list of things that we could do on xclim to improve automated parsing of xclim. For context, the two main current use cases are [finch](https://github.com/bird-house/finch/) and [custom_climate_portraits](https://github.com/Ouranosinc/custom_climate_portraits/).

 - [ ] More uniform and general indicator names (ex : not `tropical_nights` but rather `tn_days_above`)
 - [x] Better distinction between input parameter types (unit str vs int vs choice of strings)
 - [x] Related to above : a way for the parser to retrieve the different choices of a param (ex: `op = {'>', '<'}`)
 - [ ] Rewrite abstracts/other metadata when the wrapping changes it (ex: `liquid_precip_accumulation` vs `precip_accumulation`)
 - [ ] Mechanism to add alternative function names
 - [ ] Mechanism to search through indices (free text in abstract, title, function name, keywords, etc)
 - [ ] Have `indicator.json()` return a json-able object.

Feel free to add anything I missed.

### LINKED ISSUE #631: xc.atmos.daily_temperature_range Exception with `op` Argument
* xclim version:0.22.0
* Python version: 3.8.3 64-bit
* Operating System: Win 8.1 64-bit, Win 10 64-bit

### Description
Per documentation at:
https://xclim.readthedocs.io/en/stable/indices.html#xclim.indices.daily_temperature_range

`xc.atmos.daily_temperature_range` should accept an `op` argument. However, when used:
```
da = xc.atmos.daily_temperature_range(ds.tasmax, ds.tasmin, freq='YS', op='mean')
```
There is an exception:
```
Exception has occurred: TypeError
got an unexpected keyword argument 'op'
```

aulemahal:
### Pull Request Checklist:
    - This PR fixes #619 and fixes #631.

## What kind of change does this PR introduce?** 
Changes that help external program better parse xclim. May seem large, but they have in fact low impact on the computation itself.

### Changes relating to input parameters of indices and indicators:
- ~`xclim.core.utils.UnitStr` type annotation, so that type checking can differentiate between string parameters (quantity+unit vs other)~ EDIT : The next item was more useful than this one, I decided to keep only one.
- `xclim.core.units.declare_units` stores the input unit dictionary into `indice.in_units`. The indicator constructor parses that and adds them to `indicator.parameters[x]['units']`.
- New type hints  in `xclim.core.utils`: `DateStr` and `DateOfYearStr`. To be used when in need of a string representing a date (YYYY-MM-DD, may contain a time HH:MM:SS) or a date of year (MM-DD, no time).
- `xclim.core.utils.InputKind` is an enumeration of constants, one for each input type recognized by xclim. The information is stored in `Indicator.parameters[<param>]['kind']`. It is not used by xclim itself, but is there to be parsed by external tools.
- `xclim.core.formatting.parse_doc` now parses the sets of choices from the docstring type annotation. When documenting a parameters like : ` op : {'mean', 'max'}`, the choices will be stored in `indicator.parameters[x]['choices']`. For now, it _evaluates_ the string, so only very basic Python objects are available : strings, numbers, boolean and None. ~Additionally, I allowed the use of `nan`, mapped to `numpy.nan`, even though it is not used currently.~

### Others
- `Indicator._format` and `Indicator.json` default to using the **default arguments** when  `args` is not given. (Instead of doing nothing and passing raw objects/formatting strings). Moreover, some objects are converted to simpler objects so that `Indicator.json` is _always_ serializable.
- Indicators have a generated docstrings! In `xclim.core.formatting`, I reconstructed the docstring from the parsed/given attributes.
- Indicators have their own API page. The "Climate Indicators" links to that page.
- Removal of all "UNSET" mentions in quantity strings.
- Added default values to some indices. Empty defaults can be difficult to manage by external tools (because json serialization forbids the use of `inspect._empty`, so we put `None` instead, but it doesn't mean the same thing.)
- Edited the `Fit` indicator so that it _fits_ what finch wants. (`missing=skip`)
- Move several indicator methods from public to private (renamed with a "_" prefix). So that the method exposed by an indicator object make sense to the user and that the online doc is not filled with unnecessary info for the users. Similarly, moved `_nvar` to `nvar`.

### Other
Yes. 

* **Other information**:
This is in sync with bird-house/finch#151.


aulemahal:
Maybe the `UnitStr` annotation is not needed, since units are written to an attribute of the indice itself?
Are there cases where we need a quantity as a string but do not use `declare_units`?

xclim/core/formatting.py
```diff
@@ -142,8 +143,17 @@ def parse_doc(doc: str) -> Dict[str, str]:
     return out
 
 
+annotation_patt = re.compile(r".*\{(.*)\}.*")
```
huard, line 146: Define inside function if not used elsewhere. rename `type_set_pattern` ?

aulemahal: I'm defining it outside because that's the point of `compile` : do it only once. However, the performance gain is negligeable, might juste be more elegant to do it inside.

aulemahal:
@huard  I added an item to #619 and to this PR: the `json()` method of most indicators was returning a non-jsonable dictionary (especially when passed `args=None`). I fixed this with some tweaks, especially : passing `args=None` to `json()` and `format()` will use the default arguments, instead of not formatting the attrs.

aulemahal:
I reconsidered and removed the UnitStr annotation. Indicators will base the unit inference solely on the use of the `declare_units` decorator (which adds a `in_units` dict to the indice).

I made a docstring generation function for indicator as to fix ambiguities like in #631. Now the docstring doesn't contain wrapped parameters.

Using those new docstrings, I documented all indicators within the docs. A new "indicators_api" page + links in the "Climate Indicators" list. This was more difficult than it seemed as `autodoc` refused to document any object that are not subclasses of builtin types. I used a fancy trick with a patch extension.  Took time, but I learned something.  
More specifically, the patch creates an "Autodocumenter" for indicator instances and a new "py:indicator" directive. They both inherit from the "function" base object, but this could allow us to customize the way the documentation are parsed and converted to html. (Like with the `py indicator` css class).

However, docstrings are still messy because of repeated info and incoherent syntax. We can fix that in a later PR.

This PR only lacks documentation addition for the `parameters` indicator attribute and maybe some tests.

aulemahal:
@tlogan2000 @Zeitsperre @huard 
One thing I'd like to have your opinion on in this PR is about indicators / indices docstrings.
Example : 
Docstring of indice `xclim.indice.tas`
```
Average temperature from minimum and maximum temperatures.

We assume a symmetrical distribution for the temperature and retrieve the average value as Tg = (Tx + Tn) / 2

Parameters
----------
tasmin : xarray.DataArray
    Minimum (daily) temperature [℃] or [K]
tasmax : xarray.DataArray
    Maximum (daily) temperature [℃] or [K]

Returns
-------
xarray.DataArray
    Mean (daily) temperature [same units as tasmin]
```
Docstring of indicator `xclim.atmos.tg`:
```
Average temperature from minimum and maximum temperatures. (realm: atmos)

We assume a symmetrical distribution for the temperature and retrieve the average value as Tg = (Tx + Tn) / 2

This indicator will check for missing values according to the method "from_context".
Based on indice `xclim.indices._conversion.tas`.

Parameters
----------
tasmin : DataArray or str
  Minimum (daily) temperature [℃] or [K]
  Default : `ds.tasmin`.[Required units : [temperature]]
tasmax : DataArray or str
  Maximum (daily) temperature [℃] or [K]
  Default : `ds.tasmax`.[Required units : [temperature]]
ds : Dataset, optional
  Input dataset.
  Default: None.

Returns
-------
tg : DataArray
  Daily mean temperature (air_temperature) [K]
  cell_methods: time: mean within days
  description: Estimated mean temperature from maximum and minimum temperatures
```

Notice the duplicated (and somewhat different) information about units of inputs.

I would suggest to remove explicit units specification in the description of indice's parameters, to avoid mismatching info from the `declare_units` decorator and that string?

Similarly for the default values that are sometimes there, sometimes not. I force them in the generated docstring, but that means that the info is sometimes duplicated.

Easy solution would also be to not add the parsed info and only leave the description.

xclim/core/indicator.py
```diff
@@ -249,12 +255,65 @@ def _parse_compute_and_docstring(cls, kwds):
...
+
+        for name in list(params.keys()):
+            if name not in kwds["_parameters"]:
+                params.pop(name)
+        kwds["parameters"] = params
+        return kwds
+
+    @classmethod
+    def _update_and_inject_arguments(cls, compute):
+        """Update comput argument and inject those shared by all indicators."""
```
huard, line 311:
```suggestion
        """Update compute argument and inject those shared by all indicators."""
```

huard (approved):
I was able to build the docs, looks very nice. 

Had to do
```
-SPHINXBUILD   = python -msphinx
+SPHINXBUILD   = sphinx-build
```
though, but probably related to my environment.

aulemahal:
Ouf! Well, coding this in sync with finch was a good idea @huard ! The top message of the PR was edited to summarize the numerous changes done here.

The most complex is the use of type hints (annotations) in order to assign a "kind" to each input parameters. This introduced two new types "DateStr" and "DateOfYearStr" and makes writing indices a bit more difficult since a strict type hinting syntax must be followed. 

Esthetic fixes to the doc must still be done.

huard (approved):

xclim/core/formatting.py
```diff
@@ -368,6 +363,22 @@ def unprefix_attrs(source, keys, prefix):
...
 
+KIND_ANNOTATION = {
+    InputKind.VARIABLE: "str or DataArray",
+    InputKind.OPTIONAL_VARIABLE: "str or DataArray, optional",
+    InputKind.QUANTITY_STR: "quantity (string with units)",
+    InputKind.FREQ_STR: "offset alias (string)",
+    InputKind.NUMBER: "number",
+    InputKind.NUMBER_SEQUENCE: "number or sequence of numbers",
+    InputKind.STRING: "str",
+    InputKind.DATE_OF_YEAR: "date (string, MM-DD)",
```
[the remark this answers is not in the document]
aulemahal: Indeed, `DAY_OF_YEAR` makes most sense, even with the ambiguity over the integer format.

Issue #632, opened 2021-01-27
Issue with percentile_doy


ChrisDrozdowski:
* Win 8.1 64-bit, Win 10 64-bit
* Python 3.8.3 64-bit
* numpy 1.18.4
* xarray 0.16.2
* dask 2021.1.1
* xclim 0.22.0

### Description
`percent_doy` function throws many warnings during call and takes what seems to be an excessive amount of time (>15 minutes) to run.

NC files for below example are available here: https://code.dented-planet.net/Indices.zip

#### Example code:
```
import xarray as xr
import xclim as xc
from xclim.core.calendar import percentile_doy

ds = xr.open_mfdataset(['tasmin.nc', 'tasmax.nc'])
tas = xc.indices.tas(ds.tasmin, ds.tasmax) 
t10 = percentile_doy(tas, per=0.1)
```
#### Repeated warnings during call:
```
C:\Users\C Drozdowski\AppData\Local\Programs\Python\Python38\lib\site-packages\dask\array\core.py:1428: FutureWarning: The `numpy.nanpercentile` function is not implemented by Dask array. You may want to use the da.map_blocks function or something similar to silence this warning. Your code may stop working in a future release.
```

tlogan2000:
Hello @ChrisDrozdowski thanks for posting the issue. 
Will try to take a look at your file and the `doy` calculation problems ... keep you posted

tlogan2000:
I might have a solution here but will need a bit of testing to be 100% certain

If interested : Our original implementation of `percentile_doy` used an `xr.reduce(np.nanpercentile, )` call.  Changes in dask seem to be the source of the FutureWarning and major slowdown 

However I see that since our original work on `percentile_doy` there is now a  native xarray `quantile()` method that can receive a skip nan  parameter ... seems to work and is not painfully slow

Will validate that calculations are identical to orginal and try to get a fix implemented and into the master relatively quickly

tlogan2000:
@ChrisDrozdowski PR #633 has your test code executing in <2 seconds on my vm ...  Seems more reasonable :)  
(Wasn't patient enough to let finish with the original code so can't give a true comparison)
Thanks again for flagging.

Fix will be merged to master branch in a few minutes

Zeitsperre:
@tlogan2000 @huard Would it make sense to push a new patch release? I can publish one pretty quickly before I end today if we want.

aulemahal:
[...]

PR #638, opened 2021-01-28
Fix 632 b

### LINKED ISSUE #635: Explore xr.quantile() performance
> Oh goodness, I responded to the wrong Issue!!!! I am truly sorry. Multitasking never works well for me!
> I meant to respond to the closed "Issue with percentile_doy #632 ". That's the one I'm pushing for a release on.
> Please forgive me.

_Originally posted by @ChrisDrozdowski in https://github.com/Ouranosinc/xclim/issues/631#issuecomment-768607590_

tlogan2000:
### Pull Request Checklist:
    - This PR fixes #635

Performance enhancement

Yes see history.rst

* **Other information**:
This continues on the percentile_doy performance issues but hopefully provides much better real world performance


tlogan2000:
@aulemahal might get your thoughts on whether this totally makes sense?   In particular the `fill_ends` param pads nans the first and last window with the mean of the window... not sure this is the best choice..  

Even if we discard the `fill_ends` idea performance is typically much faster than a straight xr.quantile(skipna=True) that we had before but the combined is extra impressive

xclim/core/calendar.py
```diff
@@ -453,10 +459,40 @@ def percentile_doy(
     #  This is supported by DataArray.quantile, but not by groupby.reduce.
     rr = arr.rolling(min_periods=1, center=True, time=window).construct("window")
 
-    # Create empty percentile array
-    g = rr.groupby("time.dayofyear")
+    # rolling automatically adds some nans to the windows on the time-series edges
+    # Optionally fill nans in first and last n=floor(window/2) time steps with average window values for these
+    if fill_ends:
+        nstep = floor(window / 2)
```
tlogan2000, line 465: This is not my proudest piece of work... is there a more elegant way to pad the data with the mean

tlogan2000:
Small test script in case interested
````
import xarray as xr
from xclim.core.calendar import percentile_doy
import time
import xclim.subset as subset
from dask.diagnostics import ProgressBar

url = 'http://pavics.ouranos.ca:8083/twitcher/ows/proxy/thredds/dodsC/datasets/simulations/bias_adjusted/cmip5/ouranos/cb-oura-1.0/day_CanESM2_historical+rcp85_r1i1p1_na10kgrid_qm-moving-50bins-detrend_1950-2100.ncml'

tas = subset.subset_bbox(xr.open_dataset(url, chunks=dict(time=256, lon=16)), lon_bnds=[-90, -60], lat_bnds=[45, 48], start_date='1981',end_date='2010').tasmin

#add a nan for a day of year
#tas = tas.where((tas.time.dt.dayofyear != 21))

## Default should produce exactly as before
time1= time.time()

t10 = percentile_doy(tas, per=0.1, window=5)
t10.load()


print(time.time() - time1)


time1= time.time()
## Using fill ends dayofyear 1 and 365 (366?) could change slightly do to padding

t10 = percentile_doy(tas, per=0.1, window=5, fill_ends=True)
t10.load()


print(time.time() - time1)
````

aulemahal:
I have a branch on `xclim-benchmark`, work is being done as we speak (don't tell my supervisor).

tlogan2000:
> I have a branch on `xclim-benchmark`, work is being done as we speak (don't tell my supervisor).

On my end  : Original file provided in #632 now running in 19s (with a .load()  :)  )
3s if I apply my `fill_ends` trick..

xclim/core/calendar.py
aulemahal: This is assuming we _always_ want to skip NaNs.

tlogan2000: I was trying to avoid the case of filling all nans... These ones are a simple artifact resulting from the n=5 window construction.  Not sure if is reflected in the tests for `calendar` but in an application with 30 years of data :  dayofyear[0] there will have two nans tacked on to the beginning of window dimension but only for the first year of the `rr` variable.  My idea was to replace nan with only the mean of the 3 valid days of single year/dayofyear window and not the mean of all years.   Originally thought of just a bfill() for the begginning of the timeseries and ffill() for the end but the mean seemed like a better choice.

tlogan2000: In a case where there are other nans present in the data ... i.e. actual missing day in the original timeseries the masking will catch it and use xr.quantile(skipna=True) but again only where it really needs it.  Overall making things much faster as long as the data is relatively complete

tlogan2000: In a 3d case (time, lat, lon) I think this means that even with "perfect data" (i.e. no nans) there is minimally `n=(floor(window/2) * 2) * len(ds.lon) * len(ds.lat)` cases where the slower `xr.quantile(skipna=True)` call is used.  As spatial domain gets bigger the number of these cases grows quickly and there will likely be increased difference between `fill_ends=True`  and `fill_ends=False`

aulemahal (changes requested):
This solution is not lazy. If I am not mistaken, direct calls to the numpy API are sometimes lazy, but most often not.
I'll push another solution making use of the fact that we _always_ want to ignore NaNs.

aulemahal:
Here is my solution. I think the fillna-with-the-mean trick was reasonable for long timeseries and only when done at the ends, but my first idea of doing it everywhere was conceptually wrong.

As always, it all finally ended with the inevitable `xr.apply_ufunc`. I reused the same `_calc_perc` as in  `ensembles_percentiles`, so I moved that snippet to `xclim.core.utils`. In my benchmark (which I will push tomorrow), I have worse performance than this thread's initial solution, but : 
- it is lazy
- it doesn't do the fill-the-ends-with-the-mean trick

However:
- I modified the `per` arg to accept a sequence as well as a scalar. As such, the output as a `percentiles` axis, which could be a breaking change.
- I modified `per` to be 0-100, not 0-1. Which is a breaking change, but is more coherent with (1) our use of the word "percentile", (2) the same argument in `ensemble_percentiles`.
- The use of the MultiIndex-unstack trick is quite nice, but it requires daily datasets. The groupby method could accept any sampling. Another breaking change.


So. I suggest we wait and discuss this before releasing a patch.

aulemahal:
@tlogan2000 this is your PR and I can't ask for a review, but please do!

tlogan2000:
> @tlogan2000 this is your PR and I can't ask for a review, but please do!

I'll try to look this morning...
Quick questions:
* Did you try with hourly data? ... I didn't consider this at all but would have thought that the unstack would give us an array with repeated indexes (24 of the same dayofyear*year)?  Maybe a way to adapt for hourly without too much effort
* The non lazy part is due only to the if `np.any(~mask)` check right?  This is not even necessary if we remove the `fill_ends`
* Can we simply use the apply_ufunc  on the previous groupby object?

aulemahal:
The MultiIndex trick limits it to daily data (or coarser). We might be able to get it working with hourly data, but it would be yet another workaround... I'll try groupby+apply_ufunc to fix that, but I expect performance to go down.
You are right about the non-laziness source, once it is hidden in the apply_ufunc, the issue is fixed. It is still needed, since the goal is to re-perform nanpercentile only where there are NaNs.

aulemahal:
@tlogan2000 Groupby+apply_ufunc does work, removing the sampling limitation. It currently have a version that is 5-6x slower than either our implementation. 
Would it be too much to have the 2 versions? With a switch like  `if xr.infer_freq(ds) in daily_or_coarser`.

tlogan2000:
> @tlogan2000 Groupby+apply_ufunc does work, removing the sampling limitation. It currently have a version that is 5-6x slower than either our implementation.
> Would it be too much to have the 2 versions? With a switch like `if xr.infer_freq(ds) in daily_or_coarser`.

I'm ok with it  but how generic do we have to be do think?

If we only want to support hourly and coarser I think we could maybe simply try to build the index via :
````
if infer_freq() is 'subdaily'  :

   ind = pd.MultiIndex.from_arrays(
        (rr.time.dt.year.values, rr.time.dt.dayofyear.values, rr.time.dt.hour.values),
        names=("year", "dayofyear","hour"),
    )

   rrr = rr.assign_coords(time=ind).unstack("time").stack(stack_dim=("year", "window","hour"))
````

aulemahal:
I'm not sure. Maybe @huard and @Zeitsperre have opinions on this.
I was thinking of a hypothetic use case where you would compute the percentile_doy from messy observational data : lots of missing values, inconsistent time sampling, etc. And then compute the indicators using those percentiles?
Might be very uncommon.

My argument would be: if we are to have more then one version, let's stick to :  The most common use case + The most general algorithm.

tlogan2000:
> I'm not sure. Maybe @huard and @Zeitsperre have opinions on this.
> I was thinking of a hypothetic use case where you would compute the percentile_doy from messy observational data : lots of missing values, inconsistent time sampling, etc. And then compute the indicators using those percentiles?
> Might be very uncommon.
> 
> My argument would be: if we are to have more then one version, let's stick to : The most common use case + The most general algorithm.

Ok lets go for the more genreric method.  I agree the majority of users will have / be thinking of daily inputs for a dayofyear calc ... 
We should minimally update the docs to say the window is the number of timesteps but there a way we can add an adjustment to the window parameter to make sure it is applying a rolling in days and not timesteps? Something like:
````
infreq = xr.infer_freq(arr.time)
infreq = infreq if len(infreq)>1 else f"1{infreq}"

window *=  round(pd.to_timedelta('1D') / pd.to_timedelta(infreq))
````
This won't be great for  handling data at a coarser frequency than daily but I don't really see when you would use it that way

Zeitsperre:
> I'm not sure. Maybe @huard and @Zeitsperre have opinions on this.
> I was thinking of a hypothetic use case where you would compute the percentile_doy from messy observational data : lots of missing values, inconsistent time sampling, etc. And then compute the indicators using those percentiles?
> Might be very uncommon.

What I will say is that if we do aim to use this data on subdaily time-steps, obs data, regardless of sources, is FILLED with holes. That was one of the things that prevented us from using raw data in our analyses vis-à-vis ScenGen. I can see enormous value from being able to send this stuff into Xclim to get something reasonable out of it. I'm not sure how badly this complicates things but right away, that would be a valuable functionality for us, I think.

I wish I could weigh with more, but this discussion is a little too in the weeds for me. I trust the approach you guys take will be reasonable (and documented!).

tlogan2000:
@aulemahal new updated tests added if you want to run tox

tlogan2000:
Last comment for discussion.  Got super confused and had a discussion with @aulemahal and I think neither of us is 100% convinced that this calculation is really valid on a subdaily frequency.  E.g. The most clear use case is to get a daily mutliyear climatolgy or normal for individual days (hence the window operation to reduce noise). So what does this value represent if calculated on say hourly data?  I.m not sure it is invalid but it seems a bit strange.  

So decision is to implement the fix with mandatory daily or coarser frequency as input.  Performance is greatly improved using the current code and should respond to the majority of applications...  can explore the possibility of introducing sub-daily support if /when needed.

aulemahal:
As always, there were surprises. Sadly, comparing offset strings is more complicated than it looks, so I created a function for it.

Tests for T90p were broken? I have no idea how they were passing before. (see changes to `test_temperature.py`).

All tests passing.

aulemahal (approved):

ChrisDrozdowski:
@developers,

I installed 0.23.2b0 last night and the speed of `percentile_doy` for my story is vastly improved. Thanks for the effort.

PR #642, opened 2021-02-02
Standardize units managment in indices

### LINKED ISSUE #639: Using xr.set_options(keep_attrs=True) and Units Parsing Issue
* xclim version: 0.23.2b0
* Python version: 3.8.3
* Operating System: Win 8.1 and Win10 64-bit 

### Description
Using `xr.set_options(keep_attrs=True)` causes an exception in indicator call(s). See script below. If removed, no exception.

Is this expected and, as far as you know, will removing `xr.set_options(keep_attrs=True)` produce proper results?

Please download this file and put into same folder as script:
https://code.dented-planet.net/pr.nc

```
import xarray as xr
import xclim as xc

xr.set_options(keep_attrs=True)

ds = xr.open_mfdataset(['pr.nc'], engine='netcdf4')
da = xc.atmos.wetdays(ds.pr, thresh='1 mm/day', freq='YS')
```
Result: Exception

aulemahal:
### Pull Request Checklist:
    - This PR fixes #639

This explicitly set the 'units' attribute for several indices where it was assumed that the computation would remove them. When using `xr.set_options(keep_attrs=True)`, simple xarray operation will keep the original attributes (duh) and that was problematic for some of the indices. The most common example is all those "threshold-count" indices : the "compare - resample - sum" chain of operation will retain attributes when the option is activated. However, units change from whatever the input was in to "days".  In the `declare_units` decorator, if units are set on the output, it converts the output to the required units. The error was coming from there : it received "degK" or "mm/day" instead of nothing.  When units are not set on the output, it simply sets the required ones.   This PR fixes those cases, and some others.

No.

* **Other information**:
I did not find a way to properly test this without adding a test for _all_ indices, which I was to lazy to do. Thus, if xarray changes its behaviour, the issue could happen again with the remaining indices that do not set the units explicitly.

While fixing this, I noted a few things:

### 1) Do we really want `declare_units` to set the units AND test if they are already set?

I think it sould do either, but not both. I.e. : Either _declare_ units and override whatever is there or not. **Or** act a as metadata check : the unit-setting action has to be done in the indice.

### 2) Threshold-count methods could be more flexible.

Now that we have `xr.infer_freq`, we could open all threshold-count indicators to accept other sampling (regular) frequencies than daily. `infer_freq` could give the base unit and when the sum in done we put that as units (H => "hours", D => "days", etc). Offsets with multiples (ex: "3H") could be managed by multiplying the output by that factor. 
Not sure if there are many use cases though.


aulemahal:
I'm currently running tox.

@Ouranosinc/xclim-core I raised two issues in this PR, please chime in if you have time and opinions.

huard:
So we have two categories of indices: 
- unit changing (UC)
- non unit changing (NUC)

and two contexts:
- keep_attrs=True (KA)
- keep_attrs=False (NKA)
 
UC requires explicit unit specification. 
- NKA: Done by declare_units
- KA: Failure -> requires change to indice function

NUC
-NKA: Done by declare units
- KA : OK. Declare units can check dimensionality, but this is typically redundant since we're checking input units already. 

I suspect all the degree day indices will need the same approach, no ?

Your question 1 is still open. Probably worth a discussion.

aulemahal:
@tlogan2000 @Zeitsperre surtout

After discussion with David, decision has been made to make the following conceptual change to indices and the `declare_units` decorator: 

Indices **must** take care of assigning the proper units to their output. The `declare_units` decorator only checks the input data and stores information about the required input units/dimension. It doesn't modify the output.

The problem was that `declare_units` was acting on 2 levels : checking for user errors and for developer errors, the latter being undesirable. It's role in assigning units when they were not present on the indice output was prone to error in some cases, especially with all the  `keep_attrs` business, as raised above. To keep the unit flexibility* in indices, implementing a "smart" unit management code outside of the individual indice seemed to complex a task, so we decided to leave it to the indice developper to think, test and solve all unit cases.

\* By unit flexibility, I mean the possibility, for example, to input tas in °F and retrieve °F, in a simple temperature indice. Or even input "°F" and retrieve "delta_F days" in the output of a degree-days indice. 

What that means is numerous little changes to ensure good units on indices ouputs and the removal of the `out_units` argument of `declare_units`. Output units will have to be explicitly explained in the docstring.

tlogan2000:
I'm not sure I completely agree with this .. but might need a more concrete example. 

Suppose I set a xr.set_options(keep_attrs=True) globally then proceed to calculate xci.wetdays on precip. What is in the output as a units attr?   If it remains mm/s then I think this is a bit of a step back for the code.

aulemahal:
The goal is to be independent of the `keep_attrs` setting. Your example currently fails if `keep_attrs` is True. We want to force the unit attribution in the indice so that doesn't happen. The wetdays example is simple because that means only a simple : `out.attrs['units'] = "days"`, as done in the PR. For more complex cases, like degree-days, a bit more thought must be put into the code, in order to keep the flexibility in input units.

tlogan2000:
> The goal is to be independent of the `keep_attrs` setting. Your example currently fails if `keep_attrs` is True: `declare_units` raises an Error as it receives "mm/s" units.
> We want to force the unit attribution in the indice so that doesn't happen. The wetdays example is simple because that means only a simple : `out.attrs['units'] = "days"`, as done in the PR. For more complex cases, like degree-days, a bit more thought must be put into the code, in order to keep the flexibility in input units.

Ok I think I have my head around it now.

aulemahal:
Ready for review!

This PR will break everything that expected a specific unit in the output of indices, that might be a lot! The indicators are not impacted, and that API should be considered more stable anyway.

So now indicators take care of attributing the final units themselves. In a lot of cases, this mean more flexibility than before, often more than the title suggest. I added a few functions to help:

 ####  `xclim.core.units.to_agg_units` 
Infers the sampling frequency of the input and uses that to get the appropriate units of the aggregation. Example : if a "count" operation was applied along a daily dataset, like say "tx_days_above", the output unit is "day". The same operation on hourly data returns "h". It also covers the "6H" case by multiplying the data by 6 and setting "h" as units.

Similarly for "delta_prod", for operations like `(a - thresh_a).integrate('time')`, such as degree­ days. Daily temperature in "K", gives "K d". Or 6-hourly temperature in "°C" gives "h delta_degC".  Note that I modified `pint2cfunits` so that `delta_degree` units are kept as such. It should be cf-compliant and reduce unit issues.

This function allows a lot of indices that were written for daily data to be more flexible and work with any sampling frequency that is convertible to a unit (ex:  QS-DEC doesn't work, "season" is not a temporal unit.). 

For now, it defaults to assuming daily input when the frequency inference fails (`xr.infer_freq(da) is None`). As such, it doesn't change the previous behaviour, but that seems inelegant at best for me...  Should we raise an error instead? That would mean that badly constructed timeseries would fail. Main case : data converted to a different calendar using `convert_calendar` but without padding missing dates with NaN. I.e : a "365_day" converted to "366_day" will miss the 29th of Feb and its frequency will not be inferrable.

#### `xclim.core.units.rate2amount`
Useful to go from precipitation rate to amount. Multiplies by the sampling period (daily => `* 1 d`). If the sampling period cannot be found (`xr.infer_freq -> None`), computes it point by point (`da.time.diff('time')`). Units are reduced on the output (` "mm / d" * 1 h => "mm").


 ### Docstrings
Docstrings have been modified so that the output dimension or units are listed in the "Returns" section, after the type annotation. I didn't write these in a parsable way, but rather a the most specific indicator I can give given the new flexibility.
Ex:
```
Returns
----------
xarray.DataArray, [dimensionless]
   <Long name ....>
# or
xarray.DataArray, [same as tasmin]
# or
xarray.DataArray, [temperature]
"""

**Comment by aulemahal** [comment]
Passes all tests on all flavors, except py37-bottleneck (which is broken), py37-windows and macOS (duh).
@Zeitsperre  in the last commit, I changed the config of pydocstyle to skip the whole /tests/ directory. Also changed the "black" tox env so that it installs xclim normally, pylint was raising import-error issues everywhere otherwise. Oh and excluded the `tests` directory in black.

**Comment by Zeitsperre** [comment]
> @Zeitsperre in the last commit, I changed the config of pydocstyle to skip the whole /tests/ directory. Also changed the "black" tox env so that it installs xclim normally, pylint was raising import-error issues everywhere otherwise. Oh and excluded the `tests` directory in black.

Thanks for running `tox` against a bunch of environments. It's been very helpful to have that in the absence of Travis CI. For `pylint`, given that tests are installed as a pseudo-module, it requires that Xclim be installed. I can see that it's in the `doctests` build already, so rather than install `xclim` in the `black` build, just remove `pylint` from `black`. Surface-level code style checks shouldn't require `xclim` to be installed.

Excluding tests from `black` isn't what I'd do, but I don't see an issue with doing that. LGTM.

**Comment by aulemahal** [comment]
> Excluding tests from `black` isn't what I'd do,

Woups you are right, I didn't think this completely through... Last commit also removed pylint from the black run and skipped the installs. Thanks!

**Comment by aulemahal** [comment]
@huard Review when you have time, also I think I'll prefer to merge this first before finishing #634 as they both act on the same code. I'll start testing against finch after merging this also.

As a summary of difference between the two PRs: 

 -  This PR tries to maximize unit flexibility. It removes output unit declaration and makes sure units are handled and set explicitly (or almost) in the indice function. The output unit information is moved to the "Returns" section of the docstring, but in a human-readable manner only.
 - #634 propagates input units info to the indicator. It removes all input units info from the indice docstring but includes them in new generated indicator docstrings.

**Review by tlogan2000 — APPROVED** [review]
This seems good to me. Working as expected I beleive.  Might still be good to have @huard take a look as there are quite a few changes here

File: xclim/indices/__init__.py [hunk 1]
```diff
@@ -46,22 +46,35 @@
...
-     e.g. The 10th percentile of historical temperature [K].
-   freq : Optional[str]
-     Resampling frequency.
-
-   Returns
-   -------
-   xarray.DataArray
-     Output's <long_name> [units]
+   @declare_units(var1="[units dimension]", thresh="[units dimension]")
+   def indice_name(var: xr.DataArray, threshold: str = "0 degC", freq: str = "YS"):
```
huard, line 54:
```suggestion
   def indice_name(var1: xr.DataArray, threshold: str = "0 degC", freq: str = "YS"):
```

xclim/indices/__init__.py
```diff
@@ -46,22 +46,35 @@
...
+       \"\"\"
+       The first line is the title
+
+       The second paragraph is the abstract.
+
+       Parameters
+       ----------
+       <var_name> : xarray.DataArray
```
huard, line 62: Harmonize var, var1, var_name.

xclim/indices/__init__.py
```diff
@@ -46,22 +46,35 @@
...
+         <var_name> is a short name like "tas", "pr" or "sfcWind".
+       threshold : str
+         Description of the threshold (no need to specify units or the default again).
+         Parameters required to run the computation must always have a working default value.
+       freq : str
+         Resampling frequency. (the signature carries the default value information)
+
+       Returns
+       -------
+       <var_name> : xarray.DataArray, [ouput units dimension]
```
huard, line 73: You don't need to name the output.

xclim/indices/_anuclim.py
```diff
@@ -191,11 +191,11 @@ def precip_seasonality(
     with xarray.set_options(keep_attrs=True):
         seas = 100 * _anuclim_coeff_var(pr)
 
-    seas.attrs["units"] = "percent"
+    seas.attrs["units"] = "%"
```
huard, line 194: Are you sure this works ? % gave me trouble in the past.

xclim/indices/__init__.py
huard: https://numpydoc.readthedocs.io/en/latest/format.html#sections

xclim/indices/_anuclim.py
```diff
@@ -317,7 +319,7 @@ def prcptot_wetdry_quarter(
 
     Returns
     -------
-    xarray.DataArray
+    xarray.DataArray : [mm]
```
huard, line 322: This will be interpreted by sphinx as name: type

xclim/indices/_anuclim.py
```diff
@@ -376,8 +381,8 @@ def prcptot_warmcold_quarter(
 
     Returns
     -------
-    xarray.DataArray
-       Total precipitation values of the warmest/coldest quarter of each year.
+    xarray.DataArray : [mm]
```
huard, line 384: again, not sure this will play well with sphinx autodoc.

xclim/testing/tests/test_indices.py
```diff
@@ -114,7 +114,7 @@ def test_simple(self, tasmin_series):
 
         out = xci.cold_spell_duration_index(tn, tn10, freq="YS")
         assert out[0] == 10
-        assert out.units == "days"
+        convert_units_to(out, "days")
```
huard, line 117: Why remove the explicit check ?

xclim/core/units.py
```diff
@@ -302,6 +314,110 @@ def convert_units_to(
...
+    elif op == "prod":
+        out.attrs["units"] = pint2cfunits(orig_u * freq_u)
+    elif op == "delta_prod":
+        out.attrs["units"] = pint2cfunits((orig_u - orig_u) * freq_u)
+    else:
+        raise ValueError(f"Aggregation op {op} not in [count, prod, delta_prod].")
+    return out
+
+
+def rate2amount(
```
huard, line 380: I see no explicit test targeting this function. I understand it's tested through indices, but I suggest we check corner cases with a dedicated test.

xclim/core/units.py
```diff
@@ -302,6 +314,110 @@ def convert_units_to(
...
+    rate: xr.DataArray, dim: str = "time", out_units: str = None
+) -> xr.DataArray:
+    """Convert a rate variable to an amount by multiplying by the sampling period length.
+
+    If the sampling period length cannot be inferred (i.e. : it is not consistent), the rate values
```
huard, line 385:
```suggestion
    If the sampling period length cannot be inferred, the rate values
```

xclim/core/units.py
huard: Could you also include a conceptual example in the docstring to clarify what this does in practical terms.

xclim/core/units.py
```diff
@@ -302,6 +314,110 @@ def convert_units_to(
...
+    try:
+        return int(multi or "1"), FREQ_UNITS[base]
+    except KeyError:
+        raise ValueError(f"Sampling frequency {freq} has no corresponding units.")
+
+
+def to_agg_units(
+    out: xr.DataArray, orig: xr.DataArray, op: str, dim="time"
+) -> xr.DataArray:
+    """Set and convert units of an array after an aggregation operation along the sampling dimension (time).
```
huard, line 349: Is seems it also changes the data itself no ? (out = out * m)

This function should have a dedicated test. 
Also please include a conceptual example of what it does.

huard (approved):
Nice piece of work. Just some nitpicking. It's ok if my comments are handled in a another PR to avoid stalling the development flow.

xclim/indices/_anuclim.py
aulemahal: Well you are kinda right, I realize my tests worked because you (or someone else) already put the workaround in `units2pint`. I can put "percent".  I tested and cfunits recognizes both '%' and 'percent', but not 'pct'.

xclim/indices/_anuclim.py
aulemahal: Woups, should be a ",".

xclim/testing/tests/test_indices.py
aulemahal: Because the function returns "d" as units, because of the way it is converted from a pint Quantity with `f"{q:~}"`. The specific string of the unit could change if we decide "days" is the default str? I just wanted this to be a bit more flexible.

huard: but this will work even if the units is hour

aulemahal:
@huard Added doc examples where needed.
@tlogan2000  I realized I didn't finish the standardizing work in `_anuclim.py`, it is done now. I also saw two indices that were computing total precipitation using the `src_timestep` arg and multiple conditionals, all to convert rates to amounts. I replaced the blocks with my new `rate2amount` function, that should hopefully behave the same. Except that it _infers_ the source timestep OR computes it from the time axis.


And for all: I made another "breaking" change : Instead of relying to what is hardcoded in the functions, `declare_units` now reads the output's units, parses them with pint and rewrites them to string with `pint2cfunits`. This ensures more consistency. We can discuss that in a few minutes.

PR #644, opened 2021-02-15
Use GitHub CI

Zeitsperre:
### Pull Request Checklist:

This implements a Linux-only set of build checks using the GitHub CI actions. This is loosely based off of the recipes implemented in other supported projects (via @cehbrecht, thanks!). The builds here replicate those that we had working previously and leverages `tox` for all stages. Assuming that everything works relatively well out-of-the-box, we can leverage this for a handful of projects with a limited number of builds.

No.

* **Other information**:

https://github.com/roocs/clisops/pull/124#event-4331983475


.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install packages
+      run: |
+        sudo apt-get -y install pandoc
+      if: matrix.python-version == 3.6
```
tlvu, line 17: how should I read this syntax format?  Install `pandoc` only for python 3.6?

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
```
tlvu, line 19: I guess these `uses` are basically means calling a script, in this case the script is `actions/setup-python` at version `v2?

.github/workflows/main.yml
tlvu: Is the indentation wrong?  Compared to other `if` below?

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Lint
+      run: make lint
+      if: matrix.python-version == 3.6
+    - name: Test with pytest
+      run: |
+        pytest --cov xclim -m "not slow"
+    - name: Coveralls
+      uses: coverallsapp/github-action@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
```
tlvu, line 37: Where is this token defined?

tlvu (approved):
LGTM, just a few questions.

.github/workflows/main.yml
Zeitsperre: That's exactly it.

.github/workflows/main.yml
Zeitsperre: Exactly. This `if` is for GitHub actions, whereas the `if`'s below are for the shell.

.github/workflows/main.yml
Zeitsperre: It's generated on as-needed basis: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token. Each job generates a token and once completed, that token is destroyed. Coveralls can therefore not accept more runs for this build once this build has run to completion.

aulemahal (approved):
Nice! 
For the doctests and slow tests, I think I would suggest removing the normal tests from the `doctests` suite and running the slow tests in  `py38`. Just to reduce the load a bit more and have it a bit more consistent with the naming?

EDIT : And do we need the  `docs` env if we keep the RTD build?

Zeitsperre:
This is good to merge. Thanks for the reviews @aulemahal @tlvu

PR #655, opened 2021-03-11
Add snow cover indicators

### LINKED ISSUE #604: Winter indices
- [ ] **Days with Low amount of snow** (0.2 to 1.9 cm): Count of days in a cold season-centered year with
snowfall between 0.2 and 1.9 cm; suggested from studies related to snowfalls in the context of winter
road maintenance 
- [ ] **Days with Moderate amount of snow** (1.91 to 4.9 cm): Count of days in a cold season-centered year
with snowfall between 1.91 and 4.9 cm; suggested from studies related to snowfalls in the context of
winter road maintenance 
- [ ] **Days with High amount of snow** (&gt; 4.91 cm): Count of days in a cold season-centered year with snowfall
greater than 4.91 cm; suggested from studies related to snowfalls in the context of winter road
maintenance 
- [ ] **Rainfall with low temperatures**: Daily rainfall ≥ 0.4 mm and Tmin &lt; -0.2 °C 
Intense Rainfall followed by Freezing: Daily Rain &gt; 10 mm in the same day that Tmin &lt;-0.2°C 
- [ ] T**iming of Max. Snow Depth**: Date of occurrence of the maximum snow depth during a cold season-
centered year 
- [ ] **Snow cover duration**: Annual number of days in a cold season-centred year with snow depths ≥ 2 cm 
- [ ] **Start Date of Continuous Snow Cover (SDCSC)**: First day of the 14 consecutive-day period with snow
depths ≥ 2 cm for a cold season-centered year 
- [ ] **End Date of Continuous Snow Cover (EDCSC)**: First day after the SDCSC, when the snow depth &lt; 2 cm
for 14 consecutive days for a cold season-centred year 
- [ ] **Annual Max. 3-Day Snow Melt Water Equivalent**: Annual max. of the &quot;melt&quot; - decrease in snow depth -
over 3 days during period with snow cover 
- [ ] **Max. 3-Day Sum of Snow Melt and Precipitation**: Annual max of the sum of &quot;melt&quot; - decrease in snow
depth - over 3 days and precipitation during period with snow cover 
- [ ] **Blowing snow**:  Number of days (periods) with Snowfall accumulations of previous three days ≥ 5 cm
AND Wind speeds ≥ 15 km/h 
- [ ] **Winter storm**: number of events with more than 25 cm of snowfall accumulation in a 24 hr period.

huard:
### Pull Request Checklist:
    - This PR fixes #604

Add indicators:
 - snow_cover_duration
 - continuous_snow_cover_start
 - continuous_snow_cover_end
 - blowing_snow
 - winter_storm
 - snow_melt_we_max
 - melt_and_precip_max
 - snd_max_doy
 - high_precip_low_temp
 
No

* **Other information**:


huard:
@aulemahal I'd like your views on how to implement `continuous_snow_cover_end`, which is defined as a first_run_after_date, for the date found by continuous_snow_cover_start (which is different for each year).

aulemahal:
I have three ideas:

Currently we want :  first day of first run of X no-snow days after at least one run of X snow days

Simple but not the same: use `last_run`?  We'd get "last day of the last run of X snow days""

A bit more complex then your idea: compute seaon_length and add it to first_run. Then take care of the end-of-year mess.

Yet-another-utility : Implement `season_end`. This should be an easy task : we need to modify "run_end_after_date" for the case with no date.

I can do the last one quickly (I think).

huard:
1. Does match the definition we're targeting.
2. That would work, but yeah we'd need to modulo the result with the number of day in that year with that calendar. Not too bad. 
3. Note that we're still looking for the start of the run, but after a date given as a julian day. 

I'll try my hand with 2 and call for help if I stall.

huard:
Now I see what you mean. Yes, season_end is probably the best option. Looking into it.

aulemahal:
Ouf. That was more complex than we first thought. But xclim now has a `run_length.season(...)` utility that returns a Dataset with three variables : `start`, `end` and `length`. The first two are in "units" controlled by a "coord" kwarg. It acts the same as in `first_run` and friends : `True` returns the index value (a datetime) and something like `'dayofyear'`  returns `da[dim].dt.dayofyear`. `False` returns an index, which is almost useless when this is wrapped in a `groupby`.

`length` is always the number of steps (days in a daily ds) between beg and end (as [beg, end[ ).  I would have liked to have `length` in timedeltas but : pydata/xarray#5026.

huard:
Forgot to update HISTORY. Will do it with the next PR coming up shortly.

aulemahal (approved):

xclim/indices/_multivariate.py
```diff
@@ -1417,3 +1464,53 @@ def winter_rain_ratio(
...
+    freq: str = "AS-JUL",
+) -> xarray.DataArray:
+    """
+    Days with blowing snow events
+
+    Number of days where both snowfall over the last days and daily wind speeds are above respective thresholds.
+
+    Parameters
+    ----------
+    snd : xarray.DataArray
```
tlogan2000, line 1487: This calculation uses what seems like a complicated way to derive snowfall  (snd.diff() etc) ...  Could we not get it from prsn?

xclim/indices/_threshold.py
```diff
@@ -1584,3 +1722,40 @@ def _exceedance_date(grp):
...
+      Number of days per period identified as winter storms.
+
+    Notes
+    -----
+    Snowfall accumulation is estimated by the change in snow depth.
+    """
+    thresh = convert_units_to(thresh, snd)
+
+    # Compute daily accumulation
+    acc = snd.diff(dim="time")
```
tlogan2000, line 1755: Same thing here..  Does a winter storm necessarily derive from what is accumulated on the ground? I think we should be able to calculate a snowfall event from precipitation data as well

tlogan2000 (commented):
This all seems to calculate things as desired but maybe I missed some discussion but I'm not sure I totally agree with the way snowfall is being determined.... it is often derived as a diff() on snowdepth.  Why can't we use prsn directly as well?  Seems more intuitive to me

Other general comment :  Maybe not for this PR but I think with the added rl.season (very nice!), most 'season' like indices should eventually see a refactor where each one is a multi-output indicator that returns start,end, length in a single call.     i.e.  atmos.growing_season()  or   atmos.snowcover_duration() returns all 3 outputs etc.  It would reduce the number of indicators and seem simpler for a user Ithink ( is the metadata management possibly why this is trickier?)

xclim/indices/_multivariate.py
huard: Totally, that was also my first impulse, but that's how the CCCS wants it.

xclim/indices/_threshold.py
huard: We had discussions around that with @aulemahal. It could be another indice function, or just additional optional input variables, but that comes with a risk of having the same indicator identifier with different computing approaches.

tlogan2000 (commented):
10-4 si c'est fait selon les demandes des usagers je suis ok avec ca.

I have a single test failing on my local build : xclim/testing/tests/test_cli.py 
Same as in the PR in fact...

aulemahal:
Woups. The CLI parser was trying to read the "long_name" attribute of "high_precip_low_temp", which had none.
I relaxed this in CLI because I don't think we have a policy of requiring a `long_name`?
But I also added one the the indicator (through the return sections of the indices' doc) anyway.

tlogan2000 (approved):

xclim/indices/_multivariate.py
```diff
@@ -940,6 +940,11 @@ def high_precip_low_temp(
     freq : str
       Resampling frequency.
 
+    Return
+    ------
```
Zeitsperre, line 944:
```suggestion
    Returns
    -------
```

xclim/indices/_multivariate.py
```diff
@@ -717,6 +719,57 @@ def func(x, axis):
...
+    Returns
+    -------
+    xarray.DataArray, [time]
+      Count of days with high precipitation and low temperatures.
+
+    Example
+    -------
+    To compute the number of days with intense rainfall while minimum temperatures dip below -0.2C:
+    >>> pr = xr.open_dataset(path_to_pr_file).pr
+    `   >>> tasmin = x.open_dataset(path_to_tasman_file).tasmin
```
Zeitsperre, line 762: ?

aulemahal: Ach saperlipopette.
There was a bug(?) where black would move the whole docstring 4 spaces to the right, making the build crash. Looks like I missed that line...

aulemahal:
I have no idea what the RTD bug is... Works fine in my env! I'll try from scratch.

PR #656, opened 2021-03-11
sdba refactor using map_blocks

### LINKED ISSUE #652: Refactor xclim.sdba for performance improvment
Current xclim.sdba is optimized to use dask though xarray in a "under-the-hood" manner. This way, using loaded or dask-backed data works just the same.  As far as experience goes, performance seems best for datasets ranging from single timeseries to regional (~100 timeseries). Larger datasets may cause the number of dask task to explode. This makes the dask scheduler hang and use enormous amounts of RAM, before beginning any computations. Most of this is due to our extensive use of groupby's, which, when used with small groups like "Time.dayofyear", make the number of chunks explode.

The refactor I propose is the following:

Wrap most of the computation into `xr.apply_ufunc` with `dask='parallelized`.  Thus, dask will see only a single task, data will be loaded before entering the function and redaskized upon exit, reducing the number of intermediate chunks created.

To do so, I suggest we move the adjustment computation to external functions. The adjustment objects will act as "pipelines" calling the different step and storing parameters and intermediate results. The main API wouldn't change.

~However, we need xarray's features. Numpy won't be enough for our needs, especially with all the time handlings. With `dask=parallelized`, xarray passes numpy arrays to the wrapped function. We will need to reconstruct/deconstruct xarray DataArrays inside the ufunc.~
The new functions will be wrapped through `map_blocks` (and not `apply_ufunc`). Meaning some minimal overheard to merge multiple inputs into datasets.

Most of the algorithms act only along one dimension, for one group. (i.e. along 'time' for a single day of year (+window)).  To keep code codable, I suggest implementing a fancy decorator that takes care of wrapping function acting on a group and transform them in a function accepting the whole dataset.

With this, it should be still relatively easy to implement new adjustment methods, may be a bit more complex because of xclim-specific syntax. The performance should be hugely improved for large datasets as it should really be scalable.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #652

Moves most of the adjustment code to standalone functions that are wrapped in `map_blocks` call through a fancy-but-complex decorator. Also adds some numba-accelerated utilities.

Yes. The API itself hasn't changed, but there are some notable differences.

In the following I try to summarize the changes. 

- Adjustment objects in `xclim/sdba/adjustment.py` act more like pipelines : the computation functions are in `_qm.py`. Same for Detrend objects. The code is currently in a hybrid state, it is not clear to me if we want to keep that or not.
- When we have a group of operations that acts on a A) a single dask block or B) on a single group in a single block, I strongly recommend writing a single function and wrapping it with A) `xclim.sdba.base.map_blocks` or B) `map_groups`. The second inserts a  `Grouper.apply(...)` call in the `map_blocks` call.

In opposition to `xr.apply_ufunc`, `map_blocks` passes an xarray object to its children function (not a numpy array). This is what we want, because we need the coordinates (especially the "time"  one). However, instead of configuring the call and its output with keyword args, `map_blocks` wants a "template", i.e. the output dataset, but empty. Creating this is not simple, especially since I want to remove that overhead from the actual algorithm writing. Thus the two decorators. 
Both take as input a mapping from new variable name to new dimensions. See the code for examples.

Some details that might also interest users like @RondeauG :

- There is a new limitations for any input to functions wrapped by the new decorators (i.e. `ref`, `hist` and `sim`): they must have the same dimensions and be mergeable in a dataset. This means using the same calendar, for example. 
- I modified the `Grouper` object so that passing 'time' (and only for that case), groups over a dummy dimension called `group`. Before, passing 'time' acted like a passthrough : `group.apply('mean', ds)` was the exact same as `ds.mean('time')`. But, this caused generality problems downstream, so now a call to `group.apply` replaces the main dimension by the grouped dimension.

All this was in a quest for better performances and scalability. Right now, DQM without dask takes ~ half the time and with dask 20% of the time. (using two years on 1300 points, grouping by dayofyear). I'll make more tests later.

Currently, only the QM methods are transformed, will do the rest later.

### DIFF
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -9,20 +9,17 @@ from xarray.core.dataarray import DataArray
 from xclim.core.calendar import get_calendar
 from xclim.core.formatting import update_history
 
+from ._qm import dqm_scale_sim, dqm_train, eqm_train, qdm_adjust, qm_adjust
 from .base import Grouper, Parametrizable, parse_group
 from .detrending import PolyDetrend
-from .processing import normalize
 from .utils import (
     ADDITIVE,
     MULTIPLICATIVE,
     apply_correction,
     broadcast,
     equally_spaced_nodes,
-    extrapolate_qm,
     get_correction,
-    interp_on_quantiles,
     map_cdf,
-    rank,
 )
 
 __all__ = [
@@ -74,9 +71,9 @@ class BaseAdjustment(Parametrizable):
             _raise_on_multiple_chunk(ref, self.group.dim)
             _raise_on_multiple_chunk(hist, self.group.dim)
 
-            if self.group.prop == "dayofyear" and get_calendar(ref) != get_calendar(
-                hist
-            ):
+            if self.group.prop == "dayofyear" and get_calendar(
+                ref, self.group.dim
+            ) != get_calendar(hist, self.group.dim):
                 warn(
                     (
                         "Input ref and hist are defined on different calendars, "
@@ -88,6 +85,7 @@ class BaseAdjustment(Parametrizable):
                 )
 
         self._train(ref, hist)
+        self.ds.attrs["adj_params"] = str(self)
         self._hist_calendar = get_calendar(hist)
         self.__trained = True
 
@@ -124,19 +122,12 @@ class BaseAdjustment(Parametrizable):
 
         scen = self._adjust(sim, **kwargs)
         params = ", ".join([f"{k}={repr(v)}" for k, v in kwargs.items()])
+        infostr = f"{str(self)}.adjust(sim, {params})"
         scen.attrs["xclim_history"] = update_history(
-            f"Bias-adjusted with {str(self)}.adjust(sim, {params})", sim
+            f"Bias-adjusted with {infostr}", sim
         )
+        scen.attrs["bias_adjustment"] = infostr
         return scen
-
-    def _make_dataset(self, **kwargs):
-        """Set the trained dataset from the passed variables.
-
-        The trained dataset should at least have a `af` variable storing the adjustment factors.
-        Adds the adjustment parameters as the "adj_params" dictionary attribute.
-        """
-        self.ds = xr.Dataset(data_vars=kwargs)
-        self.ds.attrs["adj_params"] = str(self)
 
     def _train(self):
         raise NotImplementedError
@@ -195,31 +186,34 @@ class EmpiricalQuantileMapping(BaseAdjustment):
         )
 
     def _train(self, ref, hist):
-        quantiles = equally_spaced_nodes(self.nquantiles, eps=1e-6)
-        ref_q = self.group.apply("quantile", ref, q=quantiles).rename(
-            quantile="quantiles"
-        )
-        hist_q = self.group.apply("quantile", hist, q=quantiles).rename(
-            quantile="quantiles"
+        quantiles = np.array(
+            equally_spaced_nodes(self.nquantiles, eps=1e-6), dtype="float32"
         )
 
-        af = get_correction(hist_q, ref_q, self.kind)
+        self.ds = eqm_train(
+            xr.Dataset({"ref": ref, "hist": hist}),
+            group=self.group,
+            kind=self.kind,
+            quantiles=quantiles,
+        )
 
-        af.attrs.update(
+        self.ds.af.attrs.update(
             standard_name="Adjustment factors",
             long_name="Quantile mapping adjustment factors",
         )
-        hist_q.attrs.update(
+        self.ds.hist_q.attrs.update(
             standard_name="Model quantiles",
             long_name="Quantiles of model on the reference period",
         )
-        self._make_dataset(af=af, hist_q=hist_q)
 
     def _adjust(self, sim, interp="nearest", extrapolation="constant"):
-        af, hist_q = extrapolate_qm(self.ds.af, self.ds.hist_q, method=extrapolation)
-        af = interp_on_quantiles(sim, hist_q, af, group=self.group, method=interp)
-
-        return apply_correction(sim, af, self.kind)
+        return qm_adjust(
+            xr.Dataset({"af": self.ds.af, "hist_q": self.ds.hist_q, "sim": sim}),
+            group=self.group,
+            interp=interp,
+            extrapolation=extrapolation,
+            kind=self.kind,
+        ).scen
 
 
 class DetrendedQuantileMapping(EmpiricalQuantileMapping):
@@ -253,8 +247,6 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
     group : Union[str, Grouper]
       The grouping information used in the quantile mapping process. See :py:class:`xclim.sdba.base.Grouper` for details.
       the normalization step is always performed on each day of the year.
-    norm_window : 1
-      The window size used in the normalization grouping. Defaults to 1.
 
     In adjustment:
 
@@ -264,12 +256,6 @@ class DetrendedQuantileMapping(EmpiricalQuantileMapping):
       The method to use when detrending. If an int is passed, it is understood as a PolyDetrend (polynomial detrending) degree. Defaults to 1 (linear detrending)
     extrapolation : {'constant', 'nan'}
       The type of extrapolation to use. See :py:func:`xclim.sdba.utils.extrapolate_qm` for details. Defaults to "constant".
-    normalize_sim : bool
-      If True, scaled sim is normalized by its "dayofyear" mean and then detrended using `group`.
-        The norm is broadcasted and added back on scen using `interp='nearest'`, ignoring the passed `interp`.
-      If False, scaled sim is detrended per "dayofyear".
-      This is useful on large datasets using dask, in which case "dayofyear" is a very small division,
-        because normalization is a more efficient operation than detrending for similarly sized groups.
 
     References
     ----------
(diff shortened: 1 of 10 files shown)


aulemahal:
I think this is ready for review.

Tests pass with the new method. It also has been used by people at the Rhodium Group, plus some performance tests on my side.

The notebooks have been modified to reflect the change and add some comments on optimizing with dask.

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -8,6 +8,33 @@
...
+    "\n",
+    "A good first read on the use of dask within xarray are the latter's [Optimization tips](http://xarray.pydata.org/en/stable/dask.html#optimization-tips).\n",
+    "\n",
+    "Some xclim.sdba-specific tips:\n",
+    "\n",
+    "* Most adjustment method will need to perform operation on the whole `time` coordinate, so it is best to optimize chunking along the other dimensions. This is often different from how public data is shared, where more universal 3D chunks are used. \n",
+    "\n",
+    "    Chunking of outputs can be controlled in xarray's [`to_netcdf`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.to_netcdf.html?highlight=to_netcdf#xarray.Dataset.to_netcdf), but we recommend using [Zarr](https://zarr.readthedocs.io/en/stable/) files because of their better ability for parallel I/O and clarity in the chunking. See [`Dataset.to_zarr`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.to_zarr.html?highlight=to_zarr#xarray.Dataset.to_zarr) and this useful [rechunking package](https://rechunker.readthedocs.io).\n",
```
huard, line 21: Do we have enough experience to back this up ?

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -8,6 +8,33 @@
...
+    "\n",
+    "\n",
+    "* One of the main bottleneck for adjustments with small groups is that dask needs to build and optimize an enormous task graph. This issue has been greatly reduced with xclim 0.27 and the use of `map_blocks` in the adjustment methods. However, not all adjustment methods use this optimized syntax.\n",
+    "\n",
+    "    In order to help dask, one can split the processing in parts. For splitting traning and adjustment, see [the section below](#Initializing-an-Adjustment-object-from-a-training-dataset).\n",
+    "\n",
+    "\n",
+    "* Another massive bottleneck of parallelization of xarray is the thread-locking behaviour of some methods. It is quite difficult to isolate and avoid those locking, so one of the best workaround is to use Dask configurations with many _processes_ and few _threads_. The former do not share memory and thus are not impacted when the other is locked. However, this adds many memory transferring operations and, by experience, reduces dask's the ability to parallelize some pipelines. \n",
```
huard, line 29:
```suggestion
    "* Another massive bottleneck of parallelization of xarray is the thread-locking behaviour of some methods. It is quite difficult to isolate and avoid those locking, so one of the best workaround is to use Dask configurations with many _processes_ and few _threads_. The former do not share memory and thus are not impacted when the other is locked. However, this adds many memory transfer operations and, by experience, reduces dask's ability to parallelize some pipelines. \n",
```

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -8,6 +8,33 @@
...
+    " \n",
+    "   This is done usually by creating a dask client with a large `n_workers` and a small `threads_per_worker`. The last section of this notebook goes into a bit more details by suggesting a templace for large-scale bias-adjustment pipelines."
```
huard, line 31:
```suggestion
    "   This is done usually by creating a dask client with a large `n_workers` and a small `threads_per_worker`. The last section of this notebook goes into a bit more details by suggesting a template for large-scale bias-adjustment pipelines."
```

docs/notebooks/sdba.ipynb
```diff
@@ -164,7 +172,9 @@
     "- adjustment `scen = Adj.adjust(sim, **kwargs)`\n",
     "- post-processing on `scen` (for example: re-trending)\n",
     "\n",
-    "The train-adjust approach allows to inspect the trained adjustment object. The training information is stored in the underlying `Adj.ds` dataset and always has a `af` variable with the adjustment factors. Its layout and the other available variables vary between the different algorithm, refer to their part of the API docs.\n",
+    "The train-adjust approach allows to inspect the trained adjustment object. The training information is stored in the underlying `Adj.ds` dataset and often has a `af` variable with the adjustment factors. Its layout and the other available variables vary between the different algorithm, refer to their part of the API docs.\n",
+    "\n",
+    "For heavy processing, this seperation allows the computation and writing to disk of the training dataset before perfoming the adjustment(s). See the [advanced notebook](sdba-advanced.ipynb).\n",
```
huard, line 177:
```suggestion
    "For heavy processing, this separation allows the computation and writing to disk of the training dataset before performing the adjustment(s). See the [advanced notebook](sdba-advanced.ipynb).\n",
```

xclim/sdba/_adjustment.py
```diff
@@ -0,0 +1,177 @@
...
+
+from . import nbutils as nbu
+from . import utils as u
+from .base import map_blocks, map_groups
+
+
+@map_groups(
+    af=["<PROP>", "quantiles"], hist_q=["<PROP>", "quantiles"], scaling=["<PROP>"]
+)
+def dqm_train(ds, *, dim, kind, quantiles):
```
huard, line 17: Are these public or private methods ? If public, I suggest writing complete docstrings, otherwise prepend name with _.

huard: Hum, the entire module is private, right ? So no need for full docstrings.

xclim/sdba/adjustment.py
```diff
@@ -51,7 +53,23 @@ def _raise_on_multiple_chunk(da, main_dim):
...
-    """Base class for adjustment objects."""
+    """Base class for adjustment objects.
+
+    Children classes should implement:
+
+    __init__(**kwargs)
+      For fully functioning `Parametrizable` subclasses, the init should only pass kwargs the
+      base class. It could set some, after the super().__init__(), but only through bracket
+      access (self['abc'] = abc). All parameters should be simple python literals or other
+      `Parametrizable` subclasses instances.
```
huard, line 64: I don't understand.

xclim/sdba/adjustment.py
```diff
@@ -51,7 +53,23 @@ def _raise_on_multiple_chunk(da, main_dim):
 
 
 class BaseAdjustment(ParametrizableWithDataset):
-    """Base class for adjustment objects."""
+    """Base class for adjustment objects.
+
+    Children classes should implement:
```
huard, line 58:
```suggestion
    Children classes should implement these methods:
```

xclim/sdba/base.py
```diff
@@ -366,23 +403,198 @@ def parse_group(func: Callable) -> Callable:
...
+        content = np.empty(shape)
+    return xr.DataArray(content, dims=dims)
+
+
+def map_blocks(reduces=None, **outvars):
+    """
+    Decorator for declaring functions and wrapping them into a map_blocks. It takes care of constructing
+    the template dataset.
+
+    If `group` in the kwargs, it is assumed that `group.dim` is the only dimension reduced or modified,
```
huard, line 441:
```suggestion
    If `group` is in the kwargs, it is assumed that `group.dim` is the only dimension reduced or modified,
```

xclim/sdba/base.py
```diff
@@ -366,23 +403,198 @@ def parse_group(func: Callable) -> Callable:
...
+    and that some other dimensions might be added, but no other existing dimension will be modified.
+
+    Arguments to the decorator are mappings from variable name in the output to its *new* dimensions.
+    Dimension order is not preserved.
+    The placeholders "<PROP>" and "<DIM>" can be used to signify `group.prop` and `group.dim` respectively.
```
huard, line 446: This `<PROP>` and `<DIM>` thing doesn't feel pythonic. I assume there weren't other options?

xclim/sdba/base.py
```diff
@@ -366,23 +403,198 @@ def parse_group(func: Callable) -> Callable:
...
+
+            # Create the output dataset, but empty
+            tmpl = xr.Dataset(coords=coords)
+            for var, dims in outvars.items():
+                # Out variables must have the base dims + new_dims
+                dims = base_dims + [placeholders.get(dim, dim) for dim in dims]
+                # duck empty calls dask if chunks is not None
+                tmpl[var] = duck_empty(dims, sizes, chunks)
+
+            def _transpose_on_exit(dsblock, **kwargs):
```
huard, line 553: This is where the actual work is done right ? The function name suggests it's doing a reshape operation.

huard: Isn't it surprising that we have to deal with dimension order within xarray ?

huard (approved):

xclim/sdba/base.py
aulemahal: There are other options I guess. I could hide this in constants like `ADDITIVE`, so the decorator looks like: 
```
@map_blocks(reduces=[sdba.PROP, "quantiles"], scen=[])
```

We need a simple flag. The decorator needs to be understandable by the developper and the flag to be unambiguous for the code.  It is never used outside the definition of sdba algorithms.

xclim/sdba/base.py
aulemahal: True, will give a clearer name.

Yes, a bit surprising. Raised the issue at pydata/xarray#5326.

aulemahal: Issue on xarray closed. Conclusion:
As we pass a `template` that is supposed to look like the result, it is expected (but not checked)  that the result fits the template perfectly : sames sizes, same variables and _same dimension order_. We could also try to guess the order, but I think it is simpler to transpose the result. The performance impact should be quite low.

xclim/sdba/base.py
aulemahal: @huard  Would like your opinion on the new syntax. I made class-constants `Grouper.PROP` and `Grouper.DIM`, that are used the same way as `<PROP>` and `<DIM>`. It does feel a bit more pythonic.

huard (approved):
LGTM.

aulemahal:
Bumped the version, will merge when tests pass!

PR #664, opened 2021-03-19
Replicate cffdrs in xc.indices.fwi

### LINKED ISSUE #643: Modify fire weather index calcs to reproduce R library functionality
### Description
The official Canadian Forest Service R package for Canadian Forest Fire Danger Rating System (includes FWI) has functionality that differs form our current FWI implementation based on GFWED.  

R package : https://cran.r-project.org/web/packages/cffdrs/index.html
Docs pdf : https://cran.r-project.org/web/packages/cffdrs/cffdrs.pdf

In particular cffdrs FWI allows for:

- [ ]  different temperature thresholds for season startup / shutdown
- [ ]  for 'overwintering' of the drought code 

The first case is likely relatively simple to fix 

The second could be more work we could look into https://cfs.nrcan.gc.ca/pubwarehouse/pdfs/29152.pdf or contact Alan Cantin (contact for R Package)

aulemahal:
### Pull Request Checklist:
    - This PR fixes #643.

Synchronizes the code with the R library (cffdrs).  This means:

- Addition of the overwintering process
- Addition of a `fire_season` function
- Removal of the "dry start" cases.

The API has changed because the "start up" and "shut down" are not computed the same way as before. There is not "start_date", rather a boolean "season_mask", which can be computed on-the-fly or with `fire_season` (new indice/indicator).

All FWI related indices were moved inside the `fwi` submodule. Most docstring now point to `fire_weather_ufunc` (even if we do not recommend using this method).

### Other information
xclim doesn't match the GFWED data used before anymore because, I assume, of the different fire season determination process. 
I plan to do a notebook showing of to compute the FWI while also comparing it to the output of https://essd.copernicus.org/articles/12/1823/2020/.


xclim/indices/_multivariate.py
```diff
@@ -497,6 +498,105 @@ def drought_code(
...
+@declare_units(
+    tas="[temperature]",
+    snd="[length]",
+    temp_end_thresh="[temperature]",
+    temp_start_thresh="[temperature]",
+    snow_thresh="[length]",
+)
+def fire_season(
+    tas: xarray.DataArray,
+    snd: xarray.DataArray = None,
```
Zeitsperre:
```suggestion
    snd: Optional[xarray.DataArray] = None,
```

aulemahal:
This is ready for review!
I think @laura-vanvliet and @JeremyFyke could be interested. There is no notebook showing how it works for now, but I hope the docstring can give you an idea! The API hasn't changed much, except for the overwintering part, see `fire_weather_ufunc`.

JeremyFyke:
@aulemahal This is great to see!  Well done.  Have you managed to run any side-by-side tests to confirm that for an identical input, one gets identical output (to some level of acceptability given machine precision, etc.) between xclim and cffdrs?

@laura-vanvliet

tlogan2000:
> @aulemahal This is great to see! Well done. Have you managed to run any side-by-side tests to confirm that for an identical input, one gets identical output (to some level of acceptability given machine precision, etc.) between xclim and cffdrs?
> 
> @laura-vanvliet

@aulemahal  FYI I had created a test CanLEAD dataset of input variables for a few locations for this type of comparison  (ERA5 notebook a good idea as well I think) :   https://github.com/ECCC-CCCS/CanLEAD-FWI/blob/Ournanos_tests/data/testsites_r1i1p1_FWIinput.nc

aulemahal:
@JeremyFyke  Yes! I converted the `test_fwi` dataset that ships with R to netcdf and put it in our [testdata repo](https://github.com/Ouranosinc/xclim-testdata/tree/cffdrs-data/FWI). There is a unit test that checks that outputs are equal to a tolerance of 1e-6. Same for the fire season determination and the "test_wDC" dataset, and for the overwintering DC (there I just ran the R code with some data, not a full dataset).

@tlogan2000 Thanks, I'll try to use that!

tlogan2000 (approved):
This is looking good. 
From my end I think we can merge then see if there is fine-tuning needed after some further testing with Jeremy and Laura

JeremyFyke:
Hi all, thanks again for working on this!  This is great, and I can imagine that ensuring consistency with CFFDRS (while also being apparently MUCH faster!) could attract some interest in xclim from the Canadian wildfire research community.

We have it on our radar to try out this updated code and if it is workable and reproduces our current results, make a swap.  To this end, should we wait until this updated FWI stuff comes out in the next xclim release, or perhaps grab it directly as a clone/fork from this github repo..?

aulemahal:
Well the next release is due next week (wednesday, hopefully), so the wait wouldn't be long! But feel free to use xclim's master if you want. I expect that there might be issues with the new code or API adjustment to be made, so using the master could allow us to have some minimal iterative interaction in the next months!

laura-vanvliet:
Hi all! After some brief testing, I think there is a small error in stand-alone overwintering_drought_code function; the output_core_dims specified below are incorrect. 
```
wDC = xr.apply_ufunc(
        _overwintering_drought_code,
        last_dc,
        winter_pr,
        carry_over_fraction,
        wetting_efficiency_fraction,
        min_dc,
        input_core_dims=[[]] * 5,
        output_core_dims=[[]] * 5,
        dask="parallelized",
        output_dtypes=[last_dc.dtype],
```

aulemahal:
Woups. That's what happens when I don't write unit tests for every function. Will fix in a upcoming PR.

PR #673, opened 2021-03-24
Create Parametrizable classes from strings and datasets

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Implements `xclim.sdba.base.Parametrizable.from_string` and `xclim.sdba.base.ParametrizableWithDataset.from_dataset`, allowing to instantiate Grouper, Adjustment and Detrending objects from a string or a dataset they created. This is meant to be using in larger pipelines where the training dataset is not created in the same python session as the adjustment process.

Also, Adjustment and Detrending objects now expose a `set_dataset` method to change their `.ds` dataset attribute. This is for the case where both training and adjustment are done in the same session. For example, when one wants to trigger dask computation and write the training to disk, then reload it into the object.

Shouldn't.


HISTORY.rst
```diff
@@ -13,6 +13,7 @@ New features and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 * `generic.count_domain` used to count values within low and high thresholds.
 * Fire Weather indices now support dask-backed data.
+* Objects from the `xclim.sdba` submodule can be created from their string repr or from dataset they created.
```
huard, line 16:
```suggestion
* Objects from the `xclim.sdba` submodule can be created from their string repr or from the dataset they created.
```

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -149,6 +149,100 @@
...
+    "ds = xr.open_dataset('QDM_training.nc')\n",
+    "QDM2 = QuantileDeltaMapping.from_dataset(ds)\n",
+    "QDM2"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the case above, creating a full object from the dataset doesn't make the most sense since we are in the same python session, with the \"old\" object still available. This method effective when we reload the training data in a different python session, say on another computer. **However, take note that there is retrocompatiblity insurance.** If the QuantileDeltaMapping class was to change in a new xclim version, one would not be able to create the new object from a dataset saved with the old one.\n",
```
huard, line 229: no insurance

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -149,6 +149,100 @@
...
+    "\n",
+    "For the case where we stay in the same python session, it is still useful to write the dataset, in order to trigger the dask computations. In that case, we could also do:"
```
huard, line 231: In this case, wouldn't you recommend `compute()` instead ?

xclim/sdba/base.py
```diff
@@ -16,6 +18,64 @@ class Parametrizable(dict):
...
+        --------
+        >>> obj = Parametrizable(anint=1, anobj=Parametrizable(astr='hello world'))
+        >>> ss = str(obj)
+        >>> obj2 = Parametrizable.from_string(ss)
+        >>> obj
+        Parametrizable(anint=1, anobj=Parametrizable(astr='hello world'))
+        >>> obj2
+        Parametrizable(anint=1, anobj=Parametrizable(astr='hello world'))
+        """
+        # regex pattern that fully matches a string produces by Parametrizable._repr__
```
huard, line 50:
```suggestion
        # regex pattern that fully matches a string produced by Parametrizable._repr__
```

xclim/sdba/base.py
```diff
@@ -16,6 +18,64 @@ class Parametrizable(dict):
...
+        ):
+            dd = argmatch.groupdict()
+            if dd["vcls"] is not None:
+                params[dd["name"]] = cls._parametrizable_classes[
+                    dd["vcls"]
+                ].from_string(dd["vstr"])
+            else:
+                params[dd["name"]] = literal_eval(dd["vstr"])
+        obj.update(params)
+        return obj
```
huard, line 77: Not sure I understand what you're trying to do here... could you add some comments ?

Maybe this is unrelated, but if you're trying to cast keyword arguments to a class, you could look into pydantic. It can be used to automatically parse objects and cast them to the class implementing a data model compatible with the inputs.

huard (approved):

docs/notebooks/sdba-advanced.ipynb
aulemahal: Not wrong. I guess this depends on the size of the data! I'll add that.

xclim/sdba/base.py
aulemahal: I'll add comments.
Overall this parses the string that it generates in the `__repr__`.  The most difficult thing was to split the string between the arguments and distinguish between normal python literals and our custom "Parametrizable" classes. Thus the regex.
I'm not sure how pydantic could help us here? Does it have string parsing mechanisms?

aulemahal: More generally, we are looking for human-readable object serialization. Usual serialization is covered by pickle, but that would create unreadable strings. JSON is an option, but again that string would complex. There is [PyON](https://github.com/lagmoellertim/pyon) that can handle the custom classes in JSON, but it is not very active and not on conda.

Seeing that we do not need to cover many cases and can completely control the custom classes that we need to parse, I think having our own in-house method is ok, and secure enough.

But maybe @Zeitsperre, @tlvu or @cjauvin have ideas about this problem? TL;DR: We want to serialize a python object to a human-readable string and be able to reconstruct it. For now, object are made of simple python literals OR other instances of the same base class, thus sharing the (de-)serialization mechanisms.

However, we could split this in 2 things : normal human-readable repr + string serialization. The "problem" here is simply that it would mean adding a completely unreadable attribute to the datasets we make, an attribute that does not make sense outside of xclim (while the current kinda does).

docs/notebooks/sdba-advanced.ipynb
aulemahal, line 229:
```suggestion
    "In the case above, creating a full object from the dataset doesn't make the most sense since we are in the same python session, with the \"old\" object still available. This method effective when we reload the training data in a different python session, say on another computer. **However, take note that there is no retrocompatiblity insurance.** If the QuantileDeltaMapping class was to change in a new xclim version, one would not be able to create the new object from a dataset saved with the old one.\n",
```

docs/notebooks/sdba-advanced.ipynb
aulemahal, line 231:
```suggestion
    "For the case where we stay in the same python session, it is still useful to trigger the dask computations. For small datasets, that could mean a simple `QDM.ds.load()`, but sometimes even the training data is too large to be full loaded in memory. In that case, we could also do:"
```

xclim/sdba/base.py
Zeitsperre: @aulemahal creating a JSON file with both a human-readable portion and a pickle bytestring makes sense to me, honestly. My only concern would be ensuring that some sort of safety mechanism is in place to evaluate that the human-readable portion of this object/file corresponds with the pickled string values and maybe exit early if there's an inconsistency? My biggest worry would be that they don't agree and the values are wrong (beyond all the other concerns from pickles).

aulemahal: True that is also a concern : if the human readable attribute and the machine-readable could differ. But to check that goes back to parsing the human readable one! Also, just to precise, we want to store that information in the attribute of a netCDF. So, while a json file is readable, a json _string_ scores lower.

Zeitsperre: This is probably a terrible suggestion, but if we were outputting `zarr` data, this would be really simple. https://zarr.readthedocs.io/en/stable/spec/v2.html#attributes. Again, not saying we do throw the baby out with the bathwater, but this would solve the JSON problem.

PR #681, opened 2021-04-01
sdba - Fix from_string constructor

aulemahal:
### Pull Request Checklist:
    - This PR fixes an issue raised in slack.

My misunderstanding of python's `__new__`. I though that it was executed upon class _declaration_, but it is rather called upon class _instantiation_. When `sdba.base.Parametrizable.from_string` encounter a class name, it would look in a registry populated in the `__new__`, so only when subclasses were created. Thus, in a brand new session, if the first thing you call is a `from_string` or `from_dataset`, the registry is empty because no objects have been created yet. Resulting in an error.

Rather, here we use the `__subclasses__` python method to list all subclasses (recursively). No need for instantiation.

@huard Should we change the `Indicator` class to use this instead?   The main bonus is that classes that do not have instances would still be listed in the registry (ex: Daily, Hourly, Tasmax, ...).

No

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,8 +2,16 @@
 History
 =======
 
-0.25.0 2021-03-31)
-------------------
+0.26.0 (unreleased)
+-------------------
+
+Bug fixes
+~~~~~~~~~
+* Fix the `from_string` object creation in sdba.
+
+
+0.25.0 (2021-03-31)
+-------------------
 
 Announcements
 ~~~~~~~~~~~~~
--- a/xclim/sdba/base.py
+++ b/xclim/sdba/base.py
@@ -18,13 +18,14 @@ class Parametrizable(dict):
     :py:meth:`Parametrizable.parameters` dictionary, the copy method and the class representation.
     """
 
-    _parametrizable_classes = {}
-
-    def __new__(cls, *args, **kwargs):
-        """Add subclass to registry."""
-        # This way, the from_string constructor can match what it finds in repr strings to a class.
-        cls._parametrizable_classes[cls.__name__] = cls
-        return super().__new__(cls, *args, **kwargs)
+    @classmethod
+    def _get_subclasses(cls):
+        """Return a dict of all the subclasses of the current class and recursively for those subclasses."""
+        subcls = {}
+        for sub in cls.__subclasses__():
+            subcls[sub.__name__] = sub
+            subcls.update(sub._get_subclasses())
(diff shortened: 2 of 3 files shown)


huard (approved):
To be frank, I'm not sure I understand what this code really does, and I suspect there's a cleaner way to solve this. I understand this is essentially a serialization-deserialization problem, and I expect there are standard solutions to this. 

Do we want non-instances to be listed in the registry ?

aulemahal:
The registry is there so it's easy to subclass indicators, mainly because we create classes on-the-fly, so they are not importable and classic subclassing is not possible. Currently, if you want to subclass from a non-instance, you import it and do it normally. This change would only allow to harmonize the process and allow subclassing always to be done from the registry. Also, we would have the possibility to generate the registry "on demand", it wouldn't be a (mutable) variable.

All those things are quite weak arguments and anyway nobody uses this functionality currently (as far as I know). Let's not fix what already works.

As for the serialization-deserialization, it would indeed be nice if we used a more standard solution, but I don't know any, especially because I wanted the result to be readable, to carry information. `pickle` is the most common solution, but serializes to bytes. JSON alone doesn't carry python object information and is not as readable as the current solution. I found this : https://marshmallow.readthedocs.io/en/stable/, it would make JSON serialization easy to use.

For example.
current:
```
QuantileDeltaMapping(nquantiles=15, kind='+', group=Grouper(dim='time', add_dims=[], prop='dayofyear', name='time.dayofyear', window=1, interp=False))
```
json:
```
"{'nquantiles':15, 'kind':'+', 'group': {'dim': 'time', 'add_dims': [], 'prop': 'dayofyear', 'name': 'time.dayofyear', 'window': 1, 'interp': False}}"
```

huard:
This may be closer to what you want: https://jsonpickle.github.io/
```python
In [1]: class Thing(object):
   ...:     def __init__(self, name):
   ...:         self.name = name
   ...: 
   ...: obj = Thing('Awesome')

In [2]: 

In [2]: obj
Out[2]: <__main__.Thing at 0x7fc4956a3700>

In [3]: import jsonpickle
   ...: frozen = jsonpickle.encode(obj)

In [4]: 

In [4]: frozen
Out[4]: '{"py/object": "__main__.Thing", "name": "Awesome"}'

In [5]: thawed = jsonpickle.decode(frozen)

In [6]: thawed
Out[6]: <__main__.Thing at 0x7fc494d0eca0>
```

PR #683, opened 2021-04-06
Indicators from yaml files - cf-index-meta

### LINKED ISSUE #501: Integrate cf-index-meta to xclim
The [cf-index-meta project](https://bitbucket.org/cf-index-meta/cf-index-meta/src/master/) standardizes common climate indicators and their algorithms.

What we could do:

1. Parse metadata from the table and match them to corresponding indicators
2. Implement the generic functions and populate a xclim virtual module on-the-fly (like icclim or anuclim)

Integrating the cf-index-meta indicators directly in xclim might be problematic as we define more metadata fields (especially the translations!).

aulemahal:
### Pull Request Checklist:
    - This PR fixes #501

- Implements the generic index definitions from [cf-index-meta](https://bitbucket.org/cf-index-meta/cf-index-meta/) into `xclim.indices.generic
- New `xclim.core.yaml` submodules that handles creation of indicators from yaml-parsed dictionaries. But I plan to also add methods to create fake modules from yaml files (i.e. icclim, anuclim and friends)
- Moved "xclim/locales" to "xclim/data" and addition of a `variables.yml" with definitions of "official" xclim variables.
- Internal changes in the indicator creation to allow passing a `parameters` dict. This dict is used to override values parsed from the compute's docstring.

I copied and extended the  `variables.yml` file from `cf-index-meta`. The problem is that they do not define the dimensionality (units) and don't give descriptions of the variables, which we need to properly create the indicator and it's docstring.

Overall, the yaml file generated by `cf-index-meta` is relatively complete. But, for a complete xclim integration, I think the following features are missing:

- There is a way to define a parameters passed to the generic function, but there is no explicit way for defining the _default value_ of such a parameter, *while letting the user decide*. In addition to the default value, this is important for the docstring: right now if a parameter of the index function is mentionned in the yaml, it is _injected_, otherwise the user sees the name, default and description from the generic index function's docstring, so nothing informative.
- Multi-output indicators (FWI, wind_speed_from_vectors, etc) are not supported. Maybe that's ok.
- `cf-index-meta` defines which frequencies are allowed for a given indicator, and gives a default one. xclim doesn't have such a control. Easy to implement.
- `cf-index-meta` doesn't define the following fields : `abstract`, `title`, `keywords`, `comment`, `realm`, `missing`. The last two are not so important, but missing the first two gives uninformative docstrings.

I moved the locales json to a new folder, but it shouldn't impact users.


aulemahal:
Last commits pushed something working! It can parse the yaml file that `cf-index-meta` outputs!

Most of the doc is in `xclim.core.indicator`. I moved the module creation function there. 

* Added a mechanism to create cfcheck functions "on-the-fly", needs parsing of the `variables.yml` file, but this might be down even more interactively downloading stuff from the cf conventions website. (Also there seems to be an issue with this right now...)
* Added a mechanism to restrict allowed freq strings. Indicators now have a "allowed_periods" attibute, a list of base offsets to check against (A is the same as Y). The check is performed in the `__call__`.

TODO:

* Find a way to elegantly define indicators in yaml from existing indicators. I currently is broken when you "inject" a parameter that appears in one of the metadata strings. (Ex: long name is "When X is over {thresh}" but you define thresh = 3 in the yaml). We would need a way to partially format metadata strings?
* Find a way to reduce the number of indicator registry overwritings? It is possible to have similar (but maybe different) indicators in two submodules, with the same identifier. I think we could prepend the identifier by the submodule name in the registry for non-xclim indicators. (This way "ICCLIM.GD4" and "CF.GD4" would refer to different objects).
* More testing and documenting.

aulemahal:
Another large set of changes. 

- Fine tuning of the yaml importation, fix some tests.
- Added `cf.yml`, `anuclim.yml` and `icclim.yml` to define these virtual submodules!

I think @tlogan2000 might be interested in looking into this. I went back to the documents of ECAD (icclim) and Anuclim to  be sure the metadata was correct, so there were some modifications made to the `_anuclim.py` file + a bit more metadata added into the yml files than was before in the build_module call.

Breaking change : `xclim.icclim` will not work anymore. Only `xclim.indicators.icclim` (same for `cf` and `anuclim`).

aulemahal:
@tlogan2000 @Zeitsperre  I have partially rewritten the "mappings" notebook to introduce the new YAML way.  As now mappings are actually _Indicators_ mapping, and not indices, the old way was transformed in something a bit more complex. I didn't have the courage, nor to I think it is useful, to convert the PCC mapping. However, it could be made into a YAML file and added permanently to xclim?

@huard Earlier in this PR I decided to introduce a mechanism so that indicator subclass that are NOT declared within xclim.indicators.<realm> would have their module name prepended in the registry. In order to prevent overwriting official indicators with eponymous copies of, say, icclim.  Now, that means that the indicators in live sessions (like a notebook) are prefixed by `__main__`, which ain't cute. Also, one can pass `module` to force a fake module. As you can see in the new `mappings` notebook, that looks a bit odd, passing `module` to each indicator creation before even creating that module... 
Thoughts on this?

xclim/core/indicator.py
```diff
@@ -835,3 +1026,243 @@ class Hourly(Indicator):
...
+    doc: Optional[str] = None,
+) -> ModuleType:
+    """Create a module from imported objects.
+
+    The module is inserted as a submodule of `xclim.indicators`.
+
+    Parameters
+    ----------
+    name : str
+      New module name. If it alreay exists, the module is extended with the passed objects,
```
huard, line 1147:
```suggestion
      New module name. If it already exists, the module is extended with the passed objects,
```

docs/notebooks/example.yml
```diff
@@ -0,0 +1,56 @@
...
+      standard_name: number_of_days_with_air_temperature_below_threshold
+      long_name: "Number of Frost Days (Tmin < 0C)"
+      units: "1"
+      cell_methods:
+        - time: minimum within days
+        - time: sum over days
+    input:
+      data: tasmin
+    index_function:
+      name: count_occurrences
```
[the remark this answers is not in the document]
huard: I suggest we leave the code as is and wait until there is a demand for this feature.

xclim/core/indicator.py
```diff
@@ -835,3 +1026,243 @@ class Hourly(Indicator):
...
+    base: Indicator subclass
+      The Indicator subclass from which the new indicators are based. Superseeded by
+      the class given in the yaml file or in individual indicator definitions (see submodule's doc).
+    doc : str, optional
+      The docstring of the new submodule. Defaults to a very minimal header with the submodule's name.
+    mode: {'raise', 'warn', 'ignore'}
+      How to deal with broken indice definitions.
+    realm: str, optional
+    keywords: str, optional
+    references str, optional
```
huard, line 1210:
```suggestion
    keywords: str, optional
       Comma separated keywords.
    references str, optional
        Source citations. 
```

xclim/indices/generic.py
```diff
@@ -290,3 +295,303 @@ def daily_downsampler(da: xr.DataArray, freq: str = "YS") -> xr.DataArray:
...
+      Quantity.
+    freq: str
+      Resampling frequency.
+    """
+    # Convert units to low_data
+    high_data = convert_units_to(high_data, low_data)
+    threshold = convert_units_to(threshold, low_data)
+
+    lower = compare(low_data, "<", threshold)
+    higher = compare(high_data, ">", threshold)
```
huard, line 326: Not >= ?

xclim/indices/generic.py
```diff
@@ -290,3 +295,303 @@ def daily_downsampler(da: xr.DataArray, freq: str = "YS") -> xr.DataArray:
...
+    out.attrs["units"] = ""
+    return out
+
+
+def count_occurrences(data: xr.DataArray, threshold: str, condition: str, freq: str):
+    """Calculates the number of times some condition is met.
+
+    First, the threshold is transformed to the same standard_name and units as the input data.
+    Then the thresholding is performed as condition(data, threshold),
+    i.e. if condition is <, data < threshold.
```
huard, line 338:
```suggestion
    i.e. if condition is `<`, then this counts the number of times `data < threshold`.
```

xclim/indices/generic.py
```diff
@@ -290,3 +295,303 @@ def daily_downsampler(da: xr.DataArray, freq: str = "YS") -> xr.DataArray:
...
+    return out
+
+
+def diurnal_temperature_range(
+    low_data: xr.DataArray, high_data: xr.DataArray, freq: str
+):
+    """This index function takes two inputs, low_data and high_data, i.e. daily minimum and maximum temperature.
+
+    From this it calculated the average diurnal temperature range.
+
```
huard, line 366: Use NumPy doc standard for docstring. First line should say what the function does.

xclim/indices/generic.py
```diff
@@ -290,3 +295,303 @@ def daily_downsampler(da: xr.DataArray, freq: str = "YS") -> xr.DataArray:
...
+    Parameters
+    ----------
+    low_data : xr.DataArray
+    high_data : xr.DataArray
```
huard, line 370: add parameter description.

xclim/indices/generic.py
```diff
@@ -290,3 +295,303 @@ def daily_downsampler(da: xr.DataArray, freq: str = "YS") -> xr.DataArray:
...
+    return to_agg_units(out, data, "delta_prod")
+
+
+def interday_diurnal_temperature_range(
+    low_data: xr.DataArray, high_data: xr.DataArray, freq: str
+):
+    """This index function takes two inputs, low_data and high_data, i.e. daily minimum and maximum temperature.
+
+    From this it calculates the average absolute day-to-day difference in diurnal temperature range.
+
```
huard, line 559: Review docstring.

huard (approved):
Are the indicators created from cf.yml explicitly tested ?
This is a solid piece of work. Nice job.

docs/notebooks/example.yml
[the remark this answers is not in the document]
aulemahal: I thought of it, but as I didn't want to touch the yaml output of clix-meta (except the header), I had to have a way their name get picked up, this this two-step check : look in `xclim.indices.generic` and then in `xclim.indices`. We could open it to accept full names, when needed.

tlogan2000 (approved):
Assuming this passes the build I think this looks good.  As discussed with @aulemahal moving forward we may want to consider splitting out some of the new material in the `customizing` notebook into a 'xclim development 101 : How to code and add new indicators' or something along those lines

aulemahal:
Hopefully, the only remaining test failures would be those related to the inconsistencies between xclim's and clix-meta's unit handling.

xclim/indices/generic.py
huard: Up for discussion, but I suggest we take the liberty of using < and >=.

aulemahal:
@huard  @tlogan2000  I will merge this at the end of my day. The only real change since your reviews is the addition of a `atmosds` pytest fixture and the submodules testing where we simply execute everything we can.

PR #689, opened 2021-04-14
sdba - Extreme Values Adjustment

### LINKED ISSUE #670: Bias correction method managing extreme values
### Description

Implement bias-correction method dealing with the tail of the distribution. 

Release target: ~july 2021

aulemahal:
### Pull Request Checklist:
    - This PR fixes #670
- [ ] Merge Ouranosinc/xclim-testdata#9 and remove  `branch` args in `open_dataset` calls.

New `ExtremeValues` adjustment thing! 

Not ready yet, I have a few questions for @Balinus before. 

A few implementation details to discuss:

1. I haven't decided what this new class _is_. It only changes _extreme_ values, the _real_ bias-adjustment is done by another method. (In the julia code it's QQM, similar to our EQM) Thus, is it a children of EQM? Or do we want a more general approach, like currently coded, where the adjustment method is a parameter of that object? The latter doesn't exist for now and it's internal mechanics have to be thought through (like the serialization part).

@Balinus : Is the extremes adjustment method dependent on QQM? I.e could one decide to use a totally different adjustment method as the principal one before performing the extremes adj?

2. I added a `v_thresh` parameter, where the julia code had hardcoded `1.0`.  I'm not sure how those 1s make sense? May be I didn't get how the `biascorrect_extremes` function is supposed to be called?

@Balinus : Specifically here:
https://github.com/JuliaClimate/ClimateTools.jl/blob/2048f7428b899eddf8efdbadc9e93d3d7ce09518/src/biascorrect.jl#L315
and there
https://github.com/JuliaClimate/ClimateTools.jl/blob/2048f7428b899eddf8efdbadc9e93d3d7ce09518/src/biascorrect.jl#L332

I tried the julia code with some precipitation data from nrcan and CanESM2 : without surprise it fails, since 1 is _huge_ value of precipitation (kg m-2 s-1). Clearly, I missed something.

3. Optimization. I haven't found a way to code this in a vectorized way, so I wrote is for 1D arrays and throw that to `xr.apply_ufunc` with `vectorize=True`. Never a good option, but haven't found a clean way.

4. @huard At first, I was coding the thing vectorized, so I implemented `xc.indices.stats.dist_method`. From distribution parameters (i.e. found by `fit`) and a argument, it computes the given distribution function (`cdf`, `ppf`, etc). Sadly, I don't need it anymore, but I guess it could be useful elsewhere? 

I also split `fit` and `_fitfunc` so that the latter can be reused by my code.

No.


huard:
Oui, garde ça, je suis certain que ça va servir plus tard.

Balinus:
> Not ready yet, I have a few questions for @Balinus before.

Nice!

> A few implementation details to discuss:
> 
> I haven't decided what this new class is. It only changes extreme values, the real bias-adjustment is done by another method. (In the julia code it's QQM, similar to our EQM) Thus, is it a children of EQM? Or do we want a more general approach, like currently coded, where the adjustment method is a parameter of that object? The latter doesn't exist for now and it's internal mechanics have to be thought through (like the serialization part).
> @Balinus : Is the extremes adjustment method dependent on QQM? I.e could one decide to use a totally different adjustment method as the principal one before performing the extremes adj?

No, the method is not dependent on qqm. For the purpose of the paper, it was just simpler to call `qqmap` inside the `biascorrect_extremes` and be done with it (no performance cost). The proper way would be to split it cleanly so that people could use another method and develop a function that links them with a defualt or custom interpolation methods.

Note that since the method is a mix of empirical (non-parametric) and parametric methods, not sure if EQM is a proper parent class (assuming E stands for Empirical).

> I added a v_thresh parameter, where the julia code had hardcoded 1.0. I'm not sure how those 1s make sense? May be I didn't get how the biascorrect_extremes function is supposed to be called?
> @Balinus : Specifically here:
> https://github.com/JuliaClimate/ClimateTools.jl/blob/2048f7428b899eddf8efdbadc9e93d3d7ce09518/src/biascorrect.jl#L315
> and there
> https://github.com/JuliaClimate/ClimateTools.jl/blob/2048f7428b899eddf8efdbadc9e93d3d7ce09518/src/biascorrect.jl#L332
> 
> I tried the julia code with some precipitation data from nrcan and CanESM2 : without surprise it fails, since 1 is huge value of precipitation (kg m-2 s-1). Clearly, I missed something.

This is the hardcoded threshold for precipitation. Right now the function is defined on daily (mm) values. Proper way would be to add Unitful.jl dependency but I no longer have time for that in my free time. (decided to buy a guitar instead!)

So, the way would be to simply convert the 1mm to 1.157 x 10^-5 kg/m-2 s-1 as the default value for input with kg/m-2 s-1 units.

> Optimization. I haven't found a way to code this in a vectorized way, so I wrote is for 1D arrays and throw that to xr.apply_ufunc with vectorize=True. Never a good option, but haven't found a clean way.

It is not necessary in Julia to vectorize since it's JIT compiled (e.g. loops and vectorized code leads to the same speed, hence the simplest approach for the problem is used).

Cheers!

aulemahal:
Super! Merci!

> No, the method is not dependent on qqm. 

Nice. So I'll try to think of a good OOP structure for this.

> This is the hardcoded threshold for precipitation. 

I see! Ok, I'll add unit handling for this then,

Balinus:
Note que le 1.0mm réfère au seuil du clustering. Un nouveau cluster est défini quand la précipitation passe en-dessous de ce seuil de 1.0mm.

Aussi, quel erreurs as-tu eu lorsque tu as tenté de lancer le code en Julia? Ça fait longtemps que je n'ai pas travaillé dessus. Lorsque le papier sur la méthode d'extrême sera publié, je compte nettoyer et généraliser correctement l'approche au niveau programmation.

aulemahal:
Oui! Je vais renommer le paramètre qqch comme `cluster_thresh` je pense.

De mon bord `biascorrect_extremes` travaille un peu au début, mais ne termine jamais... Ce que je fais:
```julia
using ClimateTools

Dsim = load("xclim-testdata/sdba/fermont_CanESM2_1950-2100.nc", "pr", data_units="mm") 
Dobs = load("xclim-testdata/sdba/fermont_nrcan_1950-2013.nc", "pr", data_units="mm")

ref = temporalsubset(Dobs, (1980, 1, 1), (2009, 12, 31));
hist = temporalsubset(Dsim, (1980, 1, 1), (2009, 12, 31));
sim = temporalsubset(Dsim, (2040, 1, 1), (2069, 12, 31));

scen = biascorrect_extremes(ref, hist, sim, detrend=false, window=15, rankn=50, thresnan=0.1, keep_original=false)
```
Après un temps je vois: `corrupted double-linked list`, mais ensuite ça ne fait plus rien, mais ne termine pas non plus. C'est peut-être un problème de configuration de l'environnement par contre. J'ai suivi les instructions ici : https://juliaclimate.github.io/ClimateTools.jl/stable/installation/#Approach-no.-2.-Use-a-Conda-virtual-environment-and-link-PyCall.jl-to-it-1

Les données tests sont ici : https://github.com/Ouranosinc/xclim-testdata/tree/sdba-testfiles/sdba

Balinus:
Merci! Je vais regarder. Je suis dû pour nettoyer un peu tout ça. Ça fait au moins 1 an que j'ai pas touché au code!

Balinus:
Salut, je ne suis pas en mesure d'ouvrir les fichiers netCDF.

```
ncdump -h fermont_CanESM2_1950-2100.nc
ncdump: fermont_CanESM2_1950-2100.nc: fermont_CanESM2_1950-2100.nc: NetCDF: Unknown file format

```

aulemahal:
Étrange! Heureusement, ils sont assez petits pour github:
[sdba-testing.zip](https://github.com/Ouranosinc/xclim/files/6329925/sdba-testing.zip)

Balinus:
Très étrange... je suis capable de reproduire le même problème.... parfois. J'ai utilisé le même environnement Python, mais avec différentes façon d'installer ClimateTools, donc ce n'est pas le problème. C'est donc une dépendance qui fait planter le tout.... et je pense avoir de la chance. Première dépendance différente testée entre mon env qui fonctionne et l'autre qui ne fonctionne pas est la librairie [Ipopt.jl](https://github.com/jump-dev/Ipopt.jl). 

Il faut la version `0.6.5` si je me fie à mes tests. Je vais m'assurer de spécifier cette version minimum dans mon fichier `Project.toml`.

Je serais curieux de voir si c'Est le cas de ton côté. Sans tester, peux-tu imprimer le résultat de la commande suivante (dans le mode Package de Julia? (au cas où... pour rentrer dans le mode package, il suffit de peser sur `]` dans le REPL (vide))

```julia
pkg>st --manifest
```

aulemahal:
Oh, j'ai 0.6.3, ça pourrait être ça!

Sortie de `manifest`:

<details>
      Status `~/.julia/environments/v1.6/Manifest.toml`
  [79e6a3ab] Adapt v3.3.0
  [dce04be8] ArgCheck v2.1.0
  [7d9fca2a] Arpack v0.4.0
  [4fba245c] ArrayInterface v2.14.17
  [77b51b56] AverageShiftedHistograms v0.8.6
  [13072b0f] AxisAlgorithms v1.0.0
  [39de3d68] AxisArrays v0.4.3
  [6e4b80f9] BenchmarkTools v0.5.0
  [9e28174c] BinDeps v1.0.2
  [b99e7846] BinaryProvider v0.5.10
  [179af706] CFTime v0.1.1
  [336ed68f] CSV v0.6.1
  [49dc2e85] Calculus v0.5.1
  [324d7699] CategoricalArrays v0.7.7
  [35604d93] ClimateBase v0.6.0
  [4f4ee721] ClimateTools v0.22.0
  [523fee87] CodecBzip2 v0.7.2
  [944b1d66] CodecZlib v0.7.0
  [861a8166] Combinatorics v1.0.2
  [bbf7d656] CommonSubexpressions v0.3.0
  [34da2185] Compat v3.27.0
  [8f4d0f93] Conda v1.5.1
  [a9693cdc] CondaBinDeps v0.2.0
  [a8cc5b0e] Crayons v4.0.4
  [75c7ada1] DBFTables v0.2.3
  [9a962f9c] DataAPI v1.6.0
  [a93c6f00] DataFrames v0.20.2
  [864edb3b] DataStructures v0.17.20
  [e2d170a0] DataValueInterfaces v1.0.0
  [ab46fb84] DensityRatioEstimation v0.4.2
  [01453d9d] DiffEqDiffTools v1.7.0
  [163ba53b] DiffResults v1.0.3
  [b552c78f] DiffRules v1.0.2
  [3c3547ce] DiskArrays v0.2.7
  [b4f34e82] Distances v0.8.2
  [31c24e10] Distributions v0.22.6
  [da5c29d0] EllipsisNotation v1.0.0
  [e2ba6199] ExprTools v0.1.3
  [fe3fe864] Extremes v0.1.1
  [8f5d6c58] EzXML v1.1.0
  [48062228] FilePathsBase v0.7.0
  [1a297f60] FillArrays v0.8.14
  [59287772] Formatting v0.4.2
  [f6369f11] ForwardDiff v0.10.18
  [cf35fbd7] GeoInterface v0.5.5
  [dcc97b0b] GeoStats v0.11.8
  [323cb8eb] GeoStatsBase v0.10.4
  [cd3eb016] HTTP v0.8.19
  [7073ff75] IJulia v1.23.2
  [83e8ac13] IniFile v0.5.0
  [a98d9a8b] Interpolations v0.12.10
  [8197267c] IntervalSets v0.5.3
  [d8418881] Intervals v1.5.0
  [6a8e3230] InverseDistanceWeighting v0.3.7
  [41ab1584] InvertedIndices v1.0.0
  [b6b21f68] Ipopt v0.6.3
  [c8e1da08] IterTools v1.3.0
  [82899510] IteratorInterfaceExtensions v1.0.0
  [692b3bcd] JLLWrappers v1.2.0
  [682c06a0] JSON v0.21.1
  [7d188eb4] JSONSchema v0.3.3
  [4076af6c] JuMP v0.20.1
  [d293930c] KrigingEstimators v0.3.10
  [7f8f8fb0] LearnBase v0.4.1
  [d3d80556] LineSearches v7.1.1
  [30fc2ffe] LossFunctions v0.6.2
  [e80e1ace] MLJModelInterface v0.3.4
  [1914dd2f] MacroTools v0.5.6
  [b8f27783] MathOptInterface v0.9.20
  [fdba3010] MathProgBase v0.7.8
  [739be429] MbedTLS v1.0.3
  [e1d29d7a] Missings v0.4.5
  [78c3b35d] Mocking v0.7.1
  [d8a4904e] MutableArithmetics v0.2.15
  [85f8d34a] NCDatasets v0.10.4
  [d41bc354] NLSolversBase v7.5.0
  [77ba4419] NaNMath v0.3.5
  [b8a86587] NearestNeighbors v0.4.4
  [30363a11] NetCDF v0.10.3
  [6fe1bfb0] OffsetArrays v1.6.2
  [429524aa] Optim v0.20.6
  [bac558e1] OrderedCollections v1.4.0
  [90014a1f] PDMats v0.9.12
  [d96e819e] Parameters v0.12.2
  [69de0a69] Parsers v1.1.0
  [e61b41b6] PointPatterns v0.1.5
  [f27b6e38] Polynomials v1.2.1
  [2dfb63ee] PooledArrays v0.5.3
  [85a6dd25] PositiveFactorizations v0.2.4
  [92933f4c] ProgressMeter v1.5.0
  [438e738f] PyCall v1.92.3
  [1fd47b50] QuadGK v2.4.1
  [b3c3ace0] RangeArrays v0.3.2
  [c84ed2f1] Ratios v0.4.0
  [3cdcf5f2] RecipesBase v1.1.1
  [189a3867] Reexport v0.2.0
  [ae029012] Requires v1.1.3
  [79098fc4] Rmath v0.7.0
  [b0e4dd01] RollingFunctions v0.6.2
  [321657f4] ScientificTypes v0.8.1
  [8e980c4a] Shapefile v0.6.3
  [b85f4697] SoftGlobalScope v1.1.0
  [a2af1166] SortingAlgorithms v0.3.1
  [276daf66] SpecialFunctions v0.9.0
  [90137ffa] StaticArrays v0.12.5
  [2913bbd2] StatsBase v0.32.2
  [4c63d2b9] StatsFuns v0.9.7
  [3783bdb8] TableTraits v1.0.1
  [bd369af6] Tables v1.4.1
  [f269a46b] TimeZones v1.5.3
  [3bb67fe8] TranscodingStreams v0.9.5
  [30578b45] URIParser v0.4.1
  [3a884ed6] UnPack v1.0.2
  [b8865327] UnicodePlots v1.3.0
  [04a0146e] Variography v0.5.10
  [81def892] VersionParsing v1.2.0
  [ea10d353] WeakRefStrings v0.6.2
  [efce3f68] WoodburyMatrices v0.5.3
  [c2297ded] ZMQ v1.2.1
  [a5390f91] ZipFile v0.9.3
  [ae81ac8f] ASL_jll v0.1.2+0
  [68821587] Arpack_jll v3.5.0+3
  [6e34b625] Bzip2_jll v1.0.6+5
  [9cc047cb] Ipopt_jll v3.13.1+1
  [94ce4f54] Libiconv_jll v1.16.0+7
  [d00139f3] METIS_jll v5.1.0+5
  [d7ed1dd3] MUMPS_seq_jll v5.3.5+0
  [656ef2d0] OpenBLAS32_jll v0.3.12+1
  [efe28fd5] OpenSpecFun_jll v0.5.3+4
  [f50d1b31] Rmath_jll v0.3.0+0
  [02c8fc9c] XML2_jll v2.9.11+0
  [8f1865be] ZeroMQ_jll v4.3.2+6
  [a9144af2] libsodium_jll v1.0.20+0
  [0dad84c5] ArgTools
  [56f22d72] Artifacts
  [2a0f44e3] Base64
  [ade2ca70] Dates
  [8bb1440f] DelimitedFiles
  [8ba89e20] Distributed
  [f43a241f] Downloads
  [7b1f6079] FileWatching
  [9fa8497b] Future
  [b77e0a4c] InteractiveUtils
  [b27032c2] LibCURL
  [76f85450] LibGit2
  [8f399da3] Libdl
  [37e2e46d] LinearAlgebra
  [56ddb016] Logging
  [d6f4376e] Markdown
  [a63ad114] Mmap
  [ca575930] NetworkOptions
  [44cfe95a] Pkg
  [de0858da] Printf
  [3fa0cd96] REPL
  [9a3f8284] Random
  [ea8e919c] SHA
  [9e88b42a] Serialization
  [1a1011a3] SharedArrays
  [6462fe0b] Sockets
  [2f01184e] SparseArrays
  [10745b16] Statistics
  [4607b0f0] SuiteSparse
  [fa267f1f] TOML
  [a4e569a6] Tar
  [8dfed614] Test
  [cf7118a7] UUIDs
  [4ec0a83e] Unicode
  [e66e0078] CompilerSupportLibraries_jll
  [deac9b47] LibCURL_jll
  [29816b5a] LibSSH2_jll
  [c8ffd9c3] MbedTLS_jll
  [14a3606d] MozillaCACerts_jll
  [4536629a] OpenBLAS_jll
  [83775a58] Zlib_jll
  [8e850ede] nghttp2_jll
  [3f19e933] p7zip_jll
</details>

Balinus:
Excellent! 

oui, le warning est causé par mon utilisation d'une version de Extremes.jl qui date un peu. À l'époque, Extremes.jl était un WIP mais l'API est maintenant finalisé.  Ma prochaine étape est de mettre à jour cette dépendance.

Merci d'avoir testé! :)

aulemahal:
Je confirme que ça fonctionne quand même bien!

Après certains changements qui s'en viennent, j'ai des sorties assez proches entre le code julia et le python. Mais c'est dur de comparer la correction des extrêmes seulement, parce que y a des différences dans le QQM aussi... 

Question @Balinus : Est-ce que la méthode est gelée pour "Generalized Pareto"? Est-ce que quelqu'un pourrait vouloir changer de distribution?

Balinus:
J'imagine que si tu compares les extrêmes annuels (ou simplement le maximum de la série), tu devrais pouvoir comparer les résultats? Ces extrêmes ne sont en principe pas affectés par la méthode `qqmap` utilisée.

Pour l'aspect GPD, on pourrait imaginer vouloir utiliser des GEV (dans ce cas-ci, on modifierait seulement les extrêmes annuels, sans transition). Sinon, on développe présentement des techniques de correction qui utilise les modèles de [Naveau et al.  2016](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015WR018552). Ceux-ci sont entièrement paramétrique (donc pas besoin de lier le qqmap avec une méthode paramétrique).

Aussi, dans tes tests de comparaison, si ça adonne que tu crées des fonctions Julia et que le code change, [Revise.jl](https://github.com/timholy/Revise.jl) est très utile (et [OhMyREPL.jl](https://github.com/KristofferC/OhMyREPL.jl) aussi, mais pour d'autres raisons). Simplement les initilialiser avec un fichier `/home/user/.julia/config/startup.jl`. Ça rend l'expérience de développement beaucoup plus fluide.

Par exemple, voici mon fichier startup.jl:

```julia
import REPL
import Pkg

atreplinit() do repl

    # Custom Julia prompt (indique si l'on est dans un environnement de projet)
    repl.interface = REPL.setup_interface(repl)
    repl.interface.modes[1].prompt =  () -> begin
        str = Pkg.REPLMode.promptf()
        occursin(r"(v[0-9].[0-9])", str) ? "julia> " : str[1:end-6] * " > "
    end

    # Revise.jl (permet de changer les définition des fonctions sans redémarrer Julia)
    # Installer Revise.jl en premier
    @async try
        sleep(0.1)
        @eval using Revise
        @async Revise.wait_steal_repl_backend()
    catch
        @warn("Could not load Revise.")
    end

    # OhMyREPL.jl (autocomplétion des parenthèses, brackets, etc..)
    # Préalablement installer OhMyREPL.jl
    # Workaround pour VS Code
    @eval using OhMyREPL
    @eval OhMyREPL.enable_autocomplete_brackets(true)
end
```

Balinus:
Super! 

Voici la citation provisoire pour la documentation.


>Roy, P., Rondeau-Genesse, G., Jalbert, J., Fournier, É. 2021. Climate Scenarios of Extreme Precipitation Using a Combination of Parametric and Non-Parametric Bias Correction Methods. *Submitted to Climate Services, April 2021*.

Note qu'en général, la méthode ne fait pas de miracles lorsqu'on utilise cette méthode sur NRCan ou autre datasets non-appropriés pour les extrêmes. L'adage *Garbage-in, garbage-out* s'applique très bien. 😄

aulemahal:
Ajout de tests et documentation, c'est prêt pour l'examen!

xclim/sdba/adjustment.py
```diff
@@ -406,6 +459,168 @@ def _adjust(self, sim, interp="nearest", extrapolation="constant"):
...
+            # Clusters of large values of sim
+            _, _, sim_posmax, sim_maxs = get_clusters_1d(sim, thresh, cluster_thresh)
+
+            new_scen = scen.copy()
+            if sim_posmax.size == 0:
+                # Happens if everything is under `cluster_thresh`
+                return new_scen
+
+            # Fit the dist, force location at thresh
+            sim_fit = stats._fitfunc_1d(
```
huard, line 580: I guess this assumes a detrended sim series ? Worth mentioning in the docstring ?

xclim/sdba/adjustment.py
```diff
@@ -406,6 +459,168 @@ def _adjust(self, sim, interp="nearest", extrapolation="constant"):
...
+        )
+        self.set_dataset(ds)
+
+    def _adjust(
+        self,
+        scen,
+        sim,
+        frac: float = 0.25,
+        power: float = 1.0,
+    ):
```
huard, line 569: Docstring. I know this is a private function, but it's not clear what power does.

xclim/sdba/adjustment.py
```diff
@@ -406,6 +459,168 @@ def _adjust(self, sim, interp="nearest", extrapolation="constant"):
...
+            new_sim = dist.ppf(sim_cdf, *ref_params) + thresh
+
+            # Get the transition weights based on frac and power values
+            transition = (
+                ((sim_maxs - sim_maxs.min()) / ((sim_maxs.max()) - sim_maxs.min()))
+                / frac
+            ) ** power
+            np.clip(transition, None, 1, out=transition)
+
+            # Apply smotth linear transition between scen and corrected scen
```
huard, line 596:
```suggestion
            # Apply smooth linear transition between scen and corrected scen
```

xclim/sdba/utils.py
```diff
@@ -586,3 +586,124 @@ def best_pc_orientation(A, Binv, val=1000):
...
             err = new_err
     return orient
+
+
+def get_clusters_1d(data: np.ndarray, u1: float, u2: float):
+    """Get clusters of a 1D array.
+
+    A cluster is defined as a sequence of values larger than u2 with at least one value larger than u1.
+
+    Inpsired from julia's Extremes package, function getcluster.
```
huard, line 596: Move to Notes.

xclim/sdba/utils.py
```diff
@@ -586,3 +586,124 @@ def best_pc_orientation(A, Binv, val=1000):
...
+
+    Parameters
+    ----------
+    data: 1D ndarray
+      Values to get clusters from.
```
huard, line 601: I'm worried using the term "cluster" will create confusion. We're using cluster in the ensembles function, and here it is used to mean something else entirely. 

I suspect this is the word used in the Julia code, but I suggest once Phil has reviewed the PR, we switch the name to something else. 

Elsewhere we use the term "run" or "sequence", no ? We could call those evs (extreme value sequences) ?

Related question: why does it matter that these are sequences ? In my mind, the Pareto describes the distribution of `x - thresh` for `x > thresh`, so it's not clear where `u1` comes from, and why the concept of sequence/clusters is needed.

xclim/sdba/adjustment.py
```diff
@@ -146,6 +150,52 @@ def _adjust(self, sim, **kwargs):
         raise NotImplementedError
 
 
+class AdjustmentCorrection(BaseAdjustment):
+    """Base class for Adjustment correcting objects."""
+
+    def adjust(self, scen: DataArray, sim: Optional[DataArray] = None, **kwargs):
+        """Modifies bias-adjusted data. Refer to the class documentation for the algorithm details.
```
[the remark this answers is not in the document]
huard: Second order adjustment ?

xclim/testing/tests/test_sdba/test_adjustment.py
```diff
@@ -515,6 +519,100 @@ def dist(ref, sim):
...
+        EX.train(ref, hist)
+
+        qv = (ref.thresh + hist.thresh) / 2
+        np.testing.assert_allclose(EX.ds.fit_params, [-0.1, qv, 2], atol=0.5, rtol=0.1)
+        np.testing.assert_allclose(EX.ds.thresh, qv, atol=0.15, rtol=0.01)
+
+        scen2 = EX.adjust(scen, sim, frac=frac, power=power)
+
+        # What to test???
+        # Test if extreme values of sim are still extreme
```
huard, line 567: My question as well. @Balinus Are you confident we're testing the right thing, here?

huard (approved):

xclim/sdba/utils.py
Balinus: `clusters` word comes mainly from Coles (2001) and is the appropriate word in the statistical context (but not within the xclim nomenclature). 

The usefulness of declustering comes from the assumption that the convergence assumes independence between extracted exceedence. This assumption does probably not hold if you get multiple exceedence from a single weather event. So the clusters are defined by a lower threshold and we only extract the maximum value within each cluster.

Perhaps this assumption can be relaxed though and be linked to a keyword argument `decluster=True`.

huard: Ok, je comprend mieux ! On garde ça comme ça.

xclim/sdba/adjustment.py
Balinus: Right now the transition function is hardcoded and we played a bit with the `power` factor and settled on `1.0` in the end. What I'd like to do ultimately is provide the user with the tools so that they can provide their own definition of a transition function to be applied.

I guess that here we simply need a definition of the transition function somewhere in the docstrings. (I admit the function is a little convoluted).

xclim/sdba/adjustment.py
```diff
@@ -483,23 +483,28 @@ def __init__(
...
         Notes
         -----
-
-
-        Correction strength:
+        Extreme values are extracted from `ref`, `hist` and `sim` by finding all "clusters",
+        i.e. runs of consecutive values above `cluster_thresh`. The `q_thresh`th percentile
+        of these values is taken on `ref` and `hist` and becomes `thresh`, the extreme value
+        threshold. The maximal value of each cluster of `ref`, if it exceeds that new threshold,
+        is taken and a `dist` (default : Generalized Pareto) distribution is fitted to them.
```
Balinus, line 504: Here, if you take the exceedence approach, the only proper default is a Generalized Pareto. So, if you change the default, the approach should not be an exceedence approach.

huard: I suggest that, for now, we force the pareto and disable the option to change the distribution. Opening to other distributions would require more testing, changes to the _clustering_ algorithm and possibly other more subtle changes.

xclim/sdba/adjustment.py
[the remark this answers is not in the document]
aulemahal: I thought of Bias Adjustment Adjustment?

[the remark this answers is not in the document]
aulemahal: Lol, no "Second Order Adjustment" sounds good!

xclim/testing/tests/test_sdba/test_adjustment.py
Balinus: Well, they way I tested it was to create a synthetic array where I manually tested what the new extreme value should be based on the cdf quantile estimations. The result of this test was then transfered .

I admit this is quite *ad hoc* and not the best way to do it. Other tests should be developed in the coming week and months as I finalize the coding (the algo will remain stable).

xclim/sdba/adjustment.py
```diff
@@ -406,6 +459,168 @@ def _adjust(self, sim, interp="nearest", extrapolation="constant"):
...
+    def _train(self, ref, hist):
+        cluster_thresh = convert_units_to(self.cluster_thresh, ref)
+        hist = convert_units_to(hist, ref)
+
+        # Extreme value threshold computed relative to "large values".
+        # We use the mean between ref and hist here.
+        thresh = (
+            ref.where(ref >= cluster_thresh).quantile(self.q_thresh, dim="time")
+            + hist.where(hist >= cluster_thresh).quantile(self.q_thresh, dim="time")
+        ) / 2
```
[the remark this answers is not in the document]
aulemahal: Hum, our definition of "hist" is the timeseries of _raw_ model data, so there might be a difference between the two thresh.

There's some ambiguity here on whether `ExtremeValue.train` happens after or before the "first-order" Adjustment.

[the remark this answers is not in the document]
Balinus: Right, if already adjusted. The original algo estimate it on non-adjusted vector. Either way should do the trick, since minor difference will not affect significantly the convergence (I think!).

Balinus:
Perhaps the only thing missing is a way to bypass the estimated GPD parameters. The way it was designed in the first place (well, *why* it was developed) was to use gridded datasets such as NRCan, without using the extremes from the dataset (no real extreme information in NRCan), but by using *external* parameters (in our case, it was actually GEV parameters, hence the function to convert from GEV to GPD parameters somewhere in the code).

Of course, if you use station data, it would work better for the extreme part, but you do not get the benefit of a gridded dataset.

The external GEV parameters should be published in the coming months and will be public as far as I can tell.

xclim/sdba/adjustment.py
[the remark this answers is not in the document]
aulemahal: But, is there a reason that the thresh is not recomputed for the future data? I mean, this way the "extreme values" of a given array would be those with  `quantile(arr[arr > cluster_thresh]) > q_thresh`.

Here we transform the quantile thresh into a data thresh on ref+hist, and then use it on sim. Could make sense to stick to the quantile thresh on both?

xclim/sdba/adjustment.py
aulemahal: @huard, like in other adjustment object, everything is explained in the class docstring. The second part of the notes explains `frac` and `power`. In the same wy as we explain `interp` in EQM's docstring, not in `EQM._adjust`.

aulemahal:
I added the GEV to GPD conversion in the  `train`, fit params can be passed in `ref_params`.

The signature of the `train` method is now different from the base "SecondOrderAdjustment" and there are no other subclasses of this one. So I decided to scrap that and just have ExtremeValues be a child of BaseAdjustment and override `train` and `adjust` directly. We can recreate a `SecondOrderAdjustment` thing when it becomes usefull.

Balinus:
Awesome!

xclim/testing/tests/test_sdba/test_adjustment.py
Balinus: Perhaps one test that could be designed is to test the bias-corrected factor for the tail parameters? I'm thinking out loud here. Perhaps it does not work.

Something like (pseudo-code)
```python
scen2 = EX.adjust(scen, sim, frac=0.01, power=power, gpparam) # 0.01 so that the transition function application domain is small and we provide gpparams here.
# sim is over the same historical period 

# Find q97 (so that we are pretty sure that the transition function of scen2 is not applied
q97 = quantile(scen2, 0.97)
exceedence = scen2[scen2>q97] - q97

gpparam_scen = gpfit( exceedence )
gpparam_scen === gpparam # triple "=", assuming it exist in Python. Or test the values of the parameters with a tolerance.
```

Does that make sense? Intuitively, the tails should be similar over the historical period.

===
Another point: does the quantile/GP parameters from the simulation are estimated over a qqmap corrected vector or from the raw/original simulation? Because the original algo is over the non-corrected values. Reading the code quickly seems like it is applied to the corrected values.

aulemahal: Cool thanks! I'll try that test.

We call ""ref", "hist" and sim" the uncorrected values, "scen" is the adjusted version of "sim". So yes both distributions are fitted over uncorrected values ("ref" in the `train` and "sim` in  `adjust`). Well, if the user decides so, we're not testing anything and the first order correction happens outside `ExtremeValues`'s knowledge.

aulemahal:
Oh and note that I didn't add a mechanism to find the closest parameters vector when `ref_params` is passed (while ClimateTools.jl does that). I leave this to the user, as this is easy with xarray. (The only restriction is that `ref_params` is "alignable" with `sim` and `scen`; it can be broadcasted to their shape and coordinates. )

xclim/testing/tests/test_sdba/test_adjustment.py
aulemahal: May be something I didn't understand, but the test doesn't work... Tried it with the julia code as well to be sure, same output : very different fit parameters. But isn't that normal? Now thinking about it : the gp params are from ref, fitted on the clusters maximums with an "extreme value" threshold at quantile 0.95. Why would they be the same as gp params fitted on values of `scen` above quantile 0.97? It's not the same number of data points, no the same minimum and there is no clustering.

Balinus: (using your nomenclature)
Yes indeed, good point!

So another way would be to see if, **without** providing gp params, if you have the same gpfit over `scen` and `ref` (both with quantile 0.97). By not providing gp param, the method falls back on estimated the gp param from `ref` and thus, the tails should be similar?

But this is a long shot, since the data points are not the same and there is potential compensation between distribution's parameters.

aulemahal: Well. Am I doing something wrong?

```julia
using ClimateTools

Dsim = load("xclim-testdata/sdba/fermont_CanESM2_1950-2100.nc", "pr", data_units="mm");
Dobs = load("xclim-testdata/sdba/fermont_nrcan_1950-2013.nc", "pr", data_units="mm");

ref = temporalsubset(Dobs, (1980, 1, 1), (2009, 12, 31));
hist = temporalsubset(Dsim, (1980, 1, 1), (2009, 12, 31));
sim = temporalsubset(Dsim, (2040, 1, 1), (2069, 12, 31));

scen = biascorrect_extremes(ref, hist, sim, detrend=false, window=15, rankn=50, thresnan=0.1, keep_original=false, frac=0.000000001);

refvec = ref[1].data[1,1,:];
histvec = hist[1].data[1,1,:];
scenvec = scen[1].data[1,1,:];

P = 0.95;
thres = ClimateTools.get_threshold(refvec, histvec, thres=P);

refclusters = getcluster(refvec, thres, 1.0);
GPD_ref = gpdfit(refclusters[!,:Max], threshold = thres)
# >>> Distributions.GeneralizedPareto{Float64}(μ=14.445050263404848, σ=5.912177063426245, ξ=-0.032137958612208126)

scenclusters = getcluster(scenvec, thres, 1.0)
GPD_scen = gpdfit(scenclusters[!,:Max], threshold = thres)
# >>>Distributions.GeneralizedPareto{Float64}(μ=14.445050263404848, σ=15.727722763690819, ξ=0.0)
```

xclim/sdba/adjustment.py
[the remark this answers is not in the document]
Balinus: What you modelize with the GPD is an exceedence distribution over a given threshold, not over a quantile. So that's why the threshold remain constant. You could however modelize the evolution of GPD parameters, but this is somewhat more hard to do and clearly beyond the scope of the developed method. 

However, it is easier to develop a non-stationairy approach with GEV parameters and this could be used as input more easily and is the next phases I was envisioning when we did the method.

xclim/testing/tests/test_sdba/test_adjustment.py
aulemahal: I wrote the parameters as comments after the call, here they are:
GPD_ref = `Distributions.GeneralizedPareto(μ=14.445050263404848, σ=5.912177063426245, ξ=-0.032137958612208126)`
GPD_scen = `Distributions.GeneralizedPareto(μ=14.445050263404848, σ=15.727722763690819, ξ=0.0)`.

Using 1950-2009, adjusting `hist` instead and using quantile 0.97.
GPD_ref = `Distributions.GeneralizedPareto(μ=16.861221017837533, σ=5.71010332586595, ξ=0.04983672744022191)`
GPD_scen = `Distributions.GeneralizedPareto(μ=16.861221017837533, σ=14.653519054140373, ξ=0.0)`

Balinus: Interesting. I'll take a look when I have time. Can you extract the annual maximum boxplots easily? 

I guess that since we have different data point in the end, the different parameters reflect the uncertainty of the estimation.

aulemahal: Hum. So here are the box plots for `data.resample(time='YS').max()`, and all timeseries. `scen` is simple EQM (qqmap) and `scen2` is the ExtremeValues second-order adjusted data. AFAIK, the julia and python calls use the same arguments.

image

Balinus: Interesting! I'll have to take a dive in the coming days and compare. The previous image showing difference around 1e-4 was based only all data over Q95? 

For the sake of simplicity, what is your code for a MWE on the Python side? I assume the Julia code is the one above. 

Thanks!

aulemahal: The previous image of differences was between the results. I.e. distance between adjusted extremes in julia and python (`scen2[scen2 != scen]`). Q95 indeed.

I realize that my ref is from nrcan and you did say that extremes were badly represented there. I redid the whole thing with new data. `ref` is now AHCCD (ECCC station data, so better?). Redid the extraction for nrcan and CanESM2, for the same locations : Vancouver (BC, great rain), Kugluktuk (NU, very dry), Amos (QC, in between).

Python code:

```python
import numpy as np
import xarray as xr
import xclim as xc

from xclim import sdba

dsim = open_dataset('sdba/CanESM2_1950-2100.nc', branch='sdba-testfiles')
dref = open_dataset('sdba/ahccd_1950-2013.nc', branch='sdba-testfiles')

ref = xc.core.units.convert_units_to(dref.sel(time=slice('1950', '2009')).pr, 'mm/d') 
hist = xc.core.units.convert_units_to(dsim.sel(time=slice('1950', '2009')).pr, 'mm/d')

quantiles = np.linspace(0.01, 0.99, num=50)

    ref = ref + sdba.processing.uniform_noise_like(ref, low=1e-6, high=1e-3)
    hist = hist + sdba.processing.uniform_noise_like(hist, low=1e-6, high=1e-3)
    
EQM = sdba.EmpiricalQuantileMapping(group=sdba.Grouper('time.dayofyear', window=31), nquantiles=quantiles)
EQM.train(ref, hist)

scen = EQM.adjust(sim, interp='linear', extrapolation='constant')

EX = sdba.adjustment.ExtremeValues(cluster_thresh='1 mm/day', q_thresh=0.97)
EX.train(ref, hist)
scen2 = EX.adjust(scen, hist, frac=0.00000001)
```

Results are far more interesting!: 

image

Data is here https://github.com/Ouranosinc/xclim-testdata/tree/sdba-testfiles/sdba

But I prepared a ClimateTools.jl-ready version because it is a bit more strict on the layout : 
[sdba-testing.zip](https://github.com/Ouranosinc/xclim/files/6367881/sdba-testing.zip)

Balinus: Great! So `scen2` and `scen2-JL` are quite similar then?

aulemahal: Yes! They were already almost the same with the previous data. But the new one also gives more interesting results overall. Reading the codes, I can't see why the EQM/qqmap scens are so different, but that doesn't really matter here.

Balinus: Nice! 

for the qqmap aspect, I guess it is simply a matter of the choice made. I have not read deeply your code, hence can't comment on this aspect.

For the tests we mentioned, it does not make sense in the end. We can't expect to have the same GPD parameters since they are not the same (quantiles) samples.

Balinus (approved):
Great addition :)

PR #691, opened 2021-04-15
Implement doy to days since and inverse

### LINKED ISSUE #688: Implement a 'doy_to_days_after' and reverse functionality
### Description
Certain indices outputting 'day of year' or `doy` values are not directly usable for ensemble percentiles and statistics due to its circular nature.   

Averaging `doy` results can potentially be handled by users via https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.stats.circmean.html.  However I have not easily found a percentiles option.

The other solution is to  create offset functions allowing units to become a 'days after %mm%dd' or similar which will allow valid ensemble statistics to be calculated..   Final results (mean, percentiles etc) could potentially be restored to original `doy` via a reverse function.

To streamline wps calls in `finch` it would be potentially interesting to have an `units` parameter for the calculation of indicators that currently return to `doy` to output `days_after` instead.  Non exhaustive list of indices include  `degree_day_exceedance`, `first_day_*`, `last_day_*`  

Note I think xarray may interpret 'days_since' as a cftime flag and try to automatically 'help' in certain cases. This is why I suggest a `days_after` unit for now but I am open to debate / validation

 

 

```

### PR DESCRIPTION
**Pull request by aulemahal**
### Pull Request Checklist:
    - This PR fixes #688

Adds `xc.core.calendar.doy_to_days_since` and inverse `days_since_to_doy`. It converts between data in day of year to a number of days since a given date. By default, that date is the time coordinate of the value, but it can be changed with `start=MM-DD`. Also, passing `calendar` is possible to change the calendar in which the computation is done, otherwise the calendar of the time axis is used. In the inverse function, `start` and `calendar` can be forced, but they are read from the attributres otherwise.

No.

### DIFF
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -968,7 +968,7 @@ def days_since_to_doy(
     out = dac + start_doy
     out = xr.where(out > doy_max, out - doy_max, out)
 
-    out.attrs.update(da.attrs)
-    del out.attrs["calendar"]
-    del out.attrs["units"]
+    out.attrs.update(
+        {k: v for k, v in da.attrs.items() if k not in ["units", "calendar"]}
+    )
     return convert_calendar(out, base_calendar)
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -908,9 +908,9 @@ def doy_to_days_since(
     out = xr.where(dac > base_doy, dac, dac + doy_max) - start_doy
     out.attrs.update(da.attrs)
     if start is not None:
-        out.attrs.update(units=f"days since {start}")
+        out.attrs.update(units=f"days after {start}")
     else:
-        out.attrs.update(units="days since time coordinate")
+        out.attrs.update(units="days after time coordinate")
 
     out.attrs.update(calendar=calendar)
     return convert_calendar(out, base_calendar)
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -910,7 +910,11 @@ def doy_to_days_since(
     if start is not None:
         out.attrs.update(units=f"days after {start}")
     else:
-        out.attrs.update(units="days after time coordinate")
+        starts = np.unique(out.time.dt.strftime("%m-%d"))
+        if len(starts) == 1:
+            out.attrs.update(units=f"days after {starts[0]}")
+        else:
+            out.attrs.update(units="days after time coordinate")
 
     out.attrs.update(calendar=calendar)
     return convert_calendar(out, base_calendar)
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -26,7 +26,7 @@ from xarray.coding.cftime_offsets import (
 from xarray.coding.cftimeindex import CFTimeIndex
 from xarray.core.resample import DataArrayResample
 
-from xclim.core.utils import _calc_perc
+from xclim.core.utils import DayOfYearStr, _calc_perc
 
 # cftime and datetime classes to use for each calendar name
 datetime_classes = {"default": pydt.datetime, **cftime._cftime.DATE_TYPES}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tlogan2000** [comment]
@aulemahal  the negative `days_since` (examples) would typically not ever happen following the resample in the indices calculations right?  I was expecting a range of say 0-365 almost all the time but I think this will be the case for the vast majority of our index calculations

**Comment by aulemahal** [comment]
Generally speaking they wouldn't happen because `start=None`. They happen if the start date is given and the doy value falls beween the time coordinate and that start date. Example : data resampled with YS, doy of 15 and start=02-01, makes `days_since = -15`.

For the simple use case of performing statistics, one doesn't need to bother with `start`, so,  yes, "days_since" will always be in the [0, 365] range.

**Comment by tlogan2000** [comment]
Here's where I found the cftime bug .... doy_to_days_since then tried to build and ensemble from outputs
````
da_list = []
for dsince in ['08-01','07-01','08-15']:
    out = doy_to_days_since(da,start=dsince)
    #out.attrs['units'] = out.attrs['units'].replace('days since','days after')
    da_list.append(out.to_dataset(name='dsince'))
ens.create_ensemble(da_list)

ValueError: unable to decode time units 'days since 08-01' with "calendar 'default'". Try opening your dataset with decode_times=False or installing cftime if it is not installed.


````

**Comment by aulemahal** [comment]
Oh! I thought it would simply skip on invalid units... I'll change it.

**Comment by aulemahal** [comment]
@Zeitsperre Do you know why the Travis CI builds are being triggered?

**Comment by tlogan2000** [comment]
> Here's where I found the cftime bug .... doy_to_days_since then tried to build and ensemble from outputs
> 
> ```
> da_list = []
> for dsince in ['08-01','07-01','08-15']:
>     out = doy_to_days_since(da,start=dsince)
>     #out.attrs['units'] = out.attrs['units'].replace('days since','days after')
>     da_list.append(out.to_dataset(name='dsince'))
> ens.create_ensemble(da_list)
> 
> ValueError: unable to decode time units 'days since 08-01' with "calendar 'default'". Try opening your dataset with decode_times=False or installing cftime if it is not installed.
> ```



> Oh! I thought it would simply skip on invalid units... I'll change it.

Is it new that you have to provide a name to `to_dataset()`?  Seems to break `create_ensemble` when using a list of DataArrays

**Comment by aulemahal** [comment]
`to_dataset` takes  `dataarray.name` and asks for it if not set, may be `doy_to_days_since` destroys the previous name?

**Comment by tlogan2000** [comment]
> @Zeitsperre Do you know why the Travis CI builds are being triggered?

@tlvu managed to get us free credits I think!

**Comment by Zeitsperre** [comment]
Yeah, looks like it. If we want, I can disable them here since we've migrated.

**Comment by aulemahal** [comment]
Well for now it's mostly noise because the `travis.yml` file is commented out. Haha, would there be a way to trigger one only if the minutes of the other are emptied?

**Comment by Zeitsperre** [comment]
@aulemahal They're separate services, so that could get **very** complicated. 

My suggestion would be to officially disable Travis CI on Ouranosinc and re-enable the repositories that we want it on, one by one. @tlvu thoughts?

File: xclim/core/calendar.py [hunk 1]
```diff
@@ -807,3 +807,168 @@ def time_bnds(group, freq):
...
+    # 2cases:
+    # val is a day in the same year as its index : da - offset
+    # val is a day in the next year : da + doy_max - offset
+    out = xr.where(dac > base_doy, dac, dac + doy_max) - start_doy
+    out.attrs.update(da.attrs)
+    if start is not None:
+        out.attrs.update(units=f"days after {start}")
+    else:
+        out.attrs.update(units="days after time coordinate")
+
```
tlogan2000, line 914:
```suggestion
out.name = da.name
```
Can confirm da.name is lost somewhere along the line

xclim/testing/tests/test_calendar.py
```diff
@@ -377,3 +379,42 @@ def test_datetime_to_decimal_year(source_cal, exp180):
...
+
+def test_doy_to_days_since():
+    # simple test
+    time = date_range("2020-07-01", "2022-07-01", freq="AS-JUL")
+    da = xr.DataArray([190, 360, 3], dims=("time",), coords={"time": time})
+
+    out = doy_to_days_since(da)
+    np.testing.assert_array_equal(out, [7, 178, 186])
+
+    da2 = days_since_to_doy(out)
```
tlogan2000, line 392: line 392 :  same calculation with the valid `start` via `doy_to_days_since(da,start='07-01')`   gives different results from original `da = [190, 360, 3]`  but `da2 = [190, 360, 368]`

tlogan2000 (changes requested):
A few things to think about still I think.  Reference year for lack of a better term seems to get lost sometimes ... see my examples

tlvu:
> @aulemahal They're separate services, so that could get **very** complicated.
> 
> My suggestion would be to officially disable Travis CI on Ouranosinc and re-enable the repositories that we want it on, one by one. @tlvu thoughts?

Yes out of the blue Travis-CI replied they gave us more OSS credits and once they run out we request again.  That's bogus !  I don't want to spend time running after them every month.  Moreover their support is not so "fast".  I was questioning them whether they will recognize our OSS status and it took them 4 months to reply.\

So I say we disable Travis-CI for good or at least for repos that already migrate.

Zeitsperre:
I've suspended Travis CI access for now. We'll know if this disables the builds on next push.

tlogan2000:
> > @aulemahal They're separate services, so that could get **very** complicated.
> > My suggestion would be to officially disable Travis CI on Ouranosinc and re-enable the repositories that we want it on, one by one. @tlvu thoughts?
> 
> Yes out of the blue Travis-CI replied they gave us more OSS credits and once they run out we request again. That's bogus ! I don't want to spend time running after them every month. Moreover their support is not so "fast". I was questioning them whether they will recognize our OSS status and it took them 4 months to reply.\
> 
> So I say we disable Travis-CI for good or at least for repos that already migrate.

They might have started to see a backlash over their recent move? Who knows.   I agree that ping ponging between services makes no sense. Repos that we have moved to github actions or other services should stay as is unless we start to see major issues as we move forward

xclim/testing/tests/test_calendar.py
aulemahal: Woups. That's a ">" that should be a ">=" (calendar.py line 848)...

tlogan2000: Just figured this out too!

tlogan2000 (approved):

tlogan2000:
@aulemahal this looks good now.  we could merge and maybe make a secondary issue for providing a mechanism for output units from the `xclim.indices` calls  preference?

aulemahal:
Agreed! I'll merge when all tests complete.

PR #694, opened 2021-04-23
* add attributes to indices that return 'dayofyear'

### LINKED ISSUE #688: Implement a 'doy_to_days_after' and reverse functionality

tlogan2000:
### Pull Request Checklist:
    - This PR fixes #688  -  addtionally adds attributes to help recognize `doy` variables in WPS ensemble statistics processes
- [X] Documentation has been added / updated (for bug fixes / features)

no 

* **Other information**:

### DIFF
--- a/xclim/core/calendar.py
+++ b/xclim/core/calendar.py
@@ -872,7 +872,7 @@ def doy_to_days_since(
       Passing `start` only makes sense if `da` has a yearly sampling frequency.
     calendar: str, optional
       The calendar to use when computing the new interval.
-      If None (default), the calendar of the `time` axis is used.
+      If None (default), the calendar attribute of the data or of its `time` axis is used.
       All time coordinates of `da` must exist in this calendar.
       No check is done to ensure doy values exist in this calendar.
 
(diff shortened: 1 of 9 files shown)


aulemahal (approved):
More explicit this way!

I'll add a small change to `doy_to_days_since` to make use of this!

aulemahal:
Test fails are my fault. Will fix today.

@huard @Zeitsperre  to keep you in the loop: as the CF standard do not provide a way to mark a doy variable as such (units are "", which is uninformative, and there is no standard_name for it), we decided to add a `is_dayofyear` custom attribute. At least inside xclim and family we'll be able to use it. Additionally, a  `calendar`  attribute so that if the dataset's calendar gets converted, there is a way to notice that the doys were computed with another calendar.

tlogan2000:
> Test fails are my fault. Will fix today.

ok thx

PR #704, opened 2021-05-05
FWI bug fixes and re-implement "GFWED" specifics.

### LINKED ISSUE #676: FWI notebook with DC overwintering
Make a notebook using era5 data to compute the FWI indices, with and without overwintering, in order to compare with https://essd.copernicus.org/articles/12/1823/2020/.

I think we could extract the same years and points as in our "era5 can cities" data from the public datasets they provide. But because of the data differences (mean temp vs. temp at noon) we might need to create a new testdata set anyway.
This notebook would also act as a more advances xclim/xarray example and integrated test for the whole FWI submodule.

### LINKED ISSUE #700: Reimplement "wet startup" in FWI
Last FWI PR (#664) removed some "wet startup" code for DC and DMC because it wasn't covered in the cffdrs R package. It appears that this switch is however quite commonly used in the field, so it makes most sense to have it in xclim. 

Furthermore, there is a small bug in the standalone `overwintering_drought_code`. `output_core_dims` should be `[[]]` and not `[[]] * 5`. And it is missing a test.

Ref ECCC-CCCS/CanLEAD-FWI#6 (see comments at the end).

aulemahal:
### Pull Request Checklist:
    - This PR fixes #700.

- Reimplements the specific options and mechanisms that are in GFWED's matlab code but not in the CFS R package.
  * New `season_method="GFWED"` fire season method. It is quite similar to "LA08", but thresholds are compared to the _mean_ snow and temp, instead of the extreme of the last X days. Also, temperature is also involved in the start up threshold.
  * New `dry_start={None, "CFS", "GFWED"}` argument. If not None, DC and DMC start up values are function of the number of days since the last precipitation event. There are two sligthly different ways, the "GFWED" including a condition where if snow cover was important enough, the normal start up values are used (no matter the precipitation).
  * A few new numerical parameters controlling the thresholds of "dry start".
- Fixed a bug in the overwintering process where the precipitation on the day of the start up would be added to the accumulated winter precip when estimating the overwintered DC (but it shouldn't it's already included in the day's DC calc).
- Fixes a bug in the `overwintering_drought_code` indice, adds a test for it.
- Small reorganization of how numerical parameters arguments are managed in the code. All defaults are now set in `xc.indices.fwi.default_params`, instead of being set in multiple places. This had the negative effect of removing some kwargs from the indice and indicator calls. They are clear if you read the doc, but do not appear explicitly in the signature (and thus are not supported by external parsers like finch).

TODO:
- [ ] Add a notebook with call examples for CFS-like and GFWED-like tests.
- [x] Merge the test data into the main xclim-testdata branch.

I don't think so?

## Other information

@laura-vanvliet Do you have time to run your tests again against this branch? I re-implemented the "dry" start mechanism (DMC overwintering). For your case, I think the call becomes:

```python3
fire_weather_ufunc(
    tas = tasmaxAdjust,
    pr = prAdjust,
    ws = sfcWindAdjust,
    rh = hursAdjust,
    lat = lat,
    season_mask = xr.where(mask['mask_winter'] == 1, False, True),
    overwintering = True,
    dry_start = "CFS",
    prec_thresh=1.5,
    dmc_dry_factor=1.2,
)
```
The default values I put for the numerical parameters are those from GFWED, so we need to override them here. The "dry 
start" mechanism works with both DC and DMC, but overwintering overrides the DC part.

I wasn't sure how to name the "dry_start" methods. Before it was just a boolean, but I wanted to have a GFWED-compliant function, so both methods are implemented here with names "CFS" and "GFWED"...


aulemahal:
Sorry @tlogan2000, you have been chosen to review.

As said in the discussion in #700, I decided to go on with this PR even if we are 100% sure it fits the needs of ECCC-CCCS. It still does what is expected and can replicate both cffdrs AND GFWED.

For now, I did not go on with a notebook, but I did add an `example` section in the module's doc, which shows two basic-but-complete examples for CFS-like and GFWED-like calculations. The notebook will come later on.

The only test failing is CodeFactor. Sorry about that function, which has more if-statements than an undergrad's _machine learning_ code. While it may have removed the warning, I didn't want to cut it into pieces simply for the sake of it.

tlogan2000 (approved):

aulemahal:
Will merge when tests pass.

Issue #505, opened 2020-07-17
Consecutive threshold spanning across intervals


pbreach:
The climdex.pcic package provides options to specify whether indices that are computing consecutive days above or below a threshold can span across that interval. For example: https://rdrr.io/cran/climdex.pcic/man/climdex.cdd.html using `spells.can.span.years` option allows for consecutive dry days spanning across two years to be counted and then assigned to the year in which the dry spell began

Is this something that would be useful to include here or is there a workaround in xclim? I wrote some code on this using itertools.groupby to break out all the consecutive days and then grouped into intervals in the past. Any thoughts on if a PR would be welcome for this?

huard:
Hi @pbreach, 

A PR would certainly be welcome !

I *think* some of our utilities already support this, but maybe not all. This is something we've looked at in the past and I seem to remember we hit memory issues. xclim uses dask and so we need to create memory "halos" around each chunk to achieve this while preserving dask's parallelization. 

The relevant pieces of code are in `xclim/indices/run_length.py`. You'll see that some functions first create a `rolling` object that  spans across periods. Extending or improving on this concept for the indices that you're interested in would be appreciated.

aulemahal:
I had a quick look at the problem and @huard is indeed right: the reason this is not implemented in xclim is because of high memory usage by the run length methods.

Right now for any "run length" method, xclim resamples (groups) the data to the requested frequency and then, on each groups, applies the method. Example with an "annual longest heat spell" returning the length of the longest series of consecutive days with tmax above a threshold:
1) Cut the data into years
2) Get the boolean array of tmax > thresh
3) resample to yearly groups 
For each group:
  4) Apply `run_length.rle`, which puts the length of a series of consecutive `True` at the start of that series.
  5) Get the max of that array.

In order to manage the spells spanning across periods we could invert 3 and 4, i.e.: apply `rle` on the whole dataarray and then resample and take the max of each group. However, even if `rle` is dask-ready and quite optimized, it uses more memory than the current way. Also, that change would need us to rethink the API a bit.

For reference, a graph comparing the two methods on generic test data:
![overlaps_midchunks](https://user-images.githubusercontent.com/20629530/88727504-c5ad6580-d0fd-11ea-8bc2-2e29908c0898.png)
[...]

PR #716, opened 2021-05-19
Freeze-thaw spells statistics

aulemahal:
### Pull Request Checklist:
- [ ] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #xyz
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] HISTORY.rst has been updated (with summary of main changes)
- [ ] `bumpversion (major / minor / patch)` has been called on this branch
- [ ] Tags have been pushed (`git push --tags`)

* **What kind of change does this PR introduce?** 
Adds indice `freezethaw_spell` and indicators `freezethaw_spell_frequency`, `freezethaw_spell_mean_length` and `freezethaw_spell_max_length`. 
Adds "translations"  in the default formatter for common arguments to `op` ( or `reducer`).
Adds a `window` argument to `rle_statistics`.

* **Does this PR introduce a breaking change?** 
No. I didn't touch at `daily_freezethaw_cycles` , even though it could be obtain with `freezethaw_spell(..., window=1, op='sum')`.

* **Other information**:
To follow work of previous PR, I should have named this with a more general name, but I couldn't think of a good one. "tn_below_tx_above_spell"?

I'm also not completly sure about the wording, both in french and english.

This for the MTQ project.

PR #718, opened 2021-05-20
Improvements in run-length logic and tests

### LINKED ISSUE #657: Many run length methods are not respecting their `dim` argument
* xclim version: master

Methods in `run_length` accept a `dim` argument to control over which dimensions the runs are sought for. However, a lot of them have hardcoded uses of `'time'`, instead of using that `dim` argument.

In the vast majority of cases, this is not an issue, but it happens. In another project, I used `first_run` over a spatial dimension, but had to code a workaround.

### LINKED ISSUE #717: Run length - dask-version not accepting integers.
The non-ufunc version of `run_length` (`xc.indices.run_length.rle`) doesn't accept integer arrays, while the `ufunc` one does (`xc.indices.run_length.rle_1d`).

On one side, this makes sense because the latter does more than the former : it finds all runs of _different values_. Not only runs of True values. However, this was the cause of a bug in #716 that was really hard to pin down. Passing integers (0 and 1) to `rle` results in no runs found (not np.NaN, but 0 runs).

Suggestion : Cast the input to boolean in `rle`. 
Suggestion #2 : Make the `ufunc_dim` argument controllable xclim-wide, through a `set_options` argument, facilitating testing and debug.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #717 and fixes #657

    - Adds an option for controlling the version of the run length algorithms (ufunc vs broadcasting).
    - Uses that option to run all run length tests twice, once for each version.
    - Makes the algo version switch smarter by regrouping everything in a single function. I used the same logic than the Indicator class where "from_context" means the option is read from the options (that's the default).
    - Removes some repetive tests (I might have removed too many?)
    - Removes all hardcoded mentions of `"time"` in run length functions.
    - ~Tiny little, completely unrelated change in `calendar.py` to silence a DeprecationWarning.~
    - Slightly related : added `set_options` to the API page of the documentation

No.

* **Other information**:
Right now, the default is to switch to the "broadcasting" version for: 1) large arrays and 2) dask arrays. Actually, the "ufunc" version does work with dask arrays, just not when the main dimension ('time') is chunked. In reality, the other version also shouldn't really work in that case, but we implemented a workaround that rechunks the data on-the-fly.

I suggest that we leave it as-is for now, but consider spending some time in real benchmarking. If it is more efficient (and I think it is), then the ufunc version could be used for single-chunk dask arrays too. Also, we could get a better idea of the optimal value for `npts_opt`. Finally, it could give us ideas for #505.


xclim/core/calendar.py
```diff
@@ -340,7 +340,7 @@ def _convert_datetime(
     datetime: Union[pydt.datetime, cftime.datetime],
     new_doy: Optional[Union[float, int]] = None,
     calendar: str = "default",
-) -> Union[cftime.datetime, pydt.datetime, np.int]:
+) -> Union[cftime.datetime, pydt.datetime, int]:
```
Zeitsperre, line 343: I addressed this in #716 (and np.nan is technically a float!)

aulemahal: Oh woups. I'll revert this. Thanks!

aulemahal:
For context, dear code readers, please notice the use of a super-dooper-fancy fixture at the head of `test_run_length.py`.  Because it has `scope='module'` and `autouse=True`, it gets executed for ALL tests. So they ALL run twice.

HISTORY.rst
```diff
@@ -9,10 +9,11 @@ New features and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 * Rewrite of nearly all adjustment methods in ``sdba``, with use of ``xr.map_blocks`` to improve scalability with dask. Rewrite of some parts of the algorithms with numba-accelerated code.
 * "GFWED" specifics for fire weather computation implemented back into the FWI module. Outputs are within 3% of GFWED data.
+* Addition of the `run_length_ufunc` option to control which run length algorithm gets run. Defaults stay the same (automatic switch dependent of the input array).
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
-* The `tropical_nights` indice is being deprecated in favour of `tn_days_above` with `thresh="20 degC"`. The indicator remains valid, now wrapping this new indice.
+* The `tropical_nights` indice is being deprecated in favour of `tn_days_above` with ``thresh="20 degC"``. The indicator remains valid, now wrapping this new indice.
```
huard, line 16: Did `set_options` remove the update method from the public interface ?

HISTORY.rst
```diff
@@ -9,10 +9,11 @@ New features and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 * Rewrite of nearly all adjustment methods in ``sdba``, with use of ``xr.map_blocks`` to improve scalability with dask. Rewrite of some parts of the algorithms with numba-accelerated code.
 * "GFWED" specifics for fire weather computation implemented back into the FWI module. Outputs are within 3% of GFWED data.
+* Addition of the `run_length_ufunc` option to control which run length algorithm gets run. Defaults stay the same (automatic switch dependent of the input array).
```
huard, line 12: Explain the criteria for the automated switch.

huard (approved):

HISTORY.rst
aulemahal: Yes, but it was never intended to be public I think. Users are not supposed to use the `set_options` instance.

PR #719, opened 2021-05-21
Standardize to CMIP6 variable names

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Standardizes xclim to use CMIP6 variable names.
  
  - The `variables.yml` file was updated with all variable xclim should understand. Some are not CMIP6 and have been marked as such. There is a `xclim.testing.update_variable_yaml` func, but it still needs to be update by hand because of fields like "description" and "cell_methods". The latter exists in the CMIP6 table, but it is now unique and covers more cases than we need.
  - Almost all official indicator's cfchecks were rewritten to use `cfchecks.generate_cfchecks(varnames)`, less lines to write, the metadata is all concentrated in `variables.yml`.
  - `check_units` was modified. I believe there was an error? I think that in the last part, `start` and `end` where inverted. As a note `variables.yml` defines _canonical units_, instead of _dimensionalities_. So the check is if a transformation _exists_.

**Variables that were not CMIP6-compliant were renamed EVERYWHERE**. This will break many things, I believe. The complete list is:

    - `swe` is now `snw` (`snw` is the snow amount [kg / m²] and `swe` the liquid water equivalent thickness [m])
    - `rh` is now `hurs`
    - `dtas` is now `tdps`  (dew point temperature)
    - `ws` (in FWI) is now `sfcWind`
    - `sic` is now `siconc`
    - `area` (of sea ice indicators) is now `areacello`


xclim/core/units.py
```diff
@@ -525,38 +525,50 @@ def check_units(val: Optional[Union[str, int, float]], dim: Optional[str]) -> No
...
+    #     tu = "m s-1"
+    # else:
+    #     raise NotImplementedError(f"Dimension `{dim}` is not supported.")
+
+    # try:
+    #     units.Quantity(1, units=units2pint(val)).to(tu, "hydro")
+    # except (pint.UndefinedUnitError, pint.DimensionalityError):
+    #     raise ValidationError(
+    #         f"Value's dimension `{val_dim}` does not match expected units `{expected}`."
+    #     )
```
huard, line 572: If not needed, just remove.

HISTORY.rst
```diff
@@ -14,6 +14,14 @@ Breaking changes
...
 * The `tropical_nights` indice is being deprecated in favour of `tn_days_above` with `thresh="20 degC"`. The indicator remains valid, now wrapping this new indice.
 * Results of ``sdba.Grouper.apply`` for ``Grouper``s without a group (ex: ``Grouper('time')``) will contain a ``group`` singleton dimension.
+* CMIP6 variable names have been adopted whenever possible in xclim. Changes are:
+
+    - `swe` is now `snw` (`snw` is the snow amount [kg / m²] and `swe` the liquid water equivalent thickness [m])
+    - `rh` is now `hurs`
+    - `dtas` is now `tdps`
+    - `ws` (in FWI) is now `sfcWind`
+    - `sic` is now `siconc`
+    - `area` (of sea ice indicators) is now `areacello`
```
huard, line 24: Some indicators have changed their name. Should probably be highlighted in the breaking changes.

xclim/testing/tests/test_indices.py
```diff
@@ -1546,52 +1546,6 @@ def test_indice_against_icclim(self, cmip3_day_tas):
         np.testing.assert_array_equal(icclim, ind)
 
 
-@pytest.mark.skip("Fire season computation is not the same as GFWED")
```
huard, line 1549: Why is this gone ?

huard (approved):
Nice clean-up. Let's merge this quick.

xclim/testing/tests/test_indices.py
aulemahal: Nothing to do with this PR, sorry. I just realized it was there when working on it. These tests are now in `test_fwi.py`, since #704.

aulemahal:
@tlogan2000 You might want to take a look here. No need for full review, but this will break many things, I'd want your opinion if possible.

tlogan2000 (approved):
Looks good. I'm ok with breaking some stuff as I think it's better long term

PR #720, opened 2021-05-21
[sdba] Additionnal diagnostics 

### LINKED ISSUE #698: sdba - "verbose" mode?
As suggested in slack by @dgergel and also discussed this morning with Patrick, it could be interesting to output more than just the training dataset / scenario timeseries. I am thinking of some kind of "verbose" or "debug" mode, where sdba trains and adjusts would provide ancillary "diagnostics" variables to help the user understand what happened.

What would be the implementation? A class/method keyword argument? A new option in `xclim.set_options`?

Non exhaustive list of interesting diagnostics:
- EQM/DQM/QDM
    * `quantiles` : Which quantiles of the traning was used by each element of sim
    * `extrapolated` : Boolean flag True for elements where extrapolation was used
- ExtremeValues
   * `nclusters` : The number of clusters (and extreme values) re-adjusted.
- PrincipalComponents
   * The "best" orientation vector.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #698

Adds a `sdba_diagnostics` option. When activated, some adjustment objects will return extra variable in their training or adjustment. 

No. Default behaviour is unchanged.

* **Other information**:
@dgergel I realized this was a bit more complicated than I thought! The basic functionality works, but the "diagnostic" variable you asked for is a bit complicated to get. AFAIU, you wanted `quantiles` which would be the quantiles from `QDM.ds.quantiles` used for each value of sim, when computing the final adjustment factors. However, I guess this only makes sense when `interp='nearest'`?  And anyway, those are computed by `scipy` and are deeply buried, I don't think there are easy ways to get them when calling `interp1d` or `griddata` (see `xclim.sdba.utils.interp_on_quantiles`...

What do you think?


Zeitsperre:
https://gifs.cackhanded.net/the-lego-movie/first-try.gif

dgergel:
@aulemahal this looks like exactly what we need, thanks so much for adding this feature! I agree that it only makes sense when `interp='nearest'`.

aulemahal:
@dgergel Not sure if I understand correctly : currently the "extra diagnostics" returned in QDM is NOT the quantiles of the training dataset, but rather the ones of `sim` that will be interpolated further in the computation. Is this what you are looking for?

dgergel:
@aulemahal that is exactly what we need - the quantiles of `sim` and _not_ the training dataset. Sorry if I wasn't clear before!

aulemahal:
@huard The basic functionality (the option) is here and 2 cases are implemented.  I suggest we merge this and implement other diags as we go, when need arises.
I have ideas for extra diags in PCA, but will wait a proper PR where I update the whole thing.
See my comments above about the limitations for *QM for other interesting diags.

HISTORY.rst
```diff
@@ -12,6 +12,7 @@ New features and enhancements
 * Addition of the `run_length_ufunc` option to control which run length algorithm gets run. Defaults stay the same (automatic switch dependent of the input array : the 1D version is used with non-dask arrays with less than 9000 points per slice).
 * Indicator modules built from YAML can now use custom indices. A mapping or module of them can be given to ``build_indicator_module_from_yaml`` with the ``indices`` keyword.
 * Virtual submodules now include an `iter_indicators` function to iterate over the pairs of names and indicator objects in that module.
+* Addition of te `sdba_diagnostics` option to adds extra diagnostics variables to the outputs of Adjustment objects. Implementation of `sim_q` in QuantileDeltaMapping and `nclusters` in ExtremeValues.
```
huard, line 15:
```suggestion
* Addition of the `sdba_diagnostics` option to adds extra diagnostics variables to the outputs of Adjustment objects. Implementation of `sim_q` in QuantileDeltaMapping and `nclusters` in ExtremeValues.
```

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -278,6 +278,35 @@
...
     "QDM2.adjust(sim)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Retrieving extra diagnostics\n",
+    "\n",
+    "To fully understand what is happening during the bias-adjustment process, sdba can output more variables than what is needed for the adjustment. This behaviour is controlled by the `sdba_diagnostics` option, set with `xclim.set_options`. When True, some `.train` calls may add a few more variables to the training datasets and the `.adjust` calls will always output a dataset, with `scen` and, maybe, other diagnostics variables. See the documentation of each `Adjustment` objects to see what extra are available.\n",
```
huard, line 288:
```suggestion
    "To fully understand what is happening during the bias-adjustment process, sdba can output _diagnostic_ variables, giving more visibility to what the adjustment is doing behind the scene. This behaviour, a `verbose` option, is controlled by the `sdba_diagnostics` option, set with `xclim.set_options`. When `True`, `train` calls are instructed to include additional variables to the training datasets. In addition, the `adjust` calls will always output a dataset, with `scen` and, depending on the algorithm, other diagnostics variables. See the documentation of each `Adjustment` objects to see what extra variables are available.\n",
```

xclim/core/options.py
```diff
@@ -148,6 +151,11 @@ class set_options:
     - ``run_length_ufunc``:
       Whether to use the 1D ufunc version of run length algorithms or the dask-ready broadcasting version.
       Default is ``'auto'`` which means the latter is used for dask-backed and large arrays.
+    - ``sdba_diagnostics``: Whether to add diagnostic variables to output of train and adjust methods
+        of sdba' bias-adjustment objects, as well as some `processing` function. Additionnal variables
+        are listed in the object's docstring. When activated,  `adjust` will return a Dataset with `scen`
+         and those extra diagnostics. For `processing` function, see the doc, the output type might change, or not.
+      Default: ``False``.
```
huard, line 158:
```suggestion
    - ``sdba_diagnostics``: Whether to add diagnostic variables to outputs of sdba's `train`, `adjust` 
       and `processing` operations.  Details about these additional variables are given in the object's 
       docstring. When activated,  `adjust` will return a Dataset with `scen`  and those extra diagnostics. 
       For `processing` functions, see the doc, the output type might change, or not depending on the 
       algorithm.
      Default: ``False``.
```

huard (approved):
Rename the option to `sdba.verbose` ?

I'm concerned `diagnostics` could create confusion with other kinds of performance `diagnostics`.

aulemahal:
I find that "verbose" sounds to much like it related to logging and printing. "sdba_extras" ?

huard:
Similar ideas: 
## `full_output`
-  https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.odeint.html
## `extra_output`
- https://scikit-image.org/docs/dev/api/skimage.restoration.html#calibrate-denoiser
- https://pythonhosted.org/flownetpy/reference/flownetpy.html#module-flownetpy.kuramotonetwork

aulemahal:
I wouldn't say "full" because that's assuming too much about the completeness of the implementation ;)
And would keep the "sdba" word in the option name for explicitness
Thus chose : "sdba_extra_output".

PR #722, opened 2021-05-25
Add the function aggregate_between_dates

marielabonte:
### Pull Request Checklist:
    - This PR fixes #xyz

Add aggregate_between_dates function

* **Other information**:


xclim/indices/generic.py
```diff
@@ -594,3 +594,56 @@ def extreme_temperature_range(
     u = str2pint(low_data.units)
     out.attrs["units"] = pint2cfunits(u - u)
     return out
+
+
+def aggregate_between_dates(data, start, end, op="sum", freq=None):
```
aulemahal, line 599: I think we should add annotations here (`data: xr.DataArray` and the such), I know we don't do it everywhere, but the goal would be to have them on all functions with a potential to be used externally.

xclim/indices/generic.py
```diff
@@ -594,3 +594,56 @@ def extreme_temperature_range(
...
+    """Aggregate the data over a period between start and end dates and apply the operator on the aggregated data.
+
+    Parameters
+    ----------
+    data : xr.DataArray
+      Data to aggregate between start and end dates.
+    start
+      Start dates for the aggregation periods.
+    end
+      End dates for the aggregation periods.
```
aulemahal, line 609:
```suggestion
      Start dates (as day-of-year) for the aggregation periods.
    end
      End dates (as day-of-year) for the aggregation periods.
```

xclim/indices/generic.py
```diff
@@ -594,3 +594,56 @@ def extreme_temperature_range(
...
+    op : str
```
aulemahal, line 610:
```suggestion
    op : {'min', 'max', 'sum', 'mean', 'std'}
```
A user could try with other "operators", but those are the ones we usually list for that case. Anyway, it's only an annotation.

aulemahal (changes requested):
I think a note in the docstring about what happens with NaN in start and end could help.
Also, here we mask whenever one is NaN. What about the option of defaulting to the boundary?
Ex: When start is NaN, but end is not, start = 0 (days since).

I think we should also add a check and "workaround" for when data has a longer time acis then start or end. I.e. if `start.sel(time=base_time)` raises an error, put NaN in res instead.

marielabonte:
> I think a note in the docstring about what happens with NaN in start and end could help.
> Also, here we mask whenever one is NaN. What about the option of defaulting to the boundary?
> Ex: When start is NaN, but end is not, start = 0 (days since).
> 
> I think we should also add a check and "workaround" for when data has a longer time acis then start or end. I.e. if `start.sel(time=base_time)` raises an error, put NaN in res instead.

Do you think if we set start by default to the boundary when it is NaN, it could create a season length that is wrong and then be misinterpreted?

xclim/indices/generic.py
tlogan2000: I agree.  This could end up being a pretty useful function for expert users who want to customize the resampling period(s)

tlogan2000 (changes requested):
I agree with most comments from @aulemahal .   I think we should discuss the default `start` `end` behavior this pm wrt to `nan`  values etc

FYI I did some testing on actual data to see if I could break something but might be useful for the `nan` questions : 

````
tasmax = open_dataset("ERA5/daily_surface_cancities_1990-1993.nc").tasmax

end = atmos.first_day_above(tasmin=tasmax, thresh= '3 degC',after_date='01-01', window=5, freq='AS-JUL')
start = atmos.first_day_below(tasmin=tasmax, thresh='10 degC', freq='AS-JUL')
out = aggregate_between_dates(tasmax, start, end, op='mean', freq='AS-JUL')
````

Everything makes sense even with AS-JUL frequencies and 'incomplete'  first and last years etc

**Comment by aulemahal** [comment]
Thinking about it more, I think I agree. The behaviour I was suggesting could anyways be easily done by the user with a `fillna` call. We can discuss it further this pm.

**Comment by tlogan2000** [comment]
> Do you think if we set start by default to the boundary when it is NaN, it could create a season length that is wrong and then be misinterpreted?

I would maybe side with caution here.  We can discuss in our meeting later

**Review by tlogan2000 — APPROVED** [review]
I think this is looking good. @aulemahal do you want to give a last look?

File: xclim/testing/tests/test_generic.py [hunk 4]
```diff
@@ -142,3 +143,39 @@ def test_doyminmax(q_series):
             assert attr in da.attrs.keys()
         assert da.attrs["units"] == ""
         assert da.attrs["is_dayofyear"] == 1
+
+
+def test_aggregate_between_dates():
```
aulemahal, line 148: As said above, please add a test where time axes of `data` and  `start` or `end` do not entirely overlap.
Also, I think some tests for further edgecases could be nice : 
- `data` and `start` not in the same calendar
- `start` and `end` not at the same freq (test if the error is the good one)
- Test other frequencies (QS-DEC, for instance)

aulemahal (approved):
Sorry if I am too picky, but I think this deserves a few more tests because of all the possible edgecases.

tlogan2000 (commented):
Tests are making sense to me. Good job. I noticed a small thing that the `std` variables that are created actually have `gregorian` calendars?  Not sure this is a big deal as they are equivalent but something I noticed.  @aulemahal  Can you take a minute from your flurry of xclim merges to give a final look?

xclim/testing/tests/test_generic.py
```diff
@@ -142,3 +145,177 @@ def test_doyminmax(q_series):
...
         assert da.attrs["units"] == ""
         assert da.attrs["is_dayofyear"] == 1
+
+
+class TestAggregateBetweenDates:
+    def test_calendars(self):
+        # generate test DataArray
+        time_std = date_range("1991-07-01", "1993-06-30", freq="D", calendar="standard")
+        time_365 = date_range("1991-07-01", "1993-06-30", freq="D", calendar="noleap")
+        data_std = xr.DataArray(
```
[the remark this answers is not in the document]
aulemahal: "gregorian" and "standard" are synonyms! They are the exact same thing, so everything's cool!

tlogan2000:
@marielabonte I don't seem to see the `assert raises` test discussed on the slack?  Did it get pushed to github?

aulemahal:
All good for me! LGTM.

PR #724, opened 2021-05-25
Wind Chill Index

### LINKED ISSUE #666: Add wind chill indicator
### Description

Add wind chill indicator

- https://www.canada.ca/en/environment-climate-change/services/weather-health/wind-chill-cold-weather/wind-chill-index.html
- https://weather.gc.ca/windchill/wind_chill_e.html
- https://en.wikipedia.org/wiki/Wind_chill

aulemahal:
### Pull Request Checklist:
    - This PR fixes #666 (:smiling_imp: )

Adds the Wind Chill Index (indices+indicator).

No.

* **Other information**:
The canadian (ECCC) computation is slightly different from the american (NOAA-NWS) one. The main difference is for weak winds (sfcWind < 5 km/h) : the ECCC defines a second equation while the NWS simply says that (sfcWind < 3 mph) is invalid. Moreover, the maximum valid temperature is 0 °C and jumps to 50 °F when you pass the border.

Instead of having named "methods", I splitted the whole behaviour into three keyword arguments. Might be to complex?
Also, the `min_wind_thresh` might be superfluous as 3 mph is near 5 km/h (but not equal). 

I found no academic references for this, which is quite strange. Googling only led me to FAQs and glossaries, documents orientend for the general public and un-permalinkable web pages...


xclim/indices/_conversion.py
```diff
@@ -538,3 +539,98 @@ def rain_approximation(
...
+      If True (default), points where tas > max_temp_thresh are masked in the output.
+      If `slow_wind_calc` is also False, points where sfcWind < min_wind_thresh are masked.
+    max_temp_thresh : str
+      The maximal allowed temperature when computing the index. If mask_invalid is False,
+      points where tas is above this threshold are masked.
+    min_wind_thresh : str
+      The minimal wind speed for using the standard computation. If `slow_wind_calc` is True,
+      a slow-wind version of the index is used for points where sfcWind is below this threshold.
+      If it is False and `mask_invalid` is True, those points are simply masked.
+
```
huard, line 583: I feel this is a tad too complicated. It's not yet clear to me how to simplify this, but I think we should come up with something less flexible but simpler, ie method={"US", "CAN"}

xclim/testing/tests/test_atmos.py
```diff
@@ -155,3 +155,22 @@ def test_high_precip_low_temp(pr_series, tasmin_series):
         pr, tas, pr_thresh="1 kg m-2 s-1", tas_thresh="1 C"
     )
     np.testing.assert_array_equal(out, [1])
+
+
+def test_wind_chill_index(atmosds):
+    out = atmos.wind_chill_index(ds=atmosds)
+
+    np.testing.assert_allclose(
```
huard, line 163: Expected values from the online calculator ? Just add a comment about where you got them.

huard (changes requested):
Looks good, but let's wait for Emilia's feedback before merging.

xclim/indices/_conversion.py
aulemahal: Ok. I was trying not to have this kind of switch, but it's true that 3 kwargs is a bit complicated.

I'd keep `mask_invalid`, but can replace the other 3 by `method`.

xclim/testing/tests/test_atmos.py
aulemahal: The online calculator rounds the output, so it's hard to use it directly here. I check by hand that xclim was returning values corresponding to the calculator, but the numbers themselves are self-computed, so it's more of a continuity test (to check if a change breaks it)...

aulemahal: Ok found a way to alter the online calculator and remove the rounding. `test_indices` uses those expected values. The atmos test is left as-is.

aulemahal:
@huard  Changes were made following the advice of Emilia.

xclim/indices/_conversion.py
```diff
@@ -544,21 +544,17 @@ def rain_approximation(
...
     mask_invalid: bool = True,
-    max_temp_thresh: str = "0 degC",
-    min_wind_thresh: str = "5 km/h",
 ):
     """Wind chill index.
 
     The Wind Chill Index is an estimation of how cold the weather feels to the average person.
-    It is computed from the air temperature and the 10-m wind. As defined by the Environment and Climate Change Canada ([ECCC]_),
+    It is computed from the air temperature and the 10-m wind. As defined by the Environment and Climate Change Canada ([MVSZ15]_),
     two equations exist, the conventionnal one and one for slow winds (usually < 5 km/h), see Notes.
```
huard, line 558:
```suggestion
    two equations exist, the conventional one and one for slow winds (usually < 5 km/h), see Notes.
```

huard (approved):

PR #733, opened 2021-06-03
sdba Moving window tools

### LINKED ISSUE #712: sdba - Moving window for adjustments
It could be interesting to have a "moving window" feature for sdba's adjustments.

Here I mean that a long timeseries would be bias-adjusted on sections of X-years at a time. The goal and use of this is that for any given point of sim, the correction was done using the same number of points in ref/hist and sim. For example, in QuantileDeltaMapping, if the length of ref/hist and sim is different, the statistical properties of the result are not the same (Cannon goes over this real quick in the paper). Similarly, with ExtremeValues.

This is not a _real_ moving window, because the steps are full years.

Implementation idea:  A function that takes the input timeseries and cuts it into those X-year (overlapping) sections. For each section, the time axis is normalised (ex: `section[i].time = section[0].time`). We stitch everyting back along a "year" dimension and adjust.

A second function takes the result and rebuilds the full timeseries. I'm less certain here. Which value do we take?

This was discussed with @dgergel and recently with @RondeauG (because it appears in @Balinus's extreme values correction). The latter julia code could be used as an inspiration also.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #712

Adds tools for performing adjustment with a moving window. Please see the last section of the "SDBA Advanced" notebook for all the info (I'm too lazy to rewrite everything).

Also, various minor fixes, two that I found while coding and testing this and another found by Christian while testing raven:

- in ``QDM.adjust``, fix bug occuring with coords of 'object' dtype and ``interp='nearest'``. This is a continuation of #567. I just moved the fix from `qdm_adjust` to `utils.broadcast` so it is _not_ applied in the `interp='nearest'` case.
- in ``nbutils.quantiles``, fix dtype bug when using ``float32`` data. This is new from the `map_blocks` feature : to save lines, I coerced the `quantiles` array to the same dtype as the data in the `nbutils` function, but that resulted in a "quantiles" coordinate that was slightly different from what the `map_blocks` wrapper expected (because of the float64 -> float32 -> float64 roundtrip).
- raise a proper error when ``ref``and ``hist`` have a different calendar for map_blocks-backed adjustments. As issued by Christian in slack, the error raised when ref and hist do not have the same calendar is rather cryptic. It was known that this is not supported in 0.27, but now the error will be more explicit. The check is not perform for adjustment methods where this is supported (those not using `map_blocks`).

No.

## Other information
I did not go with the loop way, but rather I stack the windows unto a new dimension. This result in all windows having the same "time" axis when passed to `adjust`. The month and day (and day-of-year) information is kept, but the year itself is incorrect. Would that be a problem anywhere?

Also, I only did 1 test,  with 30 training years,

QDM on 150 simulated years => 4s
QDM on 65 windows of 21 years (steps of 2 years between each window) (= 1365 years , x9) => 31s (x8)
QDM looping by hand on the same number of windows => 5m48s (x87)


aulemahal:
@RondeauG  @dgergel Please throw an eye on the notebook when you have time! I was thinking adding some way to keep a bit more years by allowing incomplete windows, but I'm not sure on the best way.

Also, this will adjust the full window and _then_ throw the overlapping years, which is highly inefficient. However, the `unpack` method is agnostic to the real original window length, so we could independently implement some way to only adjust part of the data (as an additional arg to `.adjust`) and it should still work the same.

Finally, I am open to name propositions, I feel the current ones are a bit long...

aulemahal:
@huard  The MBCn PR is based on this one, so I made the notebook edits you suggested here.

There are no unit tests (for now at least), but the documentation is an integration test in itself.

huard (approved):

PR #735, opened 2021-06-07
Add Biologically Effective Growing Degree Days

Zeitsperre:
### Pull Request Checklist:
    - This PR partially addresses #355 

This adds (or begins the process of adding) Biologically Effective Growing Degree Days to xclim's agroclimatic indices.

No, it just adds an indice.

* **Other information**:

The title and source of the equation is not available outside of print (or a ebook ($)), but here's an R-based implementation of the formula for reference: https://rdrr.io/cran/agroclim/src/R/bedd.R


xclim/indices/_agro.py
```diff
@@ -113,3 +96,80 @@ def corn_heat_units(
...
+
+    def tasmax_limited(tx, thresh_tx):
+        return xarray.where((tx > thresh_tx).any(), thresh_tx, tx)
+
+    date_mask = tasmin.sel(time=sel_months(tasmin["time.month"]))
+
+    bedd = (
+        (
+            (
+                xarray.where(mask_tasmin, (tasmin - thresh_tasmin), 0)
```
tlogan2000, line 161: Not sure I follow the exact sequence of criteria but could be more efficient to use `xr.where`  a single time at the end?  Also I think use of `xr.where`  will end up filling even the "all nan" slices with zeros (i.e. ocean cells, etc) 

or could simply doing the masking at the beginning of the function work?  This would assign `nan` everywhere mask_tasmin is False
````
mask_tasmin = tasmin > thresh_tasmin
tasmin = tasmin.where(mask_tasmin)
tasmax = tasmax.where(mask_tasmin)
````

File: xclim/indices/_agro.py [hunk 2]
```diff
@@ -113,3 +96,80 @@ def corn_heat_units(
...
+    tasmax = convert_units_to(tasmax, "degC")
+    thresh_tasmin = convert_units_to(thresh_tasmin, "degC")
+    thresh_tasmax = convert_units_to(thresh_tasmax, "degC")
+
+    mask_tasmin = tasmin > thresh_tasmin
+
+    lat_mask = (abs(tasmin[lat_dim]) >= 40) & (abs(tasmin[lat_dim]) <= 50)
+    lat_constant = xarray.where(lat_mask, (abs(tasmin[lat_dim]) / 50) * 0.06, 0)
+
+    def sel_months(time):
```
tlogan2000, line 145: Not sure about this part... with xclim we typically want to give the users a choice via a `resample` frequency.  We could maybe see if we can somehow use the new  `aggregate_between_dates` that Marie Pier implemented? .

xclim/indices/_agro.py
```diff
@@ -113,3 +97,65 @@ def corn_heat_units(
...
+    thresh_tasmax="[temperature]",
+)
+def biologically_effective_degree_days(
+    tasmin: xarray.DataArray,
+    tasmax: xarray.DataArray,
+    thresh_tasmin: str = "10 degC",
+    thresh_tasmax: str = "19 degC",
+    lat_dim: str = "lat",
+    start_date: str = "4",
+    end_date: str = "10",
```
aulemahal, line 115:
```suggestion
    lat : xr.DataArray,
    thresh_tasmin: str = "10 degC",
    thresh_tasmax: str = "19 degC",
    start_date: DayOfYearStr = "04-01",
    end_date: DayOfYearStr = "10-01",
    freq: str = "YS"
```
So that `lat` needs not be a dimension. (think unstructured or rotated pole grids).

`DayOfYearStr` is recognized by the `Indicator` constructor.

I would have a `freq` argument still. In part because the Indicator class behaves differently wether it is present or not and in part to leave all these decisions to the user.

There is the `allowed_periods` attribute that can be given to the Indicator constructor if you want to restrict to yearly. (`allowed_periods=['A']`).

aulemahal (changes requested):
You're right about extending `aggregate_between_dates`!

xclim/indices/_agro.py
Zeitsperre: Thanks for the comments! 

For the end date, I believe it goes until "10-31", so this could complicate 360-day calendars. Not sure how best to proceed.

Typically, this type of indice is only valid at the yearly scale. I don't see how this info would be useful at the `'M'` frequency, but I don't see the harm in opening it up.

aulemahal: Thing is, the missing value checks need to know `freq` to run, which is one of the reason to leave them in the signature. Also, there still a need for the "YS" <-> "AS-JUL" distinction!

Would "11-01" work? If needed, a "endpoint=True/False" could be added to `aggregate_between_dates`?

xclim/indices/_agro.py
```diff
@@ -113,3 +98,66 @@ def corn_heat_units(
...
+    tasmin = convert_units_to(tasmin, "degC")
+    tasmax = convert_units_to(tasmax, "degC")
+    thresh_tasmin = convert_units_to(thresh_tasmin, "degC")
+    thresh_tasmax = convert_units_to(thresh_tasmax, "degC")
+
+    mask_tasmin = tasmin > thresh_tasmin
+    tasmin = tasmin.where(mask_tasmin)
+    tasmax = tasmax.where(mask_tasmin)
+
+    lat_mask = (abs(tasmin.lat) >= 40) & (abs(tasmin.lat) <= 50)
```
aulemahal, line 150: Sorry, last comment wasn't complete, but I really think having `lat` has an input is best because is removes all assumptions on how the data is structured from the xclim. I.e. : it is considered a variable like the others, only needs to be broadcastable to the shape of the other inputs.

xclim/indices/_agro.py
Zeitsperre: > Thing is, the missing value checks need to know `freq` to run, which is one of the reason to leave them in the signature. Also, there still a need for the "YS" <-> "AS-JUL" distinction!

Sounds good. If we can handle North and South hemisphere, all the better.

> Would "11-01" work? If needed, a "endpoint=True/False" could be added to `aggregate_between_dates`?

I think that would work, yes. I can think of two other indices (not presently coded, but easily added) that could benefit from this functionality as well.

xclim/indices/_agro.py
Zeitsperre: Would it make the most sense to send it in as a field, or is leveraging something like `cf-xarray` to get the lat field the way to go?

aulemahal: I say send it in as a field. Leveraging `cf-xarray` would be cool, but seems to me its narrowing down the problem to non-standard datasets. Except `time`, we have no other real constrains on the datasets' metadata and structure in xclim, which is why I prefer doing the same as in `FWI` and having it as an additional variable.

Zeitsperre:
This PR is good for another review. Thanks to @aulemahal, we can now pass `DayOfYear` strings to `aggregate_between_dates`. Thanks for the help !

One thing I'm wondering is if we encounter a year of data that ends too early to be used (i.e.: `start_i` is valid, but `end_i` does not exist), should we append a `NaN` to the resulting array? My feelings say yes, but want to confirm first.

marielabonte:
> This PR is good for another review. Thanks to @aulemahal, we can now pass `DayOfYear` strings to `aggregate_between_dates`. Thanks for the help !
> 
> One thing I'm wondering is if we encounter a year of data that ends too early to be used (i.e.: `start_i` is valid, but `end_i` does not exist), should we append a `NaN` to the resulting array? My feelings say yes, but want to confirm first.

Yes, if end_i does not exist aggregate_between_dates should append NaN for that year.

Zeitsperre:
I'm just wondering here, but if let's say we have a very short year (ending "01-31", for example) and we ask for a start date that occurs after this date, should we also return NaN? This acknowledges that the data extends to this year, but that the year is still invalid? I'll move forward with that approach for both.

tlogan2000:
> I'm just wondering here, but if let's say we have a very short year (ending "01-31", for example) and we ask for a start date that occurs after this date, should we also return NaN? This acknowledges that the data extends to this year, but that the year is still invalid? I'll move forward with that approach for both.

I would add a `nan` value in this case as well

Zeitsperre:
> this might come for free via `aggregate_between_dates`?

We've rewritten a good chunk of that function, so it needs to be added. Strings and datetime array approaches aren't the same presently. This function will need to be made more resilient in the future, but it seems to be working as intended for now.

aulemahal:
@Zeitsperre  I gave it a look and compared with the R code, for the lack of access to the book. It seems to me that you were doing something different than in R? I'm not exactly sure, but I rewrote the `bedd` line so it matches the R function. Also, edited the equation so it matches that too (mostly inverting min and max).

Two things:
1. Is there a reason you are doing this masking of Tx and Tn by `Tn >= thresh_tn`? The R code doesn't do it.
2. I believe the 10 and 13 from the DTR adjustment could be considered parameters too, no? If we open up thresh_tasmin and thresh_tasmax, I would open these up too.

Zeitsperre:
@aulemahal Thanks for the rewrite!

> Two things:
> 
>     1. Is there a reason you are doing this masking of Tx and Tn by `Tn >= thresh_tn`? The R code doesn't do it.

 I think the masking approach was something I did in my QGIS models to remove zones that were irrelevant. It doesn't translate that well here. Better to leave it out. 

>     2. I believe the 10 and 13 from the DTR adjustment could be considered parameters too, no? If we open up thresh_tasmin and thresh_tasmax, I would open these up too.

Honestly, I'm not sure if it makes sense to open them up.  I can't think of any reason to (unless this indice is used in a modified way for other domains). What if we open it (`high_dtr` and `low_dtr`?) and set the `Indicator` to not modify this?

aulemahal:
> Honestly, I'm not sure if it makes sense to open them up. I can't think of any reason to (unless this indice is used in a modified way for other domains). What if we open it (high_dtr and low_dtr?) and set the Indicator to not modify this?

I don't know the indice, so I could be wrong, but it seems to me that there are 4 thresholds here and I see no reason to have one fixed and the other not.

While looking up something for the bootstrap thing, I found that ECAD defines BEDD! https://www.ecad.eu/documents/atbd.pdf
They have `thresh_tasmin` customizable, but they fix `thresh_tasmax - thresh_tasmin`. Also, both the lat adjustment and the dtr adjustment are absent... Argh. Seeing this, I guess that we would be better of if we had an option (at least in the indice) to imitate ICCLIM / ECAD.  Ouf, so many ifs...

Then to simplify the indicators, I would go with what you suggest: inject the options in xclim's official version to have most "up to date" calculation and wrap it again with the ECAD option in `icclim.,yaml`.

tlogan2000:
> Then to simplify the indicators, I would go with what you suggest: inject the options in xclim's official version to have most "up to date" calculation and wrap it again with the ECAD option in `icclim.,yaml`.

I think this is what I would do as well (if I'm following correctly)...  `atmos.bedd`  would have all params are available with defaults of 'normal' values but remains fully customizable.  `icclim.bedd`  created via wrapped partial with documented param vals 'hard-coded'

Zeitsperre:
> I don't know the indice, so I could be wrong, but it seems to me that there are 4 thresholds here and I see no reason to have one fixed and the other not.

Makes sense. I'm going forward with that.

> While looking up something for the bootstrap thing, I found that ECAD defines BEDD! https://www.ecad.eu/documents/atbd.pdf
> They have `thresh_tasmin` customizable, but they fix `thresh_tasmax - thresh_tasmin`. Also, both the lat adjustment and the dtr adjustment are absent... Argh. Seeing this, I guess that we would be better of if we had an option (at least in the indice) to imitate ICCLIM / ECAD. Ouf, so many ifs...

Looks like we need to make lat be optional then. I've never seen BEDD without `k` and the `dtr` adjustment, but OK. Adding this in. 

> Then to simplify the indicators, I would go with what you suggest: inject the options in xclim's official version to have most "up to date" calculation and wrap it again with the ECAD option in `icclim.,yaml`.

Sounds good. I don't get why this version is so different, but we should support both.

Zeitsperre:
I just realized, the ECAD version also ends in September, while the Gladstones original formula goes to October. Something else to note.

Zeitsperre:
I've added an Indicator version of the new, more flagrant (emphasis on "flag") version for BEDD. For the ICCLIM version, I would be fine with using the `indicators.icclim.BEDD` to call the simplified formula. For the French translation (and even the English description) I wouldn't mind some help.

One thing I'm not sure about is how to deal with the `lat` call signature in the indicator. What am I missing? Thanks!

xclim/indices/_agro.py
```diff
@@ -113,3 +101,132 @@ def corn_heat_units(
...
+
+@declare_units(
+    tasmin="[temperature]",
+    tasmax="[temperature]",
+    thresh_tasmin="[temperature]",
+)
+def biologically_effective_degree_days(
+    tasmin: xarray.DataArray,
+    tasmax: xarray.DataArray,
+    lat: Optional[Union[xarray.DataArray, np.array]] = None,
```
aulemahal, line 114:
```suggestion
    lat: Optional[xarray.DataArray] = None,
```
Would there be a reason to accept a ndarray? Could the user not always wrap it in xarray themself? The Indicator parser wouldn't know what to do with this annotatio, whereas with `Optional[DataArray]` it detects an "OPTIONAL_VARIABLE".

xclim/data/fr.json
```diff
@@ -463,6 +463,13 @@
     "comment": "",
     "abstract": "Cumul des degrés-jours pour les jours où la température moyenne est au-dessus d'un seuil."
   },
+  "BIOLOGICALLY_EFFECTIVE_DEGREE_DAYS":{
+    "long_name": "",
+    "description": "",
+    "title": "Degré-jour de croissance effective pour la productivité biologique",
+    "comment": "Formule originale pris de Gladstones 1992.",
+    "abstract": ""
```
aulemahal, line 471:
```suggestion
    "long_name": "Degrés-jours biologiquement actifs",
    "description": "Accumulation des degrés-jours borné par Tmin > {thresh_tasmin} et Tmax-Tmin < {max_daily_degree_days}. Les degrés jours quotidiens sont corrigés selon Tmax-Tmin et selon un coefficient de longueur du jour dépendent de la laittude.",
    "title": "Degrés-jours biologiquement actifs",
    "comment": "Formule originale prise de Gladstones 1992.",
    "abstract": "Indice basé sur une relation non linéaire entre la température et les processus physiologiques. Il intègre une borne supérieure du cumul, un coefficient lié à la longueur du jour et une correction liée à l'écart journalier des températures.",
```

After moults recherches, I found this : https://tel.archives-ouvertes.fr/tel-01968041/document , p.17
Haha second page of google search, first french publication discussing BEDD :)

aulemahal (approved):
I have an idea to reduce complexity in `aggregate_between_dates`, I'll push it this afternoon.

xclim/indices/_agro.py
Zeitsperre: I think I figured that supporting any array-type object would be fine. In the test for it, passing a `numpy.array([45])` was fine. I suppose expecting a DataArray makes the most sense though.

xclim/indicators/atmos/_temperature.py
```diff
@@ -786,3 +787,28 @@ class TasminTasmax(Daily2D):
...
+            thresh_tasmin="10 degC",
+            method="gladstones",
+            low_dtr="10 degC",
+            high_dtr="13 degC",
+            max_daily_degree_days="9 degC",
+            start_date="04-01",
+            end_date="11-01",
+        ),
+    ),
+)
```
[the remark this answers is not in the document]
Zeitsperre: Thanks for the explanation. I wasn't sure if this should be its own indicator class, given that lat is a DataArray/coordinate.

Zeitsperre:
> EDIT: And there's also the icclim addition missing!

I'll add this into the ICCLIM set of Indicators!

xclim/indicators/atmos/_temperature.py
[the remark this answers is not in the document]
aulemahal: Currently your question makes sense, but with #736 it will all be caduc because that PR will lower the need for intermediate indicator classes a lot. You are right that you could have made a `TasminTasmaxLat` class and added datachecks and cfchecks for `lat` (as it is considered a variable input), but that will be done automatically with the other PR. And, we don't have any checks for lat currently anyway.

xclim/data/icclim.yml
```diff
@@ -62,6 +62,7 @@ indices:
       long_name: Biologically effective growing degree days (Summation of min((max((Tmin + Tmax)/2 - {thresh_tasmin}, 0)), 9°C), for days between 1 April and 30 September)
       comment: Revised formula published by ECAD/KNMI for ICCLIM, 2013.
     index_function:
+      name: biologically_effective_degree_days
```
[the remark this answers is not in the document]
aulemahal: I agree... Yours would have work without my suggestion of injecting the `method` parameters in the xclim indicator! Because of that we had to go back to the indice and inject everything... Also first time I inject "null" (None), happy that it works!

Zeitsperre:
Are we good to go?

aulemahal:
I think we are! LGTM.

Zeitsperre:
Just going to update the indicators in the history and will merge! Thanks enormously for your help in this. Much more complicated than I first envisioned. My memory is certainly rusty when it comes to this indice, but thankfully the others I have in mind (LTI, Huglin, etc.) are all simpler than this.

PR #736, opened 2021-06-08
Refactor how indicators know what are variables and what are parameters

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Internal changes in how `Indicators` decide which inputs are "variables" and which are "parameters". 

Changes:

- "Variables" are those with a "VARIABLE" or "OPTIONAL_VARIABLE" _kind_ so those with a `xr.DataArray` annotation (or optional). "Parameters" are the others. Before we had `Indicator.nvar` that would say the number of parameters (in signature order) which were "variables". This is dropped, `nvar` doesn't exist anymore.
- ONLY "Variables" are send to cf, data and missing checks. Optional variables that were not given (= None) are not sent.
- To simplify the internals, the default `cfcheck` will perform a check based on the CF metadata available in xclim/data/variables.yml, looking up the name of the variable (which should all be conform to CMIP6 now). No need to manually write `cfchecks` for indicators declared within xclim (except a few special cases).
- The missing method defaults to `skip` if the indice has no `freq` parameters. This inverts the previous behaviour where we had to explicitly pass `missing='skip'` when no resampling was performed.
- The last two have reduced the need for custom indicator base class in `indicators/*.py`.
 
Hopefully not.

* **Other information**:
This small afternoon fix took the full afternoon. I am sorry for the reviewer.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,15 @@
 History
 =======
 
+0.28.0 (unreleased)
+-------------------
+
+Internal Changes
+~~~~~~~~~~~~~~~~
+* The indicator call sequence now considers "variable" the inputs annoted so. Dropped the ``nvar`` attribute.
+* Default cfcheck is now to check metadata according to the variable name, using CMIP6 names in xclim/data/variable.yml.
+* ``Indicator.missing`` defaults to "skip" is ``freq`` is absent from the list of parameters.
+
 0.27.0 (2021-05-28)
 -------------------
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,7 +9,7 @@ Internal Changes
 ~~~~~~~~~~~~~~~~
 * The indicator call sequence now considers "variable" the inputs annoted so. Dropped the ``nvar`` attribute.
 * Default cfcheck is now to check metadata according to the variable name, using CMIP6 names in xclim/data/variable.yml.
-* ``Indicator.missing`` defaults to "skip" is ``freq`` is absent from the list of parameters.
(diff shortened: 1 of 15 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,15 @@
...
 =======
 
+0.28.0 (unreleased)
+-------------------
+
+Internal Changes
+~~~~~~~~~~~~~~~~
+* The indicator call sequence now considers "variable" the inputs annoted so. Dropped the ``nvar`` attribute.
+* Default cfcheck is now to check metadata according to the variable name, using CMIP6 names in xclim/data/variable.yml.
+* ``Indicator.missing`` defaults to "skip" is ``freq`` is absent from the list of parameters.
```
huard, line 12:
```suggestion
* ``Indicator.missing`` defaults to "skip" if ``freq`` is absent from the list of parameters.
```

xclim/indicators/land/_snow.py
```diff
@@ -13,35 +12,22 @@
 ]
 
 
-class SnowDepth(Daily):
-    cfcheck = staticmethod(generate_cfcheck("snd"))
+# We need to declare the base class here so the `land` module is detected automatically.
```
huard, line 15: Why is this necessary?

huard (approved):
Beau ménage de printemps.

tlogan2000 (approved):

xclim/indicators/land/_snow.py
aulemahal: To remove the need of passing `module='land'` and `realm='land'`, the Indicator's `__new__` checks where the class was declared (with the default `module`).  Because we don't know where instances are created, only where classes are (explicitly) defined.

PR #738, opened 2021-06-09
Custom translations and load locales at import

### LINKED ISSUE #565: Translated metadata for custom-made indicators
With all the new improvements to the Indicator class that allow users to create their own Indicators, we forgot to think about how to pass translated versions of the metadata.

A way to do this would be to create a translation dictionary:
```
fr2 = {'NEWIND': {'long_name': 'Nouveau long nom'}, 'attrs_mapping': {'modifiers': []}}  # attrs_mapping is needed in translation dict, even if empty.
```
and use it like `xc.set_options(metadata_locales=['fr', ('fr', fr2)])`.  We don't have mechanisms forbidding two locales with the same name so this should work.
Not only it is not very elegant, it's also currently broken. A poorly written line (mea culpa) in `locales.py` raises errors when passing dictionaries in `set_options`.

I'm not sure of the best way for this. May be I'd like input of other users on this? @tlogan2000  How would passing translated metadata make the most sense?

aulemahal:
### Pull Request Checklist:
    - This PR fixes #565 

- Automatic load of "official" locales at import time, instead of each time an indicator is called. This also adds the ability to edit locales during runtime.
- Possibility to pass custom translations when creating virtual modules.
- "Smarter" module making : passing the "stem" of the filename to `build_indicator_module_from_yaml` makes it automatically detect any python module with custom indices and custom translations. (see edited notebook)
- When an indicator is missing translations for certain fields, they are sought for in it's parent indicator class. Makes it easier to  only rename indicators in virtual modules (like in icclim).
- Fixed a small bug when creating indicators from dict (virtual module from yaml) : when adding metadata to an existing parameter, the default value was turning into None, now it stays the same if not specified.
- Small refactor on how the indicators retrieve translated metadata. Doesn't make it really simpler... but at least reduces code duplication.

Shouldn't.


xclim/core/locales.py
```diff
@@ -234,6 +229,38 @@ def __init__(self, locale):
         )
 
 
+def read_locale_file(filename, module=None):
+    with open(filename, "r", encoding="utf-8") as f:
```
huard, line 233: docstring

huard (approved):
Note that I'm not familiar enough with the internal of the translation stuff to do a good review.

PR #740, opened 2021-06-10
Feature/bootstrapping

### LINKED ISSUE #732: Feature request: Add a way to bootstrap the calculation of percentiles
This feature request is based on the information available in "Avoiding Inhomogeneity in Percentile-Based Indices of Temperature Extremes" _Zhang et.al_ ([doi]( https://doi.org/10.1175/JCLI3366.1))

Xclim already provides multiple indices/indicators based on a percentile comparison such as _atmos.tg90p_.
However, if we want to calculate the percentiles on a specific period (base period) and then use these percentiles for comparison in the whole studied period (in base + out of base) we cannot rely on the current implementation.

In short, using directly the percentiles computed for the base period would introduce an artificial discontinuity at the edges of this base period.
The paper above explains in details the issue.

To fix this, Zhang et.al suggest a bootstrapping algorithm which iterates over the base period in order to construct a computation context similar to the rest of the calculation.
For a base period of _x_  years, it iterates over theses years and create an in-base of _x-1_  years and an out of base period of a single year. Then it calculates the exceedance rate of the out of base year using the percentiles of the in-base period.
This is the approach implemented by [icclim](https://github.com/cerfacs-globc/icclim) and [RClimDex](https://github.com/ECCC-CDAS/RClimDex/search?q=bootstrap).

**I'm willing to work on this issue.**
However, I will need help to see how can this be included in the existing code base, especially regarding the API.

bzah:
### Pull Request Checklist:
    - This PR fixes #732 
---------
- [ ] Validate the output with icclim 4.x and/or climdex (R library)
This PR adds:
  - The @percentile_bootstrap decorator
  - the bootstrapping algorithm
  - the decoration of percentile based indices with @percentile_bootstrap

The API of percentiles based indices has been modified. 
They now have a parameter bootstrap_configuration: BootstrapConfig.
It is not a breaking change because bootstrap_config is always optional (default to NO_BOOTSTRAP).

* **Other information**:
This is still a wip.


xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,236 @@
...
+    return xr.concat([in_base_exceedance_rates, out_of_base_exceedance], dim="time")
+
+
+def _bootstrap_period(
+    ds_in_base_period: DataArray,
+    config: BootstrapConfig,
+    exceedance_function: ExceedanceFunction,
+) -> DataArray:
+    period_exceedance_rates = []
+    for year_ds in ds_in_base_period.groupby("time.year"):
```
bzah, line 142: We can probably parallelize this with dask, but I'm not sure how. 
We can't simply chunk the ds by time and let it run, it won't change the performances because each iteration needs the whole ds in base period to perform the computation of percentile.
Maybe by calling dask delay instead of appending result in the list ?

Zeitsperre:
I can't speak to the issue of the PR, but in order to pass the linter, I would suggest installing `pre-commit` (https://pre-commit.com/#usage) to perform the checks and adjustments (black +  flake8 + a bunch of other linters). It will automate the process entirely!

xclim/core/bootstrapping.py
aulemahal: AFAIU, this would be already parallelized, no? I mean, assuming each call to `_bootstrap_year` returns a lazy DataArray and then `out` is the concatenation of them. When computation is triggered, dask will construct it's tree and _then_ parallelize it as much as it can.

On another note, I think I would use:
```suggestion
    for year in np.unique(ds_in_base_period.time.dt.year):
```
I have a mistrust of `groupby` when it comes to dask performance. Here you are only using the year values, so no need to invoke the whole groupby machinery!

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,236 @@
...
+    out_base = ds_in_base_period.sel(time=str(out_base_year))
+    exceedance_rates = []
+    for year_ds in in_base.groupby("time.year"):
+        print(year_ds[0])
+        replicated_year = in_base.sel(time=str(year_ds[0]))
+        # it is necessary to change the time of the replicated year
+        # in order to not skip it in percentile calculation
+        replicated_year["time"] = replicated_year.time + pd.Timedelta(
+            str(out_base_year - year_ds[0]) + "y"
+        )
```
aulemahal, line 169: Same comment for `year_ds[0]` as above.

Also, what if the replicated year is a leap year and this addition brings it to a non-leap year? After testing, I think everything will be offset by a day, which ain't good!

Idea : convert everything to the closest uniform calendar ( usually 'noleap', thus dropping all 29th Feb). In that case we will be able to _replace_ the time axis :

```python
replicated_year['time'] = ds_in_base_period.time.sel(time=str(out_base_year))
```

Not sure how this calendar conversion will impact the computation, Are the percentiles of day 366 really trustable anyway?

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,236 @@
...
+        completed_in_base = xr.concat([in_base, replicated_year], dim="time")
+        thresholds = _calculate_thresholds(completed_in_base, config)
+        exceedance_rate = _calculate_exceedances(
+            config, exceedance_function, out_base, thresholds
+        )
+        exceedance_rates.append(exceedance_rate)
+    if len(exceedance_rates) == 1:
+        return exceedance_rates[0]
+    return xr.concat(exceedance_rates, dim="time").groupby("time").mean()
```
aulemahal, line 178: I'm not sure what is happening here. Why would you `groupby('time')`? I understand that `exceedance_rate` is an array at sampling freq `freq` _only_ for the `out_base_year`, no? So, there are no overlap of time elements and `groupby("time").mean()` shouldn't have any effect, no?

aulemahal (commented):
A few comments on things I saw. I didn't test anything.

As for `dask`, I have the feeling a bottleneck here will be the enormous quantity of tasks in the tree. There may be a way to use `map_blocks`, which can reduce this a lot.

xclim/core/bootstrapping.py
aulemahal: So it has the same time coordinate repeated multiple times because it is computed over the `out_base_period` for each in_base years.  I would then suggest:

```suggestion
    return xr.concat(exceedance_rates, dim="in_base_year").mean('in_base_year')
```
Again because of my chronic avoidance of `groupby` ;).

xclim/core/bootstrapping.py
bzah: I'm very new to Dask but I think it will parallelize between the different chunks if the DataArray is chunked, but will not parallelize this for loop because we treat multiple time the whole DA, no ?
I tried something like :
```
for year_ds in ds_in_base_period.groupby("time.year"):
        period_exceedance_rates.append(
            dask.delayed(_bootstrap_year)(
                ds_in_base_period.copy(deep=True),
                year_ds[0],
                config,
                exceedance_function,
            )
        )
    out = xr.concat(dask.compute(*period_exceedance_rates), dim="time")
```
It seems to work fine on small DA but goes into deadlock with bigger ones. I still have to investigate the exact reason why.
I will also have a look at map_blocks, maybe it can help here.

xclim/core/bootstrapping.py
bzah: I share your concern with the pertinence of 29th feb percentiles, I _guess_ we can skip it entirely, I'm confirming this with Christian. 
So, you suggest that we do something like ``da = calendar.convert_calendar(da, "noleap")`` at the very beginning of bootstrapping ?

bzah: I discussed the matter with Christian P. and for icclim there is no special treatment for 29th Feb. 
The idea is that, when the **out of base year is a non-leap year** and we replicate a leap year, (thus we have an extra day in the dataset), this is compensated by when the **out of base year is a leap year** and we replicate a non leap year (thus a day is lacking).
So ideally, we should copy as is the year but make sure the time dimension is not offsetted.

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,256 @@
...
+        bound_args = signature(func).bind(*args, **kwargs)
+        config = NO_BOOTSRAP
+        indice_window = None
+        for name, val in bound_args.arguments.items():
+            if name == "window":
+                indice_window = val
+            elif name == "freq":
+                freq = val
+            elif isinstance(val, DataArray):
+                da = val
```
huard, line 53: Won't this catch both tas and t90 ?

bzah: Yes it could. In my tests I always explicitly put t90=None to avoid this.
I'm not quite sure how it could be handled better because the _name_ of the main parameter (here "tas") change between functions so I cannot rely on this.
We could check that there is **one and only one** DataArray in arguments when there is a bootstrap_config maybe ?

I feel like this makes the API a bit confusing to have some parameter excluding one another (t90 and bootstrap_config), but I don't see an other way of doing that with decorators.

bzah: Actually it doesn't work anymore to do `t90=None`, since this commit beba80d5379d9d0aa4833341d3f8e93f878700ec I think. It gives this error now:
```
File "/.../indicator.py", line 621, in __call__
    self._update_attrs(
  File "/.../indicator.py", line 813, in _update_attrs
    attrs["cell_methods"] = merge_attributes(
  File "/.../formatting.py", line 264, in merge_attributes
    if attribute in in_ds.attrs or missing_str is not None:
AttributeError: 'NoneType' object has no attribute 'attrs'
```
Now, it works if I change the signature of tx90p to have a default None for t90:
```
@declare_units(tasmax="[temperature]", t90="[temperature]")
@percentile_bootstrap
def tx90p(
    tasmax: xarray.DataArray,
    t90: xarray.DataArray = None,
    freq: str = "YS",
    bootstrap_config: BootstrapConfig = NO_BOOTSRAP,
) -> xarray.DataArray:  # noqa: D401
```

huard: Hum. Ok, we'll try to think of something on our end and come back with a suggestion.

huard:
Hi @bzah 

After discussion with @aulemahal, what we would suggest is the following: 

In `calendar.percentile_doy`, add attributes needed by the bootstrap functionality to the output:
- base_period
- window
This partially replaces the BootstrapConfig object. 

In the indices, themselves, 
-  replace bootstrap_config by a boolean switch (bootstrap=True/False)

In the decorator, we should be able to detect the percentile input by the fact that it has a `percentiles` coordinate. From there we can extract the configuration attributes (base_period, window) and run the bootstrap code if bootstrap is True. 

Does that make sense to you ?

bzah:
> In the decorator, we should be able to detect the percentile input by the fact that it has a percentiles coordinate. From there we can extract the configuration attributes (base_period, window) and run the bootstrap code if bootstrap is True.
> Does that make sense to you ?

Hello @huard,
Yes I think I see what you mean, I will try to implement it on Monday.

huard:
Excellent, i'll set aside some time Monday PM to work on this. I'd like to get the basic functionality merged next week and then add some docs about it before the next release.

bzah:
> Hi @bzah
> 
> After discussion with @aulemahal, what we would suggest is the following:
> 
> In `calendar.percentile_doy`, add attributes needed by the bootstrap functionality to the output:
> 
>     * base_period
> 
>     * window
>       This partially replaces the BootstrapConfig object.
> 
> 
> In the indices, themselves,
> 
>     * replace bootstrap_config by a boolean switch (bootstrap=True/False)
> 
> 
> In the decorator, we should be able to detect the percentile input by the fact that it has a `percentiles` coordinate. From there we can extract the configuration attributes (base_period, window) and run the bootstrap code if bootstrap is True.
> 
> Does that make sense to you ?

I hope I did it right, while implementing I was not so sure I properly understood what you suggested.

In any case, I think it now looks more like the initial API. 
The API main changes are now on `calendar.percentile_doy` which now returns a PercentileConfig object.
So it looks like before: when we want to compute a percentile based indices, percentile_doy must be called and its results must used in parameter of indices.
Example :
```
t90 = percentile_doy(
        ds.tmax,
        window=5,
        per=90,
        in_base_slice=slice("1991-01-01", "2000-12-31"),
        out_of_base_slice=slice("2001-01-01", "2010-12-31"),
    )
t90.in_base_percentiles = t90.in_base_percentiles.sel(percentiles=90)

result = xc.atmos.tx90p(
    tasmax=ds.tmax,
    t90=t90,
    freq="MS",
)
```

But it still needs a configuration object and I can't see how to completely avoid it.

xclim/core/bootstrapping.py
bzah: Resolving this because with the new implementation of @huard it is now outdated

Zeitsperre:
@bzah 

From what I can see, the failing tests are from the breaking API changes. There are a few `slow` tests and a jupyter notebook that need to be updated. The relevant command for the notebook is `pytest --nbval docs/notebooks`, and for the slow tests, try `pytest --cov xclim -m slow`. If the failures aren't related to your PR, please let us know!

bzah:
@Zeitsperre
I tried to understand what the problem is in theses tests and on the jupyter notebook but I really don't get it.

For the tests, I don't know if that's normal but all the tests in test_icclim are skipped but the ones that are failing.
And when I remove the `@percentile_bootstrap` decorator from the indices in error, the tests are skipped again.
I really don't get how the annotation can trigger the tests.

As for the notebook the error is similar to the test's error. 
it's a `KeyError: 'freq'` where `freq` is supposed to be used to format the indicator's description.
However I can't even reach this cell when trying to run the notebook locally, it fails above this one.

I'm a bit lost here, do you have a clue on where to look ?

aulemahal:
@bzah The `KeyError` failures were due to some features missing from `functools.wraps` : it doesn't update the signature and doesn't hide the wrapped signature. Thus, the "official" indicators in `atmos` were working because they were parsing the signature of the underlying function, not the signature of the wrapper (no hiding). 

However, indicators _injecting_ parameters into the indice (like the ones failing in icclim and in the doc), use `xclim.core.utils.wrapped_partial`. This function, in order to work properly, _hides_ the wrapped func, exposing only the signature of the uppermost function in the stack : the `percentile_bootstrap` wrapper, whose signature has not been updated.

Fortunately, we have `boltons.funcutils.wraps` which does this signature copying job!

Also, the tests in `test_modules.py::test_icclim` were skipped because of a faulty tests. It was badly written, skipping almost everything. I rewrote the tests in `test_module.py` and hopefully now it works correctly. Still, some tests in icclim are skipped because the percentile variable doesn't exist in `atmosds`. Currently, I think this concerns the precipitation tests, where the percentile variable is named "per", no matter what percentile. We could create a `per` and add it to `atmosds` by editing `xclim/testing/tests/conftest.py::atmosds`...

bzah:
@aulemahal Thanks for the help !
FYI, `boltons.funcutils::wraps` seems to shuffle `kwags` and `args` so I had to bind them with the function signature in `percentile_bootstrap`, before retrieving bootstrap argument.

I guess the PR can leave the draft and becoming active now ?
The only thing still blocking is the validation. As I said to @huard, I found quite some differences between the results of icclim and climpact. I will (hopefully today) send you a report to summarize this by mail.

bzah:
Hello,
I wanted to fix the xdoctest error, but I'm kind of confused (again) with the error I'm getting.

I'm using ` >>> tas = xr.open_dataset(path_to_tas_file).tas` in the bootstrapping example.
However, I get a **FileNotFoundError** for `path_to_tas_file`.
Instead of doing a HTTP call to get the file, like in other tests, in bootstrapping.py it is trying to find a local file.

I guess there is a configuration to force the http call but I can't find it.

Zeitsperre:
@bzah 

There's some magic going on in the background for the doctests. I'll take a look at it.

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,191 @@
...
+    @percentile_bootstrap
+    def tg90p(
+        tas: xarray.DataArray,
+        t90: xarray.DataArray,
+        freq: str = "YS",
+        bootstrap: bool = False
+    ) -> xarray.DataArray:
+
+    Example when called:
+    >>> import xarray as xr
```
Zeitsperre, line 29:
```suggestion
```
The magic going on here is that the `xr` used in the doctests is not `xarray`, but actually points to `xclim.testing._utils`:
https://github.com/Ouranosinc/xclim/blob/675bee51a51e29298d349578d8532e35f9686457/xclim/testing/_utils.py#L82
This function lets us set up examples using a custom `open_dataset` that looks for our test datasets on your disk or pulls it from our testdata repo. Removing the import will fix this. Sorry for the confusion!

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,190 @@
...
+        tas: xarray.DataArray,
+        t90: xarray.DataArray,
+        freq: str = "YS",
+        bootstrap: bool = False
+    ) -> xarray.DataArray:
+
+    Example when called:
+    >>> tas = xr.open_dataset(path_to_tas_file).tas
+    >>> t90 = percentile_doy(tas, window=5, per=90)
+    >>> tg90p(tas=tas, t90=t90, freq="YS", bootstrap=True)
```
Zeitsperre, line 31:
```suggestion
    Examples
    --------
    >>> from xclim.core.calendar import percentile_doy
    >>> from xclim.indices import tg90p
    >>> tas = xr.open_dataset(path_to_tas_file).tas
    >>> t90 = percentile_doy(tas, window=5, per=90)
    >>> tg90p(tas=tas, t90=t90, freq="YS", bootstrap=True)
```

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,193 @@
+from inspect import signature
+from typing import Any, Callable, Dict
+
+import numpy as np
+import xarray as xr
```
Zeitsperre, line 5:
```suggestion
import xarray
```

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,193 @@
...
+            else:
+                da_key = name
+
+    # Extract the DataArray inputs from the arguments
+    da: DataArray = kwargs.pop(da_key)
+    per: DataArray = kwargs.pop(per_key)
+
+    # List of years in base period
+    clim = per.attrs["climatology_bounds"]
+    per_clim_years = xr.cftime_range(*clim, freq="YS").year
```
Zeitsperre, line 108:
```suggestion
    per_clim_years = xarray.cftime_range(*clim, freq="YS").year
```

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,193 @@
...
+        bootstrap = ba.arguments.pop("bootstrap", False)
+        if bootstrap is False:
+            return func(*args, **kwargs)
+
+        return bootstrap_func(func, **ba.arguments)
+
+    return wrapper
+
+
+def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xr.DataArray:
```
Zeitsperre, line 52:
```suggestion
def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xarray.DataArray:
```

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,193 @@
...
+            kw[per_key] = percentile_doy(bda, **pdoy_args)
+            value = compute_indice_func(**kw).mean(dim="_bootstrap")
+
+        # Otherwise run the normal computation using the original percentile
+        else:
+            kw[per_key] = per
+            value = compute_indice_func(**kw)
+
+        out.append(value)
+    out = xr.concat(out, dim="time")
```
Zeitsperre, line 143:
```suggestion
    out = xarray.concat(out, dim="time")
```

xclim/core/bootstrapping.py
```diff
@@ -0,0 +1,193 @@
...
+    # TODO Make sure that's okay to insert the unit like that
+    out.attrs["units"] = "d"
+    return out
+
+
+# TODO: Return a generator instead and assess performance
+def bootstrap_year(
+    da: DataArray, groups: Dict[Any, slice], label: Any, dim="time"
```
Zeitsperre, line 151:
```suggestion
    da: DataArray, groups: Dict[Any, slice], label: Any, dim: str = "time"
```

Zeitsperre (commented):
Figured out the issue: The `xr` imported at the top-level was replacing the `xclim.testing` import in the doctest. This should work now.

bzah:
Hello !
IMO the pull request is, at last, ready to be merged.
I'm still working on comparison with other libraries but I'm pretty confident in the implementation we have here.

Thanks a lot for your help !

Be aware that:
- I have not completed some of the PR checklist bullet points, because I'm not sure how do you wish to handle them (tags, Hystory.rst  and bump version)
- I would like to have your advices on a "todo" in bootstrap_func.
`# TODO Make sure that's okay to insert the unit like that`
`out.attrs["units"] = "d"`
I'm not sure if it's okay to assign the unit there, but without this, the unit is missing in the bootsrapped DataArray.
- David wrote a todo to transform `bootstrap_year` and make it return a generator. I have too few knowledge about generator to see what is exactly expected here but I could work on it if necessary.

huard:
I'm made a change to assign the units from the output of a single call to the compute_indice_func, because it's not always "days". 
The generator TODO is an optimization task, no need to do it right now. 
Write a blurb in HISTORY. 

For some reason I can't push changes directly to this branch, so opened another one. I think this is because you didn't allow edits from maintainers when you created the PR. No big deal.

huard (approved):

PR #746, opened 2021-06-14
Add a .zenodo.json config and a CFF file

### LINKED ISSUE #721: Make xclim more FAIR
Possibly add 
- richer software metadata (see https://codemeta.github.io/)
- citation information (see https://citation-file-format.github.io/)

### References
- https://www.sciencedirect.com/science/article/pii/S2666389921000362?via%3Dihub
- https://content.iospress.com/articles/data-science/ds190026

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #721 

Adds a CFF file and a Zenodo configuration file. Lots of missing information still (mostly ORCID info). I'm unsure if names should priority reflect user metadata for searchability. If so, then those that haven't set a first and last name will not have their real names associated with the library. Thoughts?

Only for our Zenodo citation. We need to be very certain that we're happy with it, as once we publish, we can't (shouldn't) be modifying any DOI information after publishing a new version (bad practice).

* **Other information**:

https://citation-file-format.github.io/


huard:
What are the best practices for authors? I feel that including every contributor is overkill. Also, we want to be able to update that list to reflect future meaningful contributions. I suggest we draw a line - somewhere -.

aulemahal (approved):
Nice!
Maybe we could automate the update of the version in the CFF file with bumpversion?

Relating to the problem with linking the up-to-date zenodo, maybe a compromise solution would be to put the generic doi : "10.5281/zenodo.2795043"?  As it always point to the most recent version, this could be a problem...

Also, would : https://github.com/citation-file-format/cffconvert-github-action be useful?

aulemahal:
>  I feel that including every contributor is overkill.

The main citation method for xclim, Zenodo, already lists everyone, ordered by commit number.  So, the CFF at least lists the same people? I agree that it could be interesting to reduce this list for the sake of readability, but at the same time, most citations will look like "XXXXX et al." anyway?

Zeitsperre:
> What are the best practices for authors? I feel that including every contributor is overkill. Also, we want to be able to update that list to reflect future meaningful contributions. I suggest we draw a line - somewhere -.

I pulled this list and ordering (making @tlogan2000 the first author) from Zenodo. The same people are listed beyond that. The only major departure I've made is using some real names of people. In the case of @sbiner, I don't imagine him taking issue. For one-time contributors, maybe I should revert to the names found on Zenodo.

aulemahal:
Just thought of this, @Zeitsperre, could we add a "How to cite" section in the doc? May be in the "credits" page? Also, this could be in the main README?

Zeitsperre:
> Just thought of this, @Zeitsperre, could we add a "How to cite" section in the doc? May be in the "credits" page? Also, this could be in the main README?

This raises the question of how we want to cite our library. Are we the `Ouranos - xclim development team` or individual authors? Zenodo suggests listing the first few authors and ending with the last contributor's GitHub handle for good measure. How extensive do we want to be?

aulemahal:
Hum, my reflex would be to go with what Zenodo does, as others do... Until we have something in JOSS. We could also use the Earthcube Meeting's citation (which are not yet published, but soon).

Zeitsperre:
@huard I think it would be good to try and get this merged in the next version. Thoughts?

huard (approved):

PR #747, opened 2021-06-15
Agroclimat - add rolling_drydays_events and rolling_drydays_count

marielabonte:
### Pull Request Checklist:

Add the agro indices rolling_drydays_events and rolling_drydays_count

* **Other information**:


xclim/indices/_agro.py
```diff
@@ -341,3 +344,71 @@ def water_budget(
 
     out.attrs["units"] = pr.attrs["units"]
     return out
+
+
+@declare_units(pr="[precipitation]", thresh="[precipitation]")
+def rolling_drydays_events(
```
aulemahal, line 350: What about `dry_spells_frequency` and `dry_spells_total_length` ?
To mirror the ``hot_spells`, `heat_wave`, etc,  names. @tlogan2000

xclim/indices/_agro.py
```diff
@@ -341,3 +344,71 @@ def water_budget(
 
     out.attrs["units"] = pr.attrs["units"]
     return out
+
+
+@declare_units(pr="[precipitation]", thresh="[precipitation]")
```
aulemahal, line 349: To make clear that the threshold is on the _accumulated_ precipitation, I believe it should be `thresh="[length]"`, there is no error because we are very permissive for "hydro" units!

You might want to use `xc.core.units.rate2amount` on `pr` for the same reason.

xclim/indicators/atmos/_precip.py
```diff
@@ -308,3 +310,23 @@ class HrPrecip(Hourly):
         indices.liquid_precip_ratio, suggested={"tas": _empty}, prsn=None
     ),
 )
+
+
+rolling_drydays_events = Precip(
+    identifier="rolling_drydays_events",
+    description="The number of periods of minimum {window} days during which the accumulated precipitation is under"
+    "{thresh}, for each resampling period.",
```
aulemahal, line 318:
```suggestion
    description="The {freq} number of periods of minimum {window} days during which the accumulated precipitation is under {thresh}.",
```
Formatting trick: common `freq` values are translated to an adjective when inserted in the attributes. Allows for less words and more precision!

xclim/indices/_agro.py
tlogan2000: > 
> 
> What about `dry_spells_frequency` and `dry_spells_total_length` ?
> To mirror the `hot_spells`, `heat_wave`, etc, names. @tlogan2000

Ok for me. I think I would say `dry_spell_frequency` and `dry_spell_total_length`...   i,e, not "spells"

xclim/indicators/atmos/_precip.py
tlogan2000: `description="The {freq} number of periods of at least {window} days during which accumulated {window} day precipitation is under {thresh}.",`

Would this suggested change clarify the nuance that say a 9 or 10 day period is a single event even if total precip over all days is greater than thresh?   Or does it make it more difficult?

xclim/data/fr.json
```diff
@@ -891,5 +891,17 @@
     "abstract": "Précipitations moins l'évapotranspiration potentielle en tant que mesure approximative d'un bilan hydrique de surface, où l'évapotranspiration potentielle est calculée avec une méthode donnée.",
     "description": "Précipitations moins l'évapotranspiration potentielle en tant que mesure approximative d'un bilan hydrique de surface, où l'évapotranspiration potentielle est calculée avec la méthode {méthode}.",
     "long_name": "Bilan hydrique."
+  },
+  "ROLLING_DRYDAYS_EVENTS": {
+    "title": "Événements de périodes séches.",
+    "abstract": "Le nombre de périodes de n jours minimum pendant lesquelles les précipitations accumulées sont inférieures à un seuil donné, pour chaque période de rééchantillonnage.",
```
tlogan2000, line 897: See my comment on the n day nuance but I think this is technically more accurate (even if it becomes wordy):
Applies to `ROLLING_DRYDAYS_COUNT` as well I think
````
"abstract": "Le nombre de périodes de n jours minimum pendant lesquelles les précipitations accumulées sur n jours sont inférieures à un seuil donné, pour chaque période de rééchantillonnage."

"description": "Le nombre de périodes de {window} jours minimum pendant lesquelles les précipitations accumulées sur {window} jours sont inférieures à {thresh}, pour chaque période de rééchantillonnage."
````

File: xclim/indicators/atmos/_precip.py [hunk 5]
```diff
@@ -308,3 +310,23 @@ class HrPrecip(Hourly):
...
+    units="",
+    cell_methods="",
+    compute=indices.rolling_drydays_events,
+)
+
+
+rolling_drydays_count = Precip(
+    identifier="rolling_drydays_count",
+    description="The number of days in periods of minimum {window} days during which the accumulated precipitation "
```
tlogan2000, line 327: Same comments here regarding `{freq}` and `{window}`  as for 'events' calculations

xclim/indices/_agro.py
tlogan2000: +1 for this

xclim/indices/_agro.py
```diff
@@ -341,3 +344,71 @@ def water_budget(
...
+
+    Returns
+    -------
+    xarray.DataArray
+      The number of periods of minimum {window} days for each resampling period.
+    """
+    thresh = convert_units_to(thresh, pr, "hydro")
+
+    out = (
+        (pr.rolling(time=window, center=True).sum() < thresh)
```
tlogan2000, line 376: if we follow @aulemahal suggestion and decide to use `xc.core.units.rate2amount` should this become a rolling mean?

tlogan2000 (changes requested):
A few suggestions and questions

xclim/indices/_agro.py
marielabonte: We still want the accumulated precipitation amount over the whole period, so we need to keep the rolling sum.

xclim/data/fr.json
marielabonte: Do you mean that it is better to keep the wording "pour chaque période de rééchantillonnage" instead of "Fréquence {freq:f} des périodes sèches"?

xclim/indices/_agro.py
tlogan2000: Running the code/tests now & I see how this works

xclim/indicators/atmos/_precip.py
marielabonte: Are you suggesting to replace "for each resampling period" by {freq}?

xclim/data/fr.json
tlogan2000: No. Sorry. "Fréquence {freq:f} des périodes sèches" looks good. My comment was wrt to the second inclusion of {window}   in descritpion in the `accumulées`  portion.  Idea is to indicate that even when an event is longer than `window` the sum is always performed on `n==window` days

xclim/indicators/atmos/_precip.py
tlogan2000: > Are you suggesting to replace "for each resampling period" by {freq}?

nope see above

xclim/indices/_agro.py
```diff
@@ -368,12 +368,13 @@ def rolling_drydays_events(
     Returns
     -------
     xarray.DataArray
-      The number of periods of minimum {window} days for each resampling period.
+      The number of dry periods of minimum {window} days for each resampling period.
     """
-    thresh = convert_units_to(thresh, pr, "hydro")
+    pram = rate2amount(convert_units_to(pr, "mm/day"))
```
aulemahal, line 373:
```suggestion
    pram = rate2amount(pr, out_units='mm')
```
I think this is more explicitly what you are looking for? At the same time, I don't think there's a real need to force units here?

xclim/indices/_agro.py
```diff
@@ -404,11 +405,12 @@ def rolling_drydays_count(
     Returns
     -------
     xarray.DataArray
-      The number of days in periods of minimum {window} days for each resampling period.
+      The total length in days of dry periods of minimum {window} days for each resampling period.
     """
-    thresh = convert_units_to(thresh, pr, "hydro")
+    pram = rate2amount(convert_units_to(pr, "mm/day"))
```
aulemahal, line 410: Same comment as above.
```suggestion
    pram = rate2amount(pr, "mm")
```

aulemahal (approved):
This is good!
Minor comments on unit conversions.

xclim/indices/_agro.py
marielabonte: It's true, I don't need to force the units. 

I used convert_units_to() to update the units of pram in mm. This way I can use convert_units_to() on thresh to make sure it has the same units as pram.

I could update pram units separately.

xclim/indices/_agro.py
```diff
@@ -341,3 +344,75 @@ def water_budget(
...
+      Resampling frequency.
+
+    Returns
+    -------
+    xarray.DataArray
+      The {freq} number of dry periods of minimum {window} days.
+    """
+    pram = rate2amount(pr, "mm")
+    pram.attrs["units"] = "mm"
+    thresh = convert_units_to(thresh, pram)
```
aulemahal, line 375: Hum, not sure this will work? The 2nd argument to `rate2amout` is `dim='time'`, `out_units` is third.
Also, not sure I understood your answer to the other comment and this change : why force "mm"?
Why not:
```suggestion
    pram = rate2amount(pr)
    thresh = convert_units_to(thresh, pram)
```
?

marielabonte: What I meant: I use rate2amount(pr) to convert pr values from kg/m2/s into mm. However, the attribute of pr reads like kg/m2 and not mm. So, thresh = convert_units_to(thresh, pram) returns an error, saying it cannot convert mm into kg/m2.

But I see now that if I specify the out_units like you said: pram = rate2amount(pr, out_units="mm"), pram units will be in mm. This way the thresh = convert_units_to(thresh, pram) will work.

tlogan2000 (approved):
Looks good to me now. @aulemahal do you need a final look?

xclim/indices/_agro.py
aulemahal: Oh! True, I didn't think of this first issue with kg/m²...

aulemahal:
Everything good for me! LGTM!

PR #748, opened 2021-06-15
Jeremyfyke/clausius clapeyron precipitation scaling

JeremyFyke:
### Pull Request Checklist:
    - This PR fixes #xyz

Adds a new indice calculator that calculates a Clausius-Clapeyron-based scaling of present precipitation rates for a given precipitation metric, to estimate future precipitation rates for the same metric.

No, not to my knowledge.

* **Other information**:

### DIFF
--- a/xclim/indices/_conversion.py
+++ b/xclim/indices/_conversion.py
@@ -762,7 +762,7 @@ def clausius_clapeyron_scaled_precipitation(
     if "time" in tmean_baseline.coords.keys():
         if tmean_baseline.coords["time"].size != 1:
             raise ValueError(
-                "Temperature baseline needs to be a single time slice (e.g., of a common climatologal period)."
+                "Temperature baseline needs to be a single time slice (e.g., of a common climatological period)."
             )
         # else:
         #    tmean_baseline = tmean_baseline.squeeze(dim=["time"], drop=True)
--- a/xclim/indices/_conversion.py
+++ b/xclim/indices/_conversion.py
@@ -17,6 +17,7 @@ __all__ = [
     "snowfall_approximation",
     "rain_approximation",
     "wind_chill_index",
+    "clausius_clapeyron_scaled_precipitation",
 ]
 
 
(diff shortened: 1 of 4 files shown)


aulemahal:
Hi! Interesting addition!

A few suggestion before a real review:
1. Here : https://xclim.readthedocs.io/en/stable/contributing.html, you'll find some steps for testing and linting. If you install pre-commit in your copy of xclim, the linting will be done automatically, and the black test won't fail!
2. I'd suggest moving this to `_conversion.py`, I prefer putting there those calculations that do not perform resampling.
3. I'd suggest testing with real data (that has a `time` axis) I believe the current code wouldn't work?

I'm saying the last thing because of the subtraction of array with different time coordinates (one future, one baseline), this will not work as expected in xarray!

JeremyFyke:
Hi @aulemahal,

Great, thanks for comments.  I did install/run pre-commit and it did pass on my end so a bit surprising to see it fail here.  Are there particular xclim-specific tests to apply?
( image )

OK, sounds good, I'll move to _conversion.py

Thanks for heads up on time coordinate/xarray issue.  What would you suggest doing in this case (to calculate essentially a dT value, from two values separated by a dt)?  I think it might be just fine to use base numpy here..?

Zeitsperre:
@JeremyFyke Thanks for the addition! Since you've already pushed changes, `pre-commit` won't look at anything in particular. If you want to force it to look over all existing files, run `pre-commit run --all-files`.

JeremyFyke:
OK great thanks.  BTW I'm super happy to learn about these pre-commit/etc. tools, which are new to me.  Super useful.  I'll go back and do some more work here as per comments, and get back to you with a revised (tested) version of this.

JeremyFyke:
Also, @huard, note work on this IDF scaling front as per our earlier discussions.

JeremyFyke:
Ya good question @aulemahal ... I was imagining, that the user would take an appropriate time average earlier (with another tool, like xclim.indices.generic.aggregate_between_dates, or whatever), and then feed in this to the scaling (which, at it's core, requires climatological averages).  So, I'm imagining (and perhaps could enforce via a code check) that the input to this tool would either be scalar or have spatial dimensions, but no time dimensions.  What do you think?

huard:
I think this is worth including in the test, ie take time series, compute their means and feed them to the function.

aulemahal:
If made clear in the docstring, I think the xarray error could suffice, no need for a check!

JeremyFyke:
Hmm, also rethinking a bit here.  It might be too dogmatic for me to say 'no' to time dimensions, as previously.  Say we have a time series of climatological averages as input (e.g. a boxcar moving average or something).  In this case, it could be useful to retain ability to have a time dimension accepted in the tool.  Hmm.

JeremyFyke:
Hi all,

I've pushed a few additional commits (one being a merge conflict resolution between this work, and PET work).  I *think* these changes address the issues noted above.  In particular, have moved code to _conversions.py, and added more logic to check/alter/deny time dimension structure.  I've also updated the unit test to be compatible with this code (and test it by providing input data with toy time/lat/lon dimensions. Perhaps see code itself (in _conversion.py) for the best description.  Thoughts/comments/etc/ welcome!  I'll be playing around with this code in practice for an ongoing project, so I will certainly will you know on my end if I see any problems arising.

huard (approved):
Add something like: 
`* New function `xclim.indices.clausius_clapeyron_scaled_precipitation` can be used to scale precipitation according to changes in mean temperature.`
to HISTORY and this is good to go I think. 

@tlogan2000 are you ok with this ? One possible issue would be the naming of variables: `tmean_baseline`, `tmean_future`, `pr_baseline`. Do we have a convention to refer to climatic periods, climatological means, etc ?

tlogan2000:
> `tmean_baseline`, `tmean_future`, `pr_baseline`. Do we have a convention to refer to climatic periods, climatological means, etc ?

We don't really have these type of variables so far in xclim (climatologies).  However, in general we've been trying to move towards CMIP naming conventions, so I think this variable would still be `tas` and not 'tmean'. But I think we still need the indication of a climatology so  `tas_baseline`, `tas_future` would seem ok if that seems reasonable to all concerned.   I don't have a massive objection to `tmean` but I think `tas` would be more coherent with the rest of the package.  `pr` is already as in CMIP

huard:
@JeremyFyke I don't think the broadcasting is working as you expect it. 

Also, do you want to automatically change the time of pr_baseline so the dates are added the time delta between historical and future ? If you do, what happens when, as in your example, you have multiple future periods ?

xclim/indices/_conversion.py
```diff
@@ -706,6 +707,76 @@ def wind_chill_index(
...
+            raise ValueError(
+                "Precipitation baseline needs to be a single time slice (e.g., of a common climatological period)."
+            )
+        # else:
+        #    pr_baseline = pr_baseline.squeeze(dim=["time"], drop=True)
+
+    if "time" in tmean_baseline.coords.keys():
+        if tmean_baseline.coords["time"].size != 1:
+            raise ValueError(
+                "Temperature baseline needs to be a single time slice (e.g., of a common climatologal period)."
```
huard, line 765:
```suggestion
                "Temperature baseline needs to be a single time slice (e.g., of a common climatological period)."
```

JeremyFyke:
> @JeremyFyke I don't think the broadcasting is working as you expect it.

How so? 
 
> Also, do you want to automatically change the time of pr_baseline so the dates are added the time delta between historical and future ? If you do, what happens when, as in your example, you have multiple future periods ?

I think the main thing is that the future pr values have the same time dimension absolute values as the future tmean (ts) time values (i.e. these are estimated pr values for the future time periods corresponding to the future temperature estimates).   Does this make sense?

JeremyFyke:
Also - I'm off for vacation as of now - back in 2+ weeks and happy to pick this conversation up then.  Alternatively, please feel 100% free to push this forward more unilaterally on your end in the interim - I'm quite sure I'll be fine with whatever you/Travis/etc. decide here!

huard:
TestClausiusClapeyronScaledPrecip:test_simple does not pass. 

Ok, will see what I can do.

huard:
I've rethought about this and I would like to suggest that this function accepts a delta_tas and pr. We'd leave the responsibility to the user to compute this delta_tas, instead of trying to compute the difference ourselves. In most case it's trivial, but if you want to do something a little fancy (for example compute it for different future time periods), then you have to decide how to define the dimensions of the delta_tas input.

PR #752, opened 2021-06-22
Allow requirement of optional variable

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Fixes a bug in `core.utils.infer_kind_from_parameter` and  in `Indicator._assign_named_args` that was forbidding the requirement of optional variables (i.e. : change an indicator through yaml so that an optional variable is not optional anymore.)

No

### Other information:
The way to make a variable non-optional in a yaml def, is the following:
```yaml
WATER_BUDGET:
  base: water_budget
  input:
    tasmin: tasmin
    tasmax: tasmax
  index_function:
    parameters:
      method: BR64
```
In this example, the variables `tasmin` and `tasmax`, which are optional in `atmos.water_budget` are now semi-optional. There is still a "bug" that makes the annotation unchanged (i.e.: `WATER_BUDGET.parameters['tasmin']['kind'] == OPTIONAL_VARIABLE`), but the default value has passed from `None` to `"tasmin"`. Thus a call with `WATER_BUDGET(ds=ds)` will be successful (it would fail earlier). And a call like `WATER_BUDGET(pr=ds.pr)` will fail because tasmin wasn't None, it was "tasmin". However, the error raised will be unclear. I'll fix this in a further PR.

### DIFF
--- a/docs/notebooks/example.yml
+++ b/docs/notebooks/example.yml
@@ -72,3 +72,14 @@ indices:
       name: extreme_precip_accumulation
       parameters:
         perc: 99
+  LPRatio:
+    base: liquid_precip_ratio
+    input:
+      tas: tas
+      pr: pr
+    output:
+      long_name: Liquid precip to total precip ratio.
+      description: Precipitation is estimated to be solid when tas is under 0.5°C
+    index_function:
+      parameters:
+        thresh: 0.5 degC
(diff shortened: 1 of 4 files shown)


tlogan2000 (approved):
This seems ok to me not sure why tests are failing

aulemahal:
Well, there was a mighty bug in there!  `wrapped_partial` stores injected arguments in a `_injected` dictionary, attribute to the new function. Well, we were not copying it, so wrapping a wrapped function would _update_ the _same_ dictionary. That's why `atmos.liquid_precip_ratio` was getting the injected `thresh` of `example.LPRatio`. Woups. 

While fixing this, I realized there was a way to simply update the "kind" field in `Indicator.parameters` and to change an annotation when an optional variable is being made non-optional. So this fixes the error I discussed in the header comment!

PR #753, opened 2021-06-22
Add Huglin Index

Zeitsperre:
### Pull Request Checklist:
    - This PR addresses goals in #355 

* Adds the Huglin Index with three (3) different methods for performing the day-length modification due to latitude:
   * `smoothed`: a progressive value from 1 to 1.06 for latitudes from 40 deg to 50 deg
   * `icclim`: mimics the original stepwise calculation from Huglin, identical to the ECAD/ICCLIM method.
   * `jones`: a modified version using the revised formula from Hall & Jones, 2010 (DOI:10.1111/j.1755-0238.2010.00100.x)
 * Adds a day-lengthy multiplier that return day-length in hours/per day or total day-hours between two dates (using aggregate_between_dates). 
 * The `jones` day-length multiplier (with adjustments for the BEDD `k`-curve) has also been added as an option for BEDD calculation.

The var_name for the `corn_heat_units` indicator is now `chu`.


xclim/indices/generic.py
```diff
@@ -776,6 +780,59 @@ def degree_days(tas: xr.DataArray, thresh: str, condition: str) -> xr.DataArray:
...
+
+    References
+    ----------
+    Modified day-length equations for Huglin heliothermal index published in Hall, A., & Jones, G. V. (2010). Spatial
+    analysis of climate in winegrape-growing regions in Australia. Australian Journal of Grape and Wine Research, 16(3),
+    389‑404. https://doi.org/10.1111/j.1755-0238.2010.00100.x
+
+    Examples available from Glarner, 2006 (http://www.gandraxa.com/length_of_day.xml).
+    """
+    warnings.warn("This is untested.", UserWarning, stacklevel=4)
```
aulemahal, line 826: A simple test should be easy to add, no?

xclim/indices/generic.py
```diff
@@ -776,6 +780,59 @@ def degree_days(tas: xr.DataArray, thresh: str, condition: str) -> xr.DataArray:
...
+    axis = xr.where(lat > 0, obliquity, -obliquity)
+    m_lat_dayofyear = 1 - np.tan(np.radians(lat)) * np.tan(
+        axis * (np.cos((2 * np.pi * dates.dt.dayofyear) / 365.25))
```
aulemahal, line 829: Would it make sense to change the year length according to the calendar and year?

aulemahal (approved):
Nice! Is there a reason there is no Indicator?

xclim/indices/generic.py
Zeitsperre: Yes, that would make much more sense, given that we have different year-lengths. The original formula didn't consider this possibility, but we can likely support it easily. Is there something synonymous to `dates.dt.days_in_year` that we could use?

xclim/indices/generic.py
Zeitsperre: True. This should be easy to do in `test_generic.py`.

Zeitsperre:
> Nice! Is there a reason there is no Indicator?

I wanted to see if I was on the right track before I went any farther. Indicator will be committed eventually.

xclim/indices/generic.py
Zeitsperre: Haha, never mind, I see that we've coded that ourselves!

Zeitsperre:
Nearly done. All that's missing is the ICCLIM indicator. Thanks for the help @aulemahal !

Zeitsperre:
@aulemahal @tlogan2000 If everything works for you, we can merge this.

xclim/indicators/atmos/_temperature.py
```diff
@@ -781,11 +782,36 @@ def cfcheck(tas, snd=None):
     description="Temperature-based index used to estimate the development of corn crops. "
     "Corn growth occurs when the minimum and maximum daily temperature both exceeds "
     "specific thresholds : Tmin > {thresh_tasmin} and Tmax > {thresh_tasmax}.",
+    var_name="chu",
```
aulemahal, line 785: haha, hého. You can't hide almost-breaking unrelated changes like this! ;)

I agree with it, let's just notify @marielabonte that the output of `corn_heat_units` will change with this PR.

xclim/indicators/atmos/_temperature.py
```diff
@@ -781,11 +782,36 @@ def cfcheck(tas, snd=None):
...
+    var_name="hi",
+    compute=wrapped_partial(
+        indices.huglin_index,
+        method="jones",
+        suggested=dict(
+            thresh_tasmin="10 degC",
+            start_date="04-01",
+            end_date="10-01",
+            lat=_empty,
+        ),
```
aulemahal, line 810:
```suggestion
```

This is unnecessary, as those are _already_ the default of the indice!

xclim/indices/_agro.py
```diff
@@ -297,10 +300,11 @@ def biologically_effective_degree_days(
       Latitude coordinate.
```
aulemahal, line 300: Should have given this comment on the other PR, but I would specify that lat is not optional when method is in [gladstones, jones], while it is unused with icclim.

aulemahal (approved):
a few minor comments, but I approve!

xclim/indicators/atmos/_temperature.py
Zeitsperre: Good to know!

xclim/indicators/atmos/_temperature.py
Zeitsperre: Woops, fair enough. Breaking changes mentioned above!

PR #764, opened 2021-07-06
Release 0.28.0-beta

huard:
Updated HISTORY, marked some new features as *experimental.* 

Bumpversion to 0.28.0-beta. 

Should we bump to 0.28.0 or leave it at beta ?


Zeitsperre:
> Should we bump to 0.28.0 or leave it at beta ?

Either one would be fine for me, my question to you is if we tag a beta version, should we make it available on PyPI and conda-forge?

huard:
I suppose there is no point in making a release but not distributing it. I suggest we bump to 0.28.0 and deal with bugfixes in patch releases if needed. We merged a few PRs that I consider still experimental, but they're identified in the release notes.

tlogan2000:
> 
> 
> I suppose there is no point in making a release but not distributing it. I suggest we bump to 0.28.0 and deal with bugfixes in patch releases if needed. We merged a few PRs that I consider still experimental, but they're identified in the release notes.

ok for me too

HISTORY.rst
```diff
@@ -2,21 +2,21 @@
 History
 =======
 
-0.28.0 (unreleased)
+0.28.0 (2021-07-06)
```
Zeitsperre, line 5:
```suggestion
0.28.0 (2021-07-07)
```

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.27.5-beta
+current_version = 0.28.0-beta
```
Zeitsperre, line 2: `bumpversion release`, we can tag from master and release this is merged.

HISTORY.rst
```diff
@@ -2,21 +2,21 @@
...
 -------------------
 
 New features and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 * Automatic load of translations on import and possibility to pass translations for virtual modules.
-* New ``xclim.testing.list_datasets`` function listing all available test datasets in repo `xclim-testdata`.
-* `spatial_analogs` accepts multi-indexes as the `dist_dim` parameter and will work with candidates and target arrays of different lengths.
-* `humidex` can be computed using relative humidity instead of dewpoint temperature.
+* New ``xclim.testing.list_datasets`` function listing all available test datasets in repo ``xclim-testdata``.
```
Zeitsperre, line 11: I think we need some guidelines on how best to use the tick-formatting. Should we opt for always doubled (``) for call signatures, files, functions, etc.?

Zeitsperre (approved):

HISTORY.rst
huard: That's my understanding. single ticks are for links. But happy to be proven wrong.

PR #775, opened 2021-07-14
Add optional argument to assign run lengths on the last day instead of the first one

### LINKED ISSUE #774: Add option to assign run length values to last day of run period rather than first one
At the moment, for a dataset where we want to find the run lengths of cells with values of 1, for example:
[ 0, 1, 1, 1, 1, 1, 0, 1, 0, 1, 1, 0, 0, 0]

`rle` is producing an output where the values of the run lengths are saved on the cell of the first occurence:
[ 0., 5., nan, nan, nan, nan, 0., 1., 0., 2., nan,  0., 0.,  0.]

However, it might be needed that the value of the run length needs to be saved on the last day, like this:
[ 0., nan, nan, nan, nan, 5., 0., 1., 0., nan, 2.,  0., 0.,  0.]

In order to do this, we have added a `lastday` parameter to `rle`. I'll open a PR in a minute. 

@clairbarnes is the main author of these changes and can give more information about why this functionality is useful.

raquel-ucl:
### Pull Request Checklist:
    - This PR fixes #774

`rle` can now take a `lastday` flag to save the value of the run length on the position of the last occurrence rather than the first.
`rle` tests have been modified accordingly.

No, but it's likely other `rle` related methods also need changing, e.g. `rle_statistics`. 

### Other information:
@clairbarnes is the main author of this modification.


raquel-ucl:
CodeFactor is complaining about 2 TODO sentences. Not sure how to make the CodeFactor tests pass!

Zeitsperre:
Hi @raquel-ucl, no need to worry about the CodeFactor check failing (it's complaining about some `TODO` comments that need to eventually be addressed). The docs build failure needs to be addressed. Will look into it.

xclim/indices/run_length.py
```diff
@@ -38,7 +37,7 @@ def use_ufunc(
     Parameters
     ----------
     ufunc_1dim: {'from_context', 'auto', True, False}
-    da : xarray.DataArray
+    da : .DataArray
```
huard, line 40: ?

xclim/indices/run_length.py
```diff
@@ -121,10 +128,13 @@ def rle(
 
     # back fill nans with first position after
     z = b.bfill(dim=dim)
+
     # calculate lengths
     d = z.diff(dim=dim) - 1
     d = d.where(d >= 0)
```
huard, line 134: If there is no significant performance penalty from the reindexing operation, I'm happy with the proposed solution. If there is, then I think you'd need to something like
```
d = d.where(d <= 0)
```
when `lastday is True`.

xclim/indices/run_length.py
raquel-ucl: Whoops! I was replacing references to `xarray` with `xr` and something went wrong here :)

xclim/indices/run_length.py
raquel-ucl: It doesn't look like the reindexing is impacting perfomance very much, although I've only measured the tests:

```
xclim/testing/tests/test_run_length.py::test_rle[False-True-True] lastday: True
0.0477299690246582
PASSED
xclim/testing/tests/test_run_length.py::test_rle[False-True-False] lastday: True
0.02160477638244629
PASSED
xclim/testing/tests/test_run_length.py::test_rle[False-False-True] lastday: False
0.04825186729431152
PASSED
xclim/testing/tests/test_run_length.py::test_rle[False-False-False] lastday: False
0.01996779441833496
```

Zeitsperre:
@raquel-ucl One thing that must eventually be addressed is this PR needs to be rebased/merged with the upstream `master` branch, but feel free to ignore that for now until changes here are approved.

xclim/indices/run_length.py
raquel-ucl: Fixed now.

raquel-ucl:
> @raquel-ucl One thing that must eventually be addressed is this PR needs to be rebased/merged with the upstream `master` branch, but feel free to ignore that for now until changes here are approved.

Done! :)

xclim/indices/run_length.py
huard: Excellent then.

huard (approved):

raquel-ucl:
Should this PR also include fixes for other related methods like `rle_statistics` so the `lastday` flag can get propagated to the `rle` calls?

tlogan2000:
> Should this PR also include fixes for other related methods like `rle_statistics` so the `lastday` flag can get propagated to the `rle` calls?

This makes sense to me.  `rle` is the base function for almost all functions in `run_length.py` so propagating options to higher levels makes sense.  We would also maybe want to eventually propagate to `xclim.indices` that use `run_length` (heat waves etc etc)  but I think that could be done in a separate PR

xclim/testing/tests/test_run_length.py
```diff
@@ -158,19 +165,23 @@ def test_simple(self, tas_series, coord, expected, use_dask, ufunc):
...
-        assert rl.windowed_run_events(a, 3, dim="x") == 2
+        assert rl.windowed_run_events(a, 3, dim="x", lastday=lastday) == 2
 
 
 class TestWindowedRunCount:
-    def test_simple(self):
-        a = xr.DataArray(np.zeros(50, bool), dims=("x",))
+    @pytest.mark.parametrize("lastday", [True, False])
+    def test_simple(self, lastday):
+        a = xr.DataArray(np.zeros(50, bool), dims=("time",))
```
raquel-ucl, line 179: I changed the dimension name from `x` to `time` so that tests passed, but I think the change that is actually needed is in `rle`, so that the new changes in this PR work with dimensions other than "time". 

The problematic lines are: 
https://github.com/UCL/xclim/blob/rle_lastday/xclim/indices/run_length.py#L88
`da = da.reindex(time=da.time[::-1])` 

and 

https://github.com/UCL/xclim/blob/rle_lastday/xclim/indices/run_length.py#L137
`d = d.reindex(time=d.time[::-1])`

Any ideas would be welcome! I'm not an expert on xarray, so there's hopefully a way to workaround hardcoding `time=` :)

tlogan2000: @raquel-ucl I'm pretty sure you can avoid the hard-coded `time`  and use the input `dim` param via:
````
da = da.reindex({dim:da[dim][::-1]})

File: xclim/indices/run_length.py [hunk 4]
```diff
@@ -67,18 +69,23 @@ def rle(
...
+      and are NaN on the other days of the runs.
+      If lastday is True, values are N on the last day of a run, where N is the length of that run,
       and are NaN on the other days of the runs.
     """
     use_dask = isinstance(da.data, dsk.Array)
 
     # Ensure boolean
     da = da.astype(bool)
+    if lastday:
+        da = da.reindex(time=da.time[::-1])
```
aulemahal, line 88: As noted below, but flagging it here, we should avoid hardcoding a dimension name here!

aulemahal (changes requested):
Nice! Thanks for this addition!

Problem: we have 2 versions of most run-length function "ufunc"  (`rle_1d`) and vectorized (`rle`). This PR adds the argument to the latter only.  As the switch between the two is automatic, results might be unexpected. For example : `rle_statistics` with a small dataarray will NOT respect the `lastday` argument in the current code. 

We anyway need to reconsider the usefulness of all this code duplication, but we won't have time to do this soon. For now, I suggest defaulting to the "vectorized" version when `lastday` is True. Best way to do this is to edit the check `use_ufunc` and pass `lastday` in every relevant calls.

aulemahal:
Hi @raquel-ucl, we are planning to release xclim 0.29 at the end of next week. Did you need this in that release? If so, would you like some help in completing the PR?

raquel-ucl:
Hello! Sorry it took a while to reply. I have some local changes I can upload but not sure I can make all changes for the next release. We also have a few more indices we'd like to add so it might be better we wait until the release after this one. Do you know when that'd be, approximately?

xclim/indices/run_length.py
```diff
@@ -435,7 +471,9 @@ def _get_indices(arr, *, N):
     return xr.concat((starts, ends), "bounds")
 
 
-def keep_longest_run(da: xr.DataArray, dim: str = "time") -> xr.DataArray:
+def keep_longest_run(
+    da: xr.DataArray, dim: str = "time", lastday: bool = False
+) -> xr.DataArray:
```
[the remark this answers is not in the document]
raquel-ucl: You are right. I've removed it now from the tests and the method implementation.

raquel-ucl:
@aulemahal I just discussed with my colleagues and it'd be much better if we could get this PR released end of this week. There's just one more change I need to get done to submit for review again.

xclim/indices/run_length.py
raquel-ucl: Done :)

raquel-ucl:
@aulemahal @huard @tlogan2000 I have now implemented all the requested changes, and tests pass locally. Hopefully these changes can soon be merged :)

huard:
@raquel-ucl Unless you have an objection, I'll make changes to your PR to change the argument from `last_day (bool)` to `index (str)` taking values `first` and `last`. I feel this will be easier to understand, especially when dealing with non-daily data. Also, it will allow for a future implementation for the `center` index.

raquel-ucl:
Sure @huard, no objections :)

xclim/indices/run_length.py
```diff
@@ -41,18 +44,23 @@ def use_ufunc(
...
-    if ufunc_1dim == "from_context":
-        ufunc_1dim = OPTIONS[RUN_LENGTH_UFUNC]
-
-    if ufunc_1dim == "auto":
-        return not uses_dask(da) and (da.size // da[dim].size) < npts_opt
+    if not lastday:
+        if ufunc_1dim == "from_context":
+            ufunc_1dim = OPTIONS[RUN_LENGTH_UFUNC]
+        if ufunc_1dim == "auto":
+            return not uses_dask(da) and (da.size // da[dim].size) < npts_opt
```
aulemahal, line 63:
```suggestion
    if ufunc_1dim == "from_context":
        ufunc_1dim = OPTIONS[RUN_LENGTH_UFUNC]
    if ufunc_1dim == "auto":
        ufunc_1dim =  not uses_dask(da) and (da.size // da[dim].size) < npts_opt
    # Now ufunc_1dim has to be a bool.
    # return True only if lastday is False.
    return (not lastday) and ufunc_1dim
```

aulemahal: In the previous code, if lastday was True, there was a risk of returning a string. This didn't cause any bugs because non-empty strings evaluate to True. Thus, this change shouldn't change the behaviour, only make it cleaner.

aulemahal (approved):
Thanks for the changes! One suggestion and there we go.
We'll merge @huard's PR directly after this.

raquel-ucl:
Thanks! :)

PR #777, opened 2021-07-16
Indices/calm windy days

raquel-ucl:
### Pull Request Checklist:
    - This PR fixes #xyz

* ,,,

### DIFF
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -26,6 +26,7 @@ from .generic import domain_count, threshold_count
 # -------------------------------------------------- #
 
 __all__ = [
+    "calm_days",
     "cold_spell_days",
     "cold_spell_frequency",
     "daily_pr_intensity",
@@ -68,8 +69,45 @@ __all__ = [
     "maximum_consecutive_wet_days",
     "sea_ice_area",
     "sea_ice_extent",
+    "windy_days",
 ]
 
+
+@declare_units(pr="[speed]", thresh="[speed]")
+def calm_days(
+    pr: xarray.DataArray, thresh: str = "2 m/s", freq: str = "MS"
+) -> xarray.DataArray:
+    r"""Calm days.
+
+    The number of days with average windspeed below threshold.
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+      Daily windspeed.
+    thresh : str
+      Threshold average windspeed on which to base evaluation.
+    freq : str
+      Resampling frequency.
+
+    Returns
+    -------
+    xarray.DataArray, [time]
+      Number of days with average windspeed below threshold.
+
+    Notes
+    -----
+    Let :math:`WS_{ij}` be the windspeed at day :math:`i` of period :math:`j`. Then
+    counted is the number of days where:
+
+    .. math::
+
+        WS_{ij} < Threshold [m/s]
+    """
+    thresh = convert_units_to(thresh, pr)
+    out = threshold_count(pr, "<", thresh, freq)
+    out = to_agg_units(out, pr, "count")
+    return out
 
 @declare_units(tas="[temperature]", thresh="[temperature]")
 def cold_spell_days(
@@ -1758,6 +1796,43 @@ def sea_ice_extent(
     t = convert_units_to(thresh, siconc)
     out = xarray.dot(siconc >= t, areacello)
     out.attrs["units"] = areacello.units
+    return out
+
+
+@declare_units(pr="[speed]", thresh="[speed]")
+def windy_days(
+    pr: xarray.DataArray, thresh: str = "10.8 m/s", freq: str = "MS"
+) -> xarray.DataArray:
+    r"""Windy days.
+
+    The number of days with average windspeed above threshold.
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+      Daily average windspeed.
+    thresh : str
+      Threshold average windspeed on which to base evaluation.
+    freq : str
+      Resampling frequency.
+
+    Returns
+    -------
+    xarray.DataArray, [time]
+      Number of days with average windspeed above threshold.
+
+    Notes
+    -----
+    Let :math:`WS_{ij}` be the windspeed at day :math:`i` of period :math:`j`. Then
+    counted is the number of days where:
+
+    .. math::
+
+        WS_{ij} > Threshold [m/s]
+    """
+    thresh = convert_units_to(thresh, pr)
+    out = threshold_count(pr, ">", thresh, freq)
+    out = to_agg_units(out, pr, "count")
     return out
 
 
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -109,6 +109,7 @@ def calm_days(
     out = to_agg_units(out, pr, "count")
     return out
 
+
 @declare_units(tas="[temperature]", thresh="[temperature]")
 def cold_spell_days(
     tas: xarray.DataArray,


Zeitsperre:
@raquel-ucl just a suggestion, but if you run `pre-commit install`, the environment should be setup to perform the linting hooks for you on commit (it will fail once if changes are made; committing again should succeed). It's less bothersome than having to manually run `black` over your code after every commit.

xclim/indices/_threshold.py
```diff
@@ -68,9 +69,47 @@
...
+def calm_days(
+    pr: xarray.DataArray, thresh: str = "2 m/s", freq: str = "MS"
+) -> xarray.DataArray:
+    r"""Calm days.
+
+    The number of days with average windspeed below threshold.
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
```
Zeitsperre, line 86: I would call this signature `wnd` or `sfcWnd` to not confuse it with precipitation

xclim/indices/_threshold.py
```diff
@@ -1763,6 +1802,43 @@ def sea_ice_extent(
...
+def windy_days(
+    pr: xarray.DataArray, thresh: str = "10.8 m/s", freq: str = "MS"
+) -> xarray.DataArray:
+    r"""Windy days.
+
+    The number of days with average windspeed above threshold.
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
```
Zeitsperre, line 1815: Same issue.

xclim/indices/_threshold.py
raquel-ucl: But shouldn't these methods have the same signature style as others like "dry_days" or "wet_days"?

Zeitsperre: We try to make it so that the signature style (IE: the variable being passed) makes reference to the CMIP variable name that it's examining (for examples: https://pcmdi.llnl.gov/mips/cmip3/variableList.html). Generic indices that work on run-lengths of any value tend to use `da`, temperature-based indicators use `tas`, precipitation uses `pr`, etc.

We have some wind-based indices that consider eastward and northward wind (`uas` and `vas`). Since this is examining wind-speed (not direction), I'd suggest using `sfcWind` (CF-1.0 name: "Near-Surface Wind Speed"; units: "m s-1"; standard name: "wind_speed").

raquel-ucl: Ah, I thought you were referring to the method name part of the signature rather than the parameter name. Sure, `sfcWind` sounds good. I'll update that.

Zeitsperre: My mistake! Thanks!

xclim/indices/_threshold.py
```diff
@@ -1763,6 +1802,43 @@ def sea_ice_extent(
...
+    -----
+    Let :math:`WS_{ij}` be the windspeed at day :math:`i` of period :math:`j`. Then
+    counted is the number of days where:
+
+    .. math::
+
+        WS_{ij} > Threshold [m s-1]
+    """
+    thresh = convert_units_to(thresh, sfcWind)
+    out = threshold_count(sfcWind, ">", thresh, freq)
```
huard, line 1837: I suggest using >= so calm_days + windy_days = total number of days`

xclim/indices/_threshold.py
```diff
@@ -1763,6 +1802,43 @@ def sea_ice_extent(
...
+      Number of days with average near-surface wind speed above threshold.
+
+    Notes
+    -----
+    Let :math:`WS_{ij}` be the windspeed at day :math:`i` of period :math:`j`. Then
+    counted is the number of days where:
+
+    .. math::
+
+        WS_{ij} > Threshold [m s-1]
```
Zeitsperre, line 1834:
```suggestion
        WS_{ij} >= Threshold [m s-1]
```

Zeitsperre (approved):
LGTM

raquel-ucl:
I still need to commit tests for these new indices though!

raquel-ucl:
The [tests for calm/windy days](https://github.com/UCL/xclim/blob/indices/calm_windy_days/xclim/testing/tests/test_indices.py#L1680) are failing and I can't quite tell why, although it's probably a silly mistake on my side somewhere.

I have 2 problems:
* For the windy days, I get an error about the units (speed) not being as expected (temperature). However in the method definition I declared the units to be speed and not temperature, so not sure what's going on in there. [You can inspect the test fail message](https://github.com/Ouranosinc/xclim/pull/777/checks?check_run_id=3422021906#step:5:1107).
* For both calm days, the threshold seems to remain as 10.7999, and that value is not the threshold I'm passing on the tests, and it's also not the default one in the method signature. Where could that be coming from? I have the threshold value printed out locally not to commit debug `print` sentences here.

Any help would be really appreciated.

huard:
@raquel-ucl 
`test_windy_days` call 
```
out = xci.cold_spell_days(da, thresh="12 m s-1", freq="M")
```

Zeitsperre:
> @raquel-ucl
> `test_windy_days` calls
> 
> ```
> out = xci.cold_spell_days(da, thresh="12 m s-1", freq="M")
> ```

I just realized this as well. I'm taking a look at the other test right now.

Zeitsperre:
I think I know the issue. The `sfcWind_series is instantiating a DataArray with `km h-1` as the units. For calm days, 2 km h-1 and 3.1 km h-1 are well below 3 m s-1.

I'm not sure why we're using km h-1 in sfcWind, as I believe the canonical/CMIP units are m s-1. Looking into that now.

Zeitsperre:
@raquel-ucl To fix the test bug, my advice is to simply convert your test units to match those of sfcWind_series. While the standard unit for sfcWind is m s-1, having something a bit different is good for testing pint conversion for wind units. Once that's been updated on your side, we can move ahead with merging this PR.

Thanks again for the contribution!

raquel-ucl:
Whoops, the dangers of copy pasting :) Thanks for spotting that! I've now corrected the call to `windy_days`.

For the unit conversion I just replace "m s-1" to "km h-1". Or should I be using `convert_units_to` instead? If so, would that be in the test itself or the method? 

Last thing: after my changes, I keep on getting this assert failure:

```
>       assert out.units == "km h-1"
E       AssertionError: assert 'd' == 'km h-1'
E         - km h-1
E         + d
```

I don't quite get why that "d" is coming from. There are no temperatures anywhere here. This is happening in both tests.

Zeitsperre:
This is working as intended. We want to collect monthly calm/windy days aggregated by month. The unit then is days (`d`).

xclim/testing/tests/test_indices.py
```diff
@@ -1677,6 +1677,27 @@ def test_1d(self, tasmin_series):
...
 
+class TestWindIndices:
+    def test_calm_days(self, sfcWind_series):
+        a = np.full(365, 20)  # all non-calm days
+        a[10:20] = 2  # non-calm day on default thres, but should count as calm in test
+        a[40:50] = 3.1  # non-calm day on test threshold
+        da = sfcWind_series(a)
+        out = xci.calm_days(da, thresh="3 km h-1", freq="M")
+        np.testing.assert_array_equal(out, [10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
+        assert out.units == "km h-1"
```
Zeitsperre, line 1688:
```suggestion
        assert out.units == "d"
```

xclim/testing/tests/test_indices.py
```diff
@@ -1677,6 +1677,27 @@ def test_1d(self, tasmin_series):
...
+
+    def test_windy_days(self, sfcWind_series):
+        a = np.zeros(365)  # all non-windy days
+        a[10:20] = 10.8  # windy day on default threshold, non-windy in test
+        a[40:50] = 12  # windy day on test threshold
+        a[80:90] = 15  # windy days
+        da = sfcWind_series(a)
+        out = xci.windy_days(da, thresh="12 km h-1", freq="M")
+        np.testing.assert_array_equal(out, [0, 10, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0])
+        assert out.units == "km h-1"
```
Zeitsperre, line 1698:
```suggestion
        assert out.units == "d"
```

Zeitsperre (approved):
All this is missing is an update to the history! Thanks again!

raquel-ucl:
Ah! My colleague @clairbarnes actually mention 'd' could actually be 'days' which is what these methods are counting. Is that correct?

PR #779, opened 2021-07-19
Feature/compound indices

### LINKED ISSUE #765: Implement missing ICCLIM indices
A few icclim indices were never included in early xclim implementation  @bzah indicates he can spend some time on this

List of indices:

- [x] SD : Snow depth
- [ ] CD : Cold and dry days
- [ ] CW: Cold and wet days
- [ ] WD: Warm and dry days
- [ ] WW: Warm and wet days

bzah:
### Pull Request Checklist:
    - This PR fixes #765 

Add a few compound indices based on the mean temperature and the mean precipitation:
- cold_and_dry_days
- warm_and_dry_days
- warm_and_wet_days
- cold_and_wet_days

Nope

### Other information:
There was a comment on the former implementation of cold_and_dry stating:
"There is an issue with the 1 mm threshold. It makes no sense to assume a day with < 1mm is not dry."
According to https://www.ecad.eu/documents/atbd.pdf, to compute the wet percentiles we should ignore all value below 1mm. This make sure all days with precipitation below this threshold are considered "dry".

### Example of use:
```
ds = xr.open_dataset("climpact.sampledata.gridded.1991-2010.nc")
ds_per = ds.sel(time=slice("1991-01-01", "2000-12-31"))

tas_per = percentile_doy(ds_per.tmax, window=5, per=25).sel(percentiles=25)

precip = ds_per.precip.where(ds_per.precip > 1, drop=True)
pr_per = percentile_doy(precip, window=5, per=75).sel(percentiles=75)

return xc.atmos.cold_and_wet_days(
    tas=ds.tmax, tas_25=tas_per, pr=ds.precip, pr_75=pr_per, freq="MS",
)
````

### CODE REVIEW

File: xclim/indices/_multivariate.py [hunk 1]
```diff
@@ -1477,3 +1662,24 @@ def blowing_snow(
...
+def _check_common_sampling_freq(tas, pr):
+    tas_freq = infer_sampling_units(tas)
+    pr_freq = infer_sampling_units(pr)
+    if tas_freq != pr_freq:
+        raise ValidationError(
+            f"The sampling frequencies of the two datasets are different, tas frequency is {tas_freq} whereas pr frequency is {pr_freq}"
+        )
+
+
+def _da_below_per_thrsh(da: xarray.DataArray, da_per: xarray.DataArray):
```
Zeitsperre, line 1676:
```suggestion
def _da_below_percentile_threshold(da: xarray.DataArray, da_per: xarray.DataArray):
```

xclim/indices/_multivariate.py
```diff
@@ -1477,3 +1662,24 @@ def blowing_snow(
...
+    da_per = convert_units_to(da_per, da)
+    thresh = resample_doy(da_per, da)
+    return da < thresh
+
+
+def _da_above_per_thrsh(da: xarray.DataArray, da_per: xarray.DataArray):
```
Zeitsperre, line 1682:
```suggestion
def _da_above_percentile_threshold(da: xarray.DataArray, da_per: xarray.DataArray):
```

xclim/indices/_multivariate.py
```diff
@@ -157,21 +175,188 @@ def cold_and_dry_days(
...
+    Formula to be written [cold_wet_days]_.
+
+    References
+    ----------
+    .. [cold_wet_days] Beniston, M. (2009). Trends in joint quantiles of temperature and precipitation in Europe
+        since 1901 and projected for 2100. Geophysical Research Letters, 36(7). https://doi.org/10.1029/2008GL037119
+    """
+    _check_common_sampling_freq(tas, pr)
+    tg25 = _da_below_per_thrsh(tas, tas_25)
+    pr75 = _da_above_per_thrsh(pr, pr_75)
```
Zeitsperre, line 357:
```suggestion
    tg25 = _da_below_percentile_theshold(tas, tas_25)
    pr75 = _da_above_percentile_theshold(pr, pr_75)
```

xclim/indices/_multivariate.py
```diff
@@ -157,21 +175,188 @@ def cold_and_dry_days(
...
+    Formula to be written [warm_wet_days]_.
+
+    References
+    ----------
+    .. [warm_wet_days] Beniston, M. (2009). Trends in joint quantiles of temperature and precipitation in Europe
+        since 1901 and projected for 2100. Geophysical Research Letters, 36(7). https://doi.org/10.1029/2008GL037119
+    """
+    _check_common_sampling_freq(tas, pr)
+    tg75 = _da_above_per_thrsh(tas, tas_75)
+    pr75 = _da_above_per_thrsh(pr, pr_75)
```
Zeitsperre, line 301:
```suggestion
    tg75 = _da_above_percentile_theshold(tas, tas_75)
    pr75 = _da_above_percentile_theshold(pr, pr_75)
```

xclim/indices/_multivariate.py
```diff
@@ -157,21 +175,188 @@ def cold_and_dry_days(
...
+    Formula to be written [warm_dry_days]_.
+
+    References
+    ----------
+    .. [warm_dry_days] Beniston, M. (2009). Trends in joint quantiles of temperature and precipitation in Europe
+        since 1901 and projected for 2100. Geophysical Research Letters, 36(7). https://doi.org/10.1029/2008GL037119
+    """
+    _check_common_sampling_freq(tas, pr)
+    tg75 = _da_above_per_thrsh(tas, tas_75)
+    pr25 = _da_below_per_thrsh(pr, pr_25)
```
Zeitsperre, line 245:
```suggestion
    tg75 = _da_above_percentile_theshold(tas, tas_75)
    pr25 = _da_below_percentile_theshold(pr, pr_25)
```

xclim/indices/_multivariate.py
```diff
@@ -157,21 +175,188 @@ def cold_and_dry_days(
...
         since 1901 and projected for 2100. Geophysical Research Letters, 36(7). https://doi.org/10.1029/2008GL037119
     """
-    raise NotImplementedError()
-    # There is an issue with the 1 mm threshold. It makes no sense to assume a day with < 1mm is not dry.
-    #
-    # c1 = tas < convert_units_to(tgin25, tas)
-    # c2 = (pr > convert_units_to('1 mm', pr)) * (pr < convert_units_to(wet25, pr))
+    _check_common_sampling_freq(tas, pr)
+    tg25 = _da_below_per_thrsh(tas, tas_25)
+    pr25 = _da_below_per_thrsh(pr, pr_25)
```
Zeitsperre, line 189:
```suggestion
    tg25 = _da_below_percentile_theshold(tas, tas_25)
    pr25 = _da_below_percentile_theshold(pr, pr_25)
```

xclim/indices/_multivariate.py
```diff
@@ -1477,3 +1662,24 @@ def blowing_snow(
...
     return out
+
+
+def _check_common_sampling_freq(tas, pr):
+    tas_freq = infer_sampling_units(tas)
+    pr_freq = infer_sampling_units(pr)
+    if tas_freq != pr_freq:
+        raise ValidationError(
+            f"The sampling frequencies of the two datasets are different, tas frequency is {tas_freq} whereas pr frequency is {pr_freq}"
+        )
```
Zeitsperre, line 1673: I would rewrite this as a decorator for indices. This could serve any multivariate indicator, regardless of variables.

xclim/indices/_multivariate.py
```diff
@@ -1477,3 +1662,24 @@ def blowing_snow(
...
+def _da_below_per_thrsh(da: xarray.DataArray, da_per: xarray.DataArray):
+    da_per = convert_units_to(da_per, da)
+    thresh = resample_doy(da_per, da)
+    return da < thresh
+
+
+def _da_above_per_thrsh(da: xarray.DataArray, da_per: xarray.DataArray):
+    da_per = convert_units_to(da_per, da)
+    thresh = resample_doy(da_per, da)
+    return da > thresh
```
Zeitsperre, line 1685: These two functions seem a bit out of place. These operations are typically written into the code of the indices for clarity. If we went with this pattern, I would want to see it applied to the rest of the percentile-based indices as well (ie:`tg/tn/txXXp`).

Zeitsperre (changes requested):
Thanks for the PR! I'm not so certain about placing the logic of some indicators into helper functions, but I'll let my colleagues weigh in with their thoughts. One pattern we like is to decorate the indices with the input variable checks, so I'd like to see that with `_check_common_sampling_freq`, and have that placed in `xclim.core.utils`.

We should be able to give you more thoughts next week!

xclim/indices/_multivariate.py
bzah: It could be part of datachecks.py no ? I fell like it is the same kind of checking that is done here.
But the checks currently in datachecks are not decorators themself.

xclim/indicators/atmos/_temperature.py
```diff
@@ -72,6 +72,10 @@
     "huglin_index",
     "biologically_effective_degree_days",
     "latitude_temperature_index",
+    "cold_and_dry_days",
```
bzah, line 75: I put these 4 functions in  _temperature.py but as they are related to both temperature and precipitation you may prefer if I move them in another module ?

xclim/indices/_multivariate.py
bzah: I guess it would be better to do the refactoring of tg/tn/tx..p. in a separated PR anyway. So I will just align this code with the current standard and writes all the operations in each indice.

Zeitsperre: Yes, that would be fine for now. Thank you!

xclim/indices/_multivariate.py
aulemahal: I feel these check do not belong in `indices` : indices do not handle anything outside units and computations. This is a choice we made to limit the number of small adjustments and checks like this one in the indices code, making them easier to implement and read.  The down side to this is that when moving those checks to the _indicators_, they need to be more general, less specific to each function and case.

Here, the 4 indices were added to xclim.indicator.icclim and xclim.indicators.atmos. In both case, the base indicator class used is based on `xclim.core.indicator.Daily`. So the check you wanted is already performed (and its even stricter). 

I would simply remove it from here, a mismatch will be the user's problem.
If we really want to keep them, then I guess a decorator declared in datachecks.py would be the most elegant manner.

xclim/indicators/atmos/_temperature.py
aulemahal: Up until now, we put the tas+pr in `_precip.py`. There is no good reason for this, so you can leave them here if you want.

Though, for explicitness, I suggest creating a `PrecipTemp` class:

```python
class PrecipTemp(Daily):
    """Indicators involving temperature and precipitation."""
    context = "hydro"
```

Where the `context` attribute is a (potentially unneeded) flag for the unit handling. Note that this as virtually no real role, other than making the code coherent.

bzah: I will leave them here then. I prefer to let you choose and implement the structural changes which affect the existing code base.

Zeitsperre (approved):

xclim/data/fr.json
```diff
@@ -942,5 +942,29 @@
...
+    "long_name": "Durée totale {freq:f} en jours des périodes chaudes et pluvieuses"
+  },
+  "COLD_AND_WET_DAYS": {
+    "title": "Durée totale en jours des périodes froides et pluvieuses",
+    "abstract": "Durée totale en jours où la température est en dessous du 25eme percentile et les précipitations au dessus du 75eme percentile.",
+    "description": "Durée totale {freq:f} en jours où la température est en dessous du 25eme percentile et les précipitations au dessus du 75eme percentile. Les percentiles sont calculés à partir d'une fenêtre de 5 jours centrée sur chaque jour du calendrier à l'intérieur d'une période de référence. Concernant les precipitations les percentiles ne sont calculés que pour des valeurs au dessus de 1 mm.",
+    "long_name": "Durée totale {freq:f} en jours des périodes froides et pluvieuses"
```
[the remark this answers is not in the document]
aulemahal: Also, in the docstring it is written that the quartiles are _per month_, while here and in the ICCLIM yaml they are _daily_. 

In any case, I removed the percentile explanation from the `description` field because the indicator does not know that. The percentiles are user-provided so they could be computed in many ways, we shouldn't assume how. In theory, that information will be available through the `xclim_history` attribute. And there is no reason the french metadata assumes more than the english anyway ;) !

xclim/indicators/atmos/_temperature.py
aulemahal: I think there was a misunderstanding, I am not suggesting a structural change, only that we add that `PrecipTemp` class and use it for the 4 indicators here. The indicators subclass are supposed to reflect what variable types are involved, and `Temp ` is thus incoherent for indicators taking both `tas` and `pr` as input.

aulemahal (approved):
Nice and clean addition!

A few comments on the french translations and about the indicator class.

xclim/indicators/atmos/_temperature.py
bzah: Ok, I will create PrecipTemp class and move the 4 indices of the Pr in it.
I though you wanted to move all the existing tas+pr indices in `PrecipTemp`, but because I don't know them I was afraid to mess up.

aulemahal (approved):

bzah:
Thanks @Zeitsperre and @aulemahal !

PR #781, opened 2021-07-23
Docstring and call signature modifications

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* cleaner logic in `indices._anuclim._to_quarter()`
* many docstring style corrections
* more statically-typed function call signatures
* consistency in style

No.

### DIFF
--- a/xclim/analog.py
+++ b/xclim/analog.py
@@ -75,7 +75,6 @@ from scipy.spatial import cKDTree as KDTree
 metrics = dict()
 
 
-# FIXME: Kwargs seem not to be passed to any functions.
 def spatial_analogs(
     target: xr.Dataset,
     candidates: xr.Dataset,
@@ -150,6 +149,7 @@ def spatial_analogs(
         vectorize=True,
         dask="parallelized",
         output_dtypes=[float],
+        **kwargs,
     )
     diss.name = "dissimilarity"
     diss.attrs.update(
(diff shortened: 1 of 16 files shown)


Zeitsperre:
@aulemahal Thanks for the review! There's maybe one or two final adjustments to make from what I can see, but I've identified several `# FIXME` issues (mostly unused call signatures). Would it make sense to remove some of these in this PR, or do you think that might potentially break downstream services?

xclim/analog.py
```diff
@@ -72,9 +72,10 @@
 # based on distances. J Stat Planning & Inference 143: 1249-1272
 
 # TODO: Hellinger distance
-metrics = {}
+metrics = dict()
 
 
+# FIXME: Kwargs seem not to be passed to any functions.
```
aulemahal, line 78: Woups. True, this could be fixed here by passing it in the apply_ufunc call.

xclim/indices/_anuclim.py
```diff
@@ -403,6 +403,7 @@ def prcptot_warmcold_quarter(
     return out
 
 
+# FIXME: `src_timestep` is not used here.
```
aulemahal, line 406: These two FIXME could also be fixed here. But if removing the arg creates bugs in the tests, we can wait another PR.

xclim/indices/_multivariate.py
```diff
@@ -117,6 +117,8 @@ def cold_spell_duration_index(
     >>> tn10 = percentile_doy(tasmin, per=10).sel(percentiles=10)
     >>> cold_spell_duration_index(tasmin, tn10)
     """
+    # FIXME: Bootsrap flag is not used. Should be removed.
```
aulemahal, line 120: It is used by the decorator!

xclim/indices/_anuclim.py
Zeitsperre: Alright, let's keep the FIXME's there and wait for another PR.

PR #785, opened 2021-07-28
Fix/bootstrap

### LINKED ISSUE #784: Feedback on bootstrapping
* xclim version: 0.28
* Python version: 3.6
* Operating System: Mac os

### Description

Now that I'm using it, I found a few issues on bootstrapping.

I already made some fixes for 1. and 2. ~I will provide the PR asap.~ #785 

1. *bootstrap* parameter is overridden @percentile_bootstrap
Within @percentile_bootstrap, there is ` ba.arguments.pop("bootstrap", False)`.
The fact that we pop the argument make it (obviously)  absent in the next use of `ba.arguments`. Thus, if we want to check within a call to the indice function if we are within a bootstrap process we can because it default to False.

I would propose to use `OrderedDict::get` instead of `OrderedDict::pop` here. Thus inside the running indice function we  know if the current run is bootstrapped.

2. The `freq` parameter is having a few problems:
Freq is used in bootstrap_func to parse the anchor and to get the years of the sample before bootstrapping, but :
- The default value should be "A" instead of "Y" to accept an anchor
- It should only pickupt anchors compatible with "A"
- It should handle the "S" as in "AS" if it is present
- When there is an anchor such as "-JUL", `bootstrap_year` function is failing with **NotImplemented** due to time blocs being shorter than 365 days. 
- When there is an anchor such as "-JUL" and the above is fixed, the bootstrapping may produce duplicated values in between the years.

3. Bootstrapping should maybe be run only on temperature indices.
On older version of icclim and on climDex, the bootstrapping was run only on temperature indices.
This is because the original paper of Zhang et.al was only mentioning temperatures and not precipitation.
To be honest I don't see why it would not make sens to bootstrap the precipitation but I wanted to give you the information.

bzah:
### Pull Request Checklist:
    - This PR fixes #784

This PR fix some issues with the bootstrapping feature.
Some tests were updated to cover the cases missing especially regarding the frequency.

No.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,12 @@ New indicators
 * ``warm_and_wet_days`` indicator returns the number of days where the mean daily temperature is above the 75th percentile and the mean daily precipitation is above the 75th percentile over period. Added as ``WW`` to ICCLIM module.
 * ``cold_and_wet_days`` indicator returns the number of days where the mean daily temperature is below the 25th percentile and the mean daily precipitation is above the 75th percentile over period. Added as ``CW`` to ICCLIM module.
 
+Bug fixes
+~~~~~~~~~~~~~~
+* Various bug fixes in bootstrapping:
+   - in ``percentile_bootstrap`` decorator, fix the popping of bootstrap argument to propagate in to the function call.
+   - in ``bootstrap_func``, fix some issues with the resampling frequency which was not working when anchored.
+
 0.28.1 (2021-07-29)
 -------------------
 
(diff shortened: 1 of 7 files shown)


huard (approved):

Zeitsperre (approved):
@bzah This is looking good. You just need to update the history before we can merge!

bzah:
Thanks for the review and merge !

Zeitsperre:
Thanks for the contribution!

PR #792, opened 2021-08-13
Add polynomial calculation methods for snowfall approximation based on CLASS

### LINKED ISSUE #667: Snow approximation using the CLASS polynomial behaviour
### Description

Implement CLASS calculation method in `snowfall_approximation`

huard:
### Pull Request Checklist:
    - This PR fixes #667

* Add snowfall estimation methods found in CLASS

Yes. I changed the condition for the binary method to match CLASS behaviour. Precip is now snow when tas <= 0, instead of tas < 0.

### DIFF
--- a/xclim/indices/_conversion.py
+++ b/xclim/indices/_conversion.py
@@ -575,7 +575,7 @@ def snowfall_approximation(
       Mean, maximum, or minimum daily temperature.
     thresh : str,
       Threshold temperature, used by method "binary".
-    method : {"binary"}
+    method : {"binary", "brown", "auer"}
       Which method to use when approximating snowfall from total precipitation. See notes.
 
     Returns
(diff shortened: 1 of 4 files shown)


xclim/indices/_conversion.py
```diff
@@ -585,18 +585,74 @@ def snowfall_approximation(
...
     if method == "binary":
-        prsn = pr.where(tas < thresh, 0)
+        thresh = convert_units_to(thresh, tas)
+        prsn = pr.where(tas <= thresh, 0)
+
+    elif method == "brown":
+        # Freezing point + 2C in the native units
+        upper = convert_units_to(convert_units_to(thresh, "degC") + 2, tas)
+        thresh = convert_units_to(thresh, tas)
+
```
aulemahal, line 612:
```suggestion
        thresh = convert_units_to(thresh, tas)
        upper = convert_units_to("2 degC", tas) + thresh
```
One less `convert_units_to`  calls!

xclim/indices/_conversion.py
```diff
@@ -585,18 +585,74 @@ def snowfall_approximation(
...
+        t = np.concatenate(
+            [
+                [
+                    -273.15,
+                ],
+                np.linspace(0, 6, 100, endpoint=False),
+                [6, 1e10],
+            ]
+        )
+        t = xr.DataArray(t, dims="tas")
```
aulemahal, line 635: I believe `polyval` needs a _coordinate_, so a 1D Dataarray with the same name as its dimension and a coord that is itself.
```suggestion
        t = xr.DataArray(t, dims="tas", name="tas", coords={"tas": t})
```

With this you should be able to pass `xr.polyval(t, coeffs)`

xclim/indices/_conversion.py
```diff
@@ -585,18 +585,74 @@ def snowfall_approximation(
...
+        tmp = xr.DataArray(np.empty_like(t), dims="tas", coords={"tas": t})
+
+        # The polynomial coefficients, valid between thresh and thresh + 6 (defined in CLASS)
+        coeffs = xr.DataArray(
+            [100, 4.6664, -15.038, -1.5089, 2.0399, -0.366, 0.0202],
+            dims=("degree",),
+            coords={"degree": range(7)},
+        )
+
+        fraction = xr.polyval(tmp.tas, coeffs) / 100
```
aulemahal, line 647: When computing `fraction` on my side, I got a few values above 1. Is it normal? Did I make a mistake?

aulemahal (approved):
A few suggestions and questions, but otherwise good to go!

xclim/indices/_conversion.py
huard: Hum, non, si tu transformes 2C en Kelvin, tu vas te retrouver avec upper ~ 500 K

xclim/indices/_conversion.py
huard: Indeed. works. I had tried earlier but without "name".

xclim/indices/_conversion.py
huard: If you're within [0, 6]C, it should be bounded within [0,1].

huard:
Should we add `method` to liquid_precip_ratio, winter_rain_ratio, precip_accumulation ?

huard:
@tlogan2000 FYI, a request dating from last spring from CCCS. Shall we wait for ECCC review before merging ?

tlogan2000:
> @tlogan2000 FYI, a request dating from last spring from CCCS. Shall we wait for ECCC review before merging ?

@huard who at ECCC would want a look? Emilia?  Seems relatively straight forward so I would just merge tbh.  We can have Emilia check the new master an adjust if necessary.

xclim/indices/_conversion.py
aulemahal: Woups. C'est bien vrai tout ça, :upside_down_face: .

xclim/indices/_conversion.py
aulemahal: Sorry alors:
```python3
import xarray as xr
import numpy as np

# Create nodes for the snowfall fraction: -inf, thresh, ..., thresh+6, inf [degC]
t = np.concatenate(
    [
        [
            -273.15,
        ],
        np.linspace(0, 6, 100, endpoint=False),
        [6, 1e10],
    ]
)
t = xr.DataArray(t, dims="tas")

# I can't get polyval to work directly with a DataArray, hence the empty values serving no purpose.
tmp = xr.DataArray(np.empty_like(t), dims="tas", coords={"tas": t})

# The polynomial coefficients, valid between thresh and thresh + 6 (defined in CLASS)
coeffs = xr.DataArray(
    [100, 4.6664, -15.038, -1.5089, 2.0399, -0.366, 0.0202],
    dims=("degree",),
    coords={"degree": range(7)},
)

fraction = xr.polyval(tmp.tas, coeffs) / 100
fraction[0] = 1
fraction[-2:] = 0

fraction.sel(tas=slice(-1, 1)).plot()
plt.axhline(1)
```
index

huard: Ah ouais ! Good catch, there is a line in the fortran code that caps at 1... will add that.

PR #793, opened 2021-08-13
[sdba] Refactor adjustment objects

### LINKED ISSUE #754: sdba - Is OOP really necessary?
I am implementing Cannon's MBCn adjustment and I am confronted to an adjustment object that can not follow the train-adjust scheme : it needs to do both in a single step. Furthermore, I have part of the adjustment that could be performed by an object that doesn't need a train step. This adds to the `ExtremeValues` adjustment object that is re-implementing the `train` and `adjust` method (so like 95% of its base class...).

I think I have painted myself in a corner with this OOP structure that is `BaseAdjustment`. Internally, we still need objects, to remember parameters and to have common methods (like the json parsing stuff).

I am suggesting to move to 2 types of objects (or more if needed):

- Classic Adjustment following the train-adjust scheme, but `train` would become a _class_ method returning an instance that can perform adjustments.
- No-train adjustments : with a single `adjust` _static_ method. No object is ever returned even if the code uses one internally. This is for methods where the training is not separable from the adjust, or where there is enormous computational efficiency to be gained when doing both together.

All current algorithms would loose need for an `__init__`, at least from the point of view of the user. This will divide the kwargs in 2 categories (as it is now, except for `ExtremeValues`). So, for most cases this is merely a merge between the init and the train step, but I think it will ease my implementing and will go in the direction of Patrick's advice : Do no implement "published methods", but provide the modular tools to replicate those.

Example, using `NPDFTransfer` which is part of the MBCn process.
```python
from xclim.sdba import QuantileDeltaMapping, NPDFTransfer

## Before: 
QDM = QuantileDeltaMapping(n_quantiles=10, kind='+', group='time.dayofyear')
QDM.train(ref, hist)
scen = QDM.adjust(sim, interp='linear')

NPDFT = NPDFTransfer() # No params really
# This step doesn't really need both and the training data has the same shape as either, so it's not efficient
NPDFT.train(ref, hist)
scen = NPDFT.adjust(sim)

## After
QDM = QuantileDeltaMapping.train(ref, hist, n_quantiles=10, kind='+', group='time.dayofyear')
scen = QDM.adjust(sim, interp='linear')

scen = NPDFT.adjust(ref, sim)
```

What do our power users @RondeauG  @dgergel and @huard think?

aulemahal:
### Pull Request Checklist:
    - This PR fixes #754 

* Instead of a single `BaseAdjustment` class, Adjustment algorithm inherit from either `SingleStepAdjustment` or `TwoStepAdjustment`.  In the former, only one (class) method is exposed: `cls.adjust(ref, hist, sim, **kwargs) -> scen`. In the latter, two methods: `train` and `adjust`.

Yes. Most adjustment object are now created directly through `train`, there is no need to call the `__init__` before.
So:
```python3
# Before:
DQM = DetrendedQuantileMapping(nquantiles=15, group='time.dayofyear')
DQM.train(ref, hist)
scen = DQM.adjust(sim)

# Now:
DQM = DetrendedQuantileMapping.train(ref, hist, nquantiles=15, group='time.dayofyear')
scen = DQM.adjust(sim)
```

### Other information:
I also made minor changes to the docstrings, moved them to the class instead of the `__init__`. The only algo that doesn't fit the two-step scheme is `NpdfTransfer`, so I also made the `SingleStepAdjustment`, where the object is never actually created, it's only a `adjust` class method.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,16 @@ Announcements
 ~~~~~~~~~~~~~
 * It was found that the `ExtremeValues` adjustment algorithm was not as accurate and stable as first thought. It is now hidden from `xclim.sdba` but can still be accessed via `xclim.sdba.adjustment`, with a warning. Work on improving the algorithm is ongoing, and a better implementation will be in a future version.
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+* The Adjustment classes of ``xclim.sdba`` have been refactored into 2 categories:
+
+    - ``TwoStepAdjustment`` objects (most of the algorithms), which are created **and** trained in the same call:
+      ``obj = Adj.train(ref, hist, **kwargs)``. The ``.adjust`` step stays the same.
+
+    - ``SingleStepAdjustment`` objects (only ``NpdfTransform``), which are never initialized. Their ``adjust``
+      class method performs all the work in one call.
+
 New indicators
 ~~~~~~~~~~~~~~
 * ``effective_growing_degree_days`` indice returns growing degree days using dynamic start and end dates for the growing season (based on Bootsma et al. (2005)). This has also been wrapped as an indicator.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -13,10 +13,10 @@ Breaking changes
 ~~~~~~~~~~~~~~~~
 * The Adjustment classes of ``xclim.sdba`` have been refactored into 2 categories:
 
-    - ``TwoStepAdjustment`` objects (most of the algorithms), which are created **and** trained in the same call:
+    - ``TrainAdjust`` objects (most of the algorithms), which are created **and** trained in the same call:
       ``obj = Adj.train(ref, hist, **kwargs)``. The ``.adjust`` step stays the same.
 
-    - ``SingleStepAdjustment`` objects (only ``NpdfTransform``), which are never initialized. Their ``adjust``
+    - ``Adjust`` objects (only ``NpdfTransform``), which are never initialized. Their ``adjust``
       class method performs all the work in one call.
 
 New indicators
--- a/xclim/sdba/adjustment.py
+++ b/xclim/sdba/adjustment.py
@@ -94,7 +94,7 @@ class BaseAdjustment(ParametrizableWithDataset):
         ):
             warn(
                 "Strange results could be returned when using dayofyear grouping "
-                "on data defined in the proleptic_greogrian calendar "
+                "on data defined in the proleptic_gregorian calendar "
             )
 
     @classmethod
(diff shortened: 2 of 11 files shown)


xclim/sdba/adjustment.py
```diff
@@ -460,116 +493,99 @@ def _adjust(self, sim, interp="nearest", extrapolation="constant"):
         return out.scen
 
 
-class ExtremeValues(BaseAdjustment):
-    """Second order adjustment for extreme values."""
+class ExtremeValues(TwoStepAdjustment):
+    r"""Adjustement correction for extreme values.
```
huard, line 497:
```suggestion
    r"""Adjustment correction for extreme values.
```

HISTORY.rst
```diff
@@ -9,6 +9,16 @@ Announcements
...
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+* The Adjustment classes of ``xclim.sdba`` have been refactored into 2 categories:
+
+    - ``TwoStepAdjustment`` objects (most of the algorithms), which are created **and** trained in the same call:
+      ``obj = Adj.train(ref, hist, **kwargs)``. The ``.adjust`` step stays the same.
+
+    - ``SingleStepAdjustment`` objects (only ``NpdfTransform``), which are never initialized. Their ``adjust``
+      class method performs all the work in one call.
```
huard, line 20: Minor comment, but would it be clearer to call those TrainAdjust and Adjust instead ?

xclim/sdba/adjustment.py
```diff
@@ -173,75 +184,111 @@ def set_dataset(self, ds: xr.Dataset):
...
         raise NotImplementedError
 
     def _adjust(self, sim, **kwargs):
         raise NotImplementedError
 
 
-class EmpiricalQuantileMapping(BaseAdjustment):
-    """Conventional quantile mapping adjustment."""
+class SingleStepAdjustment(BaseAdjustment):
+    """Adjustment with a single step, not fittable into the train-adjust scheme.
```
huard, line 196:
```suggestion
    """Adjustment with no intermediate trained object. 
```

huard (approved):
This new interface is more compact, which I like. It may however be less intuitive for scientists who are not familiar with class methods. I'm wondering how we could mitigate that.

xclim/sdba/adjustment.py
```diff
@@ -49,44 +49,81 @@
...
+                "Inputs are defined on different calendars,"
+                f" this is not supported for {cls.__name__} adjustment."
+            )
+
+        if group.prop == "dayofyear" and (
+            "default" in calendars or "standard" in calendars
+        ):
+            warn(
+                "Strange results could be returned when using dayofyear grouping "
+                "on data defined in the proleptic_greogrian calendar "
```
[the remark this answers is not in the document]
aulemahal: I mean strange results _could_ happen on the greogrian calendar!

aulemahal:
I see that xclim is being used elsewhere! That is great but also tells me that we need to be careful with breaking changes like this.

aaronspring:
> I see that xclim is being used elsewhere! That is great but also tells me that we need to be careful with breaking changes like this. 

Haven't merged yet but good to know you plan to have a stable API. I will just wrap your nice bias correction methods.

aaronspring:
@aulemahal after this PR, I need to provide `group='time`, before it also worked without providing `group`

PR #800, opened 2021-08-18
Catch white space in temperature units

### LINKED ISSUE #543: implement check for whitespace in unit parameters
I accidently entered `'deg C'` as the `thresh` value to an xclim.atmos() call.  

Calculations run but give erroneous values.
example:

```
import xarray as xr
from xclim import atmos
import numpy as np

ds = xr.tutorial.open_dataset('air_temperature')
tasmax = ds.air.resample(time='D').max(dim='time',keep_attrs=True)
d1 = atmos.tx_days_above(tasmax, thresh='10 degC', freq='MS')
d2 = atmos.tx_days_above(tasmax, thresh='10 deg C', freq='MS')
plt.figure()
d1.mean(dim=['lat','lon']).plot(label='Correct')
plt.legend()
d2.mean(dim=['lat','lon']).plot(label='Error')
plt.show()
plt.legend()
```
This is clearly a user error on my part but could likely be a pretty common one.  I think it would be good to try to implement some type of check for this.  I'm not sure what the best method would be... i..e doesn't have to be a whitespace check

Not sure if this could affect input parameters than `thresh`

huard:
### Pull Request Checklist:
    - This PR fixes #543

* Additional custom check. 

No

### DIFF
--- a/xclim/core/units.py
+++ b/xclim/core/units.py
@@ -153,6 +153,19 @@ def units2pint(value: Union[xr.DataArray, str, units.Quantity]) -> units.Unit:
     if unit == "1":
         unit = ""
 
+    if unit.strip() in [
+        "deg C",
+        "degree C",
+        "degree Celsius",
+        "deg K",
+        "degree K",
+        "degree Kelvin",
+        "deg F",
+        "degree F",
+        "degree Fahrenheit",
+    ]:
+        raise ValidationError(f"Remove white space from temperature units: {unit}")
+
     try:  # Pint compatible
         return units.parse_units(unit)
     except (
--- a/xclim/core/units.py
+++ b/xclim/core/units.py
@@ -153,18 +153,24 @@ def units2pint(value: Union[xr.DataArray, str, units.Quantity]) -> units.Unit:
     if unit == "1":
         unit = ""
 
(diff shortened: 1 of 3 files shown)


xclim/core/units.py
```diff
@@ -153,6 +153,19 @@ def _transform(s):
...
+    if unit.strip() in [
+        "deg C",
+        "degree C",
+        "degree Celsius",
+        "deg K",
+        "degree K",
+        "degree Kelvin",
+        "deg F",
+        "degree F",
+        "degree Fahrenheit",
```
Zeitsperre, line 165: I would include "degrees" in this list, alternatively, we could generate this, ie:
```
deg_expected  = ["deg", "degree", "degrees"]
unit_expected  = ["C", "K", "F", "Celsius", "Kelvin", "Fahrenheit", "kelvin"]  # I've seen kelvin without capitalization before.
list_to_examine = [f"{d} {u}" for d in deg_expected for u in unit_expected]
```

Zeitsperre (approved):

xclim/core/units.py
huard: Sure, just go ahead and make the change.

PR #801, opened 2021-08-19
Update history, refactor sdba processing functions, add unit handling in sdba

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Moves unit handling sutff from #791 to here. Inputs of train and adjust methods are converted to common units. ref is used as the reference. Add unit handling in processing : adapt_freq, jitter_under/over_thresh, etc. Add unit handling to the Detrend objects.
* Splits user-facing and metadata-handling function from their computational counterparts when `map_blocks` is used. This applies to `processing.py`. This also changed the signature of the concerned functions, so that they get an explicit one. 
* Generalize the setting of a `xclim_history` attribute through a new `@update_xclim_history` decorator. Decorated functions will have a correctly formatted `xclim_history` attribute added to their first output.

Yes. Signature of `adapt_freq`, `jitter_under_thresh` and others has changed. All variables are given as DataArrays (no more dataset as input) and threshold are given as _quantities_, so strings, not numbers.

### Other information:
We have a naming problem with the `xclim_history` attribute. Currently, the `update_history` function will take the inputs and merge their *history* attributes. Then we take it and put it in `xclim_history`. If a variable has been modified before by xclim, the `xclim_history` attribute is overwritten, not appended. Should we go back to `history` to make it simpler? @Zeitsperre @huard   This applies to  the indicators too!


huard:
Remind me why we decided to have `xclim_history` in the first place.

docs/notebooks/sdba-advanced.ipynb
```diff
@@ -26,9 +26,8 @@
...
-    " \n",
-    "   This is done usually by creating a dask client with a large `n_workers` and a small `threads_per_worker`. The last section of this notebook goes into a bit more details by suggesting a template for large-scale bias-adjustment pipelines.\n",
+    "* Another massive bottleneck of parallelization of xarray is the thread-locking behaviour of some methods. It is quite difficult to isolate and avoid those lockings, so one of the best workaround is to use Dask configurations with many _processes_ and few _threads_. The former do not share memory and thus are not impacted when a lock is acitved from a thread in another worker. However, this adds many memory transfer operations and, by experience, reduces dask's ability to parallelize some pipelines. This is done usually by creating a dask client with a large `n_workers` and a small `threads_per_worker`.\n",
```
huard, line 29:
```suggestion
    "* Another massive bottleneck of parallelization of xarray is the thread-locking behaviour of some methods. It is quite difficult to isolate and avoid those lockings, so one of the best workaround is to use Dask configurations with many _processes_ and few _threads_. The former do not share memory and thus are not impacted when a lock is activated from a thread in another worker. However, this adds many memory transfer operations and, by experience, reduces dask's ability to parallelize some pipelines. This is done usually by creating a dask client with a large `n_workers` and a small `threads_per_worker`.\n",
```
This is done ... what is this ?

docs/notebooks/sdba.ipynb
```diff
@@ -233,8 +229,8 @@
    "outputs": [],
    "source": [
     "# 2nd try with adapt_freq\n",
-    "ds_ad = sdba.processing.adapt_freq(xr.Dataset(dict(sim=pr_hist, ref=pr_ref, thresh=0.05)), group='time')\n",
-    "QM_ad = sdba.EmpiricalQuantileMapping.train(pr_ref, ds_ad.sim_ad, nquantiles=15, kind='*', group='time')\n",
+    "sim_ad, pth, dP0 = sdba.processing.adapt_freq(pr_ref, pr_sim, thresh=\"0.05 mm d-1\", group='time')\n",
```
huard, line 232: I think the API is a lot more consistent now.

xclim/core/formatting.py
```diff
@@ -321,6 +323,84 @@ def update_history(
     return merged_history
 
 
+def update_xclim_history(func):
```
huard, line 326: Will this play well with the cf-xarray tracking mechanism ?

xclim/core/formatting.py
```diff
@@ -321,6 +323,84 @@ def update_history(
...
+            new_name=out.name,
+            **da_dict,
+        )
+        out.attrs["xclim_history"] = attr
+        return outs
+
+    return _call_and_add_history
+
+
+def gen_call_string(funcname: str, *args, **kwargs):
```
huard, line 364: Ah nice ! This will be useful for provenance work as well.

huard: Open question: should we pass the func object instead so it can be inspected ? The advantage I see is that we could bind *args and print input parameters as **kwargs for clarity.

xclim/sdba/__init__.py
```diff
@@ -56,10 +56,25 @@
 Most methods acting on grouped data also accept a `window` int argument to pad the groups with data from adjacent ones.
 Units of `window` are the sampling frequency of the main grouping dimension (usually `time`). For more complex grouping,
 one can pass a :py:class:`xclim.sdba.base.Grouper` directly.
+
+Notes for developers
+====================
+To be scalable and performant, the sdba module makes use of the special decorators :py:func`xclim.sdba.base.map_blocks`
+and :py:func:`xclim.sdba.base.map_groups`. However, they have the inconvenient that functions wrapped by them are unable
+to manag xarray attributes (including units) correctly and their signatures are sometime wrong and often unclear. For
```
huard, line 64:
```suggestion
to manage xarray attributes (including units) correctly and their signatures are sometime wrong and often unclear. For
```

xclim/sdba/_processing.py
```diff
@@ -0,0 +1,146 @@
...
+
+@map_groups(reduces=[Grouper.PROP], data=[])
+def _normalize(
+    ds: xr.Dataset,
+    *,
+    dim: str,
+    kind: str = ADDITIVE,
+) -> xr.Dataset:
+    """Normalize an array by removing its mean.
+    Normalization if performed group-wise.
```
huard, line 99:
```suggestion
    Normalization is performed group-wise.
```

huard (approved):
How confident are you that our test coverage would have picked potential errors introduced in this PR? 
From experience, it's easy to break unit handling, and tests were not initially covered to check this. You added unit related checks, but how tight are we overall ?

docs/notebooks/sdba-advanced.ipynb
aulemahal: "This" is "Dask configurations with many _processes_ and few _threads_."
I'll add an example.

xclim/core/formatting.py
aulemahal: I haven't studied that thing much, but I feel we are not aiming for the same goal.
`cf-xarray`'s solution is a callback executed _every time_ an _operation_ is performed. 
Here, the history is only updated after a function, so a set of regrouped operation. Something that cf-xarray won't be able to catch.

Will it play well : no. This will _set_ the attribute to something constructed from the call string + the history *of the inputs*. Thus, if cf-xarray modifies anything in the same attribute within the function, that would be discarded.

xclim/core/formatting.py
aulemahal: haha. I began writing this taking in the function (or the signature), but realized there were no useful supplementary information?

One thing I tried to do was to split "positional-only" from "keyword" arguments according to the signature (instead of whether they are in *args or **kwargs). But I wasn't convinced that was interesting enough to justify needing a function.

Also, in the Indicator, we replace the _function_ name with the indicator's ID. It was easier to do this if the name was passed.

aulemahal:
We decided to use `xclim_history` because CF conventions say that `history` is a _dataset_ attribute, but we are always dealing with DataArrays.

As for the tests, I did add more, but you are right that it's always a tricky question. Failures and bugs in the attributes will not always trigger test failures... I plan to add a bit more.

xclim/core/formatting.py
huard: Got it.

xclim/core/formatting.py
huard: What if we want to print the full path to the function, e.g. `xclim.atmost.growing_degree_days` ? Imagine we have two `average` indicators in two different modules, that could get confusing.

aulemahal: Well, this would confirm my choice of taking a "funcname" as input! So the calling function could decide on what name to use. In the indicator, using the "registry_id" fulfills the need you are indicating since it is "unique".

aulemahal: And I could update `update_xclim_history` to prepend  `func.__module__` to solve this in sdba too.

aulemahal:
@Zeitsperre  I'd like to hear your opinion on a change from `xclim_history` to `history`!

I think this PR as revealed a _faille_ in our logic of being CF-orthodox and not using `history` : If xclim performs successive actions on a variable, only the latest is recorded in the `xclim_history` attribute. Moreover, it might append the input's `history`,  as if that latest action was the only one.
 I would suggest going back to `history` as it is the one that makes the most sense and is understood by the largest crowd. And it would solve the bug. It would be only slightly non-CF, as the `history` attribute is not flagged as a variable attribute by the conventions.

Zeitsperre:
> @Zeitsperre I'd like to hear your opinion on a change from `xclim_history` to `history`!
 
> I think this PR as revealed a _faille_ in our logic of being CF-orthodox and not using `history` : If xclim performs successive actions on a variable, only the latest is recorded in the `xclim_history` attribute. Moreover, it might append the input's `history`, as if that latest action was the only one.

This is true. We should be at the very least appending some general summary of actions taken to the `history` attribute. If we want to revisit this, my feeling is that the `xclim_history` should be used for debugging purposes, IE: add the call signatures used, xclim version, whether the data was Indicator-compliant. This should have some internal provenance-like information that the general user doesn't need (and doesn't know to look for). 

> I would suggest going back to `history` as it is the one that makes the most sense and is understood by the largest crowd. And it would solve the bug. It would be only slightly non-CF, as the `history` attribute is not flagged as a variable attribute by the conventions.

From my understanding, because it isn't flagged as a variable attribute means that we are free to do it. Try adding a `history` and `xclim_history` to a variable in a NetCDF and run it through a [CF-Checker](https://pumatest.nerc.ac.uk/cgi-bin/cf-checker.pl). there shouldn't be any errors.

Zeitsperre:
Very small point, but the Notes docstring for `NpdfTransform` could use a small fix while you're fixing things up: https://xclim.readthedocs.io/en/stable/sdba_api.html#xclim.sdba.adjustment.NpdfTransform

aulemahal:
So I changed "xclim_history" to "history" throughout xclim!
@tlogan2000  (so you're in the loop too).

This is not super-breaking, but kinda. 
The output of an indicator computation before:
```
xclim_history : [2021-07-07 18:46:28] growing_degree_days: GROWING_DEGREE_DAYS(tas=<array>, thresh='10.0 degC', freq='YS') - xclim version: 0.28.0.
```
now:
```
history :  [2021-08-23 14:05:45] growing_degree_days: GROWING_DEGREE_DAYS(tas=tas, thresh='10.0 degC', freq='YS') - xclim version: 0.28.2-beta.
```
If the indicator was from a virtual module, the module name would be appended in the call string (ex: `icclim.CD(...)`).

Applies also to ensembles percentiles and stats, `percentile_doy`, and a few others.

aulemahal:
@Zeitsperre 
The CF-checker returns the following on the file produced by `usage.ipynb`:
```
CHECKING NetCDF FILE: /tmp/27024.nc
=====================
Using CF Checker Version 4.1.0
Checking against CF Version CF-1.8
Using Standard Name Table Version 77 (2021-01-19T13:38:50Z)
Using Area Type Table Version 10 (23 June 2020)
Using Standardized Region Name Table Version 4 (18 December 2018)

WARN: (2.6.1): No 'Conventions' attribute present

------------------
Checking variable: time
------------------
WARN: (3): No standard_name or long_name attribute specified

------------------
Checking variable: lat
------------------

------------------
Checking variable: lon
------------------

------------------
Checking variable: growing_degree_days
------------------
INFO: attribute history is being used in a non-standard way
WARN: (7.3): Coordinate variable time should have bounds or climatology attribute
WARN: (7.3): Coordinate variable time should have bounds or climatology attribute

ERRORS detected: 0
WARNINGS given: 4
INFORMATION messages: 1
```

Zeitsperre:
That's exactly what we should expect. LGTM.

aulemahal:
I'm turning crazy soon. The sdba notebook compiles with a failure on the first try, but passes on the second try...

Zeitsperre:
> I'm turning crazy soon. The sdba notebook compiles with a failure on the first try, but passes on the second try...

This actually sounds insane, by Einstein's definition anyway. I'll give it a try as well.

Zeitsperre:
Works for me on the first try. It may have to do with the `compute()` call. Jupyter has a habit of running through to the next cell before asynchronous processes are finished. Could it be that?

aulemahal:
:exploding_head:  So I copied only this example in a new script and ran it 7 times. At the start, I deleted the xclim testing data and the script's output. Between trial 3 and 4, I modified the code to have some more info in the traceback (the algo was untouched).

1. Fail
2. Fail
3. Pass
4. Fail
5. Pass
6. Pass
7. Fail

Funny, eh?

Zeitsperre:
What is going on?! This is really strange behaviour. Any information from the trace back (is it different from the `doctests` build)?

aulemahal:
I could get the following info:

- It always fails at iteration 0 of `npdf_transform`.
- It doesn't fail for _all_ chunks. There are three chunks in the data, so `npdf_transform` is executed thrice.
- It always fails at line `scensp = ADJ.adjust(simp, **kwargs["adj_kws"])` and shows that `simp` has no attributes at all, while `histp` and `refp` had some. And even though `sim` has some and `xr.set_options(keep_attrs=True)` was used.

Thus, it seems that, at random moments, the last matrix multiplication of `npdf_transform` doesn't respect the context under which it is executed and attrs are not copied.

aulemahal:
Added an explicit setting of units to all ADJ inputs. And a comment to note that it is weird.
Before I had a 6/10 failure rate. Now I have a 0/3 failure rate. THUS, it works.

aulemahal:
Now, that's more interesting, a real error! My bad.

PR #802, opened 2021-08-20
[#797] Update linear interpolation of percentiles

### LINKED ISSUE #797: Feature: Have other linear interpolation for percentiles
* xclim version: 0.28
* Python version: 3.x
* Operating System: Mac os

### Description
Follow up to the discussion with @huard.
In other climate indices libraries such as climdex and icclim, the percentile interpolation method used is the 8th method of[ Hyndman and Fan](https://www.amherst.edu/media/view/129116/original/Sample+Quantiles.pdf). Both library have their own C/C++ implementation.
In Xclim, the default interpolation of numpy is used, which correspond to the 7th method of Hyndman and Fan.
However **numpy does not** implement the 8th method (see https://github.com/numpy/numpy/issues/10736#issuecomment-897851315).

It would be great if xclim:
- Use the Scipy [mquantiles method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mquantiles.html?highlight=mquantiles) instead of `np.percentile`
- Have a parameter to choose the interpolation in `calendar.percentile_doy`

As usual, I can work on this issue.

bzah:
### Pull Request Checklist:
    - This PR fixes #797 

Update `_calc_perc` and `percentile_doy` functions to allow more linear interpolation methods.

I'm not sure how impactful it is but `_calc_perc` now return a numpy masked array.


bzah:
A remark about the failing tests here.
I think it is quite counter productive to have two places where the same things get tested.

For example, the test from `test_indices.TestTGXN10p.test_tg10p_simple` and  the one from `test_temperature.TestT10p.test_tg10p_simple` are both testing that tg10p is properly computed.
IMO, the test of `atmo.tg10p` should only ensure that the `Temp` object is properly created with the expected properties.

aulemahal:
@bzah good point! Indicators need to be tested, but there's no point to repeat the same tests. Recently, I think we went with this guideline:
- Indice tests (`test_indices.py`) use idealized data and really test the mathematical computation. (like edgecases, presence or not of missing data).
- Indicator tests (`test_{var kind}.py` use "real" data (usually `atmos_ds`, the ERA5 extraction) and test minimally for values, but add tests for attributes and stuff like that.

bzah:
I'm getting really good results with my implementation, I still have to write some unit tests, some documentation and clean up a bit but the PR will be reviewable soon.

bzah:
I ended up completely rewriting numpy functions ::percentile and ::nanpercentile

It would be really appreciated if you can give me your insights on this implementation.
However I will also try to submit it to numpy because:
 - both function show better performances, especially for nanpercentiles which is up to x50 in some cases.
 - they are, IMO, much more readable than current implementations.
 - they add some error correction using floating point epsilon (based on the R implementation)

So I understand if you prefer to wait for the PR on numpy before commenting.

I also wrote a script to compare performance an assert the non regression (up to 14th decimal):
https://gist.github.com/bzah/2a84d050b8a1aed1b40a2ed1526e1f12
In the gist comment you can find a report of some test runs.

huard (commented):
I suggest we prep this PR for merging, irrespective of what NumPy is doing. There are impressive speed-ups for nans that are worth having ASAP. 

I suggest adding the reference to the Hyndman paper in the docs, and add some guidance to users on which values should be chosen for which occasion. 

`calc_percentiles` loops over `_quantile` for each percentile. Is there any way to avoid that ? Once `arr` is sorted, I would think finding other percentiles is quick. I'm hopeful the slow-down you measure can be reduced by passing the array of percentiles to `_quantile`. I suspect numpy folks will make the same comment.

bzah:
@huard 

> calc_percentiles loops over _quantile for each percentile. Is there any way to avoid that ? Once arr is sorted, I would think finding other percentiles is quick. I'm hopeful the slow-down you measure can be reduced by passing the array of percentiles to _quantile. I suspect numpy folks will make the same comment.

I was expecting discussion on this .
Yes, I agree with you it would improve the performances when there are multiple quantiles, and it's more numpy style anyway. However, I find it much easier to manipulate a scalar to understand the code and the performance are not even that bad even with 100 quantiles.

I will try to see how I can make it work with a np.array.

About the doc, once in numpy the API will probably be slightly different. 
Because I'm lazy, I wanted to write docs only on numpy but I can start it here and reuse it I suppose.

bzah:
@huard  I did something that work but I'm pretty sure it is not optimal.
I tested with a smallish array (100, 180, 30), iterated 10 times and on 1000 percentiles (from 0.1 to 99.9, step 0.1).
I have these results:
```
nanpercentile: 19.904847238 secs
NEW nan_calc_percentiles: 22.494946018 secs
OLD nan_calc_percentiles: 50.227164944 secs
```
I suppose I'm not doing it properly, it should perform better than nanpercentile.
I'm open to suggestion if someone has a better idea.

I'm coming back Tuesday, I will try to improve it then.

bzah:
About the test failure:
The failing test is in *test_adjustment.py*, I'm not 100% sure but I don't think I broke it with this feature.
Moreover the test works fine locally, is there something I'm missing ?

The error is`AttributeError: 'DataArray' object has no attribute 'units'` on *TestDQM::test_mon_U[True-*-pr]  *

A part from this error, I think the feature is ready for review.

HISTORY.rst
```diff
@@ -197,7 +203,7 @@ Internal Changes
 
 Announcements
 ~~~~~~~~~~~~~
-* `xclim` no longer supports Python3.6. Code conventions and new features from Python3.7 (`PEP 537 <https://www.python.org/dev/peps/pep-0537/#features-for-3-7>`_) are now accepted.
+* `xclim` no longer supports Python3.6. Code conventions and neures from Python3.7 (`PEP 537 <https://www.python.org/dev/peps/pep-0537/#features-for-3-7>`_) are now accepted.
```
bzah, line 206: oops

aulemahal:
> The failing test is in test_adjustment.py

Not your fault! This test shows a seemingly random bug in dask...

bzah:
FYI I created the draft PR on numpy: https://github.com/numpy/numpy/pull/19857
I did quite a few changes to fit the needs of numpy API, I don't believe they are necessary in Xclim.

bzah:
I will remove the `fuzz` part, it may causes license issues because it is inspired by R and climdex which have GPL licenses.

HISTORY.rst
```diff
@@ -20,6 +20,12 @@ Internal Changes
 ~~~~~~~~~~~~~~~~
 * `xclim` code quality checks now use the newest `black` (v21.8-beta). Checks launched via `tox` and `pre-commit` now run formatting modifications over Jupyter notebooks found under `docs`.
 
+New features and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* utils.py::nan_calc_percentiles now uses a custom algorithm instead of `numpy::nanpercentiles` to have more flexibility on the interpolation method. The performances are also improved. *
+* calendar.py::percentile_doy now uses the 8th method of Hyndman&Fan for linear interpolation where alpha == beta == 1/3. This is to be consistant with other climate indices packages (climdex, icclim) *
```
huard, line 26:
```suggestion
* ``calendar.py::percentile_doy`` now uses the 8th method of Hyndman & Fan for linear interpolation (alpha = beta = 1/3). Previously, the function used Numpy's percentile, which corresponds to the 7th method. This change is motivated by the fact that the 8th is recommended by Hyndman & Fay and it ensures consistency with other climate indices packages (climdex, icclim).
```

HISTORY.rst
```diff
@@ -20,6 +20,12 @@ Internal Changes
 ~~~~~~~~~~~~~~~~
 * `xclim` code quality checks now use the newest `black` (v21.8-beta). Checks launched via `tox` and `pre-commit` now run formatting modifications over Jupyter notebooks found under `docs`.
 
+New features and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* utils.py::nan_calc_percentiles now uses a custom algorithm instead of `numpy::nanpercentiles` to have more flexibility on the interpolation method. The performances are also improved. *
```
huard, line 25:
```suggestion
* ``utils.py::nan_calc_percentiles`` now uses a custom algorithm instead of ``numpy::nanpercentiles`` to have more flexibility on the interpolation method. The performance is also improved. 
```

HISTORY.rst
```diff
@@ -20,6 +20,12 @@ Internal Changes
...
+* calendar.py::percentile_doy now uses the 8th method of Hyndman&Fan for linear interpolation where alpha == beta == 1/3. This is to be consistant with other climate indices packages (climdex, icclim) *
+* utils.py::_cal_perc is now only a proxy for utils.py::nan_calc_percentiles with some axis moves *
```
huard, line 27:
```suggestion
* ``utils.py::_cal_perc`` is now only a proxy for ``utils.py::nan_calc_percentiles`` with some axis moves.
```
and move to Internal changes section.

xclim/core/calendar.py
```diff
@@ -444,12 +444,15 @@ def days_in_year(year: int, calendar: str = "default") -> int:
...
+    per: Union[float, Sequence[float]] = 10.0,
+    alpha: float = 1.0 / 3.0,
+    beta: float = 1.0 / 3.0,
 ) -> xr.DataArray:
     """Percentile value for each day of the year.
 
     Return the climatological percentile over a moving window around each day of the year.
-    All NaNs are skipped.
+
+    Default value for alpha and beta gives type 8 interpolation method of Hyndman&Fan.
```
huard, line 455:
```suggestion
    Different quantile estimators can be used by specifying `alpha` and `beta` according to specifications given by [Hyndman&Fan]_. The default definition corresponds to method 8, which meets multiple desirable statistical properties for sample quantiles. Note that `numpy.percentile` corresponds to method 7, with alpha and beta set to 1.
```

xclim/core/calendar.py
```diff
@@ -459,12 +462,17 @@ def percentile_doy(
       Number of time-steps around each day of the year to include in the calculation.
     per : float or sequence of floats
       Percentile(s) between [0, 100]
+    alpha: float
+        used with beta to express the linear interpolation wanted.
+    beta: float
+        used with alpha to express the linear interpolation wanted.
```
huard, line 468:
```suggestion
    alpha: float
        Plotting position parameter.
    beta: float
        Plotting position parameter.
```

xclim/core/calendar.py
```diff
@@ -459,12 +462,17 @@ def percentile_doy(
...
 
     Returns
     -------
     xr.DataArray
       The percentiles indexed by the day of the year.
       For calendars with 366 days, percentiles of doys 1-365 are interpolated to the 1-366 range.
+      Nan will be masked
```
huard, line 475:
```suggestion
      Nan will be masked
      
    References
    ----------
    .. [Hyndman&Fan] Hyndman, R. J., & Fan, Y. (1996). Sample quantiles in statistical packages. The American Statistician, 50(4), 361-365.  
```

xclim/core/utils.py
```diff
@@ -208,37 +208,153 @@ def uses_dask(da):
...
-        ).ravel()
-    return out
+
+def virtual_index_formula(
+    array_size: Union[int, np.array], quantiles: np.array, alpha: float, beta: float
+) -> np.array:
+    """
+    Compute the floating point indexes of an array for the linear interpolation of quantiles.
+
+    Note:
```
huard, line 247:
```suggestion
    Notes
    -----
```

xclim/core/utils.py
```diff
@@ -208,37 +208,153 @@ def uses_dask(da):
...
+        return np.broadcast_to(result, (quantiles.size,) + result.shape)
+    # The dimensions of `q` are prepended to the output shape, so we need the
+    # axis being sampled from `ap` to be first.
+    arr = np.moveaxis(arr, axis, 0)
+    # nan_count is not a scalar
+    nan_count = np.isnan(arr).sum(0).astype(float)
+    valid_values_count = values_count - nan_count
+    # We need at least two values to do an interpolation
+    too_few_values = valid_values_count < 2
+    if too_few_values.any():
```
huard, line 302: Don't you need to call `any` on a specific axis ?

xclim/testing/tests/test_precip.py
```diff
@@ -383,10 +383,10 @@ def test_days_over_precip_thresh():
     per = percentile_doy(pr, window=5, per=80)
 
     out1 = atmos.days_over_precip_thresh(pr, per)
-    np.testing.assert_array_equal(out1[1, :, 0], np.array([87, 63, 74, 79]))
+    np.testing.assert_array_equal(out1[1, :, 0], np.array([81, 61, 69, 78]))
 
     out2 = atmos.days_over_precip_thresh(pr, per, thresh="2 mm/d")
-    np.testing.assert_array_equal(out2[1, :, 0], np.array([87, 63, 71, 79]))
+    np.testing.assert_array_equal(out2[1, :, 0], np.array([81, 61, 66, 78]))
```
huard, line 389: So there can be quite significant differences in results. Interesting.

xclim/testing/tests/test_utils.py
```diff
@@ -90,3 +95,54 @@ def test_ensure_chunk_size():
     assert out.chunks[0] == (3, 3, 3, 3, 3, 5)
     assert out.chunks[1] == (10, 11)
     assert out.chunks[2] == (20,)
+
+
+class Test_nan_calc_percentiles:
+    def test_calc_perc_type7(self):
+        # Exemple array from: https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method
+        arr = np.asarray([15.0, 20.0, 35.0, 40.0, 50.0])
+        res = nan_calc_percentiles(arr, percentiles=[40.0])
```
huard, line 104:
```suggestion
        res = nan_calc_percentiles(arr, percentiles=[40.0], alpha=1, beta=1)
```
I know these are the default values, but just for clarity.

huard (approved):
Nice !

xclim/core/utils.py
bzah: valid_values_count is derivative from the  valid_values_count which himself comes from a scalar (value_count) and a reduced array (nan_count) so I don't think we need an axis parameter.

However it might be a code smell that the execution flow is hard to follow and it needs some refactoring.

bzah:
I did quite a lot of changes in the numpy PR which greatly simplify the code of _quantile.
However, **the performance improvement has been dropped** because it was not part of the initial issue in numpy.
Thus, the xclim version performs much better but:
1. it does not handle some cases such as:
   - when dtype is Fraction (object) or bool
   - using `out` parameter to mutate an existing array
   - using `keepdims` to reshape the result to the input dimensions
   - using  `overwrite_input` to qvoid copying the input array
   - having other interpolation options
2. It is very hard to read (thus maintain) `_nan_quantile` because I did the refactoring only on numpy.

My main issue is that I don't know if the performance improvement will be merged or not on numpy.
And I think the boost is too good to be left behind, especially when I think about bootstrapping which call percentile_doy very often.

So, ideally, I would like keep the xclim implementation and refactor it, but to do it only once I'm done with numpy's version to avoid going back and forth between the two.
Is that okay for you ?

huard:
@bzah Is this ready to be merged ?

bzah:
I have not yet done the refactoring because I'm waiting for a final review on the numpy's PR.
If you would like to include this asap, it can be merged as is and I will refactor in a future PR. 
The behavior will not change.

aulemahal (approved):
I didn't dive in the `nan_calc_percentiles` code. This is great! Clearly, `numpy` is a better place for the function, but xclim can serve as a first home. 

A few nitpicky comments.

xclim/core/calendar.py
```diff
@@ -444,12 +444,14 @@ def days_in_year(year: int, calendar: str = "default") -> int:
...
-    per: Union[float, Sequence[float]] = 10,
+    per: Union[float, Sequence[float]] = 10.0,
+    alpha: float = 1.0 / 3.0,
+    beta: float = 1.0 / 3.0,
 ) -> xr.DataArray:
     """Percentile value for each day of the year.
 
     Return the climatological percentile over a moving window around each day of the year.
-    All NaNs are skipped.
+    Different quantile estimators can be used by specifying `alpha` and `beta` according to specifications given by [Hyndman&Fan]_. The default definition corresponds to method 8, which meets multiple desirable statistical properties for sample quantiles. Note that `numpy.percentile` corresponds to method 7, with alpha and beta set to 1.
```
aulemahal, line 454:
```suggestion
    Different quantile estimators can be used by specifying `alpha` and `beta` according to specifications given by [HyndmanFan]_. The default definition corresponds to method 8, which meets multiple desirable statistical properties for sample quantiles. Note that `numpy.percentile` corresponds to method 7, with alpha and beta set to 1.
```
Sphinx seems to not like the &.

huard:
I suggest to merge now, and open a new refactoring issue later in the future.

PR #806, opened 2021-08-23
Better `get_file` logic in xclim.testing._utils

### LINKED ISSUE #805: Xclim testing data gathering utilities do not perform origin branch md5 checks
@huard ran into this problem today, but I have in the past as well.

The logic for gathering xclim's test data checks to see if there is a local file and, if one is found, delivers that file via `xclim.testing.open_dataset()`. If there has been any upstream changes to the testing data, this is not checked for.

In order to evade introducing versioning to the xclim testing data local store, we need to at least attempt to retrieve the upstream md5 checksum file to compare it against our local file as an added step.

Zeitsperre:
…tion

### Pull Request Checklist:
    - This PR fixes #805 

* `xclim.testing.open_dataset()` now performs a md5 checksum file URL retrieve with every call.
  * If the local file does not match the upstream md5 checksum *and the file is deleted*.
  * If network/repo access is unavailable, a warning is shown that validity cannot be determined and continues.
  * If everything checks out, business as usual.

The comparison with upstream md5 will lead to destruction of existing local data if the files are stale/obsolete. Breaking, but in a good way?


aulemahal (approved):
Nice! I trust this code, as there aren't many ways to test it right now...

Zeitsperre:
The only way that comes to mind is to run a test while inhibiting URLrequest calls. I've seen it done before, but it might require a requests/socket mocking library (like this: https://github.com/miketheman/pytest-socket).

It works locally?

aulemahal:
> but it might require a requests/socket mocking library 

Bah. I say :crossed_fingers: .

PR #808, opened 2021-08-24
Add index argument to rle functions

huard:
See #775 for details.


AUTHORS.rst
```diff
@@ -24,3 +24,5 @@ Contributors
 * Jwen Fai Low `@jwenfai <https://github.com/jwenfai>`_
 * Marie-Pier Labonté `@marielabonte <https://github.com/marielabonte>`
 * Jeremy Fyke `@jeremyfyke https://github.com/jeremyfyke`_
+* Clair Barnes <clair.barnes.16@ucl.ac.uk> @clairbarnes <https://github.com/clairbarnes>`_
+* Raquel Alegre <raquel.alegre@gmail.com> @raquel-ucl <https://github.com/raquel-ucl>`_
```
Zeitsperre, line 28:
```suggestion
* Clair Barnes <clair.barnes.16@ucl.ac.uk> `@clairbarnes <https://github.com/clairbarnes>`_
* Raquel Alegre <raquel.alegre@gmail.com> `@raquel-ucl <https://github.com/raquel-ucl>`_
```

xclim/indices/run_length.py
```diff
@@ -12,7 +12,6 @@
 from warnings import warn
 
 import numpy as np
-import xarray
```
Zeitsperre, line 15: If at any point there are docstring examples introduced here, we'll need to resort to using `import xarray`. This is because the `xr` used in the docstrings is a reference to `import xcim.testing as xr` injected in the doctest namespace. Not an issue here right now, but an FYI.

huard: Got it, thanks.

Issue #810, opened 2021-08-25
Usage question: How to use multiple members as samples in `sdba`?


aaronspring:
### Description
(Using xr object names as in https://xclim.readthedocs.io/en/latest/notebooks/sdba.html)
I would like to use several ensemble members of `hist` to create the forecast model distribution in `sdba`.
In my understand, when `hist` has two members as dimensions, then each member is bias corrected according to its own members distribution. This makes sense for spatial dimensions `lon` and `lat`.

Do you see a way to use `member` and `time` combined? I tried stacking, but this doesnt work with `group.month`.

### What I Did and What I Received
https://gist.github.com/aaronspring/a8018af34d5e91803e0e028babfb34d6

aaronspring:
Great that this is to-be-again included: `group = sdba.Grouper('time', add_dims=['member'])` I will use this once ready.

PR #817, opened 2021-08-26
Fix window and extra dimension issues in map_blocks

### LINKED ISSUE #816: [sdba] Bug in `map_blocks` for `window` and coordinate-less dimensions.
* xclim version: 0.28.4

The recent addition of handling of `Grouper.add_dims` has a bug when a window is used : `map_blocks` expects it on the output. 
Also, there is a raise when new dimensions do not have coordinates since `map_blocks` is unable to infer the final size of that new dimension. But, that's not true if the dimension already appears on one of the inputs, there should be a way around.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #816

* Removes `"window"` from `Grouper.add_dims`. Doing this, `map_blocks` doesn't consider it when constructing the template. A call to  `Grouper.apply` still adds it, but as all this happens within the computation in `map_blocks`, the decorator doesn't need to be notified.
* Add a unit setting in `dqm_adjust`. This shoudln't be needed, but a fail seems to happen randomly when it's not there. Same thing as discussed in #810...
* When `map_blocks` encounters a "new dimension" that is in fact already present on one of the inputs BUT doesn't have a coordinate, it doesn't fail anymore.
* Added a few more tests...

No.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -187,27 +187,68 @@ The method we use is as follows::
 Deploying
 ---------
 
-A reminder for the maintainers on how to deploy.
+A reminder for the maintainers on how to prepare the library for a tagged version.
+
 Make sure all your changes are committed (**including an entry in HISTORY.rst**).
 Then run::
 
-$ bumpversion <option> # possible options: major / minor / patch / release
-$ git push
-$ git push --tags
+    $ bumpversion <option> # possible options: major / minor / patch / release
+    $ git add *
+    $ git commit -m "Bumped version to v1.2.3-XYZ"
+
+For PyPI releases/stable versions, ensure that the last version bumping command run is `$ bumpversion release`.
+These changes can now be merged to the main development branch::
+
+    $ git push
+
+With this performed, we can tag a version that will act as the GitHub-provided stable source archive::
+
+    $ git tag v1.2.3-XYZ
+    $ git push --tags
 
 Packaging
 ---------
 
-When test coverage and stability is adequate, maintainers should update the pip-installable package (wheel) on PyPI.
-In order to do this, you will need the following libraries installed:
+When a new version has been minted (features have been successfully integrated test coverage and stability is adequate),
+maintainers should update the pip-installable package (wheel and source release) on PyPI as well as the binary on conda-forge.
 
-* twine
-* setuptools
-* wheel
+TThe simple approach
+~~~~~~~~~~~~~~~~~~~
 
-.. TODO::
+The simplest approach to packaging for general support (pip wheels) requires the following packages installed:
+ * setuptools
+ * wheel
+ * twine
 
-    Finish the packaging documentation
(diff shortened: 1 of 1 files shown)


xclim/testing/tests/test_sdba/test_adjustment.py
```diff
@@ -477,6 +477,39 @@ def test_mon_U(self, mon_series, series, mon_triangular, kind, name):
...
+        hist = dsim.sel(time=slice("1981", "2010"))
+        sim = dsim.sel(time=slice("2041", "2070"))
+
+        # With add_dims, "does it run" test
+        group = Grouper("time.dayofyear", window=5, add_dims=["location"])
+        EQM = EmpiricalQuantileMapping.train(ref, hist, group=group)
+        EQM.adjust(sim).load()
+
+        # Without, sanity test.
+        group = Grouper("time.dayofyear", window=5)
```
aaronspring, line 508: Is it a breaking change now that window can only be passed inside Grouper? @aulemahal

aaronspring: I.e. shouldn’t that be also mentioned in History and the docstrings

aulemahal: Did this break? On my side, EQM works by passing them as kwargs.
In theory, both ways should work as whenever you can pass `group` as a string, you should also be able to pass `window` and `add_dims`  as extra kwargs. In my tests and code, I simply prefer to create the `Grouper` object ahead and reuse it, but we made the code so both should be possible.

aaronspring: I had to pass window in Group now which worked before outside group but I also used add_dims, so it might be due to the combination

aulemahal: Sorry about that. With the current master, I was able to create EQM, DQM, QDM, LOCI and Scaling with:
```
adj = ADJ.train(ref, hist, group='time.dayofyear', window=10, add_dims=['location'], **kwargs)
ad.group
# Grouper(add_dims=['location'], name='time.dayofyear', window=10, interp=False)
```
The kwargs parsing stuff happens before the actual training is called (even before the Adjustment object is created). 

What does your failing code look like?

aulemahal: Oh I realize now that what might have changed is the behaviour is `group` is not given! 
I see this is a breaking change and we should have noted it, sorry!

But, `window > 1` has no meaning if `group='time'`, no?  I mean it will work and results will be different than without, but it doesn't do anything statistically significative. For `add_dims`, though, this is a a clear loss.

aaronspring: I think I was confused about how to use Group instead of group and where to put the kwargs like window or add_dims. I wrongly thought that they need to be in Group but that isn’t necessary.

Issue #538, opened 2020-09-01
anuclim add docs / refactor / add indicators (?)


tlogan2000:
Currently the mapped anuclim module is not seen anywhere on the xclim docs.  These calculations are popular enough that I think an doc page would be warranted.

- [ ] Add doc page similar to ICCLIM module https://xclim.readthedocs.io/en/stable/icclim.html

- [ ] Refactor `input_freq` to match choice of decision here: https://github.com/Ouranosinc/xclim/pull/531#issuecomment-684821386

Discussion point :
Also while many anuclim calculation are simple partial mappings of existing xclim.indices there a few new methods in `indices._anuclim.py` - Should these be added to the `atmos` module?  This might be tricky as some allow for the flexible input frequencies (daily, weekly, or monthly) and would require work to accommodate checks of the Indicator class etc..

aulemahal:
The documentation of anuclim is here : https://xclim.readthedocs.io/en/stable/indicators.html#anuclim-indices
Since the coming of virtual submodules.

I suggest (héhé) removing `src_timestep`, as we can get this info with `xr.infer_freq`.

However, currently, anuclim _indicators_ are constructed using `Daily` as a base, thus can not accept weekly and monthly inputs. 

I guess this is the last part of this issue?

[ ] Allow weekly and monthly inputs for anuclim indices.

PR #823, opened 2021-08-30
Fix rate2amount, remove src_timestep in anuclim

### LINKED ISSUE #814: rate2amount assumes a month or year has a set number of seconds
* xclim version: 0.28.3-beta

### Description

`rate2amount` works well when time units are constant, but does not account for the fact that months or years can have a variable number of seconds. 

```
pr = pr_series(np.random.rand(1000))
pram = rate2amount(pr)  # Multiplies pr by number of seconds per day
a1 = pram.resample(time="MS").sum(dim="time", keep_attrs=True)
a2 = rate2amount(pr.resample(time="MS").mean(dim="time", keep_attrs=True))
```

No actual results are affected, because in xclim every usage is on daily data. 

I think we need to special case month and yr time units.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #814 and goes in the direction of #538.

* The "long" method for  `rate2amount` is now taken when the sampling frequency is monthly, seasonal or yearly. In those cases, the function is now exact even on the last element. The "S" anchor is also taken into account, or rather its absence : end-of-period frequencies are handled differently.
* `amount2rate` added for the inverse operation.
* All anuclim indices have their `src_timestep` removed.
* `infer_sampling_units` will now raise on 'M' and 'Y' frequencies as their units are non-constant.

Yes, anuclim doesn't accept the `src_timestep` argument. This constrains the time coordinate slightly more (for the indices, the indicators were already strict because of  `check_daily`).


xclim/core/units.py
```diff
@@ -506,13 +506,43 @@ def rate2amount(
     >>> pram.values
     array([7.00008327e-18, 1.63335276e-17, 1.63335276e-17])
     """
+    m = 1
+    u = None  # Default to assume a non-uniform axis
+    label = "lower"
+    time = rate[dim]
+
     try:
-        m, u = infer_sampling_units(rate.time, deffreq=None)
```
huard, line 510: Je suggère de raiser une erreur dans `infer_sampling_units` si base dans M, Q, A pour éviter de se faire prendre plus tard.

huard (approved):

aulemahal:
@huard I ran into problems when removing 'M' and 'Y' from `infer_sampling_units` because we had an indice using this. So I created `amount2rate` to solve that problem.

@tlogan2000 This change in `rate2amount` broke the anuclim tests (not the indices themselves), but I decided to go on and remove all `src_timestep` instances in this PR. Also the small changes in the tests' expected values are because 1972 has 366 days.

aulemahal:
I could merge. But I really would like @huard  to review the new changes in `units.py` (amount2rate and rate2amount). And it would also be nice if @tlogan2000 had time to throw a quick eye on the changes in anuclim, just to confirm it stills look like expected.

tlogan2000:
> I could merge. But I really would like @huard to review the new changes in `units.py` (amount2rate and rate2amount). And it would also be nice if @tlogan2000 had time to throw a quick eye on the changes in anuclim, just to confirm it stills look like expected.

Ok sorry have been occupied with other things the past week or so

HISTORY.rst
```diff
@@ -5,10 +5,21 @@ History
 0.30.0 (unreleased)
 -------------------
 
-Fixes
-~~~~~
+Breaking changes
+~~~~~~~~~~~~~~~~
+* All "Anuclim" indices and indicators have lost their ``src_timestep`` argument. Most of them were already not using it and now every function infers the frequency from the data directly. This may add stricter constraints on the time coordinate, the same as for :py:func:``xr.infer_freq``.
```
huard, line 10:
```suggestion
* All "Anuclim" indices and indicators have lost their ``src_timestep`` argument. Most of them were not using it and now every function infers the frequency from the data directly. This may add stricter constraints on the time coordinate, the same as for :py:func:``xr.infer_freq``.
```

xclim/core/units.py
```diff
@@ -461,6 +484,74 @@ def to_agg_units(
...
+            time.diff(dim, label=label)
+            .reindex({dim: da[dim]}, method="ffill")
+            .astype(float)
+        )
+        dt = dt / 1e9  # Convert to seconds
+
+        if to == "amount":
+            tu = (str2pint(da.units) * str2pint("s")).to_reduced_units()
+            out = da * dt * tu.m
+        elif to == "rate":
```
huard, line 536: You have an if/elif here and a if/else farther down to case switch on `to`. If there are only two options, it could be a bool. If you want to keep the str argument for clarity, then I suggest using if/elif/else.

xclim/testing/tests/test_indices.py
```diff
@@ -1365,62 +1365,51 @@ def test_simple(self, pr_series):
...
+        out = xci.prcptot_wetdry_quarter(p_weekly, op="wettest")
         np.testing.assert_array_almost_equal(out, [241, 241])
-        out = xci.prcptot_wetdry_quarter(p_weekly, op="driest", src_timestep="W")
+        out = xci.prcptot_wetdry_quarter(p_weekly, op="driest")
         np.testing.assert_array_almost_equal(out, [60, 60])
 
-        p_month = xci.precip_accumulation(a, freq="MS")
-        p_month.attrs["units"] = "mm month-1"
-        out = xci.prcptot_wetdry_quarter(p_month, op="wettest", src_timestep="M")
+        # Can't use precip_accumulation cause "month" is not a constant unit
```
tlogan2000, line 1400: I tried the following and get only tiny differences  (1e-14): 
````
p_month = a.resample(time="MS").mean(keep_attrs=True)
p_month_amt = xci.precip_accumulation(a, freq="MS")
out = xci.prcptot_wetdry_quarter(p_month, op="wettest")
out1 =  xci.prcptot_wetdry_quarter(amount2rate(p_month), op="wettest")
np.testing.assert_array_equal(out, out1)  # error but really small
````

However pushing the limits with more data I see larger differences ( could have made a mistake I suppose) but I think these should result in the same value: 

````
np.random.seed(43)
a = np.random.uniform(low=0, high=35, size=365*50)  # ~50 yrs of random precip between 0, 35 mm)
a = pr_series(a, start="1971-01-01", units="mm/d")
for freq in ['MS', '3MS', 'YS', '5YS', '10YS', '20YS' ,'30YS']:
      print(freq)
       a1 = a.resample(time=freq).mean(keep_attrs=True)
       a2 = amount2rate(xci.precip_accumulation(a, freq=freq), out_units=a1.units)
       print(np.abs(a1-a2).max().values)

File: xclim/testing/tests/test_indices.py [hunk 5]
```diff
@@ -1547,67 +1522,59 @@ def test_Celsius(self, tas_series):
...
     @pytest.mark.parametrize(
-        "freq,units,expected",
+        "freq,expected",
         [
-            (("D", "D"), "mm/day", [435.0, 442.0]),
-            (("7D", "W"), "mm/week", [441.0, 436.0]),
-            (("MS", "M"), "mm/month", [435.0, 442.0]),
+            ("D", [435.0, 443.0]),
+            ("7D", [441.0, 485.0]),
+            ("MS", [435.0, 443.0]),
```
huard, line 1545: Why are the values different ?

huard (approved):

tlogan2000 (changes requested):

xclim/testing/tests/test_indices.py
aulemahal: Because I added a day to the timeseries. See above, before it was 2x 365 days, but 1972 is a leap year, so the timeseries would end on Dec. 30, which I didn't like. I made it 731 days and voilà.

xclim/testing/tests/test_indices.py
tlogan2000: Ok makes sense.

tlogan2000 (approved):

PR #849, opened 2021-09-27
Stronger yaml structure

### LINKED ISSUE #828: A stronger YAML system for virtual indicators
Now that the virtual submodule system has been tested in a few in-house cases, and before it is used extensively, I think we need to rethink some of it.

Issues:

1. No validation. Extra keys do not trigger errors. We have real use case where one entry was misplaced and resulted in wrong indicators.
2. Very relax specification. In order to accept clix-meta yaml as-is AND to have a specification that was closer to xclim's internals, the code in `Indicator.from_dict` is quite complex.

Ideas:

1. Explore [yamale](https://github.com/23andMe/Yamale) or [schema](https://github.com/keleshev/schema) for YAML validation. The first as the interesting feature that the whole spec would sit in a yaml file. xclim wouldn't need to depend on a extra package, but would provide the tools needed so users can perform the validation, if they so wish. The second has more users.

2. Rethink the whole yaml spec and make it strict. Make it so it mirrors the structure of indicator objects better. Then, add a function that translates clix-meta yaml to this xclim yaml. We don't need this runtime-compatibility with clix-meta, we only need some way of updating xclim when clix-meta  is updated.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #828 

* Refactor of the yaml/dict parsing to simplify it for both users and developers. Instead of a mix between following clix-meta's syntax and opening up for xclim-specific/-simplified things, the YAML/dict specification is now stricter.
* `xclim.core.utils.adapt_clix_meta_yaml` translated the clix-meta yml files to xclim-compliant ones. A similar function is here : https://gist.github.com/aulemahal/80f5ab7f614e479a4e0c5a7eacfecd8f for yml files made with previous versions of xclim.

My goal was to make the yaml layout similar to that of the `Indicator` object itself. As the main aspect was to remove all the multiple ways to specify things and keep a single one, I can't show a perfect example, I invite reviewers to look at the changes to the yml files in `xclim/data` instead.

I tried to follow this plan to go from yaml to indicator module:

- `build_indicator_module_from_yaml` reads in the files (yml, py, json), cycles through the dicts in the `indices` section of the yaml. It sets the default according to module-wide attributes and parses the `base` and `compute` keys if they refer to something "local" (a base class from the same module, a compute function from the passed `py` file).
- `Indicator.from_dict` parses the compute function, the injected parameters, the modified parameters. Grosso modo, it performs the actions we would do within the call to the indicator's `__init__` in `xclim/indicators/*/*.py`.
- `Indicator.__new__` does the rest, as before.

This PR removes a lot of parsing from `from_dict` since the yaml now imitates `Indicator`.

Also, this add the possibility to check one's yaml with `yamale`. A schema is included in xclim (data/schema.yml) and `yamale` is added as a dev dep for testing.

Yes. All previously written yaml virtual module will break. I linked a function from my gist above to help convert them (the function is not complete, it might fail).

Also, I renamed `Indicator.cf_attrs` to `Indicator.output`. This is not the best name, I know, but it acknowledges that one can now pass non-cf attrs if they wish. I also removed the option to access attributes directly on the indicator. I.e.: `ind.standard_name` will now fail in most cases. It was a semi-working option before, with undocumented behaviour. One should use `ind.output[0]['standard_name']` instead. I expect discussion about this part.

### DIFF
--- a/docs/notebooks/example.py
+++ b/docs/notebooks/example.py
@@ -4,8 +4,10 @@ from xclim.core.units import declare_units, rate2amount
 
 
 @declare_units(pr="[precipitation]")
-def extreme_precip_accumulation(pr: xr.DataArray, perc: float = 95, freq: str = "YS"):
-    """Total precipitation accumulation during extreme events.
+def extreme_precip_accumulation_and_days(
+    pr: xr.DataArray, perc: float = 95, freq: str = "YS"
+):
+    """Total precipitation accumulation during extreme events and numbr of days of such precipitation.
 
     The `perc` percentile of the precipitation (including all values, not in a day-of-year manner)
     is computed. Then, for each period, the days where `pr` is above the threshold are accumulated,
@@ -24,11 +26,17 @@ def extreme_precip_accumulation(pr: xr.DataArray, perc: float = 95, freq: str = 
     -------
     xarray.DataArray
       Precipitation accumulated during events where pr was above the {perc}th percentile of the whole series.
+    xarray.DataArray
+      Number of days where pr was above the {perc}th percentile of the whole series.
     """
-    pr_thresh = pr.quantile(perc / 100, dim="time")
+    pr_thresh = pr.quantile(perc / 100, dim="time").drop_vars("quantile")
 
-    pr_extreme = rate2amount(pr).where(pr >= pr_thresh)
+    extreme_days = pr >= pr_thresh
+    pr_extreme = rate2amount(pr).where(extreme_days)
 
-    out = pr_extreme.resample(time=freq).sum().drop_vars("quantile")
-    out.attrs["units"] = pr_extreme.units
-    return out
+    out1 = pr_extreme.resample(time=freq).sum()
+    out1.attrs["units"] = pr_extreme.units
+
+    out2 = extreme_days.resample(time=freq).sum()
+    out2.attrs["units"] = "days"
+    return out1, out2
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -15,7 +15,7 @@ Dictionary and YAML parser
 
 To construct indicators dynamically, xclim can also use dictionaries and parse them from YAML files.
 This is especially useful for generating whole indicator "submodules" from files.
-This functionality is based on and extends the work of [clix-meta](https://github.com/clix-meta/clix-meta/).
+This functionality is inspired by the work of [clix-meta](https://github.com/clix-meta/clix-meta/).
 
 YAML file structure
 ~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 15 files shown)


xclim/core/indicator.py
```diff
@@ -318,11 +277,12 @@ class Indicator(IndicatorRegistrar):
...
     """
 
-    cf_attrs: Sequence[Mapping[str, Any]]
+    output: Sequence[Mapping[str, Any]] = None
     """A list of metadata information for each output of the indicator.
 
        It minimally contains a "var_name" entry, and may contain : "standard_name", "long_name",
-       "units", "cell_methods", "description" and "comment".
+       "units", "cell_methods", "description" and "comment" on official xclim indicators. Other
+       fields could also be present if the indicator was created from outside xclim.
```
aulemahal, line 285: We can discuss the name of this attribute.  I want it to be the same as the section in the yaml.

`cf_attrs` is false because we can now accept non-cf attrs (current example: `proposed_standard_name` from clix-meta indices).

xclim/core/indicator.py
```diff
@@ -454,157 +459,77 @@ def from_dict(
...
+                if cls is None:
+                    raise ValueError(
+                        f"Requested base class {data['base']} is neither in the indicators registry nor in base classes registry."
+                    )
+            else:
+                cls = data["base"]
+        elif cls is Indicator:
+            raise ValueError(
+                "Indicators can't be created from dict with the Indicator class as base. Please use an existing indicator as base, or another base class."
+            )
```
aulemahal, line 478: This is born from a need to use `from_dict` outside a yaml file, but without knowing before hand which class shoud be subclassed. It's strange to change the `cls` var, but I haven't found another clean way...

xclim/core/indicator.py
```diff
@@ -1107,6 +1037,10 @@ def datacheck(**das):  # noqa
             datachecks.check_freq(da, "H")
 
 
+base_registry["Hourly"] = Hourly
+base_registry["Daily"] = Daily
```
aulemahal, line 1041: Because of the need of `Indicator.from_dict` to be able to parse the  `base` string, we need to have a registry of base classes that are incomplete indicators (i.e. are not in the main registry).

docs/notebooks/example.py
```diff
@@ -4,8 +4,10 @@
 
 
 @declare_units(pr="[precipitation]")
-def extreme_precip_accumulation(pr: xr.DataArray, perc: float = 95, freq: str = "YS"):
-    """Total precipitation accumulation during extreme events.
+def extreme_precip_accumulation_and_days(
+    pr: xr.DataArray, perc: float = 95, freq: str = "YS"
+):
+    """Total precipitation accumulation during extreme events and numbr of days of such precipitation.
```
huard, line 10:
```suggestion
    """Total precipitation accumulation during extreme events and number of days of such precipitation.
```

xclim/core/indicator.py
huard: - nc_attrs 
- metadata
- var_attrs

xclim/core/indicator.py
```diff
@@ -1384,32 +1306,58 @@ def build_indicator_module_from_yaml(
 
     # Module-wide default values for some attributes
     defkwargs = {
-        # Other default argument, only given in case the indicator definition does not give them.
+        # Only usedin case the indicator definition does not give them.
```
huard, line 1309:
```suggestion
        # Only used in case the indicator definition does not give them.
```

huard (approved):
Just did a superficial review. Agree output name should be changed.

tlogan2000 (approved):
I think this is much more intuitive.  One of the first tasks that Juliette will have is to make a yaml module for the CCDP project.  If this is ready to merge today or tomorrow I can get her to work from master in order to test drive the new config and we can adjust if necessary? ..

huard:
I suggest we make the 0.30 release without this change, then merge it right after so it stays in master for a while, given people living on the bleeding edge some time to adjust.

tlogan2000:
> I suggest we make the 0.30 release without this change, then merge it right after so it stays in master for a while, given people living on the bleeding edge some time to adjust.

Ok.  Juliette can work straight from the `stronger-yaml` branch if need be

aulemahal:
Ok, at first I wanted this asap, but that might be simply to risky.

release 0.30 and then merge seems right and that can be fast: we could release today.

xclim/core/indicator.py
aulemahal: I'd go with "metadata" or "attrs".

What does @Zeitsperre think?  This is the object with the unformatted attributes of the indicator's output. It is a list of dictionaries. This is also the name that will be used in the yaml.

huard:
I missed it. Will have to think about this. This is fairly big breakage.

aulemahal:
So the `getattr` solution was quite easy to implement. With the last commit, all attributes in `self._cf_names` can be accessed as attributes, but they _always_ return a list of strings. This is different from before, where the returned value was the same a passed in the class constructor (usually a single string). I can do the trick where single-output indicator return strings and multi-output return a list if we want a minimal breaking change.

xclim/core/indicator.py
aulemahal: Thinking of it `metadata` doesn't carry the distinction between output variable attributes and general metadata. (Like `title` vs `long_name`, `abstract` vs `description`) And neither does `attrs`.

Zeitsperre: If I understand correctly, I would maybe go with `inherited_attrs`?

aulemahal: I think you are technically correct, the attrs are inherited from the indice. However, I feel that this notion will be alien for the user. 
The branch currently uses "var_attrs", which convey the meaning I want it to convey, but I am open for more discussion, it's not my favorite.

huard: Why again are we ditching cf_attrs ? Because one new attribute is not strictly_required by the CF convention ? I don't feel this is a strong enough reason to break things.

aulemahal: well. I realize that now. 
At first it's because I didn't want `cf_attrs` as the section name in the yaml, but now  `var_attrs`  is not drastically different, so why change? I am also ok with accepting non-cf attrs in `cf_attrs`.
Sorry for the noise!

huard: Pas de trouble, content quand ça arrive ici aux autres !

aulemahal:
Last commits changed the behaviour of `Indicator.standard_name`: it now returns a string when there is only one output.

aulemahal:
This is ready to merge.

huard:
I suggest we merge ASAP and make sure it gets some real-world use before the next release.

aulemahal:
Squasing because of the large number of commits and the back-and-forth.

PR #873, opened 2021-10-15
Refactor indicator parameters

### LINKED ISSUE #855: Bug: Exception raised when using positional arguments for some indicators
* xclim version: 0.30.0
* Python version: 3.x
* Operating System: mac os

### Description

When using positional arguments, computing of some indicators fail.
I think it is the indices wrapped by `wrapped_partial`.
`wrapped_partial` has a different signature than the targeted indice function and in Indicator class, `ba = self._sig.bind(*args, **kwds)` seems to dislike that. I'm not sure though.

If enforcing keyword arguments is intended, I could not find this information in the doc.

### What I Did

```
atmos.tropical_nights( da, "22 degC", "MS" )
```

### What I Received
<details>
  <summary>TypeError: too many positional arguments</summary>

  ```
  icclim/tests/unit_tests/test_indicies.py:99 (Test_TR.test_tr_default_threshold[True])
self = <icclim.tests.unit_tests.test_indicies.Test_TR object at 0x1622670a0>
use_dask = True

    def test_tr_default_threshold(self, use_dask):
        ds = Dataset()
        ds["tas"] = stub_tas(value=26 + 273.15, use_dask=use_dask)
        ds.tas[:5] = 0
        conf = IndiceConfig(
            ds=ds,
            slice_mode=Frequency.MONTH,
            var_name=["tas"],
            netcdf_version=NetcdfVersion.NETCDF4,
        )
>       res = tr(conf)

test_indicies.py:111: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../eca_indices.py:151: in tr
    return atmos.tropical_nights(
/Users/aoun/.conda/envs/icclim/lib/python3.9/site-packages/xclim/core/indicator.py:637: in __call__
    das, params, all_params = self._parse_variables_from_call(args, kwds)
/Users/aoun/.conda/envs/icclim/lib/python3.9/site-packages/xclim/core/indicator.py:727: in _parse_variables_from_call
    ba = self._sig.bind(*args, **kwds)
/Users/aoun/.conda/envs/icclim/lib/python3.9/inspect.py:3062: in bind
    return self._bind(args, kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Signature (tasmin: Union[str, xarray.core.dataarray.DataArray] = 'tasmin', *, thresh: str = '20 degC', freq: str = 'YS', ds: xarray.core.dataset.Dataset = None)>
args = (<xarray.DataArray 'tas' (time: 1826, lat: 1, lon: 1)>
array([[[  0.  ]],

       [[  0.  ]],

       [[  0.  ]],

   ...  * time     (time) datetime64[ns] 2042-01-01 2042-01-02 ... 2046-12-31
Attributes:
    units:    K, '20.0 degC', 'MS')
kwargs = {}

    def _bind(self, args, kwargs, *, partial=False):
        """Private method. Don't use directly."""
    
        arguments = {}
    
        parameters = iter(self.parameters.values())
        parameters_ex = ()
        arg_vals = iter(args)
    
        while True:
            # Let's iterate through the positional arguments and corresponding
            # parameters
            try:
                arg_val = next(arg_vals)
            except StopIteration:
                # No more positional arguments
                try:
                    param = next(parameters)
                except StopIteration:
                    # No more parameters. That's it. Just need to check that
                    # we have no `kwargs` after this while loop
                    break
                else:
                    if param.kind == _VAR_POSITIONAL:
                        # That's OK, just empty *args.  Let's start parsing
                        # kwargs
                        break
                    elif param.name in kwargs:
                        if param.kind == _POSITIONAL_ONLY:
                            msg = '{arg!r} parameter is positional only, ' \
                                  'but was passed as a keyword'
                            msg = msg.format(arg=param.name)
                            raise TypeError(msg) from None
                        parameters_ex = (param,)
                        break
                    elif (param.kind == _VAR_KEYWORD or
                                                param.default is not _empty):
                        # That's fine too - we have a default value for this
                        # parameter.  So, lets start parsing `kwargs`, starting
                        # with the current parameter
                        parameters_ex = (param,)
                        break
                    else:
                        # No default, not VAR_KEYWORD, not VAR_POSITIONAL,
                        # not in `kwargs`
                        if partial:
                            parameters_ex = (param,)
                            break
                        else:
                            msg = 'missing a required argument: {arg!r}'
                            msg = msg.format(arg=param.name)
                            raise TypeError(msg) from None
            else:
                # We have a positional argument to process
                try:
                    param = next(parameters)
                except StopIteration:
                    raise TypeError('too many positional arguments') from None
                else:
                    if param.kind in (_VAR_KEYWORD, _KEYWORD_ONLY):
                        # Looks like we have no parameter for this positional
                        # argument
>                       raise TypeError(
                            'too many positional arguments') from None
E                       TypeError: too many positional arguments

/Users/aoun/.conda/envs/icclim/lib/python3.9/inspect.py:2988: TypeError


  ```

</details>

### LINKED ISSUE #857: Harmonize compute function parameter handling
* xclim version: 0.30

### Description

It is possible to customize `compute` functions using the `wrapped_partial` utility function, but there is a second mechanism that also modifies the compute signature in `indicators._parse_indice` to : 

- add the `ds` argument and allow `str` values instead of DataArrays
- modify defaults as specified by the YAML file 

My suggestion would be to consolidate the "signature modifying logic" to one function, or indicator method. So for example, the Indicator class could take arguments such as: 
- `compute_new_defaults`
- `conpute_frozen_params`

A second related issue is the fact that the signature of indicators varies depending on whether or not `compute` is defined using `wrapped_partial`. `wrapped_partial`, due to its use of `partial`, converts all parameters following a `suggested` or `fixed` parameters into KEYWORD_ONLY parameters. If we want a consistent signature across indicators, one option would be to make all non DataArray arguments KEYWORD_ONLY. I would do this within the logic above, and leave indices as is.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #855, fixes #857

* When an indicator modifies or inject the parameters of its underlying indice, this is now done by passing a `parameters` arg to the constructor. `wrapped_partial` is thus deprecated and should not be used anymore.
    - Inside the indicator, the `compute` function is not modified anymore, except if we need to wrap it with `declare_units` (happens when using generic function directly, like in clix-meta).
    - The `input` mapping of the yaml modules is now a part of the Indicator construction : this allows to rename input arguments and fill their metadata with CMIP6-standard stuff.
   
#### New possibilities

* Injected parameter can be modified : If basing an indicator on another one that injected a parameter in its compute function, that parameter can simply be re-injected. In the yaml files, this was possible but required repeating the `compute` field : this is not needd anymore.
* Changing an input variable _name_ is now possible. Ex: passing `input={'da': 'tasmin'}` will make `tasmin` appear in the indicator's parameters and signature, with correct units and all. Even more, one can pass : `input={'tas': 'tasmin'}`, so a tas-based indicator is now tasmin-based! (Only feasible if the dimensionality doesn't change).

Yes.

* The signature of indicators is now systematically constructed : input variables (optional or not) are positional or keyword arguments and *all* other parameters are keyword only.
* Some indicators have modified signatures because we now _rename_ variables when wrapping generic indices. This is the case for the whole `cf` module, for example.
* `wrapped_partial` should not be used anymore. Pass the parameters in `parameters` instead.
* `Indicator.parameters` now includes the injected parameters, which it didn't before. If one wants to iterates on non-injected parameters, one can use `Indicator.iter_parameters()`, yielding pairs of `(name, meta)`, where meta is always dictionary.

Moreover, the PR adds many security checks when constructing indicators:

1. `ValueError("Malformed dosctring [...]")` when the docstring lists parameters absent from the compute function.
2. `ValueError("Parameter XXX was passed [...]")` when a parameter passed explicitly in `parameters` doesn't exist in the compute function.
3. `ValueError("Compute argument XXX was mapped to YYY, which is not understood [...]")` when `input` contains a mapping to a variable absent from `xclim.core.utils.VARIABLES` (`data/variables.yml`).
4. `ValueError("When changing the name of a variable [...]")` when `input` tries to map an existing variable (that has units) to another one with different units.
5. `ValueError("[...] is missing expected units")` when a [optional] variable argument is missing units, i.e. a generic function was passed but units were not inserted by passing `input` or `parameters`.

I have the feeling that 1 is too strong, because it means the docstring must be correctly formatted, which was not a requirement before, simply a bonus. Could be changed to a warning? Removed completely?

EDIT: I began this PR as a branch of `stronger-yaml`, but the PR related to this one was squashed before merging, thus explaining the very large amount of commits here. I'll squash this one too.


xclim/core/formatting.py
```diff
@@ -477,29 +478,29 @@ def _gen_parameters_section(parameters, allowed_periods=None):
 
     Parameters
     ----------
-    parameters : generator, pairs of name and parameters' dict
-      Parameters dictionary (`Ind.iter_parameters()`).
+    parameters :
```
huard, line 481:
```suggestion
    parameters : mapping
```

xclim/core/indicator.py
```diff
@@ -139,6 +141,65 @@
 _indicators_registry = defaultdict(list)  # Private instance registry
 
 
+class empty_param:
```
huard, line 144: camel case ?

xclim/core/indicator.py
```diff
@@ -139,6 +141,65 @@
...
+    """Class for storing an indicator's controllable parameter.
+
+    For retrocompatibility, this class implements a "getitem" and a special "contains".
+
+    Example
+    -------
+    >>> p = Parameter(InputKind.NUMBER, default=2, description='A simple number')
+    >>> p.units  # has not been set
+    <class 'xclim.core.indicator.empty_param'>
+    >>> 'units' in p  # Easier/retrocompatible way to test is units are set
```
huard, line 161:
```suggestion
    >>> 'units' in p  # Easier/retrocompatible way to test if units are set
```

xclim/core/indicator.py
huard: `empty_param = object()`

huard (approved):

aulemahal:
@huard  Was there a reason you preferred an object rather than a class, for `_empty` ? I tried using an object and it fails down the line because some function take a copy of `parameters`. An `param_copy.units is _empty` fails, since it's a copy. Using a class (like `inspect` does), doesn't have this problem.

Issue #843, opened 2021-09-15
Proposal of a drought duration climate index


yrouranos:
* xclim version: 0.25.0
* Python version: 3.7.6
* Operating System: LinuxMint 20.1

### Description
A drought duration climate index was developed (currently available in /scen_workflow_afr/indices.py). The algorithm is currently being tested (using climate data from West Africa). This index can be used to identify the impacts of climate change in countries where a monsoon is present.

### What I Did
A call to the developed function returns an instance of `xarray.DataArray` which contains the number of days per year that are comprised in a dry period. 2 methods were implemented to define what a dry period is (based on a daily or total precipitation amount). Parameters allow adjusting detection rules in different countries or regions.

drydurtot ('1d' method): A dry period is a sequence of _D_threshold_ consecutive days during which the daily precipitation amount (_P_day_) is lower than a threshold value (_P_threshold_).

drydurtot ('tot' method): A dry period is a sequence of _D_threshold_ consecutive days during which the total amount of precipitation (_P_total_) is lower than a threshold value (_P_threshold_).

In both cases, the user can specify the days of year to consider (between day DOY_min and DOY_max). 

### What I Received
This section is not required.

aulemahal:
How does this compare to : `dry_spell_total_length` ? It looks like this is similar to your `tot` method. Maybe, the `1d` method could be added to the existing function?
https://xclim.readthedocs.io/en/stable/indices.html?highlight=dry_spell#xclim.indices.dry_spell_total_length

It was added quite recently in xclim.

Currently, we don't have many methods with the equivalent of `DOY_min` and `DOY_max`. I would try to think of a more general way to add this functionality. It is kinda similar in idea to the problem in #649 where we decided to not implement anything in xclim. Instead, users would mask the data themselves and disable/config missing checks accordingly.

yrouranos:
@aulemahal It's indeed very similar to `dry_spell_total_length`. Adding the `1d` method to the existing function would probably be done easily by replacing `.sum() < thresh` with `.max() < thresh`.

[...]

PR #879, opened 2021-10-19
Add nr cdd

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* I have added the parameter `op: {"sum", "max"}` to `dry_spell_frequency`. 
* `sum` is the equivalent of the previous definition of the function. It checks if the accumulated precipitation over the window is below the threshold. 
* `add` is useful to calcucate `nr_cdd` for climatedata.ca. It checks if the maximum daily precipation over the window is below the threshold.

no

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.30.4-beta
+current_version = 0.30.5-beta
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+).(?P<patch>\d+)(\-(?P<release>[a-z]+))?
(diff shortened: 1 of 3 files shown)


aulemahal:
Also, @yrouranos, Correct me if I'm wrong, but I think that this, if applied to `dry_spell_total_length` would partially fix your issue #843?

yrouranos:
@aulemahal Regarding the addition of the `op` parameter suggested in the current PR, in theory, it would partially address #843. However, the consideration of start/end dates in an unusual way (searching for dry spells in a first step; masking results in a 2nd step) requires a different coding approach. Also, while performing unit tests, I found at least 3 contexts in which the current `dry_spell_total_length` returns incorrect results. The function that I developed addresses this as well. I was about to replace the current function with an improved one tomorrow. I was not aware that I was not the only one to work on this climate index.

PR #911, opened 2021-11-09
Run-length optimizations for window=1

### LINKED ISSUE #910: Rolling indicators with window=1 cause huge memory usage spikes
When running indicators that use the `run_length` functions in their backend (eg: `xclim.atmos.dlyfrzthw`), if `window=1` then this will cause enormous memory usage spikes; enough to grind servers to a halt. This likely due to `dask` running lots of `rolling()` calculations.

In order to address this, we might need to intelligently dictate to the `run_length` function that if the `window=1` to not perform rolling calculations. I say intelligently since we don't want to add a ton of if/else statements if we can avoid it.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #910

* Adds optimized pathways for the `window=1` cases in `rl.windowed_run_count`, `rl.windowed_run_events` and `rl.first_run` (and thus to most run-length functions using `window`, except `rle_statistics` and `longest_run`).

No, except that the `ufunc_1dim` argument and the `run_length_ufunc` options are now ignored when `window=1`.

Recall that the `ufunc` version is selected for small arrays that have no chunking along the dimension.

### Other information:
For `windowed_run_count` and `windowed_run_events` the performance gain is clear, see figures in next comments. In both cases, the ufunc version was slower than the optimization but faster then the non-ufunc version. The memory consumption was similar for the optimized and ufunc versions, and horrible for the non-ufunc version.

I also saw that the non-ufunc version triggered many small dask loadings before entering the main computation. I think something is not lazy in `rl.rle`.

For `first_run`, the non-ufunc  version is only slightly slower than the new optimized way, while the ufunc version is much slower. Here, the memory consumption was better in the non-ufunc than in the optimized version.

All tests performed with a 150 years daily datasets of 50 chunks of 65 points.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,10 @@ History
 -------------------
 Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan (:user:`tlogan2000`)
 
+New features and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``.
+
 Internal changes
 ~~~~~~~~~~~~~~~~
 * Removed some logging configurations in ``dataflags`` that were polluting python's main logging configuration. (:pull:`909`).
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,7 @@ Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan
 
 New features and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-* Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``.
+* Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``, (:pull:`911`, :issue:`910`).
 
 Internal changes
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


aulemahal:
# Figures for `windowed_run_events`.
x is the time in seconds, y is the total memory consumption (all threads, I was using 10 dask threads on the local scheduler).

## Non-ufunc
wre_norm_vect

## Ufunc
wre_norm_ufunc

## Optimized
wre_opt

# Figures for `first_run`

## Non-ufunc
fr_norm

## Ufunc
fr_ufunc

## Optimized
fr_opt

HISTORY.rst
```diff
@@ -6,6 +6,10 @@ History
 -------------------
 Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan (:user:`tlogan2000`)
 
+New features and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``, (:pull:`911`, :issue:`910`).
```
Zeitsperre, line 11:
```suggestion
* Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``. (:pull:`911`, :issue:`910`).
```

Zeitsperre (approved):
Thanks for the lightning-fast turnaround. Once the failing tests are addressed, LGTM.

Issue #424, opened 2020-04-24
Diagnostic tests for downscaling methods


huard:
### Description

Define generic diagnostics (e.g. trend preservation) to evaluate how various method perform.

aulemahal:
Discussions about Ouranos' internal scenarisation workflow and needs for the upcoming ESPO-R dataset have brought this issue back to the priorities.

I suggest going forward by implementing something inspired from the VALUE project (value-cost.eu). I see 2 new "submodule" : "indices" or "diagnotics" and "measures", replicating the division shown here : http://www.value-cost.eu/validationportal/app/#!indices

- `indices` : Functions reducing the time axis of the variables, very similar to the main Indicators. How much we can pull from the core xclim library is yet to be seen.
- `measures` : Functions giving means to compare the indices between different dataset, usually obs, hist and scen. They might also include further dimension reduction (time? spatial?).

As for the rest of `sdba`, these are basic tools that users must then use in their own script. We're not implementing any mega function returning a Good/Bad flag or any plotting methods.

The VALUE code is here : https://github.com/SantanderMetGroup/R_VALUE, 
However, I see that their code license is incompatible with ours... I'm not sure what that means, but I guess I'll make sure that names are different and that I implement the indices without looking at their code??

RondeauG:
Looking at the link provided, I'd say that most of the indices seem relevant. If we want to prioritise, the areas where we currently are lacking with our in-house qualitative checkups (figures) are:

- Anything spell-related (e.g. Median of wet (>=1mm) Spell Length Distribution)
- Transition probabilities & autocorrelation
- Extremes (e.g. 98th percentile, 20-year return value)  --> High and low
- Monthly variability
- Interannual variability

juliettelavoie:
Thanks for your input Gabriel! J'ai regroupé les indices de VALUE par aspect avec quelques infos supplémentaires. Veux-tu indiquer ceux que tu trouves pertinents dans la colonne Q ? https://ouranos-my.sharepoint.com/:x:/g/personal/jullav1_ouranos_ca/ETCaUuKJUz5OmjVkM1gOLJsBInomqAUA8-Llu-Ox__lIfA?e=ZAERfs

PR #914, opened 2021-11-11
[sdba] Re-write of Extreme's adjustment

### LINKED ISSUE #789: Faulty train()/adjust() behavior in sdba.ExtremeValues
* xclim version: v0.28.1

### Description
`ExtremeValues`: As currently coded, and as it exists in its Julia counterpart, `train()` achieves nothing and only calculates the GP parameters of `ref`, while `adjust()` fits `sim` to `ref` without allowances for changes in future precipitation intensity/frequency.

This does not correspond to the behaviour of the version coded and thoroughly tested in MATLAB.

### xclim implementation
(Images below)

As coded currently, `sdba.ExtremeValues.train(ref, hist)` will:
> 
> 1. Use `ref` and `hist` to find the right `tresh`
> 2. Calculate `ref_params`
> 

And `sdba.ExtremeValues.train()` will:
> 
> 1. Calculate `sim_fit`.
> 2. Use `sim_cdf = dist.cdf(sim_maxs, *sim_fit)` to find the probability of each maximum in `sim`.
> 3. Transpose that probability directly onto `ref`'s Paretto: `new_sim = dist.ppf(sim_cdf, *ref_params)`
> 4. Continue onwards (not tested yet).
> 

By definition, this creates a perfect fit between `sim` and `ref` and does not allow for changes in future precipitation intensity/frequency.

### MATLAB implementation
(Images below)

MATLAB's equivalent to `sdba.ExtremeValues.train(ref, hist)` will:
> 
> 1. Use `ref` and `hist` to find the right `tresh`
> 2. Calculate `ref_fit` and  `hist_fit`
> 3. Use `hist_cdf = dist.cdf(hist_maxs, *hist_fit)` and  `ref_cdf = dist.cdf(ref_maxs, *ref_fit)` to find the probability of each maximum in `hist` and `ref`.
> 4. Construct a multiplicative transfer function between the CDFs, based on probabilities (rather than on the actual precipitation values).
>

MATLAB's equivalent to `sdba.ExtremeValues.adjust(sim)` will:
> 
> 1. Calculate `sim_fit`
> 2. Use `sim_cdf = dist.cdf(sim_maxs, *sim_fit)` to find the probability of each maximum in `sim`.
> 3. Apply the transfer function on the maximums based on their probability (rather than on the actual precipitation values).
> 4. Continue onwards
>
The main difference here being that the comparison is between `ref` and `hist`, which allows `sim` to be different.

### Illustration
**_Illustration of train()_**
Black scatters = `ref`.  Black line = Representation of the fit.
Red scatters = `hist` (raw), according to MATLAB.
Green scatters = `hist` (corrected), according to MATLAB. --> Not actually used.
(No equivalent in xclim's implementation of `train()`)
image
X axis = Probability
Y axis = Precipitation (mm)

**_Illustration of the transfer function in MATLAB_**
(Ignore the straight line at the end. This was for capping purposes.)
image
X axis = Probability
Y axis = Correction factor

**_Illustration of adjust()_**
Black scatters = `ref`.  Black line = Representation of the fit.
Blue scatters = `sim` (raw), according to xclim.
Orange scatters = `sim` (corrected), according to xclim. --> _**Obvious overfitting to `ref`**_
Magenta scatters = `sim` (raw), according to MATLAB.
Cyan scatters = `sim` (corrected), according to MATLAB.
image
X axis = Probability (overlapping labels because it combines both a MATLAB and python figure)
Y axis = Precipitation (mm)

**_Same as above, but for a period closer to `ref`_**
To illustrate that if `sim` is close to `ref`, then the cyan scatters are close to the black line.
image

### LINKED ISSUE #790: sdba.ExtremeValues sensitive to the threshold
* xclim version: v0.28.1
**The tests were done in MATLAB's version due to the issues mentioned in #789, but this should not impact the results displayed here.**

### Description
As currently coded, the threshold used in the Paretto distribution is the average of `ref`'s and `hist`'s. For rarer events, this workaround can have a substantial impact on the results.

### Sensitivity for the higher extremes
For a case where `thresh(ref) = 17.55` and `thresh(hist) = 20.29`, I re-fit and re-calculated the value associated to a probability of 0.999 in `ref`'s CDF for `thresh in range(13, 21, 0.01)`.

image
X axis = `thresh`
Y axis = Precipitation (mm) for an extreme with a probability of 0.999 according to `ref`

### Impossible values
Same experience, but plotting the extreme associated with a probability of 1.0
image
X axis = `thresh`
Y axis = Precipitation (mm) for an extreme with a probability of 1.0 according to `ref`
(Missing scatters were associated a value of `NaN` or `Inf`)

In that latter plot `P(x) = 1.0`, there is an issue with those thresholds that find a value, since it should not be possible. Furthermore, some of these are close to the average (as is coded), or even to the univariate result. Using such a value, `thresh = 17.70`, and plotting the full range of the fit gives this:

Black scatters = `ref`. Black line = Representation of the fit.
Green scatters = `hist` (corrected)
image
X axis = Probability
Y axis = Precipitation (mm)

In this particular example, `hist` did not get fit on such extreme values, but it is in theory a possible issue (although difficult to replicate).

aulemahal:
### Pull Request Checklist:
    - This PR fixes #789 and fixes #790

* Rewrite of `sdba.ExtremeValues`, according to the Matlab version and comments from @RondeauG.

No, the `ExtremeValues` was hidden in previous versions.

### Other information:
I'm not sure what to test...


aulemahal:
One thing that gosses me here is that with dask we must know the dimension sizes _before_ the computation. In the trained `ds`, I have the `px_hist` and `af` along a "quantiles" dim (no coord). The only way I found for estimating it is `(1 - q_thresh) * ref.time.size` . This is the maximal theoritical size if cluster_thresh = 0.  But that's usually a lot larger than what we need. In my tests it was 10 times too large, so `ds.af` jas 90% NaN... yish.

Secondly, I have implemented `dist` so it could be controlled if we wanted, but `genpareto` is hardcoded. Do we want to accept other distributions?

xclim/sdba/_adjustment.py
```diff
@@ -289,3 +291,102 @@ def npdf_transform(ds: xr.Dataset, **kwargs) -> xr.Dataset:
...
+    Pcommon = (Px_hist <= Pmax) & (Px_hist >= Pmin)
+    Px_hist = Px_hist[Pcommon]
+    hist_in_ref = dist.ppf(Px_hist, *ref_params)
+
+    # Adjustment factors, unsorted
+    af = hist_in_ref / hist[Pcommon]
+    # sort them in Px order, and pad to have N values.
+    order = np.argsort(Px_hist)
+    px_hist = np.pad(Px_hist[order], ((0, N - af.size),), constant_values=np.NaN)
+    af = np.pad(af[order], ((0, N - af.size),), constant_values=np.NaN)
```
RondeauG, line 333: `constant_values=np.NaN` causes NaNs to appear in the result of `adjust` if some values in `sim` are above `max(Pcommon)`

```suggestion
    af = np.pad(af[order], ((0, N - af.size),), constant_values=af[order][-1])
```

RondeauG (changes requested):
1. I've successfully compared the results from my Matlab code and this new PR. Extremes are the same. Differences exist for other values, but this is due to differences in the transfer function and the number of corrected values (different approach between this and the older code). In short: The fix seems good.
2. My test fails when I try to use `dask`. I'll send @aulemahal my code so he can further test it out and see if he has the same issues.

My test also fails when I try to use `dask`

xclim/sdba/_adjustment.py
```diff
@@ -289,3 +291,102 @@ def npdf_transform(ds: xr.Dataset, **kwargs) -> xr.Dataset:
             "escores": escores,
         }
     )
+
+
+def _fit_on_cluster(data, thresh, dist, cluster_thresh):
```
huard, line 296: One-liner docstring.

xclim/sdba/_adjustment.py
```diff
@@ -289,3 +291,102 @@ def npdf_transform(ds: xr.Dataset, **kwargs) -> xr.Dataset:
...
+    _, _, _, maximums = u.get_clusters_1d(data, thresh, cluster_thresh)
+    params = list(
+        _fitfunc_1d(maximums - thresh, dist=dist, floc=0, nparams=3, method="ML")
+    )
+    # We forced 0, put back thresh.
+    params[-2] = thresh
+    return params
+
+
+def _extremes_train_1d(ref, hist, ref_params, *, q_thresh, cluster_thresh, dist, N):
```
huard, line 306: Two-liner docstring ?

huard (approved):
Just a superficial review. 
I suppose it will be useful to get https://github.com/Ouranosinc/xclim/issues/424 done.

xclim/sdba/_adjustment.py
aulemahal: I got this fixed by tweaking `extrapolate_qm` and `interp_on_quantiles` to ignore NaNs. 

Maybe @huard  could have a look on these changes. I modified the output of `extrapolate_qm` in the case of `method='nan'` (not really related to this PR). And `interp_on_quantiles` now drops ANY NaNs in the inputs EXCEPT for those expected extrapolated bounds.

RondeauG (approved):
Fixes appear to be working as intended.

huard (approved):
Happy with the changes. Could you just confirm that these changes are exercised by the test suite ?

aulemahal:
The last commit reactivates the Extreme's tests and removes a bit that was not necessarily true. The tests pass, but they don't do much more than checking if the method runs... I'm not sure of what to test?

@RondeauG Are the algorithms close enough that we could compare an output from the matlab version to the one here? Not value by value, but something like a general change in the distributions...

huard:
I'm happy to leave testing against a reference value to a future PR. I would maybe advertise the algo as experimental until then.

PR #915, opened 2021-11-15
Indices/heat index

### LINKED ISSUE #807: Add heat index calculation
We have some code to calculate the heat index (i.e. how the human body perceives temperature, which combines relative humidity with air temperature) that we can also add to the list of indices. It uses [this formula](https://en.wikipedia.org/wiki/Heat_index#Formula).

Original code by @clairbarnes

raquel-ucl:
### Pull Request Checklist:
    - This PR fixes #807 
  - [ ] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

* Calculation of heat index

No

### Other information:

For now I've just added the method as a threshold calculation, but we still need to change a couple things:
- [ ] The issue discussion points we shouldn't add the index calculation in `_threshold.py` but in `_conversion.py`. We can look into `humindex` which is doing something similar, so the code I have added so far might need some tweaking.
- [ ] There are no tests yet, so we can look into those for `humindex` and do something similar as well.
- [ ] Add @JamieJQuinn to list of AUTHORS.rst


xclim/indices/_threshold.py
```diff
@@ -1237,6 +1238,62 @@ def days_with_snow(
...
+      Resampling frequency.
+
+    Returns
+    -------
+    xarray.DataArray, [time][temperature]
+      Heat index.
+    """
+    thresh = convert_units_to(thresh, tasmax)
+    t = convert_units_to(tasmax, "degC")
+    t = t.where(t > 20)
```
aulemahal, line 1271:
```suggestion
    thresh = convert_units_to(thresh, "degC")
    t = convert_units_to(tasmax, "degC")
    t = t.where(t > thresh)
```
I'm guessing that hardcoded 20 was forgotten? Also, better to convert everything _explicitly_ to the same units if we need a specific one.

xclim/indices/_threshold.py
```diff
@@ -1237,6 +1238,62 @@ def days_with_snow(
...
+    r = convert_units_to(hurs, "%")
+
+    tr = np.multiply(t, r)
```
aulemahal, line 1274: I checked and the use of `np.multiply` seems to have no unwanted side effects (like with dimension matching or with dask arrays). However, I think we'd still prefer the direct: `tt = t * t` etc.

xclim/indices/_threshold.py
```diff
@@ -1237,6 +1238,62 @@ def days_with_snow(
...
+        -8.78469475556
+        + 1.61139411 * t
+        + 2.33854883889 * r
+        - 0.14611605 * tr
+        - 0.012308094 * tt
+        - 0.0164248277778 * rr
+        + 0.002211732 * ttr
+        + 0.00072546 * trr
+        - 0.000003582 * ttrr
+    )
```
aulemahal, line 1291: I'm guessing there is a source for these factors? Please add it to the docstring in a "References" section.

xclim/indices/_threshold.py
```diff
@@ -1237,6 +1238,62 @@ def days_with_snow(
     return to_agg_units(out, prsn, "count")
 
 
+@declare_units(tasmax="[temperature]", hurs="[g m-3]", thresh="[temperature]")
```
aulemahal, line 1241:
```suggestion
@declare_units(tasmax="[temperature]", hurs="[]", thresh="[temperature]")
```
According to CMIP6, which we follow for variables' meta, the canonical units for hurs are "%", thus it is "dimensionless".

Also, xclim's `convert_units_to` is unable to convert from "mass per volume" to "dimensionless" as requested at line 1272. That's because we haven't implemented this transformation in the 'hydro' context, the one where the water density (1000 kg/m³) is implicit. That can be done in this PR, if you need it.

aulemahal (commented):
Thanks for this addition!
I have a few comments for now, mostly to harmonize this indice with the others of the module, style- and convention-wise.

Also needed before merging:
- [ ] Indicator corresponding to this indice.
    * [ ] French metadata translations (we can help for this :) )

xclim/indices/_threshold.py
JamieJQuinn: @aulemahal unfortunately the only source of these *specific* coefficients are from [the wikipedia article on the heat index](https://en.wikipedia.org/wiki/Heat_index). The references within all give formulae in Fahrenheit, not Celsius and the coefficients seem to have been converted for the wiki article. We could rewrite the formula in F, converting units in/out of the function from/to Celsius, allowing us to directly reference the proper source, or we can just reference the wikipedia article. What are your thoughts?

aulemahal: I see! I would cite at least one reference and mention that factors have been converted to Celsius. Or indeed, doing this all in Fahrenheit could be even more explicit, but I think we want to have the output Celsius-like so it may be too much noise.

I see that there are more than one formulation! Something we do in xclim (see relative humidity for example) is to have a `method` arg which is a string to switch between different implementations. It would make sense here as it is mostly only the coefficients that change. But no need to do this here, it can be done later when the need arises.

xclim/indices/_threshold.py
raquel-ucl: Done!

xclim/data/fr.json
```diff
@@ -975,6 +975,12 @@
     "description": "Indice humidex décrivant la température ressentie par une personne en réponse à l'humidité relative.",
     "long_name": "indice humidex"
   },
+  "HEAT_INDEX": {
```
raquel-ucl, line 978: @aulemahal I've created this in case you want to make a commit suggestion for the french version of the description. Is there anywhere else where this is needed?

aulemahal: Good, I'll find something! That's all for the translations, thanks.

xclim/data/fr.json
```diff
@@ -975,6 +975,12 @@
...
+    "title": "",
+    "abstract": "",
+    "description": "",
+    "long_name": ""
```
aulemahal, line 982:
```suggestion
    "title": "Indice de chaleur",
    "abstract": "L'indice de chaleur est une estimation de la température ressentie par une personne à l'ombre lorsqu'on tient compte de l'humidité relative.",
    "description": "Température ressentie estimée avec l'humidité relative et la température ambiante.",
    "long_name": "Indice de chaleur"
```

xclim/indicators/atmos/_conversion.py
```diff
@@ -38,6 +39,15 @@ class Converter(Indicator):
     compute=indices.humidex,
 )
 
+heat_index = Converter(
+    identifier="heat_index",
+    units="C",
+    standard_name="air_temperature",
+    long_name="heat index",
+    description="Days in which the human body perceives temperature is hot, which combines relative humidity with the air temperature.",
```
aulemahal, line 47: I'm not sure I understand this description (and the abstract). It says "Days [...]", but the output is a temperature-like value, no?

aulemahal:
Because I was confused myself, I suggest adding a "Notes" section saying that this is the American index that tries to measure the same thing as the Humidex, which is Canadian. And a similar note in Humidex.

xclim/indices/_threshold.py
raquel-ucl: Would it be a lot of trouble to do the transformation? Another option would be to check the hurs data is already in % and not to perform the calculation otherwise.

aulemahal: The problem here would be that xclim's `declare_units` is limited to a single dimensionality per variable. A workaround for your suggestion, would be to pass `hurs=None` but that disables the unit check completely, which I find quite inelegant.

Also I'm not sure I understand how "hurs", _relative_ humidity, could be given in "g m-3"? Wouldn't it be _absolute_ humidity in that case?  I would prefer that users convert their data before. We could add examples on how to do that in the docstring or in the doc.

xclim/indices/_threshold.py
clairbarnes: @aulemahal Sorry, I didn't get a notification about this conversation... the reference for these factors is [blazejczyk2012] Blazejczyk, K., Epstein, Y., Jendritzky, G., Staiger, H., & Tinz, B. (2012). Comparison of UTCI to selected thermal indices. International journal of biometeorology, 56(3), 515-535 (I think that @raquel-ucl may have added this already?)

raquel-ucl: Yes, I added it in https://github.com/Ouranosinc/xclim/pull/915/commits/6327f32ccf0cba9e12e596e839c5d8deac39781e

xclim/indices/_threshold.py
raquel-ucl: After talking to @clairbarnes I have changed the hurs input to dimensionless as in this suggestion.

xclim/indicators/atmos/_conversion.py
raquel-ucl: I've corrected this now.

raquel-ucl:
I think I've tackled all comments now. It should be ready for final review :)

HISTORY.rst
```diff
@@ -19,6 +19,7 @@ New features and enhancements
 * New ``**indexer`` keyword args added to many indicators, it accepts the same arguments as ``xclim.indices.generic.select_time``, which has been improved. Unless otherwise specified, the time selection is done before any computation. (:pull:`934`, :issue:`899`).
 * Rewrite of ``xclim.sdba.ExtremeValues``, now fixed with a correct algorithm. It has not been tested extensively and should be considered experimental. (:pull:`914`, :issue:`789`, :issue:`790`).
 * Added `days_over_precip_doy_thresh` and `fraction_over_precip_doy_thresh` indicators to distinguish between WMO and ECAD definition of the Rxxp and RxxpTot indices. (:issue:`931`, :pull:`940`)
+* Added `heat_index` indicator.
```
Zeitsperre:
```suggestion
* Added `heat_index` indicator. (:issue:`807`, :pull:`915`)
```
It would be good to have a bit more of a description. Maybe mention the source for the methods?

Zeitsperre:
Looks like there's an issue with the regex for the release notes. I'll push a fix today!

xclim/indices/_conversion.py
```diff
@@ -116,6 +124,74 @@ def humidex(
...
+@declare_units(tasmax="[temperature]", hurs="[]")
+def heat_index(
+    tasmax: xr.DataArray,
+    hurs: xr.DataArray,
+    freq: str = "MS",
+) -> xr.DataArray:
+    r"""Daily heat index.
+
+    Perceived temperature after relative humidity is taken into account. The
+    index is only valid for temperatures above 20C.
```
aulemahal, line 136:
```suggestion
    index is only valid for temperatures above 20°C.
```

xclim/indices/_conversion.py
```diff
@@ -116,6 +124,74 @@ def humidex(
...
+      Maximum daily temperature.
+    hurs : xr.DataArray
+      Relative humidity.
+    freq : str
+      Resampling frequency.
+
+    Returns
+    -------
+    xr.DataArray, [time][temperature]
+      Heat index for days with temperature above 20C.
```
aulemahal, line 150:
```suggestion
      Heat index for days with temperature above 20°C.
```

xclim/indices/_conversion.py
```diff
@@ -116,6 +124,74 @@ def humidex(
...
+
+    References
+    ----------
+    .. [blazejczyk2012] Blazejczyk, K., Epstein, Y., Jendritzky, G., Staiger,
+    H., & Tinz, B. (2012). Comparison of UTCI to selected thermal indices.
+    International journal of biometeorology, 56(3), 515-535.
```
aulemahal, line 156:
```suggestion
    .. [blazejczyk2012] Blazejczyk, K., Epstein, Y., Jendritzky, G., Staiger, H., & Tinz, B. (2012). Comparison of UTCI to selected thermal indices. International journal of biometeorology, 56(3), 515-535.
```
I know its unelegant to have long lines, but Sphinx/rst won't parse it correctly otherwise.

aulemahal (approved):
A few nitpicks for the docstring. 
If you have time, it would be nice to have a unit test for the indicator as well. I guess `xclim/testing/tests/test_atmos.py` would be a good place. You can use the `atmosds` pytest fixture, it's a ERA5-land based dataset that should contain all variables you need. 

Otherwise, this is looking good!

Zeitsperre:
@raquel-ucl the fix is in master branch now. You just need to rebase/merge master in for the fix !

raquel-ucl:
Added tests for the indicator with the atmosds data. Not sure if that's what you were looking for! It took a while to find data that was not too cold in order to compute the index (needs values > 20C and it's cold over there!)

PR #917, opened 2021-11-15
Count cold spell days first, then resample.

### LINKED ISSUE #916: Cold spell days
* xclim version: 0.31.0
* Python version: 3.8
* Operating System: MacOS

### Description
Currently, `cold_spell_days` divides the time series into periods, then counts run lengths contained within each period. This means that spells that span two periods are treated as two separate cold spells: particularly when handling data at monthly frequency this can mean that cold spells are not identified.

### Suggested solution
We can switch the order for this calculation so we don't miss or duplicate cold spells. Original codes is by @clairbarnes.

raquel-ucl:
### Pull Request Checklist:
    - This PR fixes #916 

* The original method divided the time series into periods, then counted run lengths contained within each period. This means that spells that span two periods are treated as two separate cold spells: particularly when handling data at monthly frequency this can mean that cold spells are not identified. This PR is the proposed alternative.

### Other information:
- [ ] Although there are already tests for this method, we still need to add tests for edge cases (ie periods that used to be counted twice are counted once correctly now, and periods that were not counted at all now are).

### DIFF
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -200,14 +200,9 @@ def cold_spell_frequency(
     t = convert_units_to(thresh, tas)
     over = tas < t
 
-    group = over.resample(time=freq)
-
-    out = group.map(rl.windowed_run_events, window=window, dim="time")
-
-    # runs = rl.rle(over, index=index)
-    # cold_spells = runs >= window
-
-    # out = cold_spells.resample(time="AS-JUL").sum()
+    runs = rl.rle(over, index=index)
+    cold_spells = runs >= window
+    out = cold_spells.resample(time=freq).sum()
 
     out.attrs["units"] = ""
     return out
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -157,9 +157,13 @@ def cold_spell_days(
     t = convert_units_to(thresh, tas)
     over = tas < t
 
-    runs = rl.rle(over)
-    cold_spells = runs.ffill(dim="time") >= window
-    out = cold_spells.resample(time=freq).sum()
+    group = over.resample(time=freq)
+
+    out = group.map(rl.windowed_run_count, window=window, dim="time")
+
+    # runs = rl.rle(over)
+    # cold_spells = runs.ffill(dim="time") >= window
+    # out = cold_spells.resample(time=freq).sum()
 
     return to_agg_units(out, tas, "count")
 
@@ -170,6 +174,7 @@ def cold_spell_frequency(
     thresh: str = "-10 degC",
     window: int = 5,
     freq: str = "AS-JUL",
+    index: str = "first",
 ) -> xarray.DataArray:
     r"""Cold spell frequency.
 
@@ -186,6 +191,8 @@ def cold_spell_frequency(
       Minimum number of days with temperature below threshold to qualify as a cold spell.
     freq : str
       Resampling frequency.
+    index : str
+      If ‘first’, the run length is indexed at the first element in the run. If ‘last’, at the last element in the run.
 
     Returns
     -------
@@ -196,9 +203,16 @@ def cold_spell_frequency(
     """
     t = convert_units_to(thresh, tas)
     over = tas < t
+
     group = over.resample(time=freq)
 
     out = group.map(rl.windowed_run_events, window=window, dim="time")
+
+    # runs = rl.rle(over, index=index)
+    # cold_spells = runs >= window
+
+    # out = cold_spells.resample(time="AS-JUL").sum()
+
     out.attrs["units"] = ""
     return out
 
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -156,9 +156,11 @@ def cold_spell_days(
     """
     t = convert_units_to(thresh, tas)
     over = tas < t
-    group = over.resample(time=freq)
 
-    out = group.map(rl.windowed_run_count, window=window, dim="time")
+    runs = rl.rle(over)
+    cold_spells = runs.ffill(dim="time") >= window
+    out = cold_spells.resample(time=freq).sum().rename(cold_spell_days)
+
     return to_agg_units(out, tas, "count")
 
 
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -157,13 +157,9 @@ def cold_spell_days(
     t = convert_units_to(thresh, tas)
     over = tas < t
 
-    group = over.resample(time=freq)
-
-    out = group.map(rl.windowed_run_count, window=window, dim="time")
-
-    # runs = rl.rle(over)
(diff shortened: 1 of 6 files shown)


tlogan2000:
I believe that we tried this originally (a few years ago now) and it caused pretty major headaches with dask when using large climate projection datasets.   I suggest investigating a test calculation on something 'big' to see if changes in `runlength` or `dask` allow for soemthing stable ...

xclim/indices/_threshold.py
```diff
@@ -156,9 +156,11 @@ def cold_spell_days(
     """
     t = convert_units_to(thresh, tas)
     over = tas < t
-    group = over.resample(time=freq)
 
-    out = group.map(rl.windowed_run_count, window=window, dim="time")
+    runs = rl.rle(over)
+    cold_spells = runs.ffill(dim="time") >= window
+    out = cold_spells.resample(time=freq).sum().rename(cold_spell_days)
```
aulemahal, line 162:
```suggestion
    out = cold_spells.resample(time=freq).sum()
```
Not sure what that rename was meant to do, called with a function?

Also, except for units, indices do not set other metadata fields on their output, that is the role of the Indicators.

aulemahal (changes requested):
That makes much sense! We can do this explicitly here, but we must think of way to generalize that. I'm guessing almost all indices using  `windowed_run_count` could benefit from it.

However, could you add a test for the change? Like a test that would fail with the old method and now passes.

xclim/indices/_threshold.py
raquel-ucl: Ah, that's probably left over from running the code locally. I'll commit the suggestion.

JamieJQuinn:
@aulemahal amended existing tests so that they failed with the previous days + frequency calculations and now pass with the new calculations. I noticed there are very similar tests [in test_indices.py](https://github.com/Ouranosinc/xclim/blob/f248b5c7923ad5bc588f1ba6d04085b1cce2c8b4/xclim/testing/tests/test_indices.py#L122) and  [in test_temperature.py](https://github.com/Ouranosinc/xclim/blob/f248b5c7923ad5bc588f1ba6d04085b1cce2c8b4/xclim/testing/tests/test_temperature.py#L385) so I've fixed the tests where they broke due to the change but I've only explicitly tested the new behaviour in `test_indices.py`. 

Let me know if you want me to replicate that test in `test_temperature.py`!

JamieJQuinn:
@tlogan2000 Have you got similar tests of the performance with dask?

JamieJQuinn:
@raquel-ucl I can't edit your opening comment so could you please tick off everything bar bumpversion? I've also added you to `.zenodo.json` but I couldn't find your ORCID in the group's info so do add if you have one.

xclim/indices/_threshold.py
```diff
@@ -186,6 +187,8 @@ def cold_spell_frequency(
       Minimum number of days with temperature below threshold to qualify as a cold spell.
     freq : str
       Resampling frequency.
+    index : str
```
Zeitsperre, line 190:
```suggestion
    index : {"first", "last"}
```

bzah:
Hi, I'm also interested in the feature. 
If no one is on it, I'm going to run a benchmark on my side to assess performances of this vs former code.

bzah:
I made a few tests with 2 datasets, one quite small (53MB) and one larger (843M).
I suspect we don't need to really test on a huge dataset because the larger one already shows some limits of the new algorithm. 
However on small datasets, the new algorithm seems to perform better than the former one.

I tried 3 configuration:
1. Numpy only for reference.
2. Dask with somewhat bad chunking.
It's an "auto" chunking everywhere but on time dimension, where no chucking is done.
3. Dask with "auto" chunking on every dim.

I'm using the distributed scheduler with the following configuration:
```
 distributed.Client(memory_limit='12GB',
                             timeout=20,
                             n_workers=1,
                             threads_per_worker=10)
```

All tests were done on my laptop.

<details>
  <summary>Numpy only results</summary>

  On the larger dataset, they have about the same performances.
  ```
  data shape: (13140, 120, 180)
  old algo: 26.002015164 secs
  new algo: 26.72492288 secs
  ```
</details>

<details>
  <summary>Dask bad chunking</summary>
  
  For both algorithms here, performances are awful, but they are even worse for the new algo.
  I'm even getting some dreadful `distributed.utils_perf - WARNING - full garbage collections took 15% CPU time recently (threshold: 10%)` with the later.

  ```
data shape: (13140, 120, 180)
old algo 152.265213125 secs
new algo 332.73681010199994 secs
  ```
</details>

<details>
  <summary>Dask good chunking</summary>
  
  Performances are really good for the old algorithm and somewhat bad for the new one in this case:
  ```
data shape: (13140, 120, 180)
old algo 13.619929155999998 secs
new algo 70.769079325 secs
  ```

</details>

----

I don't think it's necessary to test it on larger data for now. I highly suspect Dask would start to get mad and flood with warnings if we try it.

Side Notes:
- I wrote on gitter that the 2nd strategy was good for rolling window indices, but it's actually only good for the bootstrapping algorithm not the actual rolling window process.
- The script gist to run these tests is [here](https://gist.github.com/bzah/f56882ec8929e0133d03bbfc8161cfc2).

raquel-ucl:
@bzah What's the next step? Do you need us to look into this?

JamieJQuinn:
@tlogan2000 @aulemahal Seems like we're getting pretty poor performance with Dask. Has this problem been solved elsewhere in the code or do we have some idea of a Dask-friendly way to do this?

aulemahal:
An untested alternative for now is to use the other version of the "rle" algorithm. The one we call "ufunc" and is applied on 1D timeseries with `xr.apply_ufunc` with `vectorize=True` (`xc.indices.run_length.rle_1d`. The issue with this algo is that it needs to have a single chunk along `time`, which is often not practical without a heavy rechunking step... Also, there is no way to use it directly currently, it's always used by the counting function themselves (`windowed_run_count` for example).

We have a xclim meeting this afternoon, we'll try to come up with a plan for the issue. Overall, a better  `rle` would be the best option, but it is not in our near future plans to spend time on this.

aulemahal:
@raquel-ucl @JamieJQuinn We decided that, for now, the performance loss was too significative to merge this right away. We'll try to improve  `rle` directly at the source before inverting the operation order like here. Also, when we do so, we'll need to invert the order everywhere, not only in `cold_spell_days`, so xclim stays coherent. 

Sorry!

JamieJQuinn:
No worries, we totally understand. Before I go and make a whole new issue + PR for this idea, would you be open to an warning printed when using the function with a fine-grained frequency like freq=="MS" or == "M", to highlight this potential issue to users?

aulemahal:
Yes, that can be a temporary solution, so its explicit. I suggest you make this warning trigger a function in `xclim.core.utils` so we can use it in a few indices, not only `cold_spell_days`. Thanks!

PR #924, opened 2021-11-15
Added jetstream strength and latitude calculation

### LINKED ISSUE #923: Strength and latitude of jetstream
@clairbarnes and @Thomasjkeel have put together some code to identify the latitude and strength of maximum smoothed zonal windspeed in the region 15-75N, -60-0E.

Not sure where this contribution should go, but I'll make a PR so you can see it and we can take it from there.

raquel-ucl:
### Pull Request Checklist:
    - This PR fixes #923 

* Calculation of strength and latitude of jetstream. This code will return two DataArrays but can easily be split into two methods or modified to return a single Dataset if that's more convenient.

### Other information:

- [x] We need to add Tom Keel (@Thomasjkeel, thomas.keel.18 at ucl.ac.uk) to the list of contributors.


xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
     return out
 
 
+@declare_units(ua="[speed]")
+def jetstream_metric_woolings(
+    ua: xarray.DataArray,
+) -> xarray.DataArray:
+    """Strength and latitude of jetstream.
+
+    Identify latitude and strength of maximum smoothed zonal windspeed in the region 15-75&deg;N, -60-0&deg;E.
```
huard, line 2069:
```suggestion
    Identify latitude and strength of maximum smoothed zonal wind speed in the region from 15 to 75°N and -60 to 0°E.
```

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+
+    Parameters
+    ----------
+    ua : xarray.DataArray
+      u-component wind velocity at between 750 and 950 hPa.
```
huard, line 2074: Maybe mention this is the eastward wind component.

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+
+    References
+    ----------
+    .. [woollings2010] Woollings, T., Hannachi, A., & Hoskins, B. (2010). Variability of the North Atlantic eddy‐driven jet stream. Quarterly Journal of the Royal Meteorological Society, 136(649), 856-868.
+
+    """
+
+    # select only relevant hPa levels, compute zonal mean windspeed
+    pmin = convert_units_to("750 hPa", ua.plev)
+    pmax = convert_units_to("950 hPa", ua.plev)
```
huard, line 2091: This relies on pressure levels being named `plev`. I suggest we use cf-xarray to identify the pressure coordinate. 
Off the top of my head, it would look like
```
import cf_xarray
...
pmin = convert_units_to("750 hPa", ua.cf["pressure"])
```

I don't know if we want to bring cf-xarray into xclim within this PR or another separate one though. @aulemahal  Thoughts ?

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+
+    ua = ua.sel(plev=slice(pmin, pmax), lat=slice(15, 75), lon=slice(-60, 0))
```
huard, line 2093: Same idea here with lat and lon. 
Also, should be check that the original grid includes the target region. And will this work if longitudes are defined from 0 to 360 ?

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
     return out
 
 
+@declare_units(ua="[speed]")
+def jetstream_metric_woolings(
```
huard, line 2064: @tlogan2000 Should this go in atmos or conversion, or an entirely new module, e.g. `synoptic`?

xclim/testing/tests/test_indices.py
```diff
@@ -1045,6 +1045,41 @@ def test_below_simple(self, tasmax_series):
...
+    time_coords = pd.date_range("2000-01-01", "2000-03-06", freq="D")
+    # make random ua data array of shape (66 days, 3 plevs, 5 lons, 5 lats)
+    np.random.seed(42)
+    da_ua = xr.DataArray(
+        np.random.rand(66, 3, 5, 5),
+        coords={
+            "time": time_coords,
+            "plev": [75000, 85000, 100000],
+            "lon": [-60, -59, -58, -57, -56],
+            "lat": [15, 16, 17, 18, 19],
```
huard, line 1059: I suggest we make sure it works with arbitrary variable names, as long as cf-xarray can recognize the variables (for example based on the standard_name).

xclim/testing/tests/test_indices.py
```diff
@@ -1045,6 +1045,41 @@ def test_below_simple(self, tasmax_series):
...
+    def test_jetstream_metric_woolings(self):
+        da_ua = self.da_ua
+        out = xci.jetstream_metric_woolings(da_ua)
+        np.testing.assert_equal(len(out), 2)
+        jetlat, jetstr = out
+        # should be 6 values that are not NaN because of 61 day moving window and 66 chosen
+        np.testing.assert_equal(np.sum(~np.isnan(jetlat).data), 6)
+        np.testing.assert_equal(np.sum(~np.isnan(jetstr).data), 6)
+        np.testing.assert_equal(jetlat.max().data, 19.0)
+        np.testing.assert_equal(jetstr.max().data, 0.5620588628647811)
```
huard, line 1078: How were the expected values determined ? Or in other words, what reference are we checking against to make sure it works as expected ?

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+    lanczos_weights = xarray.DataArray(w[0 + (window_size % 2) : -1], dims=["window"])
+
+    # apply the filter
+    ua_lf = (
+        zonal_mean.rolling(time=window_size, center=True)
+        .construct("window")
+        .dot(lanczos_weights)
+    )
+
+    jetlat = ua_lf.idxmax("lat").rename("jetlat").assign_attrs(units="degrees")
```
huard, line 2127: Should the units be those of the original latitude variable ?

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+    w = np.zeros([nwts])
+    n = nwts // 2
+    w[n] = 2 * cutoff
+    k = np.arange(1.0, n)
+    sigma = np.sin(np.pi * k / n) * n / (np.pi * k)
+    firstfactor = np.sin(2.0 * np.pi * cutoff * k) / (np.pi * k)
+    w[n - 1 : 0 : -1] = firstfactor * sigma
+    w[n + 1 : -1] = firstfactor * sigma
+
+    lanczos_weights = xarray.DataArray(w[0 + (window_size % 2) : -1], dims=["window"])
```
huard, line 2118: I suggest separating this bit of code into a separate function.

xclim/indices/_threshold.py
```diff
@@ -2059,6 +2060,76 @@ def windy_days(
...
+
+    # apply the filter
+    ua_lf = (
+        zonal_mean.rolling(time=window_size, center=True)
+        .construct("window")
+        .dot(lanczos_weights)
+    )
+
+    jetlat = ua_lf.idxmax("lat").rename("jetlat").assign_attrs(units="degrees")
+    jetstr = ua_lf.max("lat").rename("jetstr").assign_attrs(units="m s-1")
```
huard, line 2128: Same idea here. What happens if the original ua variable is in km/h ?

huard (changes requested):
This is great new addition to xclim. 

I suggest we create a new module for indices (`synoptic`) to store this algorithm, and a corresponding `indicators.atmos._synoptic` module for the corresponding indicator.

xclim/indices/_threshold.py
Thomasjkeel: I think using cf_array in this function is a good idea and would look something like:
```python
pmin = convert_units_to("750 hPa", ua.cf["air_pressure"])  
pmax = convert_units_to("950 hPa", ua.cf["air_pressure"])

ua = ua.cf.sel(air_pressure=slice(pmin, pmax), latitude=slice(15, 75), longitude=slice(lonmin, lonmax))
```
To determine `lonmin` & `lonmax` in above case, do you know if xclim or xarray already have any existing solution for checking whether longitude is -180-180 or 0-360? If not, I am happy to try and find or write something e.g. using 'lon % 180', etc.

xclim/indices/_threshold.py
tlogan2000: I am undecided here.  I don't think `conversion` is really a great fit.   Input/output variables suggest that it should stay in `atmos` but it is definitely quite different from everything else in there.   What other calculations could we imagine in a `synoptic` module  (i.e. is the jetstream calculation going to be the only calculation in that new module)

xclim/indices/_threshold.py
huard: Not sure, the spatial subsetting logic originally in xclim has all been moved out to `clisops.subset`. I suggest here that we define the range of longitudes that is expected, and raise an error if the input does not meet that requirement, and possibly document how to change the coordinates in the docstring notes.

xclim/indices/_threshold.py
huard: For now it would be, but I can imagine cyclone detection algorithms ending up there.

xclim/indices/_threshold.py
Thomasjkeel: There is an issue with the units returned by this function – automatically converted from 'degrees_north' or 'degrees' to deg

xclim/testing/tests/test_indices.py
Thomasjkeel: I haven't yet been able to solve this comment yet, so any advice would be welcome! Currently, I am taking a random array of numbers (random seed==42) of shape: 66 days, 3 plev, 5 lat, 5 lon, as this function needs at minimum of 61 days for the Lanczos filter and then I have checked the data before and results manually. One way of getting a reference to check this function might be using actual ua data (maybe the dataset used in Woolings et al. 2010 paper), but then the test might take too long to run. Still unsure...

xclim/testing/tests/test_indices.py
```diff
@@ -1082,8 +1094,8 @@ def test_jetstream_metric_woolings(self):
         np.testing.assert_equal(np.sum(~np.isnan(jetstr).data), 6)
         np.testing.assert_equal(jetlat.max().data, 19.0)
         np.testing.assert_equal(jetstr.max().data, 0.5620588628647811)
-        assert jetlat.units == "deg"
-        assert jetstr.units == "m s-1"
+        assert jetlat.units == da_ua.cf["latitude"].units
```
Thomasjkeel, line 1097: currently: "deg" != "degrees_north"

xclim/testing/tests/test_indices.py
```diff
@@ -1053,27 +1053,39 @@ class TestJetStreamIndices:
...
             "units": "m s-1",
         },
     )
-    da_ua.plev.attrs["units"] = "Pa"
-    da_ua.lat.attrs["units"] = "degrees"
+
+    da_ua.Z.attrs = {"units": "Pa", "standard_name": "pressure"}
+    da_ua.X.attrs = {"units": "degrees_east", "standard_name": "longitude"}
+    da_ua.Y.attrs = {"units": "degrees_north", "standard_name": "latitude"}
+    da_ua.T.attrs = {"standard_name": "time"}
```
Thomasjkeel, line 1071: `T` is not being noticed by cf_xarray as `time`

xclim/indices/_synoptic.py
```diff
@@ -42,31 +43,38 @@ def jetstream_metric_woolings(
...
     lon_max = 0
-    lons_within_range = any((ua["lon"] >= lon_min) & (ua["lon"] <= lon_max))
+    lons_within_range = any(
+        (ua.cf["longitude"] >= lon_min) & (ua.cf["longitude"] <= lon_max)
+    )
     if not lons_within_range:
         raise ValueError(
             "Longitude values need to be in a range between %s-%s. Consider changing the longitude coordinates to between -180.E–180.W"
             % (lon_min, lon_max)
         )
```
Thomasjkeel: Should lon_min, lon_max be a function argument? And, should this check be put into another function (i.e. following SRP)?

xclim/testing/tests/test_indices.py
huard: Then just add a comment stating that you've computed the expected answer manually. 

Would it be possible to create a very simple synthetic data set (e.g `np.ones()`) where the answer can be computed easily?

HISTORY.rst
```diff
@@ -4,7 +4,17 @@ History
 
 0.32.0 (unreleased)
 -------------------
-Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan (:user:`tlogan2000`), Trevor James Smith (:user:`Zeitsperre`)
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan (:user:`tlogan2000`), Trevor James Smith (:user:`Zeitsperre`), Abel Aoun (:user:`bzah`), David Huard (:user:`huard`), Clair Barnes (:user:`clairbarnes`), Raquel Alegre (:user:`raquel-ucl`), Jamie Quinn (:user:`JamieJQuinn`), Tom Keel (:user: `Thomasjkeel`),
```
Zeitsperre, line 7:
```suggestion
Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Travis Logan (:user:`tlogan2000`), Trevor James Smith (:user:`Zeitsperre`), Abel Aoun (:user:`bzah`), David Huard (:user:`huard`), Clair Barnes (:user:`clairbarnes`), Raquel Alegre (:user:`raquel-ucl`), Jamie Quinn (:user:`JamieJQuinn`), Tom Keel (:user:`Thomasjkeel`),
```

Zeitsperre:
The build is failing because of a new dependency (`cf_xarray`). Once you've added this new dependency to the `setup.py`,  please also be sure to add that information to the breaking changes (on the PR description and in the history - on its own line). Thanks!

xclim/testing/tests/test_indices.py
Thomasjkeel: Okay, I have made changes and added an improved test with np.ones/np.zeros test in: ff3d1dd1c7cda53dc14b2e14a85eb076c1723b53. Hope this is okay :)

xclim/testing/tests/test_indices.py
aulemahal: That's one assumption we are ok to make in xclim. The time coordinate is always "time", so it can be accessed directly as such `da_ua.time`, no need to go through `cf_xarray` for this one.

xclim/indices/_synoptic.py
aulemahal: Having those as arguments seems logical to me, unless other values do not make sense for the algorithm?
A thing that can be done to ensure a meaningful indicator is too have them are arguments on the_indice_ but inject the normal values in the _indicator_. This way, the users is able to control it, but only when using the less-rigid version.

aulemahal (commented):
The indice looks good!

For a complete xclim addition, this would be missing an indicator.
This will be the first indicator that reduces dimensions other than `time`, I would suggest creating a new `_synoptic.py` file in  `xclim/indicators/atmos` as well.

xclim/indices/_synoptic.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # select only relevant hPa levels, compute zonal mean windspeed
+    pmin = convert_units_to("750 hPa", ua.cf["pressure"])
+    pmax = convert_units_to("950 hPa", ua.cf["pressure"])
+
+    ua = ua.cf.sel(
+        pressure=slice(pmin, pmax),
+        latitude=slice(15, 75),
+        longitude=slice(lon_min, lon_max),
+    )
```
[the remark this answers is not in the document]
Thomasjkeel: I added this as a 'WARNING' in 76b674b02169e9daed3775a3def988675096f3ab. Hope this is okay?

xclim/indices/_synoptic.py
Thomasjkeel: I am unsure how the *indicator* part of the module exactly works, so I am not sure how to implement this. I believe this is related to why f5b7fbbe66ff6d64c7b9e493698fb5d27cb9c07d is failing. Happy to learn, so I can be more help, but alternatively, happy for anyone to take over and fix!

xclim/indicators/atmos/_synoptic.py
```diff
@@ -0,0 +1,22 @@
...
+"""Synoptic indicator definitions."""
+
+from xclim import indices
+from xclim.core.indicator import Daily
+
+__all__ = ["jetstream_metric_woolings"]
+
+
+class JetStream(Daily):
+    """Indicator involving daily u- and/or v-component wind series."""
```
aulemahal, line 11: I agree that this is confusing. In `xclim.core.indicator` we define multiple base `Indicator` classes. The most basic is `Indicator` and its most used child `ResamplingIndicator`. The latter is meant for indicators that resample along the `time` dimension, those that have a `freq` argument. Because that is the case of most indicators, for convenience and also for "historical" reasons (not wanting to make breaking changes), there are a few subclasses that inherit from `ResamplingIndicator` even if their name doesn't explicitly show it. It is the case of `Daily` which adds the condition that inputs have a daily frequency.

In your case, the `jetstream_metric_woolings` is _not_ resampling, so you'll want to use the `Indicator` base class directly. To have the daily input check, you can add `src_freq='D'` (default is `None` which deactivates any frequency checks).

```suggestion
from xclim.core.indicator import Indicator

__all__ = ["jetstream_metric_woolings"]


class JetStream(Indicator):
    """Indicator involving daily u- and/or v-component wind series."""
    
    src_freq = 'D'
```

xclim/indicators/atmos/_synoptic.py
```diff
@@ -0,0 +1,22 @@
...
+
+jetstream_metric_woolings = JetStream(
+    identifier="jetstream_metric_woolings",
+    units="days",
+    standard_name="daily_latitude_and_strength_of_jetstream",
+    long_name="Daily latitude and strength of maximum smoothed zonal wind speed",
+    description="{freq} daily latitude and strength of maximum smoothed zonal wind speed",
+    cell_methods="",
+    compute=indices.jetstream_metric_woolings,
+)
```
aulemahal, line 22:
```suggestion
jetstream_metric_woolings = JetStream(
    identifier="jetstream_metric_woolings",
    units=["degrees_North", "m s-1"],
    long_name=["Latitude of maximum smoothed zonal wind speed", "Maximum strength of smoothed zonal wind speed"],
    description=["Daily latitude of maximum smoothed zonal wind speed", "Daily maximum strength of smoothed zonal wind speed"],
    compute=indices.jetstream_metric_woolings,
)
```
As the "compute" function (the indice) returns 2 outputs (DataArrays), we must give metadata fields for each input. Passing list is one way, another would have been to pass a list of dictionaries like `cf_attrs=[{'units': ...}, {'units': ....}]`.

I removed the standard name as it was not from the official list. (http://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html)

The cell methods could theoretically be generated following the instructions in the [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#cell-methods), but xclim lacks a mechanism to inject the correct dimension names... Like, to be super exact we would want to have something like : "lon: mean pressure: mean lat: maximum", but with the correct dimension names instead of "lon", "lat" and "pressure". As we can't, I prefer removing "cell_methods" entirely.

xclim/testing/tests/test_indices.py
aulemahal: The problem was in the `declare_units` decorator! On exit, it converts the units to `pint` and then back to string, in order to ensure a CF-compliant output. However, `pint` thought "degrees_north" was simply an alias of "degrees". PR #959 fixes this by telling `pint` that it's a compatible but independent unit.

xclim/indices/_synoptic.py
```diff
@@ -0,0 +1,114 @@
...
+    Identify latitude and strength of maximum smoothed zonal wind speed in the region from 15 to 75°N and -60 to 0°E.
+
+    Warnings
+    --------
+    This metric expects eastward wind component (u) to be on a regular grid (i.e. Plate Carree, 1D lat and lon)
+
+    Parameters
+    ----------
+    ua : xarray.DataArray
+      eastward wind component (u) at between 750 and 950 hPa.
```
huard, line 34:
```suggestion
      Eastward wind component (u) at between 750 and 950 hPa.
```

xclim/indices/_synoptic.py
```diff
@@ -0,0 +1,114 @@
...
+    filter_freq = 10
+    window_size = 61
+    cutoff = 1 / filter_freq
+    if ua.time.size <= filter_freq or ua.time.size <= window_size:
+        raise ValueError(
+            f"Time series is too short to apply 61-day Lanczos filter (got a length of  {ua.time.size})"
+        )
+
+    # compute low-pass filter weights
+    lanczos_weights = compute_low_pass_filter_weights(
```
huard, line 85:
```suggestion
    lanczos_weights = _compute_low_pass_filter_weights(
```

xclim/indices/_synoptic.py
```diff
@@ -0,0 +1,114 @@
...
+        zonal_mean.rolling(time=window_size, center=True)
+        .construct("window")
+        .dot(lanczos_weights)
+    )
+    jetlat = ua_lf.cf.idxmax(lat_name).rename("jetlat").assign_attrs(units=lat_units)
+    jetstr = ua_lf.cf.max(lat_name).rename("jetstr").assign_attrs(units=ua_units)
+    return jetlat, jetstr
+
+
+def compute_low_pass_filter_weights(
```
huard, line 99:
```suggestion
def _compute_low_pass_filter_weights(
```

huard (approved):

xclim/indicators/atmos/_synoptic.py
```diff
@@ -0,0 +1,29 @@
...
+__all__ = ["jetstream_metric_woolings"]
+
+
+class JetStream(Indicator):
+    """Indicator involving daily u- and/or v-component wind series."""
+
+    src_freq = "D"
+
+
+jetstream_metric_woolings = JetStream(
```
huard, line 16:
```suggestion
jetstream_metric_woollings = JetStream(
```

xclim/indicators/atmos/_synoptic.py
```diff
@@ -0,0 +1,29 @@
...
+    identifier="jetstream_metric_woolings",
```
huard, line 17:
```suggestion
    identifier="jetstream_metric_woollings",
```

xclim/indicators/atmos/_synoptic.py
```diff
@@ -0,0 +1,29 @@
...
+    units=["degrees_North", "m s-1"],
+    long_name=[
+        "Latitude of maximum smoothed zonal wind speed",
+        "Maximum strength of smoothed zonal wind speed",
+    ],
+    description=[
+        "Daily latitude of maximum smoothed zonal wind speed",
+        "Daily maximum strength of smoothed zonal wind speed",
+    ],
+    compute=indices.jetstream_metric_woolings,
```
huard, line 28:
```suggestion
    compute=indices.jetstream_metric_woollings,
```

xclim/indices/_synoptic.py
```diff
@@ -0,0 +1,114 @@
...
+# ATTENTION: ASSUME ALL INDICES WRONG UNTIL TESTED ! #
+# -------------------------------------------------- #
+
+__all__ = [
+    "jetstream_metric_woolings",
+]
+
+
+@declare_units(ua="[speed]")
+def jetstream_metric_woolings(
```
huard, line 20:
```suggestion
def jetstream_metric_woollings(
```

xclim/testing/tests/test_indices.py
```diff
@@ -1045,6 +1045,62 @@ def test_below_simple(self, tasmax_series):
...
+            "units": "m s-1",
+        },
+    )
+
+    da_ua.Z.attrs = {"units": "Pa", "standard_name": "pressure"}
+    da_ua.X.attrs = {"units": "degrees_east", "standard_name": "longitude"}
+    da_ua.Y.attrs = {"units": "degrees_north", "standard_name": "latitude"}
+    da_ua.T.attrs = {"standard_name": "time"}
+
+    def test_jetstream_metric_woolings(self):
```
huard, line 1075:
```suggestion
    def test_jetstream_metric_woollings(self):
```

xclim/testing/tests/test_indices.py
```diff
@@ -1045,6 +1045,62 @@ def test_below_simple(self, tasmax_series):
...
+        da_ua = self.da_ua
+        # Should raise ValueError as longitude is in 0-360 instead of -180.E-180.W
+        with pytest.raises(ValueError):
+            _ = xci.jetstream_metric_woolings(da_ua)
```
huard, line 1079:
```suggestion
            _ = xci.jetstream_metric_woollings(da_ua)
```

xclim/testing/tests/test_indices.py
```diff
@@ -1045,6 +1045,62 @@ def test_below_simple(self, tasmax_series):
...
+        da_ua = da_ua.cf.assign_coords(
+            {
+                "X": (
+                    "X",
+                    (da_ua.cf["longitude"] - 180).data,
+                    da_ua.cf["longitude"].attrs,
+                )
+            }
+        )
+        out = xci.jetstream_metric_woolings(da_ua)
```
huard, line 1090:
```suggestion
        out = xci.jetstream_metric_woollings(da_ua)
```

Thomasjkeel:
Thanks for reviewing @huard. Just released that I should have pushed commit suggestions in one go (instead of individually in the browser), so sorry if you got sent a lot of emails just then (it is getting late in the UK!). Thanks everyone for all your help in getting this PR ready (up until this point), been a really valuable experience!

Zeitsperre (approved):

Zeitsperre:
@raquel-ucl @Thomasjkeel Thanks for the PR !

Issue #899, opened 2021-11-03
Masking and indexing as argument of indicator's call


aulemahal:
Proposition to inject the `**indexer` args to _all_ indicators. 

With this proposition, `[Resampling]Indicator`'s call would mask values along the `time` axis according to the instruction passed to the `indexer` kwargs. The masking would occur before the compute. The `missing` methods would need to be modified to take this masking into account.

If possible, the masked values could even be dropped, in order to accelerate computation.

For example:  `xc.atmos.tg_max(tas, season='MAM', freq='YS')` would return one value per year (on Jan 1st) of the maximal temperature in the spring months. Internally, it would look like `tas2 = tas.where(tas.time.dt.month.isin([3, 4, 5]))`.

But we could open up to fancier indexing: 
Ex: `xc.atmos.tg_max(tas, within_dates=['03-13', '05-14'], freq='YS')`. Again yearly values of maximal temperature but only considering days between March 13th and May 14th. 

Maybe even : 
```
start = xc.atmos.first_day_above(...)
end = xc.atmos.first_snowfall(...)
xc.atmos.tg_max(tas, within)_dates=(start, end), freq='YS')
```
?

I think this needs:

- [ ] A standalone function to mask a timeseries according to a tuple of dates (doy int , month-day str, arrays of doys). @yrouranos has a draft of this in #894. @RondeauG  has also a need for this. 
- [ ]  #867 to be done.
- [ ] A way for `missing` methods to take this into account.

It would fix a reccuring issues of users wanting to compute some indicators on clea-but-incomplete data or on subsets. It would on other hand either fill the signature with many arguments or, if we leave them as kwargs, forbid the use of kwargs by indices (which is not too bad).

RondeauG:
My need is slightly different, in the sense that I'm interested to apply a function on specific periods of the year (e.g. cumulative precipitation deficit over summer and fall), but the output should be a time series, not an aggregate.

yrouranos:
@aulemahal

I think that it would be a nice feature. In agriculture, there is often a need for masking. Weather conditions usually affect a specific part of the life cycle of a plant/culture (ex: growth, development of flowers/fruits, dormancy). Drought during rain season, too warm during dormancy, strong winds removing flowers from trees, etc.

Masking before computation may be a problem for a few climate indices. My feeling is that it could be problematic if there's a `window` involved (when the algorithm needs to consider values beyond the `within_dates` range).

[...]

PR #927, opened 2021-11-17
Resampling indicator

### LINKED ISSUE #867: ResamplingIndicator
# Proposition for a restructuring of the `Indicator` class

- Create a new `ResamplingIndicator` for indicators that resample the data (most of them), those which have a `freq` argument.
- Remove all resampling-related operations from the base `Indicator` class.  This includes : missing values handling, "allowed_periods" check and handling of the `indexer` kwargs.
- The latter is currently only implemented on a few indicators, but we could generalize its use to _all_ resampling indicator. Thus, for all indicator when it makes sense, user could pass things like: `season='DJF'`. This would resolve an issue that pops up often in different scenarios : how to reduce the computation load without making `check_daily` fail.

### LINKED ISSUE #898: Allowed input freq as a parameter
Proposition to have indicators' datachecks using the class attribute `freq` when checking input's frequency.

By this I mean scraping the class-inheritance pattern of `Daily`, `Hourly` and `DailyWeeklyMonthy`. This will allow for a finer control over the accepted input freq for specific indicators. 

For example, it could allow whole modules, like `icclim` to allow weekly and monthly data. 

The default would stay `freq = 'D'`, but could be changed from indicator to indicator without having to inherit from a new base class. Ex:

```
from xclim.core.indicator import registry

tg_max_bis = registry['TG_MAX'](
    identifier="tg_max2",
    long_name="Maximum monthly mean temperature",
    description="{freq} maximum of monthly mean temperature.",
    cell_methods="time: mean within days time: maximum over months",
)
```
This creates a new `tg_max_bis` indicator that only accepts monthly input. The previous way was:
```
tg_max_bis = Monthly(
    identifier="tg_max",
    units="K",
    standard_name="air_temperature",
    long_name="Maximum monthly mean temperature",
    description="{freq} maximum of monthly mean temperature.",
    cell_methods="time: mean within days time: maximum over months",
    compute=indices.tg_max,
)
```
But there is no such thing as `Monthly` it would have to be created.

On a side note, this proposition removes _all_  conventional class-inheritance pattern from the creation of indicators. From now on, the only way to create a new indicator is through initialization of an existing class.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #867 and fixes #898

* Split all resampling-related functionalities of the indicators into a new `ResamplingIndicator` subclass. I did this by creating private pre- and post-processing methods that are called in `__call__`. This makes it easy to "inject" functionalities in the indicator's processing. The "resampling-related functionalities" are : checking for the allowed "freq" arg and handling of missing values.
* Have the data input frequency expected by the indicator stored in a class attribute. This makes it easy to modify it when creating indicators. This improves the changes in #885.

I don't think so.

### Other information:

I think that with this change it becomes a bit clearer what are the uses cases for conventional subclassing of the `Indicator` class:
1. Add functionality by overriding one of the pre- or post-process methods.
2. Convenient way of modifying the following attributes : `context`, `src_freq`, `realm`, `title`, `abstract`, `keywords`, `references`, `notes` and `allowed_periods`. 
In both cases, the goal is to create a new _base_ class and _then_ to create multiple indicators.

I'll put this in a "developer's note" section of the doc in a further PR.

 This is also a first step towards a way to inject the `indexer` kwargs as planned in #899.

### DIFF
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -1279,7 +1279,7 @@ class ResamplingIndicator(Indicator):
         super().__init__(**kwds)
 
     def _preprocess_and_checks(self, das, params, indexer):
-        """Perform parent's checks and also checks if freq is allowed."""
+        """Perform parent's checks and also check if freq is allowed."""
         das, params, indexer = super()._preprocess_and_checks(das, params, indexer)
 
         # Check if the period is allowed:
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -92,6 +92,7 @@ import weakref
 from collections import OrderedDict, defaultdict
 from copy import deepcopy
 from dataclasses import asdict, dataclass
+from functools import reduce
 from inspect import Parameter as _Parameter
 from inspect import Signature
 from inspect import _empty as _empty_default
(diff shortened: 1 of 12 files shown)


aulemahal:
@bzah With this PR, I modified the way one can specifiy which input frequencies are accepted. See the `data/anuclim.yml` for an example. I hope it doesn't break your code too much!

The good part is that one can now inherit from an indicator created from `Daily` and still modifiy `src_freq`. `Daily` is only a convenience class, it now doesn't hardcode the frequency check, like it did before.

bzah:
Thank you for your concern (and for the feature!) @aulemahal, it won't break my code I'm using atmos daily indicators directly and I have simply disabled the daily check for now on icc v5.
After this PR I might inherit them and change `src_freq`, but there is an ongoing discussion to see which indices/indicators should accept other frequencies.

xclim/core/indicator.py
```diff
@@ -1285,47 +1232,109 @@ def injected_parameters(self):
...
+        self._missing = kls.execute
+        if self.missing_options:
+            kls.validate(**self.missing_options)
 
+        super().__init__(**kwds)
 
-class DailyWeeklyMonthly(Indicator):
-    """Indicator defined for inputs at daily, weekly or monthly frequencies.
+    def _preprocess_and_checks(self, das, params, indexer):
+        """Perform parent's checks and also checks if freq is allowed."""
```
huard, line 1282:
```suggestion
        """Perform parent's checks and also check if freq is allowed."""
```

huard (approved):
Good for me.

PR #934, opened 2021-11-24
Indexing within indicators

### LINKED ISSUE #899: Masking and indexing as argument of indicator's call

aulemahal:
### Pull Request Checklist:
    - This PR fixes #899

* Creates another base Indicator class, `ResamplingIndicatorWithIndexing`, that injects the `indexer` kwargs, as it was done in some "stats" function. The indexing is done in the `preprocess_and_check` function, so the indice (`compute`) receives an input where values outside the select period are **masked**.
* Expand `generic.select_time` with `doy_bounds` and `date_bounds`. The first takes in bounds as day-of-year integers and the second takes in dates in the month-day format (`%m-%d`). 

* Indexing now _masks_ values by default, instead of dropping them. There are issues when values are dropped, some indices do not expect those gaps. The principal culprit is `units.rate2amount` which converts by multiplying the rate by the timestep length, which is inferred from the coordinate itself. In case of a gap, the rate is assumed constant on the whole gap...
* Most indicators now have the new `**indexer`  in their signature.

### Other information:
An indice may add an `**indexer` var_keyword and handle the time selecting itself. If the indicator is constructed from `ResamplingIndicator` (the one that _doesn't_ inject this new param), the missing values handling will still understand the passed indexing parameter. This can be useful when the order between indexing and computing is important (like for #884 and many others).
This means that the order is currently "wrong" for many cases: i.e. indexing is done before computation. In a way, if it is made clear what the default order is, I think it's ok to have this caveat temporarily. But should we: 
1. Implement these cases now, in this PR.
2. Let them be wrong for now and correct them later.
3. Revert injecting `**indexer` everywhere and slowly implement it where it makes most sense afterwards.
?

### DIFF
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -1348,7 +1348,7 @@ class ResamplingIndicatorWithIndexing(ResamplingIndicator):
 
     @classmethod
     def _injected_parameters(self):
-        return super()._injected_parameters + [
+        return super()._injected_parameters() + [
             (
                 "indexer",
                 Parameter(
@@ -1366,20 +1366,20 @@ class ResamplingIndicatorWithIndexing(ResamplingIndicator):
         """Perform parent's checks and also check if freq is allowed."""
         das, params = super()._preprocess_and_checks(das, params)
 
-        indexer = params.get("indexer")
-        if indexer:
-            # do things
-            pass
+        indxr = params.get("indexer")
+        if indxr:
+            das = {k: indices.generic.select_time(da, **indxr) for k, da in das.items()}
+            print(das)
         return das, params
 
 
-class Daily(ResamplingIndicator):
+class Daily(ResamplingIndicatorWithIndexing):
     """Class for daily inputs and resampling computes."""
 
     src_freq = "D"
 
 
-class Hourly(ResamplingIndicator):
+class Hourly(ResamplingIndicatorWithIndexing):
     """Class for hourly inputs and resampling computes."""
 
     src_freq = "H"
@@ -1387,6 +1387,7 @@ class Hourly(ResamplingIndicator):
 
 base_registry["Indicator"] = Indicator
 base_registry["ResamplingIndicator"] = ResamplingIndicator
+base_registry["ResamplingIndicatorWithIndexing"] = ResamplingIndicatorWithIndexing
 base_registry["Hourly"] = Hourly
 base_registry["Daily"] = Daily
 
--- a/xclim/core/indicator.py
+++ b/xclim/core/indicator.py
@@ -1339,7 +1339,9 @@ class ResamplingIndicator(Indicator):
             )
             # Reduce by or and broadcast to ensure the same length in time
             # When indexing is used and there are no valid points in the last period, mask will not include it
-            mask = reduce(np.logical_or, miss).reindex_like(outs[0], fill_value=True)
+            mask = reduce(np.logical_or, miss)
+            if isinstance(mask, DataArray) and mask.time.size < outs[0].time.size:
+                mask = mask.reindex(time=outs[0].time, fill_value=True)
             outs = [out.where(~mask) for out in outs]
 
         return outs
@@ -1357,8 +1359,8 @@ class ResamplingIndicatorWithIndexing(ResamplingIndicator):
                     kind=InputKind.KWARGS,
                     description=(
                         "Indexing parameters to compute the indicator on a temporal "
-                        "subset of the data, see ... for details on how to use this "
-                        "parameter."
+                        "subset of the data. It accepts the same arguments as "
+                        ":py:func:`xclim.indices.generic.select_time`."
                     ),
                 ),
             )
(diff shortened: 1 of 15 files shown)


xclim/core/indicator.py
```diff
@@ -1309,30 +1331,64 @@ def _postprocess(self, outs, das, params, indexer):
...
+                self._missing(
+                    da, params["freq"], src_freq, options, params.get("indexer", {})
+                )
                 for da in das.values()
                 if "time" in da.coords
             )
-            mask = reduce(np.logical_or, miss)
+            # Reduce by or and broadcast to ensure the same length in time
+            # When indexing is used and there are no valid points in the last period, mask will not include it
+            mask = reduce(np.logical_or, miss).reindex_like(outs[0], fill_value=True)
```
aulemahal, line 1342: In the case where the last period contains no points after indexing, it doesn't appear in `mask` and so it was dropped from the result, which is a breaking change in a way. 
However, this `reindex_like` trick adds complexity, but fixes the problem.

I would prefer a breaking change than this.

Ex: If you do something with a time series from Jan 1st to Dec 31st, then pass `freq='AS=JUL'` and indexer : `month=2`. The second period, is completely dropped from the mask as it contains no points. Before, it wasn't dropped, rather it was masked out.

xclim/indices/generic.py
```diff
@@ -57,31 +59,120 @@
...
+
+        # Get doy of date, this is now safe because the calendar is uniform.
+        start = to_cftime_datetime("2000-" + start, calendar).dayofyr
+        end = to_cftime_datetime("2000-" + end, calendar).dayofyr
+
+        if start <= end:
+            doys = np.arange(start, end + 1)
+        else:
+            doys = np.concatenate((np.arange(start, 367), np.arange(0, end + 1)))
+        mask = time.time.dt.dayofyear.isin(doys)
```
huard, line 172: Could you just set `doy_bounds` and move the `elif doy_bounds is not None:` code segment below to avoid repeating ?

huard (approved):

xclim/core/indicator.py
```diff
@@ -1133,13 +1155,13 @@ def _format(
...
-            mba["indexer"] = dv
-        else:
-            mba["indexer"] = "annual"
+        # if indexer:
+        #     dk, dv = indexer.copy().popitem()
+        #     if dk == "month":
+        #         dv = "m{}".format(dv)
+        #     mba["indexer"] = dv
+        # else:
+        #     mba["indexer"] = "annual"
```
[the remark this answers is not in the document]
aulemahal: Code is semi obsolete, but functionality needs to be implemented back (that's the failing tests).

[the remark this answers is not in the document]
aulemahal: Updated, please have a look.

I leave to the AttrFormatter the role of translating the "m{X}", season or freq string into words. 
Current code has the issue that `doy_bounds` and  `date_bounds` are suboptimally stated and currently not translatable.

huard (approved):

xclim/core/indicator.py
```diff
@@ -1131,15 +1153,18 @@ def _format(
...
+            elif k == "indexer":
+                if v and v not in [_empty, _empty_default]:
+                    dk, dv = v.copy().popitem()
+                    if dk == "month":
+                        dv = f"m{dv}"
+                    elif dk in ("doy_bounds", "date_bounds"):
+                        dv = f"{dv[0]} to {dv[1]}"
+                    mba["indexer"] = dv
+                else:
+                    mba["indexer"] = args.get("freq") or indices.generic.default_freq()
```
aulemahal, line 1165: Woups, I realize the mistake. `v` is not a correct indexer, that's the point!

But then, I'm not sure of the need of `default_freq`? 
```suggestion
                    mba["indexer"] = args.get("freq") or "YS"
```

PR #935, opened 2021-11-25
Adapt to CF conventions 1.9 - remove gregorian calendar

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Deprecates the calendar name "gregorian". Uses "standard" instead, as stated in the version 1.9 of the CF conventions (Sept 2021). In this we simply follow `cftime` that made this change yesterday (Unidata/cftime#257).
* Changes to `get_calendar` and `ensure_ctime_array` to fix a bug I saw while making this change. Both were failing in some edgecases involving `xr.CFTimeIndex` objects.

Yes, `get_calendar` will never return "gregorian", but "standard".

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -13,7 +13,7 @@ New features and enhancements
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
-* Following version 1.9 of the CF Conventions, publised in September 2021, the calendar name "gregorian" is deprecated. ``core.calendar.get_calendar`` will return "standard", even if the underlying cftime objects still use "gregorian" (cftime <= 1.5.1).
+* Following version 1.9 of the CF Conventions, publised in September 2021, the calendar name "gregorian" is deprecated. ``core.calendar.get_calendar`` will return "standard", even if the underlying cftime objects still use "gregorian" (cftime <= 1.5.1). (:pull:`935`).
 
 Internal changes
 ~~~~~~~~~~~~~~~~
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -11,6 +11,10 @@ New features and enhancements
 * Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``. (:pull:`911`, :issue:`910`).
 * The data input frequency expected by ``Indicator``s is now in the ``src_freq`` attribute and is thus controlable by subclassing existing indicators. (:issue:`898`, :pull:`927`).
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+* Following version 1.9 of the CF Conventions, publised in September 2021, the calendar name "gregorian" is deprecated. ``core.calendar.get_calendar`` will return "standard", even if the underlying cftime objects still use "gregorian" (cftime <= 1.5.1).
+
 Internal changes
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -11,6 +11,10 @@ New features and enhancements
 * Added an optimized pathway for ``xclim.indices.run_length`` functions when ``window=1``. (:pull:`911`, :issue:`910`).
 * The data input frequency expected by ``Indicator``s is now in the ``src_freq`` attribute and is thus controlable by subclassing existing indicators. (:issue:`898`, :pull:`927`).
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+* Following version 1.9 of the CF Conventions, publised in September 2021, the calendar name "gregorian" is deprecated. ``core.calendar.get_calendar`` will return "standard", even if the underlying cftime objects still use "gregorian" (cftime <= 1.5.1). (:pull:`935`).
```
huard, line 16:
```suggestion
* Following version 1.9 of the CF Conventions, published in September 2021, the calendar name "gregorian" is deprecated. ``core.calendar.get_calendar`` will return "standard", even if the underlying cftime objects still use "gregorian" (cftime <= 1.5.1). (:pull:`935`).
```

huard (approved):

xclim/core/calendar.py
```diff
@@ -450,14 +454,16 @@ def ensure_cftime_array(
 ) -> Union[CFTimeIndex, np.ndarray]:
```
aulemahal, line 454:
```suggestion
) -> np.ndarray:
```

xclim/core/calendar.py
```diff
@@ -70,19 +70,23 @@ def get_calendar(obj: Any, dim: str = "time") -> str:
...
+    elif isinstance(obj, xr.CFTimeIndex):
+        obj = obj.values[0]
+    else:
+        obj = np.take(obj, 0)
+        # Take zeroth element, overcome cases when arrays or lists are passed.
     if isinstance(obj, pydt.datetime):  # Also covers pandas Timestamp
         return "default"
     if isinstance(obj, cftime.datetime):
+        if obj.calendar == "gregorian":
+            return "standard"
```
Zeitsperre, line 89: Perhaps this comment is out of scope, but would we want to emit a warning here about this change of calendar? I don't think we specify any output `Conventions` to our datasets. Should we be appending this information to our output Datasets?

Zeitsperre (approved):

xclim/core/calendar.py
aulemahal: I think we do not add this attribute, simply because we don't output Datasets. If we ever do, than I guess we could append the info. And if we do, we'll have to be even more careful about this kind of change.

It's a good question. But I believe that this change doesn't need a warning : 
  1. CF conventions have always accepted "standard" as equivalent to "gregorian", they just dropped the latter recently.
  2. Code might "break", but only with cftime <= 1.5.1. For people using the forthcoming cftime 1.5.2, this break will be generalized and we will be able to shift responsability to them ;).

Zeitsperre: If that's the case, is this an event where we should bump the required version of CFtime within xclim?

aulemahal: It's not released yet! That bug was found because of our smart tox builds using the master!

PR #940, opened 2021-11-26
Improve rxxp indices

### LINKED ISSUE #931: Feature request: Provide the ECAD definition of R75p R95p and R99p indices.
* xclim version: 0.31.x
* Python version: all
* Operating System: all

### Description

Currently, the indices R75p R95p and R99p (as per icclim.yml file) rely on `days_over_precip_thresh` function which expect in `per` parameter a DataArray of **doy** percentiles.

However, when reading the definition from [ECAD](https://www.ecad.eu/documents/atbd.pdf#subsubsection.5.3.8) documentation:
> R75p
• Days with RR > 75th percentile of daily amounts (moderate wet days)
(days)
Let RRwj be the daily precipitation amount at wet day w (RR ≥ 1.0 mm)
of period j and let RRwn75 be the 75th percentile of precipitation at wet
days in the 1961–1990 period. Then counted is the number of days where:
RRwj > RRwn75

As per this definition, the percentiles should be computed on the whole base period (above 61-90) and not for each day of year.

Thus, I would like to implement an index function similar to `days_over_precip_thresh` with a `thresh` DataArray parameter having values thresholds for each spatial point.
Or maybe simply replace `days_over_precip_thresh` but both definition could make sens I suppose.
icclim.yml should be updated to probably follow ECAD definition as icclim v4 does.

### What I Did

I compared the output of icclimv4 and found very different results.

Notes: 
- **I'm willing to work on this issue.**
- climdex/climpact use again another definition for r99p (and others indices). It sum the precipitation amounts above the percentile thresholds and the percentiles are **not** for doy.
- The subject of RxxP definition is discussed on clix-meta: https://github.com/clix-meta/clix-meta/issues/53.
- The [WMO definition](https://library.wmo.int/doc_num.php?explnum_id=9419) for R95p is 
>R95p [...] the number of days with rainfall above the 95th percentile of daily accumulations.

Thus current `days_over_precip_thresh` is following a valid definition, the one from WMO, but it differ from icclim which uses ECAD definition.

bzah:
### Pull Request Checklist:
    - This PR fixes #931

- Add new indicator `days_over_precip_doy_thresh` for the WMO definition of Rxxp indices.
The existing `days_over_precip_thresh` (notice ~doy~) description is now made for the ECA&D definiton of Rxxp indices.
Note that both indicators call the same **indice** function, only the description is changed.
- Also added some error handling when non doy percentiles are used in BS.

Not a breaking change, but those who calls `days_over_precip_thresh` with `per` computed using doy values will receive an improper description for the indicator. 
They should start using `days_over_precip_doy_thresh` for this purpose.

### Other information:
ECAD (pdf): https://www.ecad.eu/documents/atbd.pdf
WMO: https://library.wmo.int/index.php?lvl=notice_display&id=138#.YUBV6CVS9ZI

### DIFF
--- a/xclim/indicators/atmos/_precip.py
+++ b/xclim/indicators/atmos/_precip.py
@@ -316,7 +316,7 @@ high_precip_low_temp = PrTasx(
 )
 
 
-fraction_over_precip_thresh = Precip(
+fraction_over_precip_doy_thresh = Precip(
     identifier="fraction_over_precip_thresh",
     description="{freq} fraction of total precipitation due to days with precipitation above a daily percentile."
     " Only days with at least {thresh} are included in the total.",
@@ -325,6 +325,14 @@ fraction_over_precip_thresh = Precip(
     compute=indices.fraction_over_precip_thresh,
 )
 
+fraction_over_precip_thresh = Precip(
+    identifier="fraction_over_precip_thresh",
+    description="{freq} fraction of total precipitation due to days with precipitation above percentile."
+    " Only days with at least {thresh} are included in the total.",
+    units="",
+    cell_methods="",
+    compute=indices.fraction_over_precip_thresh,
+)
 
 liquid_precip_ratio = PrTasx(
     identifier="liquid_precip_ratio",
--- a/xclim/indicators/atmos/_precip.py
+++ b/xclim/indicators/atmos/_precip.py
@@ -290,13 +290,22 @@ days_with_snow = Precip(
 days_over_precip_thresh = Precip(
     identifier="days_over_precip_thresh",
     standard_name="number_of_days_with_lwe_thickness_of_precipitation_amount_above_threshold",
-    description="{freq} number of days with precipitation above a daily percentile."
+    description="{freq} number of days with precipitation above percentile."
     " Only days with at least {thresh} are counted.",
     units="days",
     cell_methods="time: sum over days",
     compute=indices.days_over_precip_thresh,
 )
 
+days_over_precip_doy_thresh = Precip(
+    identifier="days_over_precip_doy_thresh",
+    standard_name="number_of_days_with_lwe_thickness_of_precipitation_amount_above_daily_threshold",
+    description="{freq} number of days with precipitation above a daily percentile."
+    " Only days with at least {thresh} are counted.",
+    units="days",
+    cell_methods="time: sum over days",
+    compute=indices.days_over_precip_thresh,
+)
 
 high_precip_low_temp = PrTasx(
     identifier="high_precip_low_temp",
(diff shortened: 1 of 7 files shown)


xclim/indicators/atmos/_precip.py
```diff
@@ -290,13 +290,22 @@ class HrPrecip(Hourly):
...
     standard_name="number_of_days_with_lwe_thickness_of_precipitation_amount_above_threshold",
-    description="{freq} number of days with precipitation above a daily percentile."
+    description="{freq} number of days with precipitation above percentile."
     " Only days with at least {thresh} are counted.",
     units="days",
     cell_methods="time: sum over days",
     compute=indices.days_over_precip_thresh,
 )
 
+days_over_precip_doy_thresh = Precip(
```
bzah, line 300: I forgot to do the same on `fraction_over_precip_thresh`.

xclim/indices/_multivariate.py
```diff
@@ -1106,7 +1106,8 @@ def days_over_precip_thresh(
     pr : xarray.DataArray
       Mean daily precipitation flux.
     per : xarray.DataArray
-      Daily percentile of wet day precipitation flux.
+      Percentile of wet day precipitation flux. Either computed daily (one value per day
+      of year) or computed over a period (one value per spatial point).
```
huard, line 1110: I guess this should also apply to `fraction_over_precip_thresh`.

huard (approved):

xclim/indicators/atmos/_precip.py
```diff
@@ -307,7 +316,7 @@ class HrPrecip(Hourly):
 )
 
 
-fraction_over_precip_thresh = Precip(
+fraction_over_precip_doy_thresh = Precip(
     identifier="fraction_over_precip_thresh",
```
aulemahal, line 320: This identifier should be different than the one of `fraction_over_precip_thresh`.

xclim/indices/_multivariate.py
```diff
@@ -1189,6 +1195,10 @@ def fraction_over_precip_thresh(
     if "dayofyear" in per.coords:
         # Create time series out of doy values.
         tp = resample_doy(tp, pr)
+    else:
```
aulemahal, line 1198: Shouldn't it be `elif bootstrap:` ?

aulemahal (commented):
Sorry, I should have looked into this with more care before the merge. There's 2 issues I see. The first is triggering warnings each time xclim is loaded. The second I have not tested.

xclim/indices/_multivariate.py
bzah: Yes it should, oopsi

PR #941, opened 2021-11-29
[sdba] Refactor interp_on_quantiles

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

I though the changes in #914 would address a bug I saw while computing ESPO-R. However, I had forgotten about the splitting pathways for `group='time'` vs `group='time.prop'`. That also meant testing was not sufficient.  This PR changes `xclim.sdba.utils.interp_on_quantiles` so that:

* All NaNs in the coordinates are dropped. Minor modification from the Extremes' PR, but dropping everything explicitly, instead of relying on `interp1d`, has the advantage that NaNs in the new coord are not considered "out of bounds", but NaNs. They return NaN in the output, not an extrapolated value.
* All NaNs in the new coordinates are dropped. In the 2D case, with 'nearest', NaNs would cause QHull errors.

* Not really.

### Other information:
The docstring was improved.

I realized that `method='cubic'` does not respect the constant extrapolation as set by `extrapolate_qm`... I added a note to the doc, but I'm not sure what to do here. @huard ?  The logic we used works well with 'nearest' and 'linear', but as 'cubic' fetches a third point it estimates something else.

EDIT:
I compared the performance before and after the change and it is insignificant. `griddata` and `interp1d` are much slower than any numpy overhead I added, so I did not see any timing difference.

### DIFF
--- a/xclim/sdba/utils.py
+++ b/xclim/sdba/utils.py
@@ -392,6 +392,70 @@ def add_endpoints(
     return ensure_chunk_size(out, **{dim: -1})
 
 
+def _interp_on_quantiles_1D(newx, oldx, oldy, method, extrap):
+    mask_new = np.isnan(newx)
+    if np.all(mask_new):
+        warn(
+            "All-NaN slice encountered in interp_on_quantiles",
+            category=RuntimeWarning,
+        )
+        return newx
+
+    if extrap == "constant":
+        # This is not needed with nearest, as it is the default behaviour
+        fill_value = (oldy[0], oldy[-1])
+    else:  # extrap == 'nan'
+        fill_value = np.NaN
+
+    mask_old = np.isnan(oldy) | np.isnan(oldx)
+
+    out = np.full_like(newx, np.NaN, dtype=f"float{oldy.dtype.itemsize * 8}")
+    out[~mask_new] = interp1d(
+        oldx[~mask_old],
+        oldy[~mask_old],
+        kind=method,
+        bounds_error=False,
+        fill_value=fill_value,
+    )(newx[~mask_new])
+    return out
+
+
+def _interp_on_quantiles_2D(newx, newg, oldx, oldy, oldg, method, extrap):  # noqa
+    mask_new = np.isnan(newx) | np.isnan(newg)
+    if np.all(mask_new):
+        warn(
+            "All-NaN slice encountered in interp_on_quantiles",
+            category=RuntimeWarning,
+        )
+        return newx
+
+    mask_old = np.isnan(oldy) | np.isnan(oldx) | np.isnan(oldg)
+
+    out = np.full_like(newx, np.NaN, dtype=f"float{oldy.dtype.itemsize * 8}")
+    out[~mask_new] = griddata(
+        (oldx[~mask_old], oldg[~mask_old]),
+        oldy[~mask_old],
+        (newx[~mask_new], newg[~mask_new]),
+        method=method,
+    )
+
+    def first_and_last_nonnull(vector):
+        return vector[np.where(~np.isnan(vector))[0][np.array([0, -1])]]
+
+    if (method == "nearest" and extrap == "nan") or extrap == "constant":
+        igrp = np.searchsorted(oldg[:, 0], np.round(newg).astype(int))
+        toolow = newx < np.nanmin(oldx, axis=-1)[igrp]
+        toohigh = newx > np.nanmax(oldx, axis=-1)[igrp]
+        if extrap == "constant":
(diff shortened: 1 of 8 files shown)


xclim/sdba/utils.py
```diff
@@ -417,35 +422,39 @@ def interp_on_quantiles(
       If it does, interpolation is done in 2D on "quantiles" and on the group dimension.
     group : Union[str, Grouper]
       The dimension and grouping information. (ex: "time" or "time.month").
-      Defaults to the "group" attribute of xq, or "time" if there is none.
+      Defaults to "time".
     method : {'nearest', 'linear', 'cubic'}
-      The interpolation method.
+      The interpolation method. Using cubic interpolation will NOT respect the constant
+      extrapolation as added by :py:func:`~xclim.sdba.utils.extrapolate_qm`.
```
huard, line 428: You mean `interp_on_quantiles` will in fact extrapolate ? And this extrapolation is going to be cubic instead of constant. If the extrapolation used nan, then it's ok ? The cubic extrapolation will still be nan, right ?

huard (approved):

xclim/sdba/utils.py
```diff
@@ -459,31 +468,38 @@ def _interp_quantiles_1D(newx, oldx, oldy):
...
             warn(
                 "All-NaN slice encountered in interp_on_quantiles",
                 category=RuntimeWarning,
             )
             return _newx
 
-        mask = np.isnan(_oldx) & np.isnan(_oldy) & np.isnan(_oldg)
-        mask[:, np.array([0, -1])] = False  # bounds might be NaN for a good reason.
+        if method != "nearest":
+            _oldx = np.clip(_oldx, _newx.min() - 1, _newx.max() + 1)
```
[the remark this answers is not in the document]
aulemahal: With non-nearest methods,  griddata uses `qhull` and constructs a "convex hull" with the coordinates before performing the interpolation. When we extrapolate with "constant", the bounds of `_oldx` are ±inf, and the convex hull creation fails.
I will add a comment.

xclim/sdba/utils.py
```diff
@@ -459,31 +468,38 @@ def _interp_quantiles_1D(newx, oldx, oldy):
...
 
-        return griddata(
-            (_oldx[~mask], _oldg[~mask]),
-            _oldy[~mask],
-            (_newx, _newg),
+        mask_old = np.isnan(_oldy)
+        mask_old[:, np.array([0, -1])] = False  # bounds might be NaN for a good reason.
+        mask_old = mask_old | np.isnan(_oldx) | np.isnan(_oldg)
+
+        out = np.full_like(_newx, np.NaN, dtype=_oldy.dtype)
```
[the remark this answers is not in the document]
aulemahal: I'm not sure what interpolation on integer data would even mean, but I guess that is a possibility if the data is weird and the user unaware. I'll force a float dtype.

xclim/sdba/utils.py
aulemahal: Yes, it will extrapolate. 'cubic' uses 3 points, It means the fitted spline for out-of-bounds data is not flat as it includes the second-to-last real point in addition to the last and its copy. I didn't try with `nan`, will do.

aulemahal: Woups. When `extrapolate='nan', method='cubic'`, output is NaN everywhere.

Well. I'll fix that too.

aulemahal:
Last commits went deeper and refactored `interp_on_quantiles`. 

- Extrapolation is now handled by that function, without need to use `extrapolate_qm`. Changes to the adjustment methods are internal and nothing has changed in the API for the normal user.
- However, extrapolation is now better handled. With `interp='cubic'`, the constant extrapolation is fixed.
- On a side note, I removed the `interp` argument of `Grouper`'s init. It served almost no purpose, introducing more noise than doing good. It stays on `Grouper.get_index` as boolean only. It only matters with monthly grouping, so... almost never. 
- A bit of code refactor.

Should we remove `extrapolate_qm`? It is not used in the adjustment methods anymore.

huard (approved):
Does it make sense to repurpose test_extrapolate_qm to test the new extrapolation code ? I know its implicitly tested elsewhere, but still.

aulemahal:
The new extrapolation code is explicitly tested in `test_interp_on_quantiles`!  (see `expe`).

I would rather suggest to _remove_ `extrapolate_qm`? I liked the way it dealt with the situation, but sadly `griddata` demands more complexity...

huard:
All good then. 
Ok for removal, I would be surprised anyone used that function directly. If we want to do it by the book, we could deprecate it for the next release before removal.

PR #942, opened 2021-11-30
Improve quantile function

bzah:
### Pull Request Checklist:
    - This PR fixes #xyz

This is the follow up PR to benefit from the review which happened on numpy for the same feature.

- Rename to follow np naming
- Update documentation
- Refactor code to improve readability
- Handle some edge cases with np `_linear_interpolation` implementation
- May improve perfs on some cases (when axis=0)

Nope.

### Other information:
It is _still interesting_ to keep this version in xclim and not rely on np because of the performances improvements it provides.
Indeed, I was not able to include these perfs changes in np because it introduced a lot of refactoring and some design questions (in short, it merges `np.quantile` and `np.nanquantile` in a single function).

However depending on the array shape the performances are not as great as I first thought:
np.nanquantile rely on apply_along_axis which tends to be slow if all axis are quite large like `(180, 180 , 3000)`
```
np.percentile: 3.422793875 secs
xc.nan_calc_percentiles: 4.604135543 secs
np.nanpercentile: 27.327997681999996 secs
```
But np.nanpercentile can be very fast with shapes like `(18, 18, 30000)`.
It can even be faster than np.percentile (but gives different results obviously).
```
np.percentile: 2.55501401 secs
xc.nan_calc_percentiles: 7.295944888000001 secs
np.nanpercentile: 2.2845503559999987 secs
```

### DIFF
--- a/xclim/testing/tests/test_ensembles.py
+++ b/xclim/testing/tests/test_ensembles.py
@@ -145,19 +145,19 @@ class TestEnsembleStats:
             ens = ens.transpose()
 
         out1 = ensembles.ensemble_percentiles(ens, split=True)
-        np.testing.assert_array_equal(
+        np.testing.assert_array_almost_equal(
             mquantiles(
                 ens["tg_mean"].isel(time=0, lon=5, lat=5), 0.1, alphap=1, betap=1
             ),
             out1["tg_mean_p10"].isel(time=0, lon=5, lat=5),
         )
-        np.testing.assert_array_equal(
+        np.testing.assert_array_almost_equal(
             mquantiles(
                 ens["tg_mean"].isel(time=0, lon=5, lat=5), alphap=1, betap=1, prob=0.50
             ),
             out1["tg_mean_p50"].isel(time=0, lon=5, lat=5),
         )
-        np.testing.assert_array_equal(
+        np.testing.assert_array_almost_equal(
             mquantiles(
                 ens["tg_mean"].isel(time=0, lon=5, lat=5), alphap=1, betap=1, prob=0.90
             ),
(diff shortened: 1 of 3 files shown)


xclim/testing/tests/test_ensembles.py
```diff
@@ -145,19 +145,19 @@ def test_calc_perc(self, transpose):
...
         out1 = ensembles.ensemble_percentiles(ens, split=True)
-        np.testing.assert_array_equal(
+        np.testing.assert_array_almost_equal(
             mquantiles(
                 ens["tg_mean"].isel(time=0, lon=5, lat=5), 0.1, alphap=1, betap=1
             ),
             out1["tg_mean_p10"].isel(time=0, lon=5, lat=5),
         )
-        np.testing.assert_array_equal(
+        np.testing.assert_array_almost_equal(
```
bzah, line 154: The error comes from an unpacking of scalar values I have added, I will try to fix this.

bzah: Oh no it's not the unpacking, I don't get why there is ~10e-14 difference between the two...

aulemahal: That is very small, could it be just some float32-float64 conversion?

bzah: Yeah you are right, `ens["tg_mean"].dtype` is float32 but `mquantiles` result is float64 so it is very likely a conversion issue.
Thanks @Zeitsperre for the fix!

Zeitsperre (approved):
@bzah All this needs is your name on the contributors for 0.32, and it's good to go. Thanks for PR!

PR #1009, opened 2022-02-04
[sdba] Fix constant extrapolation for monthly grouping and with interpolation

### LINKED ISSUE #1008: Nans are added when interpolation of adjustment factors for quantile mapping when grouper is 'time.month'
* xclim version: 0.33.0
* Python version: 3.8.10
* Operating System: Ubuntu 20.04

### Description
Hi, thanks for this great package that I'm using a lot! I ran into a small issue with the quantile mapping routines which is most likely a bug. I'm applying the empirical quantile mapping (sdba.EmpiricalQuantileMapping) and found numerous 'NaN' values in the quantile mapped data which were not in the input data. This only happens when the group is `time.month` and I use either the 
`linear` or `cubic` interpolation method in the `.adjust` function. The nans are mostly present with values at the edge of distribution. Also when there is no trend (comment out `+ 0.1 * (t - t[0]).days / 365`) there are still 3 nans present.

### What I Did (minimum reproducable example)
Taking the example data from https://xclim.readthedocs.io/en/stable/notebooks/sdba.html

```
import numpy as np
import xarray as xr
import cftime
from xclim import sdba

# Create toy data to explore bias adjustment, here fake temperature timeseries
t = xr.cftime_range("2000-01-01", "2030-12-31", freq="D", calendar="noleap")
ref = xr.DataArray(
    (
        -20 * np.cos(2 * np.pi * t.dayofyear / 365)
        + 2 * np.random.random_sample((t.size,))
        + 273.15
        + 0.1 * (t - t[0]).days / 365
    ),  # "warming" of 1K per decade,
    dims=("time",),
    coords={"time": t},
    attrs={"units": "K"},
)
sim = xr.DataArray(
    (
        -18 * np.cos(2 * np.pi * t.dayofyear / 365)
        + 2 * np.random.random_sample((t.size,))
        + 273.15
        + 0.11 * (t - t[0]).days / 365
    ),  # "warming" of 1.1K per decade
    dims=("time",),
    coords={"time": t},
    attrs={"units": "K"},
)

ref = ref.sel(time=slice(None, "2015-01-01"))
hist = sim.sel(time=slice(None, "2015-01-01"))



QM = sdba.EmpiricalQuantileMapping.train(
    ref, hist, nquantiles=15, group='time.month', kind="+"
)
scen = QM.adjust(sim, extrapolation="constant", interp="nearest")
print(f"{scen.isnull().sum('time').values} nans for nearest interpolation")

scen = QM.adjust(sim, extrapolation="constant", interp="linear")
print(f"{scen.isnull().sum('time').values} nans for linear interpolation")

scen = QM.adjust(sim, extrapolation="constant", interp="cubic")
print(f"{scen.isnull().sum('time').values} nans for cubic interpolation")

f,(ax1,ax2) = plt.subplots(2,1)
scen.isnull().plot(ax=ax1)

sim.plot(ax=ax2,color='k')
sim[scen.isnull()].to_dataset(name='scen').plot.scatter(x='time',y='scen',color='r')
```

### What I Received

```
0 nans for nearest interpolation
78 nans for linear interpolation
78 nans for cubic interpolation
```
Screenshot from 2022-02-04 16-00-05

Thanks, Folmer Krikken

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1008

Ok, so when doing quantile mapping, the train step will create a 2D map of adjustment factors. The "x" axis are values of the variable, the CDF of the requested quantiles. and the "g" axis is the temporal group (like months or dayofyear). Note that the "x" axis is different for each temporal group. EX: the historical CDF is different for each month.

The adjustment steps takes each value of `sim` and finds the correct factor by interpolating into this map. 

When linear or cubic interpolation is requested with a "time.month" grouping, we use some kind of "decimal" month as the "g" coordinate in the interpolation. Like Jan 15th is 1, Jan 31st is 1.5 and so on.

When the "constant" extrapolation is requested, the previous behaviour was to find the max and min along the historical "x" axis for each given month. For example, if the CDF of july goes from 270 to 330 K, all values of sim in july that are < 270 are given 270 as the output value, thus the extrapolation is "constant".

However, that's not how `griddata` (who does the interpolation) works. Say the hist CDF of June goes from 280 to 330. Now say the sim value on the 1st of July (6.5) is 271. When you think of it, it is outside our 2D map, and thus griddata can't interpolated and puts `NaN`. However, our extrapolation function sees it within the CDF of july and doesn't extrapolate it.

Bam, a NaN that was not expected.

This PR interpolates the bounds linearly so that it uses the same logic that `griddata` uses to determine which value to extrapolate. In fact, it might not be the exact same "convex hull", but it's near enough.

The fix is applied on all cases (not only "time.month"), even if it is not needed. We could say it's a more general version of the previous algorithm, so there shouldn't be any change in the results. Also, the performance seems unchanged (thanks numba!).

No.

### DIFF
--- a/xclim/sdba/utils.py
+++ b/xclim/sdba/utils.py
@@ -438,7 +438,6 @@ def _interp_on_quantiles_2D(newx, newg, oldx, oldy, oldg, method, extrap):  # no
             category=RuntimeWarning,
         )
         return out
-
     out[~mask_new] = griddata(
         (oldx[~mask_old], oldg[~mask_old]),
         oldy[~mask_old],
@@ -493,7 +492,7 @@ def interp_on_quantiles(
     - 'nan' : Any value of `newx` outside the range of `xq` is set to NaN.
     - 'constant' : Values of `newx` smaller than the minimum of `xq` are set to the first
       value of `yq` and those larger than the maximum, set to the last one (first and
-      last values along the "quantiles" dimension).
+      last values along the "quantiles" dimension). When the grouping is
     """
     dim = group.dim
     prop = group.prop
--- a/xclim/sdba/utils.py
+++ b/xclim/sdba/utils.py
@@ -3,7 +3,7 @@ sdba utilities
 --------------
 """
 import itertools
-from typing import Callable, List, Mapping, Optional, Tuple, Union
+from typing import Callable, Mapping, Optional, Tuple, Union
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


xclim/sdba/utils.py
```diff
@@ -493,7 +386,7 @@ def interp_on_quantiles(
     - 'nan' : Any value of `newx` outside the range of `xq` is set to NaN.
     - 'constant' : Values of `newx` smaller than the minimum of `xq` are set to the first
       value of `yq` and those larger than the maximum, set to the last one (first and
-      last values along the "quantiles" dimension).
+      last values along the "quantiles" dimension). When the grouping is
```
huard, line 389: ?

huard (approved):
Is this error exercised by an existing test ?

xclim/sdba/utils.py
aulemahal: woupsi

PR #1017, opened 2022-02-16
Enh/map block bootstrap

### LINKED ISSUE #932: Bootstrapping quickly exhausts memory
* xclim version: 0.31.0
* Python version: 3.9.7
* Operating System: Linux 74cf36c51c91 4.18.0-305.19.1.el8_4.x86_64 #⁠1 SMP Wed Sep 15 15:39:39 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Description

Trying to calculate indices with bootstrap set to True results in a rather large amount of RAM (120 GB) being exhausted. No matter how small I make the chunks of the dataset that is used to calculate the index it fails. Is this something that's expected due to the computationally heavy nature of bootstrapping, or is this an issue with xclim?

### What I Did

Here's the code that I ran:

```python
import xarray as xr
import xclim

mpi_historical = xr.open_dataset('http://esgf-data3.ceda.ac.uk/thredds/dodsC/esg_cmip6/CMIP6/CMIP/MPI-M/MPI-ESM1-2-LR/historical/r1i1p1f1/day/tasmin/gn/v20190710/tasmin_day_MPI-ESM1-2-LR_historical_r1i1p1f1_gn_19900101-20091231.nc',
                                 chunks=dict(time=1000))

p10 = xclim.core.calendar.percentile_doy(mpi_historical.tasmin, per=10)
TN10p_mpi_historical = xclim.indicators.icclim.TN10p(mpi_historical.tasmin, t10=p10,
                                                     freq='QS-DEC', bootstrap=True)

TN10p_mpi_historical.load()
```

### What I Received

I'm doing these computations through a remote jupyterlab instance, and a short while after the code starts to execute the .load() part it crashes due to memory exhaustion. If the imported dataset is divided into smaller chunks it takes a longer time for the crash to happen. 

I also get the following warnings about chunk sizes:

```
/opt/conda/envs/my_kernel/lib/python3.9/site-packages/xarray/core/indexing.py:1227: PerformanceWarning: Slicing is producing a large chunk. To accept the large
chunk and silence this warning, set the option
    >>> with dask.config.set(**{'array.slicing.split_large_chunks': False}):
    ...     array[indexer]

To avoid creating the large chunks, set the option
    >>> with dask.config.set(**{'array.slicing.split_large_chunks': True}):
    ...     array[indexer]
  return self.array[key]
```

bzah:
### Pull Request Checklist:
    - This PR fixes #932 

This PR make an optimization to dask graph when bootstrapping percentiles using `xarray.map_block`.
There are also a few refactorings to clarify the code.

#### Before
The main concern was the intermediary results stored on disk by dasks.
In a large dataset, It was generating up 150Go of data on disk.
This had two consequences:
- Computation is slow down by i/o operations, which may even idle all worker's threads while the i/o is in progress.
- The OS may not allow an application to write so much data on disk, especially if there is no such space...
I suspect dask was storing the data broadcasted to window size which could effectively be reused in multiple steps of bootstrapping algorithm.  
The other concern was the high memory (RAM) use. This lead dask to spill even more on disk and also limited the number of threads concurrently running.
For example, on dask default scheduler the OS would often kill the python process due to the amount of memory being use.
Finally on large dataset the dask graph could be populated with a lot of tasks.

#### Now
When running bootstrap algorithm on a single file of shape `time: 3653, lat: 128, lon: 256` and stored in chunks: `chunksizes': (1, 128, 256)`, there are approximately the same number of dask tasks created as before.
However, on a zarr store of shape `lat: 256, lon: 512, time: 38716` and stored in chunks `(38716, 3, 512)`, when 10 years are bootstrapped on a 20 years study, it was generating 1M tasks where now it "only" generates 250K tasks.

Plus, there is **much** less spilling on disk going on and the memory use always stays at a very manageable level.

No.

### Other information:
There was also some cases where years were duplicated and an (ugly) post-processing was filtering them out.
The root cause was the object used to iterates. We were using the whole da instead of the overlapping da in between reference period and studied period.

Two more errors can now be raised to avoid starting bootstrap when it's unnecessary:
- When the reference period and the studied period don't overlap.
- When the reference period and the studied period **fully** overlap.
This last case may seem counterintuitive. The bootstrap algorithm is used to avoid the inhomogeneity of percentile comparison between the period where both the index and the daily percentiles are computed (reference period) and the period where **only** the index is computed.
Thus, when reference period is equal to the study period there is no inhomogeneity.

All tests were done on my laptop. CPU: i5 2GHz 4 cores (x2) ; RAM: 16GB (dask is configured to use this as a limit)

### DIFF
--- a/xclim/core/bootstrapping.py
+++ b/xclim/core/bootstrapping.py
@@ -12,6 +12,8 @@ import xclim.core.utils
 
 from .calendar import compare_offsets, convert_calendar, parse_offset, percentile_doy
 
+BOOTSTRAP_DIM = "_bootstrap"
+
 
 def percentile_bootstrap(func):
     """Decorator applying a bootstrap step to the calculation of exceedance over a percentile threshold.
@@ -55,7 +57,7 @@ def percentile_bootstrap(func):
     return wrapper
 
 
-def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xr.DataArray:
+def bootstrap_func(compute_index_func: Callable, **kwargs) -> xr.DataArray:
     """Bootstrap the computation of percentile-based exceedance indices.
 
     Indices measuring exceedance over percentile-based threshold may contain artificial discontinuities at the
@@ -65,7 +67,7 @@ def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xr.DataArray:
 
     Parameters
     ----------
-    compute_indice_func : Callable
+    compute_index_func : Callable
       Indice function.
     kwargs : dict
       Arguments to `func`.
@@ -118,9 +120,21 @@ def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xr.DataArray:
     # List of years in base period
     clim = per.attrs["climatology_bounds"]
 
-    # `da` over base period used to compute percentile
-    overlapping_da = da.sel(time=slice(*clim))
-
+    if xclim.core.utils.uses_dask(da):
+        chunking = {d: "auto" for d in da.dims}
+        chunking["time"] = -1  # no chunking on time to use map_block
+        da = da.chunk(chunking)
+    # overlap of `da` and base period used to compute percentile
+    overlap_da = da.sel(time=slice(*clim))
+    # TODO add unit tests for these exceptions
+    if len(overlap_da) == len(da):
+        raise KeyError(
+            "`bootstrap` is unnecessary when all years between in_base (percentiles period) and out_of_base (index period) are overlapping"
+        )
+    if len(overlap_da) == 0:
+        raise KeyError(
+            "`bootstrap` is unnecessary when no year overlap between in_base (percentiles period) and out_of_base (index period)."
+        )
     # Arguments used to compute percentile
     percentile = per.percentiles.data.tolist()  # Can be a list or scalar
     pdoy_args = dict(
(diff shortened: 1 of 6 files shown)


bzah:
Oopsi, I missclicked on "create pull request" too soon...

xclim/core/calendar.py
```diff
@@ -491,10 +496,13 @@ def percentile_doy(
     rrr = rr.assign_coords(time=ind).unstack("time").stack(stack_dim=("year", "window"))
 
     if rrr.chunks is not None and len(rrr.chunks[rrr.get_axis_num("stack_dim")]) > 1:
-        # Preserve chunk size
-        time_chunks_count = len(arr.chunks[arr.get_axis_num("time")])
-        doy_chunk_size = np.ceil(len(rrr.dayofyear) / (window * time_chunks_count))
-        rrr = rrr.chunk(dict(stack_dim=-1, dayofyear=doy_chunk_size))
+        chunking = {"stack_dim": -1}
+        if keep_chunk_size:
```
bzah, line 500: Not 100% sure it's useful, I was expecting better perfs when doing the chunking only once in bootstrap algo but it seems to not change anything.
I guess it's because map_block already computes things chunk by chunk and does not rechunk again each chunk

bzah:
I'm satisfied with the perfs for now.
It works especially well when the data is stored in an appropriate chunking schema, that is when time is not chunked (or not too much).
However on 100 netcdf files with 1 year per file, it performs better than before but it's not crazy good.

BTW, I'm considering making Pangeo `rechunker` a dependency of icclim to delegate to it the pre-process rechunking. If you have some experience with rechunker, I'm curious to hear what you think about this.

bzah:
> ... I first transfered the data to a zarr file and _than_ did the rechunking. It was in most cases faster and less buggy.

Ah! I was wondering why they were doing this in rechunker tutorial. Thanks for the tips!

huard (approved):
Did a quick superficial review, no objection.

xclim/core/bootstrapping.py
```diff
@@ -102,82 +108,108 @@ def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xarray.DataArray:
...
-    da_base = da.sel(time=slice(*clim))
-
-    # Arguments used to compute percentile
-    percentile = per.percentiles.data.tolist()  # Can be a list or scalar
+    # Boundary years of reference period
+    clim = per_da.attrs["climatology_bounds"]
+    if xclim.core.utils.uses_dask(da):
+        chunking = {d: "auto" for d in da.dims}
+        chunking["time"] = -1  # no chunking on time to use map_block
+        da = da.chunk(chunking)
```
aulemahal, line 124: Is this auto-rechunking a performance improvement? I understand the `time = -1`, but passing `auto` for the other dimensions might override a user-intended chunking, adding unnecessary (or at least unwanted) steps in the dask tree?

xclim/core/bootstrapping.py
```diff
@@ -102,82 +108,108 @@ def bootstrap_func(compute_indice_func: Callable, **kwargs) -> xarray.DataArray:
...
+            per = xr.map_blocks(
+                percentile_doy.__wrapped__,  # strip history update from percentile_doy
+                obj=bda,
+                kwargs={**pdoy_args, "copy": False},
+                template=per_template,
+            )
+            if "percentiles" not in per_da.dims:
+                per = per.squeeze("percentiles")
+            kw[per_key] = per
+            value = compute_index_func(**kw).mean(dim="_bootstrap", keep_attrs=True)
```
aulemahal, line 174:
```suggestion
            value = compute_index_func(**kw).mean(dim=BOOTSTRAP_DIM, keep_attrs=True)
```
A small oubli I devine.

xclim/core/bootstrapping.py
aulemahal: In any cases, I suggest raising a warning when the chunking is modified.

aulemahal (approved):
Looks good! I have yet to use this bootstrap feature, so I am not totally familiar with the issues, but this looks pretty good to optimize dask.

Zeitsperre:
@bzah Do you think it would be possible to address the final comments today or tomorrow? Our next release is scheduled for the end of the week. Thanks again!

xclim/core/bootstrapping.py
bzah: You devine well, merci.

xclim/core/bootstrapping.py
bzah: The `"auto"` is useful to rechunk the whole thing and conserve chunk size to the configured `array.chunk-size` of dask.config.
I fear that without rechunking other dims, the chunks could become very large and dask workers might overflow their memory limits. This depends of course on the original chunking.

I agree a warning is necessary.
Plus, this step can be avoided altogether if time is already unchunked, I will add it as well.

bzah: I'm not opposed to remove this "auto" rechunk but, it's a choice between:
- Preserving the overall chunk sizes to the configurable `array.chunk-size`.
- Preserving the chunked dimensions (other than time) to their original chunk sizes.

aulemahal: Your suggestion of not rechunking if the time dimensions is already in a single block is, I think, the best choice! In the other case, I'm ok with an automatic rechunking  if a warning is issued.

bzah:
If you are ok with the warning wording I think it can be merged.

aulemahal:
All good for me! LGTM

PR #1019, opened 2022-02-16
Change is_dayofyear to np.int32

### LINKED ISSUE #980: is_dayofyear attribute as int32
When creating day-of-year indices, xclim adds a special "is_dayofyear=1" attribute, in addition to `calendar=...` and `units=""`. This way, downstream functions of xclim can understand the data and handle it correctly, without messing with the units or, as previously, with the standard name (which wouldn't be CF). It is a 1 because netCDF doesn't support bools in attributes... 

Python's integers are "int64", however, thredds has a deep rooted hatred for this "long" dtype. A simple solution is to explicitly put `np.int32(1)`. xarray will respect this when writing to file and thredds will not lament.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #980

* The "is_dayofyear" attribute added by several indices is now a ``numpy.int32`` instance, instead of python's ``int``. This ensures a THREDDS server can read it when the variableis saved to a netCDF file with xarray/netCDF4.

No.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,14 @@
 History
 =======
 
+0.34.0 (unreleased)
+-------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`)
+
+Internal changes
+^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


Zeitsperre (approved):
Looks good!

tlogan2000 (approved):

Issue #886, opened 2021-10-28
Proposal: Issue Forms and/or Markdown template options


Zeitsperre:
GitHub now offers Issue Forms using Markdown (video: https://www.youtube.com/watch?v=2Yh8ueUE0oY; docs: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms).

It would be good to differentiate between Bug Reports and Feature Requests at the very least. There should be some required information fields for both of them. Nothing too fancy. (This can also be ported to our existing cookiecutter templates that are closely aligned with xclim's project template).

huard:
I'm -1 on this. This is meant to add friction for users issuing bug or feature requests. The large majority of issues are created by ourselves, so I don't see why we'd want more friction.

PR #1028, opened 2022-02-23
Use Issue Forms

Zeitsperre:
### Pull Request Checklist:
    - This PR exapands upon #886 
  - [ ] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

* This PR adds issue forms, which are a structured way of asking for required information from users/devs on bug reports, support tickets, and enhancements requests.
* Adjusted YAML formatting

No.

### Other information:

Feel free to suggest changes or simplifications to the approach. I added in what I think makes the most sense, but this is to better aid us as a group.


tlogan2000 (approved):

huard:
Is there a way we could test it in action before merging ? I'm not convinced this is going to be useful at this point in xclim's development.

juliettelavoie:
When we submit a new issue, will it be mandatory to use one of those templates or will there also be a "empty template" choice  ?

Zeitsperre:
@huard There's no way to test beforehand. We could merge, try it out, and roll back the commit, though. 

@juliettelavoie There's still the option of opening a generic issue, which will load the typical schema we have currently (https://github.com/Ouranosinc/xclim/blob/master/.github/ISSUE_TEMPLATE.md). Blank issues are permitted, but we could specify right at the top of the list that the default ticket is the ISSUE_TEMPLATE.md.

Zeitsperre:
@huard I think this is ready. I've made a generic issue form as the topmost form, with one content box, so effectively the same as what we have currently, but we have the option of opening blank forms as well. I would suggest merging this, seeing whether we like it or not, and making changes/reverting it in another PR. What do you think?

PR #1030, opened 2022-03-03
Skipna on LOESS

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Implement some basic `skipna` behaviour in `loess_smoothing` and thus in `LoessDetrend`. This doesn't remove the missing values, but at least it doesn't propagate them as it did before.

Before this change, a single NaN value would make the whole loess-window be NaN, so missing values would "bleed out".

Yes because results will change. But the previous behaviour can be obtained by passing `skipna=False`.


xclim/sdba/loess.py
```diff
@@ -75,12 +76,21 @@ def _loess_nb(
...
+      same missing values as the input.
 
     References
     ----------
     Code adapted from https://gist.github.com/agramfort/850437
     Cleveland, W. S., 1979. Robust Locally Weighted Regression and Smoothing Scatterplot, Journal of the American Statistical Association 74, 829–836.
     """
+    if skipna:
+        nan = np.isnan(y)
+        out = np.zeros(x.size) * np.NaN
```
huard, line 90: If you're the kind of person who cares about shaving nanoseconds, np.empty is faster than np.zeros. 
```python
In [2]: %timeit np.empty(10000)                                                                                                                                                                            
560 ns ± 7.78 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit np.zeros(10000)                                                                                                                                                                            
2.28 µs ± 193 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

huard (approved):

xclim/sdba/loess.py
bzah: You could even write `np.full(x.size, np.NaN)` to gain maybe half a microsecond ! 

```
In [5]: %timeit np.full(10000, np.NaN)
6.88 µs ± 766 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit np.empty(10000) * np.NaN
7.24 µs ± 569 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit np.zeros(10000) * np.NaN
8.63 µs ± 283 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

PR #1032, opened 2022-03-08
Add check to verify that JSON files are always valid

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds check to verify that JSON files are always valid
* Adds the pretty-json-format hook to remove duplicate keys and adjust whitespace

No.


aulemahal (approved):
Nice! I think this kinda fixes an issue I had when first implementing the translations. We didn't have a nice way to preserver the "style" of the json files. Here we are!

As we moved to yaml files for the modules, I was wondering if we would gain from moving the translations to yaml as well? Maybe later.

Zeitsperre:
@aulemahal For our purposes, I don't think we'd lose anything in moving to YAML for all of our internal configurations. On the contrary, we'd probably gain a lot in readability. My ideal future would have us migrating our configurations to TOML files, and anything requiring readability/schematics to YAML (or JSON, depending on data types). 

The only major reason I could see for us in closely following JSON would be to have the option of sharing our configurations with others or allowing for on-the-fly data validation (thinking of WPS applications). JSON is a lot stricter, more widely adopted across languages, and easier to validate, that's about it.

PR #1033, opened 2022-03-09
Improve spatial analogs : dask, Székely-Rizzo

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Some changes to the logic of  `spatial_analogs` that makes dask-backed arrays compatible.
* Add an `__all__` to `spatial_analogs`.
* Implementation of `szekely_rizzo` for analogs, modified to be scale-invariant.
* Fix typo "Skezely" to "Székely" in all of xclim.

No, the only api that changed is the previously non-implemented `skezely_rizzo` function. Some attributes might have changed since there was a typo in that name.

### Other information:
As you might see in the commit history, I tried to accelerate the analog computation with numba. I focused on `zech_aslan` as this is the one I use currently.  But I eventually reverted the changes, turns out that the `scipy.spatial.distance` implementations of `cdist` and `pdist` are already quite performant. My own numba implementations didn't compete, even when using numba to do the vectorization (which is currently done in `apply_ufunc` by numpy). The main performance gain from this PR comes from the usage of dask.


Zeitsperre:
Hot take incoming:

I think in the case of "Spatial Analogue", given both the fact that we borrow more from the British English (spelling), and that "analog" is more often used in American English for differentiating between digital and analog signals, I'd like to put forward that we're better off using "analogue" when describing things that are "analogous". Thoughts?

aulemahal:
Oh! I didn't know it was valid! I simply saw that the articles we were citing were using "analog". If the difference is American vs Canadian english, than I'm in for `analogue` ;).

aulemahal:
Changed it back, but to avoid making a (quite snob) breaking change the name of the function remains the same.

Zeitsperre:
That's fair. Again, it's more of an opinion. We don't have norms in xclim on the dialect of English to use, and we often come up against American English norms baked into Python (e.g. "color", instead of "colour"). Maybe we need guidelines for regional dialects of both English and French?

HISTORY.rst
```diff
@@ -4,12 +4,20 @@ History
...
 ^^^^^^^^^^^^^^
 * New indicator ``specific_humidity_from_dewpoint``, computing specific humidity from the dewpoint temperature and air pressure. (:issue:`864`, :pull:`1027`)
 
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* New spatial analogues method "szekely_rizzo" (:pull:`1033`).
+
+Bug fixes
+^^^^^^^^^
+* ``xclim.analog.spatial_analogs`` now compatible with dask-backed DataArrays (:pull:`1033`).
```
aulemahal, line 19:
```suggestion
* ``xclim.analog.spatial_analogs`` now compatible with dask-backed DataArrays (:pull:`1033`).
* Parameter ``dmin`` added to spatial analog method "zech_aslan", to avoid singularities on identical points (:pull:`1033`).
```

aulemahal:
@huard I just pushed 2 minor adjustments:
- Add a `standardize` argument to the new Székely-Rizzo method, activated by default, so that metric becomes scale-invariant.
- Comment on the scale-dependence of the methods, so the user is in connaissance de cause.

(I also updated the testdata, so the tests will fail until we merge the branch in `xclim-testdata`).

aulemahal:
There should be another PR relating to spatial analogs before the next release. With a notebook and maybe the implementation of another distance metric.

PR #1041, opened 2022-03-23
Add loguru

### LINKED ISSUE #1039: Disable logging in xclim.ensembles module?
### Generic Issue

When importing `clisops.subset.core` and `xclim.ensembles` the logging output in the xclim.ensembles is somehow always sent to the console, or worse, to the jupyter notebook output window.    

@Zeitsperre has been digging into this (see https://github.com/roocs/clisops/pull/216) but I am wondering just how critical it is to have a logger at all?  

Notes - this seems to be the only submodule where a logger is used and as far as I can tell and only keeps track of 1) whether matplotlib is installed or not (for ensemble reduction graphs) or 2) which simulation is being accessed in the loop for the ensemble creation.   

### Code of Conduct

- [X] #1040

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1039 

* Adds loguru as an installation dependency.
* Cleans up a few call signatures
* Updated contributing documentation with examples of how to enable logging in scripts/jupyter.

Yes, loguru is now a hard dependency for xclim.

### Other information:

For users and use cases that rely on gathering info from xclim using the python logging library, ~I still need to figure out how to synchronize events between the two systems so that messages are passed between them. I don't see this as being an enormous problem unless we start logging events from asynchronous processes (in which case, I would argue that we not bother passing events to standard logging at all).~ Loguru loggers are [multiprocessing-capable](https://loguru.readthedocs.io/en/stable/resources/recipes.html?highlight=enqueue#compatibility-with-multiprocessing-using-enqueue-argument). How this works with `dask`-based multiprocessing still need to be looked into.

https://loguru.readthedocs.io/en/stable/index.html

**In order to enable logging:**

For stdout:
```
import sys
from loguru import logger

logger.activate("xclim")
logger.add(sys.stdout, level=LEVEL)  # LEVEL can be str (e.g. "WARNING") or int (e.g. 30)
```
For writing to file:
```
from loguru import logger

logger.activate("xclim")
logger.add("my_logging_file.log", enqueue=True, level=LEVEL)
# 'enqueue' is a safeguard for ensuring that logged events from spawned processes are written in sequential order  - Needed for MPI
```


xclim/__init__.py
```diff
@@ -1,17 +1,43 @@
...
 __author__ = """Travis Logan"""
 __email__ = "logan.travis@ouranos.ca"
 __version__ = "0.34.2-beta"
 
 
+# On import, clisops sets a root logger for console output. This is huge annoyance to deal with.
+# For upstream fix discussion, see: https://github.com/roocs/clisops/pull/216
+if "clisops" in sys.modules:
+    root_logger = logging.getLogger()
+    root_logger.removeHandler(root_logger.handlers[0])
```
aulemahal, line 25: xclim doesn't import clisops, I don't think this is the job of xclim?

IMHO, best option is to fix it in clisops, second best is to do this in the user's script, third best is not to use clisops.

xclim/_scripting.py
```diff
@@ -0,0 +1,51 @@
+import logging
```
aulemahal, line 1: Is all this needed for a normal logging experience? I'm not sure why xclim would offer so many utilities and tools, it's a library not a framework?

xclim/sdba/_adjustment.py
```diff
@@ -378,6 +381,7 @@ def _fit_cluster_and_cdf(data, thresh, dist, cluster_thresh):
     return dist.cdf(data, *fut_params)
 
 
+# FIXME: 'group' is not used
```
aulemahal, line 384: Same a above.

xclim/__init__.py
Zeitsperre: You're right. I was trying to disable logging from clisops. This type of change is out of scope for this PR. Acceptance of the changes to their logging configuration upstream are still under consideration. Will remove.

xclim/__init__.py
```diff
@@ -1,17 +1,43 @@
...
+
+
+def showwarning(message, *args, **kwargs):
+    """Inject warnings from `warnings.warn` into `loguru`."""
+    logger.warning(message)
+    showwarning_(message, *args, **kwargs)
```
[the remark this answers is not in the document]
Zeitsperre: It's not so much a bug, but more a feature? All this monkey patch does is ensure that messages raised from `warnings.warn` (which reports events in standard logging) are passed to the loguru logger. This not only affects us, but also any other dependency-related warnings that users should be notified of. Out of all the monkey patches suggested by loguru, I would opt to keep this one. This seems like a good idea overall.

xclim/__init__.py
```diff
@@ -1,17 +1,43 @@
...
+
+
+showwarning_ = warnings.showwarning
+warnings.showwarning = showwarning
+
+# Remove default "DEBUG"-level logger and deactivate xclim logger on import
+logger.remove()
+logger.disable("xclim")
```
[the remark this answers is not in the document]
Zeitsperre: We can re-enable it. Not an issue. It'll clean up the tests quite a bit as well.

xclim/_scripting.py
Zeitsperre: These have now all been removed.

xclim/__init__.py
Zeitsperre: Removed.

xclim/__init__.py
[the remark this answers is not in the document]
Zeitsperre: See also: https://github.com/Delgan/loguru/issues/536

HISTORY.rst
```diff
@@ -21,9 +21,9 @@ New features and enhancements
 
     import sys
     from loguru import logger
-    
+
     logger.activate("xclim")
-    
+
```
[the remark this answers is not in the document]
Zeitsperre: Oh nvm, it's the whitespace checker, haha

PR #1049, opened 2022-04-06
Fix lazy_indexing when using twice the same indexes (with dask)

### LINKED ISSUE #1048: Lazy indexing can't tell tasks apart
### Setup Information

- Xclim version: master
- Python version: 3.9.10
- Operating System: Linux

### Description

When using `xc.indices.run_length.lazy_indexing` with the same index on two different arrays, dask can't tell the two indexing tasks apart. Thus, the second call generates tasks that overwrite the first. And the first output has the results of the second call.

In the example below, `ia == ib` which should not be the case! `ib` is correct, but `ia` isn't.

### Steps To Reproduce
```python3
import xarray as xr
import xclim as xc
import numpy as np
from dask import compute

da = xr.DataArray(np.arange(100), dims=('x',), name='a')
db = xr.DataArray(-np.arange(100), dims=('x',), name='b')
indx = xr.DataArray(np.random.randint(0, 99, (10, 10, 10)), dims=('w', 'y', 'z'), name='index')
indxc = indx.chunk()

iac = xc.indices.run_length.lazy_indexing(da, indxc)
ibc = xc.indices.run_length.lazy_indexing(db, indxc)
ia, ib = compute(iac, ibc)
```
### Additional context

I have not tested this extensively, may be it only happens with simplistic examples like this one. (my original script is not much more complicated).

I do have a solution. In `lazy_indexing`, we declare a 2 inputs function and then we use `partial` to fill the first argument (`da` or `db`). This partial func is passed to `index.map_blocks`. I believe xarray/dask can't see that the partial functions are different and so the resulting tasks have the same name. I think those names are formed with the hash of the objects, so I'm guessing the two partial objects are returning the same hash even though the filled arg is a different object?

Anyway, we can pass additional args through `map_blocks` and when doing so the bug disappears.  

PR coming up.

### Contribution

### Code of Conduct

aulemahal:
…h dask - upd hist

### Pull Request Checklist:
    - This PR fixes #1048

* Uses `map_blocks` in a different (better) manner so that xarray/dask is able to differentiate when the indexing function is called with the same indexes but a different array.

Nope.

### Other information:
`map_blocks` converts DataArray arguments to datasets, and this bugs when the dataarrays don't have names. I added a simple workaround.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,14 @@
 History
 =======
 
+0.36.0 (unreleased)
+-------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Invoking ``lazy_indexing`` twice in row (or more) using the same indexes (using dask) is now fixed. (:issue:`1048`, :pull:`1049`).
+
 0.35.0 (01-04-2022)
 -------------------
 Contributors to this version: David Huard (:user:`huard`), Trevor James Smith (:user:`Zeitsperre`) and Pascal Bourgault (:user:`aulemahal`).
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.35.0
+current_version = 0.35.1-beta
 commit = True
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+).(?P<patch>\d+)(\-(?P<release>[a-z]+))?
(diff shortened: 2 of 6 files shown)


Zeitsperre:
@aulemahal Since this is the first PR since last release, don't forget to bumpversion!

huard (approved):
Do we need to do an audit of the code and make sure this doesn't happen elsewhere?

aulemahal:
Good idea!

Searching `map_blocks` everywhere in the code, I found no other instances where we were using `partial`. I'm not sure how this problem could happen otherwise...

PR #1050, opened 2022-04-08
ENH: Improve percentile based indices' descriptions

### LINKED ISSUE #1047: In `{cold, warm}_spell_duration_index` indicators description, the percentile threshold is not dynamically set.
### Setup Information

- Xclim version: 0.35
- Python version: all
- Operating System: all

### Description

I tried to compute `cold_spell_duration_index` with, instead of the 10th percentile, the 5th percentile.
The result looks good however, in the description attribute, hard-coded values are used for "10th percentile" and for "a 5-day window". 
Both the percentile threshold and the window are parameters of `percentile_doy` and should not be hard-coded.

`percentile_doy` returns a DataArray, we could store these parameters in its attributes (actually the window is already in them).
Then, the Indicator formatter could trace back these and properly format the indices.

This issue exists for at least `cold_spell_duration_index` and `warm_spell_duration_index`. I haven't check if other indices could be impacted by this.

### Steps To Reproduce

```
da = xr.open_dataset("tas.nc").tasmax
per = percentile_doy(da, window=5, per=5)
csdi = cold_spell_duration_index(da, tn10=per, freq="YS", bootstrap=False)

# assert fails
assert "5th percentile" in csdi.attrs["description"]
```

### Additional context

I already tried to implement something but it needs more effort than initially expected. Thus, I prefer to make sure this feature/bug-fix is worth before continuing to work on it.

### Contribution

### Code of Conduct

bzah:
### Pull Request Checklist:
    - This PR fixes #1047

This PR makes csdi and wsdi indicators' use the proper parameterization used in `percentile_doy` to format their description.

This concerns, the base percentile value, the length of rolling window of days and the period on which percentiles are computed.

A limit of the current approach is that if percentiles are **not** computed using `percentile_doy` the description will use default, non configurable values. 

Yes, the signatures of all indices and indicators having a DataArray parameter holding percentiles have been modified.
Before, the parameter name was assuming a percentile value such as in `t90`.
Now, they are renamed into one of `tas_per`, `tasmax_per`, `tasmin_per`, `pr_per` depending on the expected variable type.
Their typing is also changed to `PercentileDataArray` which is now outputted by `percentile_doy`.

### Other information:
With `per = percentile_doy(da_per, window=2, per=[42, 27])` the description (reformatted for github) now looks like:
```
'description': 
"Annual number of days with at least 6 consecutive days where
 the daily minimum temperature is below the [42, 27]th percentile(s). 
a 2 day(s) window, centred on each calendar day in the ['2015-01-01', '2018-12-31'] period,
 is used to compute the [42, 27]th percentile(s)."
```


xclim/core/calendar.py
```diff
@@ -527,6 +527,7 @@ def percentile_doy(
     p.attrs["window"] = window
     p.attrs["alpha"] = alpha
     p.attrs["beta"] = beta
+    p.attrs["per_base_thresh"] = per
```
aulemahal, line 530: Isn't this information already available in the `percentiles` coordinate? The issue here is that I'm not sure the netCDF  backend (or others) will elegantly handle a list passed as attribute. Even more problematic is if the user passed a `ndarray` or a `DataArray`.

xclim/core/indicator.py
```diff
@@ -853,10 +857,21 @@ def _parse_variables_from_call(self, args, kwds):
...
+                        if "percentile_doy" in data.attrs.get("history", ""):
+                            base_per = data.attrs.get("per_base_thresh")
+                            if len(base_per) == 1:
+                                preformatted_params["per_base_thresh"] = base_per[0]
+                            else:
+                                preformatted_params["per_base_thresh"] = base_per
+                            preformatted_params["per_window"] = data.attrs.get("window")
+                            preformatted_params["per_period"] = data.attrs.get(
+                                "climatology_bounds"
+                            )
```
aulemahal, line 870: I am not convinced parsing the history is the safest way to check this. I think I'd rather prefer a new `InputKind` entry, that's what they were meant for. The problem here is that a "percentile threshold" is not differentiable from a normal input in the world of `typing`, and thus I think we'll need to use the same workaround as for `DateStr`.

I also believe `per` should be replaced by the input argument's name. Then it would be more intuitive for the user that if the input arg is `t90` then `t90_window` is the extracted window value. This would also allow for doing this on indicators with more than one percentile threshold (like `cold_and_dry_days`). 

Also, I think this code should reside in `format`, in a goal to centralize this parsing and translating.

xclim/indicators/atmos/_temperature.py
```diff
@@ -394,10 +394,15 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
-    "a 5-day window centred on each calendar day in the  1961-1990 period",
+    "where the daily minimum temperature is below the {per_base_thresh}th "
+    "percentile(s). A {per_window} day(s) window, centred on each calendar day in the "
+    "{per_period} period, is used to compute the {per_base_thresh}th percentile(s).",
     cell_methods="",
+    default_params={
+        "per_base_thresh": 10,
+        "per_window": 5,
+        "per_period": "1961-1990",
+    },
```
aulemahal, line 405: I understand what you are trying to do, but I think this assumes too much. I would prefer an error or an empty string, rather than false information.

xclim/indices/_multivariate.py
```diff
@@ -61,26 +61,26 @@
 ]
 
 
-@declare_units(tasmin="[temperature]", tn10="[temperature]")
+@declare_units(tasmin="[temperature]", per_da="[temperature]")
 @percentile_bootstrap
 def cold_spell_duration_index(
     tasmin: xarray.DataArray,
-    tn10: xarray.DataArray,
+    per_da: xarray.DataArray,
```
aulemahal, line 68: Oh! I wrote the comment above before seeing that you were renaming the inputs. I think the idea from above still holds, but I indeed agree with generalizing the indices by removing the percentile from the name.

I believe you missing a few cases in the examples, but the idea is good.
I think it should be done everywhere and that includes things like `cold_and_dry_days`, and thus maybe the thresholded variable's name should be included in the name? `tas_per` ? Only for the multivariate cases?

aulemahal (changes requested):
Instead of `default_params` and  `preformatted_attrs`, I think all the parsing could be done in `Indicator.format`.  This way, it would be easier to have more than one of those percentile thresholds. My best suggestion for now would be to return "N.A."  for the cases where the correct attributes can't be parsed (replacing the `default_params`), and maybe issuing a warning?

xclim/core/calendar.py
bzah: True, I forgot it's a already there.

xclim/core/indicator.py
bzah: I tend to agree that parsing a string, possibly modified by the user, is particularly unsafe. I went with it because it's already what we do in bootrapping algo.
Another approach could be to inherit DataArray in a PercentileDataArray class which would be outputted by `percentile_doy`. This way it is still a DataArray and can be manipulated as is but, it can as well be recognized as a different `InputKind`.

> Also, I think this code should reside in format, in a goal to centralize this parsing and translating.
I'll try to move these there.

xclim/indices/_multivariate.py
Zeitsperre: I would also vote for something similar `tasmin_percentile` or `tn_percentile` for this call signature. `percentile` alone give me the impression that I can pass `tas` (and not `tasmin`).

Zeitsperre (commented):
I just had a few things to mention. I trust @huard and @aulemahal on this. Be sure to mention these call signature changes as breaking changes in the History!

xclim/indices/_multivariate.py
bzah: Yep, and I forgot to rename in unit tests as well.
I'm ok with `tasmin_per` here and `tas_per`, `tasmax_per`, `pr_per` where needed. I think it's better to apply the same logic everywhere, it makes the code much easier to maintain.

BTW, I forgot to add it in the initial PR statement but renaming these is a breaking changes, is okay ?

xclim/indicators/atmos/_temperature.py
```diff
@@ -834,11 +839,18 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
 
 warm_spell_duration_index = Temp(
     identifier="warm_spell_duration_index",
-    description="{freq} total number of days within spells of at least {window} days"
-    " with tmax above the 90th daily percentile.",
+    description="{freq} number of days with at least {window} consecutive days "
```
bzah, line 842: I suppose the french translation must also be updated accordingly.

xclim/indices/_multivariate.py
Zeitsperre: Changing call signatures is OK by me, but be sure to mention them in the History, and bump the version as well!

xclim/indicators/atmos/_temperature.py
```diff
@@ -834,11 +839,18 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
+    "where the daily maximum temperature is above the {per_base_thresh}th "
+    "percentile(s). A {per_window} day(s) window, centred on each calendar day in the "
+    "{per_period} period, is used to compute the {per_base_thresh}th percentile(s).",
```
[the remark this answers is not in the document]
bzah: I thought about it and I agree, but without a proper i18n module, including for English, it will be very artisanal to do so and might not work well in French (and upcoming translations).
Still I can go the artisanal way.

bzah:
Just a note on bootstrapping (yeah I love this topic). 
It's quite unnecessary to bootstrap percentiles when using non extreme percentiles such the 25th or the 75th. 

This PR encourages users to use various percentile thresholds on bootstrappable indices. 
Given the cost this algo, it would make sense to warn users when they try to bootstrap indices on non-extreme percentiles.

I think this should be addressed in another PR though.

For ref:
> [...] This bias is relatively greater for higher percentiles. For example, there would be only one exceedance over the 99th percentile, giving an exceedance rate of 1/150 = 0.67%, which is much smaller than the nominal rate of 1%. 
[Zhang et al.](https://journals.ametsoc.org/view/journals/clim/18/11/jcli3366.1.xml)

xclim/core/indicator.py
bzah: This should be resolved by https://github.com/Ouranosinc/xclim/pull/1050/commits/12fab092876292104a9ca9e183c2c8f493fd80b6

bzah:
The last unit test failure is a bit mysterious to me.
It's on test_temperature.py, in `test_warm_spell_duration_index`.

It seems I somehow changed how the output axes are ordered because, now 
`np.testing.assert_array_equal(out[0, :, 0], np.array([np.nan, 3, 0, 0, np.nan]))` raise an error
but, 
`np.testing.assert_array_equal(out[:, 0, 0], np.array([np.nan, 3, 0, 0, np.nan]))`does not...

aulemahal:
> The last "unit test" failure is a bit mysterious to me.
To me too... I quickly checked and I can't see where a change could have modified the dimension order. This being said, we had similar issues before and the current opinion is that the dimension order _is not_ guaranteed. This unit test was written before those discussions and it is kinda wrong to assume the order.

xclim/data/fr.json
```diff
@@ -856,17 +856,29 @@
     "abstract": "Nombre de jours avec précipitation au-dessus d'un seuil et températures sous un seuil."
   },
   "DAYS_OVER_PRECIP_THRESH": {
-    "title": "Nombre de jours pluvieux où la précipitation est au-dessus d'un percentile quotidien",
-    "abstract": "Nombre de jours d'une période où la précipitation est au-dessus d'un percentile quotidien et d'un seuil fixe.",
-    "description": "Nombre {freq:m} de jours où la précipitation au-dessus d'un percentile quotidien. Seuls les jours avec au moins {thresh} sont comptés.",
-    "long_name": "Nombre de jours pluvieux où la précipitation est au-dessus d'un percentile quotidien"
+    "title": "Nombre de jours pluvieux où la précipitation est au-dessus du {pr_per_thresh}e percentile",
```
bzah, line 859: Vous dites _la_ précipitation au Québec ?

xclim/testing/tests/test_precip.py
```diff
@@ -400,21 +401,27 @@ def test_days_over_precip_doy_thresh():
...
     assert "only days with at least 2 mm/d are counted." in out2.description
+    assert "[80]th percentile" in out2.attrs["description"]
+    assert "['1990-01-01', '1993-12-31'] period" in out2.attrs["description"]
+    assert "5 day(s)" in out2.attrs["description"]
 
 
 def test_days_over_precip_thresh():
     pr = open_dataset("ERA5/daily_surface_cancities_1990-1993.nc").pr
     per = pr.quantile(0.8, "time", keep_attrs=True)
+    per = PercentileDataArray.from_da(per, build_climatology_bounds(pr))
```
bzah, line 416: This case is what user must do when their percentiles/quantiles were not computed with `percentile_doy` in order to have the proper metadata.
Without this wrapping, they will get `{unknown}th percentile` and similar placeholders in the indicators metadata.

xclim/data/fr.json
aulemahal: Haha, c'est drôle on vient d'avoir ce débat. Je ne pense pas que le choix venait d'une différence Québec-France. On se disait que "la précipitation" était synonyme de "précipitation totale", c'est à dire en cumulant toutes les formes de précipitation. On se disait que de dire "les précipitations" impliquait une distinction potentielle entre les différentes phase.

Cela étant dit, si tu penses qu'une version est meilleure que l'autre, si tu connais des références sur lesquelles se baser, nous sommes à l'écoute!

xclim/testing/tests/test_precip.py
aulemahal: What does @huard  think? Is this too laborious?

xclim/data/fr.json
bzah: Ça m'a juste fait bizarre en lisant `la`, mais je ne pense pas que mon avis soit très pertient sur ce sujet.

aulemahal: haha, je comprends. Mais j'insiste par contre, je pense qu'on aimerait savoir ce que d'autres groupes de recherche en pensent. Avez-vous des traductions (officielles ou non) pour les indices de icclim?

huard (approved):

xclim/testing/tests/test_precip.py
huard: No, I think that's reasonable.

xclim/testing/tests/test_cli.py
```diff
@@ -1,373 +1,373 @@
...
-)
-def test_release_notes_failure(method, error):
-    runner = CliRunner()
-    results = runner.invoke(
-        cli,
-        ["release_notes", *method],
-    )
-    assert error in results.output
+# #!/usr/bin/env python
+# # Tests for `xclim` package, command line interface
```
bzah, line 2: Ooooooops

xclim/data/fr.json
bzah: A priori non, on n'a jamais travaillé sur la traduction des indices.

bzah:
@Zeitsperre the coverage should be better now. I initially commented out `test_cli` because of some error I have locally on `@cli.result_callback()`. But, I didn't meant to commit that.

xclim/core/utils.py
```diff
@@ -5,6 +5,8 @@
 
 Helper functions for the indices computation, indicator construction and other things.
 """
+from __future__ import annotations
```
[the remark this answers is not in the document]
bzah: True, I got carried away by this awesome syntax it enables 😄 
I'll rollback this and create an issue.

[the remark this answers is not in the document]
Zeitsperre: Agreed that it's awesome!

bzah:
@aulemahal is there something else you would like to discuss on these changes ?

xclim/core/formatting.py
```diff
@@ -46,6 +61,13 @@ def __init__(
         self.modifiers = modifiers
         self.mapping = mapping
 
+    def format(self, format_string: str, /, *args: Any, **kwargs: dict) -> str:
+        kw = {**kwargs}
+        for k, v in DEFAULT_FORMAT_PARAMS.items():
+            if k not in kw:
+                kw.update({k: v})
+        return super().format(format_string, *args, **kw)
```
aulemahal, line 69:
```suggestion
    def format(self, format_string: str, /, *args: Any, **kwargs: dict) -> str:
        for k, v in DEFAULT_FORMAT_PARAMS.items():
            if k not in kwargs:
                kwargs.update({k: v})
        return super().format(format_string, *args, **kwargs)
```

xclim/core/indicator.py
```diff
@@ -1027,7 +1040,8 @@ def _update_attrs(cls, args, das, attrs, var_id=None, names=None):
         # In the history attr, call signature will be all keywords and might be in a
         # different order than the real function (but order doesn't really matter with keywords).
         kwargs = OrderedDict(**das)
-        for k, v in args.items():
+        filtered_args = [(k, v) for k, v in args.items() if k in cls._all_parameters]
+        for k, v in filtered_args:
```
aulemahal, line 1044: Is this filter there to remove the copy of the PercentileDataArray? If so, why does it need to be removed? If it has the same name in `das` and `args` then it just replaces its copy, so it's a no-op, no?

aulemahal (approved):
Sorry not to have followed the development of this branch! 
Now that I re-read the changes, I'm not sure I understand the need for the need for the `PercentileDataArray` class.

The annotation is useful so that the base `Indicator` class knows how to format the param, how to inject the good attributes. But if `percentile_doy` already sets all the needed attributes, why do we need a suboject entirely? The stuff in  `get_metadata`  could be directly in `Indicator.format` or in a standalone function in `formatting.py`, no?

As I understand it, right now,  the computation works if the user uses a normal DataArray, but all three metadata entries for the percentile variable are "unknown".  Could we cast the input as a `PercentileDataArray` somewhere in the indicator's pipeline to ensure all available informations are added?

Finally, I think it would useful to add the following line to all percentile parameter's docstrings:
```
Xclim expects an output from :py:func:`~xclim.core.calendar.percentile_doy`.
```
(or something in the like).

bzah:
@aulemahal 
>Now that I re-read the changes, I'm not sure I understand the need for the PercentileDataArray class. 

I'm [quoting you](https://github.com/Ouranosinc/xclim/pull/1050#discussion_r846168726) for why I added the `PercentileDataArray` class:
> I am not convinced parsing the history is the safest way to check this. I think I'd rather prefer a new InputKind entry, that's what they were meant for. The problem here is that a "percentile threshold" is not differentiable from a normal input in the world of typing, and thus I think we'll need to use the same workaround as for DateStr.

I'm not sure how else it would be possible to recognize percentiles inputs.
However, `PercentileDataArray` could probably be hidden and not exposed in the API if that's what you meant by 
"cast the input as a PercentileDataArray somewhere in the indicator's pipeline".
I'll look into that!


----

On the other hand, I feel like all InputKind logic is some sort of typing over the python typing system. I don't think it's really necessary and IMHO it could be replaced by a classes and TypedDict. But, that's my java background speaking, if I don't see classes, POJO and Singleton every now and then it gets itchy.

bzah:
Alright, I went for a in-between implementation. 
I removed the `PERCENTILE_VARIABLE` InputKind and replaced it by a simple logic check to see if a DataArray is compatible with PercentileDataArray.

The `PercentileDataArray` class still exists but, is mainly used to easily create a valid percentiles input with `from_da` method.
I think it's useful in the case where percentiles are not computed with percentile_doy, like in precipitations example:
```
per = pr.quantile(0.8, "time", keep_attrs=True)
per = PercentileDataArray.from_da(per, climatology_bounds=build_climatology_bounds(pr))
```

In that case, we need to let the user fill `climatology_bounds` parameter because in Indicator pipeline we have no way to determine this period for which percentiles were computed.

PR #1059, opened 2022-04-21
Add ends to yearly window

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* The `append_end` argument has been added to `sdba.unpack_moving_yearly_window`. 
* When set to False, we keep the original behavior of returning the middle years of each window. Hence, the reconstructed timeseries is missing (window-time)/2 time points from the original timeseries.
* When set to True, we append the year before the middle years of the first window and the years after the middle years of the last window. Hence, we can recover the full inital timeseries.
* In both case, years that are not convered by a window at the end of the timeseries are not including. The user must make sure to set the window and step sizes probably, if they want to recover the initial time series.

Yes, the default behavior is `append_ends`=`True`, but we can recover the previous behavior by setting it to `False`

### DIFF
--- a/xclim/sdba/processing.py
+++ b/xclim/sdba/processing.py
@@ -575,7 +575,7 @@ def unpack_moving_yearly_window(da: xr.DataArray, dim: str = "movingwin", append
 
     if append_ends:
         out.insert(0, da.isel({dim: 0, 'time': slice(None, left * N_in_year)}).drop_vars(dim))  # add end at the front
-        back_end = da.isel({dim: -1, 'time': slice(-(left+1) * N_in_year, None)}).drop_vars(dim)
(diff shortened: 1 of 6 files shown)


xclim/sdba/processing.py
```diff
@@ -512,12 +512,17 @@ def construct_moving_yearly_window(
     return daw
 
 
-def unpack_moving_yearly_window(da: xr.DataArray, dim: str = "movingwin"):
+def unpack_moving_yearly_window(
+    da: xr.DataArray, dim: str = "movingwin", append_ends=True
```
aulemahal, line 516:
```suggestion
    da: xr.DataArray, dim: str = "movingwin", append_ends: bool = True
```

aulemahal (approved):
Thanks!

PR #1062, opened 2022-04-26
Utci

### LINKED ISSUE #1060: UTCI (Universal Thermal Climate Index)
### Addressing a Problem?

Hi guys,
I want to contribute with a new climate index.
It is the Universal Thermal Climate Index (UTCI).  
The calcualtation I'll take from [here](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#universal-thermal-climate-index-utci).
I'll do the implemetation in your code.
Are you interested in that new index and what is the easiest way for me to contribute?
Cheers

### Potential Solution

_No response_

### Additional context

_No response_

### Contribution

### Code of Conduct

ludwiglierhammer:
### Pull Request Checklist:
    - This PR fixes #1060

* new index: ultimate thermal climate index (UTCI)
* Calculation taken from [pythermalcomfort](https://pythermalcomfort.readthedocs.io/en/latest/)

No


aulemahal (changes requested):
Thanks for the PR! 
I'm hesitating in adding a dependency for a single indicator. My main concerns are that `pythermalcomfort`, while looking like it's currently active, still relies on a single developer. Moreover, it has a dependency (`jos3`) that looks very quiet and relies also on a single person. This situation may raise issues in the future as we can't be 100% sure the packages will be kept up to date.

As `utci` is not super complicated in it's logic and is not a creation of the devs of `pythermalcomfort`, I'd have no problem reimplementing it directly in xclim, based on the original source. 

Or, I would be ok with making `pythermalcomfort` an optional dependency. It might still trigger problems for us (the dev team) in the future, but at least those problems shouldn't affect the users. 

What do @Zeitsperre and @huard  think?

For reference, the main code for utci is here : https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/0e1b5e0ebd1f0dc3f968efb54112ac2d8fea1340/src/pythermalcomfort/optimized_functions.py#L465
It's a simple equation, but accelerated with numba, which we could do to.

huard:
I'm on the fence on this one, but I would also lean toward porting the numba code and tests in xclim with proper credits, and a link to the pythermalcomfort library in the docstring.

ludwiglierhammer:
I fully understand your concerns. I think it would be great to reimplement the `pythermalcomfort` indices in xclim. How can I help you with that?

aulemahal:
@ludwiglierhammer Well, for this indice in particular, it's quite simple because it's almost only a polynomial equation. So just coding it directly should be easy. 

`pythermalcomfort` uses the MIT license while we use the Apache 2.0 license. I believe, one could simply copy-paste their code in xclim and then arrange it so it fits our internal logic, as long as we cite them as the original source. However, they already are taking the code from elsewhere. As far as I understand, devs of `pythermalcomfort` are _not_ the creators/authors of UTCI. Thus, I woud prefer if the code in xclim was inspired by the original source, which seems to be here: http://www.utci.org/utci_doku.php  It's Fortran, but the core part is easy to understand, even if it is long.

I would do as you did, but instead of importing `utci` from `pythermalcomfort`, simply code it beside. I would also move the function to `_conversion.py`. Our internal naming convention is a bit messy, but that file is where the non-resampling indices live. It is also where the saturation vapor pressure function is, which you will need if not using `pythermalcomfort`.  If you are comfortable you could use numba (if you're not, I can integrate it after a first draft).

So : an indice function in `_conversion.py` that manages the masking and the unit conversions + a `_utci` function in the same file that contains only the long equation, directly sourced from the fortran code and optimized through numba (with `numba.vectorize`). In the main function you would call it with  `apply_ufunc`, with `dask='parallelized'`.

HOWEVER, I'm sorry we went this far, but I initially did not spend enough time looking at the function... UTCI uses 4 inputs:

- air temperature (dry bulb, `tas`)
- water vapour pressure (that we get from `hurs` and the saturation vapour pressure, which in turn is a function of `tas`)
- surface wind speed (`sfcWind`)
- the mean radiant temperature

This is the first time I see that last one Reading [wikipedia](https://en.wikipedia.org/wiki/Mean_radiant_temperature), I see that's it's not really a climate variable. It depends very strongly on a particular place and position. The calculation section is pretty clear about how difficult it is to estimate it. For this reason, I am now unsure this function is in scope for `xclim`... 
What do @huard  @Zeitsperre and @tlogan2000 think?

For me, it now looks more like a case where one would uses xclim's "virtual submodules" mechanic to add it in its scripts, but it might not be fit for the core xclim library...

Sorry, I should have looked into it more deeply before!

ludwiglierhammer:
We can set mean radiant temperature to air temperature, following the paper of [Froehlich and Matzarakis (2019)](https://gmd.copernicus.org/preprints/gmd-2019-202/gmd-2019-202.pdf) (page 5 line 2; they call mean radiant temperature `Tmrt` and air temperature `Ta`). They justified their decision by referencing [Jendritzky et al. (2012)](). With this information I think we can delete tr as parameter too.
I hope I can dispel your doubts.

ludwiglierhammer:
[Here](https://pubmed.ncbi.nlm.nih.gov/22187087/) is the Jendritzky paper.

ludwiglierhammer:
@aulemahal I moved `utci` to _conversion.py and abopted the code from `pythermalcomfort`. I didn't manage the numba thing so far.

Zeitsperre:
@ludwiglierhammer Thanks again for all your work! We are set to release a version today, but there are a handful of other changes we the maintainers need to add in (such as French translation). If all goes well, we'll be able to get this in the next release (v0.37).

Just a suggestion for future commits, but if you run the following commands, the autoformatter for our coding style will be run on commit:
```
$ pip install pre-commit
$ pre-commit install
```
For more information: https://pre-commit.com/

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1181,574 @@ def potential_evapotranspiration(
...
+
+    Parameters
+    ----------
+    tas : xarray.DataArray
+        Mean daily temperature
+    hurs : xarray.DataArray
+        Relative Humidity
+    sfcWind : xarray.DataArray
+        Wind velocity
+    tr: xarray.DataArray, optional
```
aulemahal, line 1206: It seems that `tr` exists in CMIP6, I didn't know that!:

https://github.com/PCMDI/cmip6-cmor-tables/blob/cf039eb222cb0b7b92b3df38d986b8015450a828/Tables/CMIP6_Eday.json#L1997-L2006

But it has a slightly different description. Do you know how it relates to "the mean radiant temperature" this indice expects? Are they close enough so that we could use `tr` if it was available from a model?
If not, I would prefer using a non-cmip6 name to avoid any confusion.  `tmrt` and `tmr` are available.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1181,574 @@ def potential_evapotranspiration(
...
+
+        tas_valid = valid_range(tas, (-50.0, 50.0))
+        tr_valid = valid_range(tr - tas, (-30.0, 30.0))
+        sfcWind_valid = valid_range(sfcWind, (0.5, 17.0))
+        valid = ~(np.isnan(tas_valid) | np.isnan(tr_valid) | np.isnan(sfcWind_valid))
+        utci_approx = np.where(valid, utci_approx, np.nan)
+
+        return np.round_(utci_approx, 1)
+
+    tr = tas.copy()
```
aulemahal, line 1747: Only if `tr is None`!

Also, I think you're missing a conversion of `sfcWind`.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1181,574 @@ def potential_evapotranspiration(
...
+    tas = convert_units_to(tas, "degC")
+    tr = convert_units_to(tr, "degC")
+    hurs = convert_units_to(hurs, "pct")
+
+    return xr.apply_ufunc(
+        _utci, tas, tr, sfcWind, hurs, dask="parallelized"
+    ).assign_attrs({"units": "degC"})
```
aulemahal, line 1754: When wrapping with `apply_ufunc`, the inputs are already passed to `_utci` as numpy arrays, so no need for the `np.array(...)` casts above.
However, I suggest using `saturation_vapor_pressure` from this file instead of reimplementing it. Thus, it would need to be done outside the `apply_ufunc`. 

I can contribute a few other performance technicalities next week, like numba.

aulemahal (changes requested):
I discussed with @tlogan2000  and looked quickly at the articles you cited and we agree that this is indeed fit for xclim, sorry for the hesitation!

Are you able to write a test function for this? We usually have two versions for the tests:
1. One using the indice directly with idealized data, in `xclim/testings/tests/test_indices.py`
2. One using the indicator with real data, comparing with known output, in `xclim/testing/tests/test_atmos.py`

For 2, I would suggest using our pytest fixture `atmosds`, which corresponds to the dataset returned by:
```python3
ds = open_dataset('ERA5/daily_surface_cancities_1990-1993.nc')
```
it includes `tas`, `hurs` and `sfcWind`. The expected data could be generated using `pythermalcomfort` or online calculator on utci.org.

Let us know If you are ok with doing all this or if you need some help! Thanks a lot!

xclim/indicators/atmos/_conversion.py
```diff
@@ -273,3 +274,16 @@ class Converter(Indicator):
...
     compute=indices.corn_heat_units,
 )
+
+universal_thermal_climate_index = Converter(
+    identifier="universal_thermal_climate_index",
+    units="K",
+    long_name="Universal Thermal Climate Index",
+    description="UTCI is the equivalent temperature for the environment derived from a reference environment. "
+    "It is one of the most comprehensive indices for calculating heat stress in outdoor spaces."
+    "Where the applicability wind speed between 0.5 and 17 m/s is exceeded, nan applied.",
```
huard, line 284: I'd shorten the description as well. Something like 

> UTCI is the equivalent temperature for the environment derived from a reference environment. and is used to evaluate heat stress in outdoor spaces.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1181,574 @@ def potential_evapotranspiration(
...
+def universal_thermal_climate_index(
+    tas: xr.DataArray,
+    hurs: xr.DataArray,
+    sfcWind: xr.DataArray,
+    tr: xr.DataArray = None,
+) -> xr.DataArray:
+    """
+    Mean Universal Thermal Climate Index (UTCI)
+
+    The mean of daily UTCI
```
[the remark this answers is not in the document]
huard: Agreed, the explanation should include information related to the bounds imposed to the input variables.

environment.yml
```diff
@@ -49,3 +49,4 @@ dependencies:
     - xdoctest
     - yamale
     - yamllint
+    - pythermalcomfort>=2.1.0
```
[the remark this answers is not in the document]
ludwiglierhammer: I'll give it a try tomorrow.

xclim/indicators/atmos/_conversion.py
```diff
@@ -273,3 +274,16 @@ class Converter(Indicator):
...
+    cell_methods="",
+    comment="Original formula published in Zare et al. 2018",
```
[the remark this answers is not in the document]
ludwiglierhammer: I took this information directly from `pythermalcomfort`
[UTCI description](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#universal-thermal-climate-index-utci)
[references](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#id67)

xclim/indices/_conversion.py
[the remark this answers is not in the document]
ludwiglierhammer: Totally agree, the 'mean' is redundant.
For more explanation we can take something from [here](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#universal-thermal-climate-index-utci).

xclim/indices/_conversion.py
ludwiglierhammer: [Here](http://www.utci.org/utci_doku.php) you can find the description and calculation of `tr`. The CMIP6 `tr` is not the correct radiative temperature. The 'Effective radiative surface temperature' (tr), in the CMIP6 var list, is the temperature associated with a black body (Earth) and used for the radiative transfer calculations. In the UTCI equation, the 'radiative temperature' we need is the [solar radiation](https://doi.org/10.1016/j.wace.2018.01.004) that is directly hitting the person or bouncing back from buildings onto a person in the urban canon. Thus, I'll use the non-cmip6 name `tmrt`. Or should we delete `tr` fully from the paramter list?

xclim/indices/_conversion.py
ludwiglierhammer: I took the exponential function from `pythermalcomfort` because the parameters are defined slightly differnte to `saturation_vapor_pressure`. Do you think I should use `saturation_vapor_pressure` anyhow? And what method do you prefer?

ludwiglierhammer:
Are you interested in implementing more climate indices from `pythermalcomfort` into xclim? If so should we ask the `pythermalcomfort` guys if they wanna contribute? 
And are you interested in an official cooperation between Ouranos and [GERICS](https://gerics.de/index.php.en) the insitutiton where I work. I and my instituition would be glad.

xclim/indices/_conversion.py
aulemahal: As it seems an essential element of the original equation, I'd leave `tmrt` in the indice. I'm guessing that the scope of xclim often will mean that it is not available, but at least we have the option and it makes it more obvious that not passing it _is an approximation_.

xclim/indices/_conversion.py
aulemahal: Indeed, I kinda assumed that this method would correspond to one of those we implemented, but it appears it isn't the case. I suggest that we implement that method and use it there. The fortran code cites this source : https://www.thunderscientific.com/tech_info/reflibrary/its90formulas.pdf (pp. 3-4) 
It's name would be "ITS90" I guess.

xclim/indicators/atmos/_conversion.py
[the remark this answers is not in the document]
aulemahal: Unless there are good arguments for choosing another reference, I would go for the oldest, the ones on utci.org. We can have more than one, but the goal of the metadata is to cite the _original_ implementation.
 Zare et al. seems to be an article that describes and analyzes the indice? If they contributed changes to the code (like the approximation Tmrt = tas, let's cite them.
If we feel that some reference could really help a user to understand better than the original ones, we can add them to the "Notes" section of the docstring. They won't be added to the computed indicator, but will appear in the docs.

aulemahal:
We need to discuss this among the core dev team and with the other parts of Ouranos, but those are nice ideas! 
Tagging those who know more than me on these issues : @tlogan2000 @huard

Zeitsperre:
@ludwiglierhammer

Don't forget to add your author details to the `.zenodo.json`, run `$ bumpversion patch`, as well as add your name as a contributor to the next version and briefly describe your additions in the relevant sections of `HISTORY.rst`.

ludwiglierhammer:
I tried to add ITS90 to the saturation_vapor_pressure test. Unfortunately, the results do not match with the expected values (see details in the above checks). How shpuld we proceed with this?

aulemahal:
Hi!
I finally took time to dive into this.
The last commit update the `_utci` function directly from the fortran code and vectorizes it with numba. I also made some changes to the saturation vapor pressure so that the code style is consistent.   Finally, I used numpy's `f2py` module to build a module and use the fortran code directly in my tests. 

A few issues and discussion points:

- Your saturation vapor pressure was wrong because you were using  `np.log1p(tas)` instead of `np.log(tas)`. I see that this comes from `pythermalcomfort` (ptc). I'm not sure how this slipped through. The `its90` method I pushed gives results with a maximum absolute error of 1e-5 compared to the calculation in the fortran file. The version you took from ptc is exactly the one from the fortran code, but including this `log1p` error and it yields errors up to 0.23 Pa (in my tests).

- `pythermalcomfort` rounds the output to the first decimal. Why? This is not mentioned in any of the papers I saw. The web calculator does it, but it's more an "esthetic" choice than scientific, no?

- The current code, when tested with the ERA5 dataset, yields results with a maximal absolute error of 2e-5 when compared with the fortran code. The error between the `pythermalcomfort` and the fortran code goes up to 0.103, but it seems only because of 1) the error in the saturation vapor pressure, 2) the unnecessary rounding. (I haven't updated the tests, I'll do it later using the fortran code).

- Reading the papers you linked, I have not found details about approximating tmrt to tas. I wanted to mention the approximation and cite a reference in the docstring. If we can't find a paper supporting this, we'll have to put the variable non-optional again.

ludwiglierhammer:
Hi @aulemahal,
many thanks for your help and work.
* yes, I just adopted the formula from `pythermalcomfort`. Hence the `np.log1p(tas)` comes directly from there. 
* I think it is very likely for esthetic reasons, it makes more sense scientifically to keep the decimals. 
* We should change `np.log1p(tas)` to `np.log(tas)` and keep the decimals. But I think you've already done this, isn't?
* Unfortunately, I think I totally missunderstood the two papers :/ [Here](https://rmets.onlinelibrary.wiley.com/doi/epdf/10.1002/gdj3.102 ) (Eq 2, Fig 1) you can find the calculation of `tmrt` from surface radiation budget and solar angle. They produced for ERA5-reanalysis the variables MRT & UTCI on the [ECMWF / CDS page](https://cds.climate.copernicus.eu/cdsapp#!/dataset/10.24381/cds.553b7518?tab=overview ). For more details have a look on [this](https://link.springer.com/content/pdf/10.1007/s00484-020-01900-5.pdf). Do you think we should implement this in xclim too? 
Then we need:
  * the solar elevation angle (The calculation is [here](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015GL066868) in Appendix A. How to manage with daily data?)
  * the downwelling thermal radiation flux from the atmosphere (`rlds`)
  * the  upwelling thermal radiation flux from the ground (`rlus`)
  * the direct solar radiation flux from the sun (`rsds`)
  * the diffuse solar radiation flux with the latter equal to the sum of the isotropic diffuse solar radiation flux (at the moment I am not sure how to calculate)
  * the surface-reflected solar radiation flux (`rsus`)
  
What do you think about it?

aulemahal:
Hi @ludwiglierhammer, thanks for the last papers! This is quite interesting. 

I suggest that we put `Tmrt` non-optional on UTCI, but that we add a `mean_radiant_temperature` indicator, based on the equations you cited.

I can try to add the needed variables to the test dataset. I guess they should be available in ERA5[-land], since that's what the paper uses?

For the solar angle, we already have something like that in `potential_evapotranspiration`, so it could simply be promoted to its own function. This one doesn't really need to be it's own indicator. If the solar angle is not provided in the `mean_radiant_temperature` function it could simply default to using the first variable's time and lat values. Assuming they are there, otherwise it would fail with a meaningful message. That part I can do.


Also, in case you didn't see that comment from @huard :
> A more formal collaboration with GERICS would be great! I suggest we hold a meeting to share objectives and roadmaps, and discuss how we can work together. Let us know when would be a good time (after May 16 preferably).

ludwiglierhammer:
Hi @aulemahal,
there is a [package](https://github.com/ecmwf-projects/thermofeel) on github with a supporting [paper](https://www.sciencedirect.com/science/article/pii/S2352711022000176?via%3Dihub). The data for the above paper (Napoli et al. 2020) was calculated with this package. Hence the required data shoul be available in ERA5. I'll try to implement `MRT` to xclim. It would be great if you can help me with the performance again like for `UTCI`.

ludwiglierhammer:
Hi @huard, 
sorry for the late reply. We are very busy in May. What about the beginning of July, let's say on 1st or 2nd?

Zeitsperre:
FYI, this could be useful for validation: https://cds.climate.copernicus.eu/cdsapp#!/dataset/derived-utci-historical?tab=overview

ludwiglierhammer:
Hi guys,
sorry for my late, late reply and many thanks for all of you work.
We had to clearify how to calculate `mean_radiant_temperature`.
As mentioned above in `thermofeel` we need the following six parameters to calculate `mrt`:
- **surface solar radiation downwards** (`ssrd`): This is `rsds`
- **surface net solar radiation** (`ssr`): we use **surface solar radiation upwards** instead because **surface net solar radiation** is only needed to calculate **surface solar radiation upwards**. This is `rlds`
-  **Total sky direct solar radiation at surface** (`f_dir`): We can't find a way to calculate this but **the ratio of direct solar radiation**. 
[Liljegren et. al (2008)](https://www.tandfonline.com/doi/abs/10.1080/15459620802310770) say (page 648) that "`f_dir(_ratio)` is the fraction of the total horizontal solar irradiance due to the direct beam of the sun; (1 - `f_dir(_ratio)`) is the fraction due to the diffuse solar irradiance". 
Hence `f_dir` = `f_dir(_ratio)` * `rsds`
From [PyWBGT](https://github.com/QINQINKONG/PyWBGT) I took the calculation of `f_dir(_ratio)`. The package references Liljegren et. al (2008).
- **Surface thermal radiation downwards** (`strd`): This is `rlds`.
- **Surface net thermal radiation** (`strr`): We use `rlus` instead (see above explanation).
- **cosine of solar zenith angle** (`cossza`): Here again many thanks for your work. The function of the average you implemented is:
`np.sin(declination) * np.sin(lat) + np.cos(declination) * np.cos(lat) * np.sin(h_s) / h_s`,
where `h_s` is the hour angle at sunset.
As we see in [Hogan, Hirahara (2016)](https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1002/2015GL066868) (A3) the formula is, I think: 
`np.sin(declination) * np.sin(lat) + np.cos(declination) * np.cos(lat) * (np.sin(h_max)-np.sin(h_min)) / (h_max-h_min)`,  
where `h_max` is the minimum of the current hour angle and the hour angle at sunset and `h_min` is the maximum of the current hour_angle and the hour angle at sunrise. 
I implemented some new stats in `cosine_of_solar_zenith_angle`:
  * the cosine of the solar zenith angle during the sunlit period of each interval (`csza_s`; stat is `sunlit`; from `PyWBGT`) 
  * the cosine of the solar zenith angle during each interval (`csza_i` ;stat is `interval`; from `PyWBGT`)
  * the instantaneous cosine of the solar zenith angle (`csza_inst` ;stat is `instant`; from `thermofeel`). 

Now we can calculate `mrt` using to stats:
- stat is `sunlit`: 
`csza_i`  is `csza_i` 
`csza_s` is `csza_s`
- stat is `instant`:
`csza_i` is `csza_inst`
`csza_s` is `csza_inst`

I took 3-hourly CORDEX EUR-11 data to calculate `utci`. Unfortunately, I got this user warning:
Variable has a non-conforming cell_methods: Got `time: point`, which do not include the expected `time: mean`.
The result is that `utci`'s time series is empty. This is because the time cell_methods of the radiation variables is `mean` and those of the remaining ones is `point`. Do you have an idea how to solve this issue?
My next steps are implementing some test.
@huard: What about our collaboration meeting?

ludwiglierhammer:
To get the `csza_i`, `csza_s` and `csza_inst` we need the time interval between two time steps in hours.
In the `mrt` routine I defined :
`interval = ((dates[1] - dates[0]).values.astype("timedelta64[h]").astype(np.int32))`
Maybe you know more elegant solution.
My solution presumes that the time intervals between all consecutive time steps are equal.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1173,446 @@ def potential_evapotranspiration(
...
+            & (-30 < delta)
+            & (delta < 30)
+            & (0.5 < sfcWind)
+            & (sfcWind < 17.0)
+        )
+    return utci
+
+
+def _fdir_ratio(dates, csza_i, csza_s, rsds):
+    """Return ratio of direct solar radiation."""
```
aulemahal, line 1502: Please add a reference here too.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1173,446 @@ def potential_evapotranspiration(
...
+    numerical weather prediction models. Int J Biometeorol 64, 1233–1245 (2020).
+    https://doi.org/10.1007/s00484-020-01900-5
+    """
+    rsds = convert_units_to(rsds, "W m-2")
+    rsus = convert_units_to(rsus, "W m-2")
+    rlds = convert_units_to(rlds, "W m-2")
+    rlus = convert_units_to(rlus, "W m-2")
+
+    dates = rsds.time
+    hours = dates.dt.hour
```
aulemahal, line 1565:
```suggestion
    hours = ((dates - dates.dt.floor('D')).dt.seconds / 3600).assign_attrs(units='h')
```
The time since midnight in hours, at the second resolution, instead of at the hour resolution. Is this better?

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1173,446 @@ def potential_evapotranspiration(
...
+    lat = rsds.lat
+    lon = rsds.lon
+    decimal_year = datetime_to_decimal_year(times=dates, calendar=dates.dt.calendar)
+    day_angle = ((decimal_year % 1) * 2 * np.pi).assign_attrs(units="rad")
+    dec = solar_declination(day_angle)
+    if stat == "sunlit":
+        interval = (
+            (dates[1] - dates[0]).values.astype("timedelta64[h]").astype(np.int32)
```
aulemahal, line 1573:
```suggestion
            (dates.diff('time').dt.seconds / 3600).reindex(time=dates.time, method='bfill')
```
How's that? One for each timestep and a resolution of 1s.   It assumes the coordinate marks the end of the interval, which is different from the assumption in `;csza` (there it takes the hour as the middle of the interval).

xclim/indices/helpers.py
```diff
@@ -0,0 +1,358 @@
...
+    if stat == "interval":
+        sha = (hours - 12) * 15 / 180 * np.pi + lon
+        k = interval / 2.0
+        h_s = sha - k * 15 * np.pi / 180
+        h_e = sha + k * 15 * np.pi / 180
+        csza = np.sin(declination) * np.sin(lat) + np.cos(declination) * np.cos(lat) * (
+            np.sin(h_e) - np.sin(h_s)
+        ) / (h_e - h_s)
+        return np.clip(csza, 0, None)
+    if stat == "sunlit":
```
aulemahal, line 259: This only works if the interval doesn't include the night. For example, daily data is usually given at midnight (the start of the day) : `hours = 0` and `interval=24`. Thus, `h_s = noon > h_sr` and `h_e = noon < h_ss` and we are integrating over the night, as if there was light. My example is not that bad because we kinda fall back to the correct answer anyway, but it illustrates a caveat of the function, I think.

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1173,446 @@ def potential_evapotranspiration(
...
+        )
+        csza_i = cosine_of_solar_zenith_angle(
```
aulemahal, line 1575: At first, I thought we would implement a daily version of the equations, as with most of xclim. Here, you assume subdaily data, which is good, but I think the daily case should be supported. I think we could check if the data is dail with `xr.infer_freq(dates) == 'D')` and if it is, use `method='average'` in `csza`. What do you think?

xclim/indices/_conversion.py
```diff
@@ -1180,3 +1173,446 @@ def potential_evapotranspiration(
...
+    rsds_direct = fdir_ratio * rsds
+    rsds_diffuse = rsds - rsds_direct
+
+    gamma = np.arcsin(csza_i)
+    fp = 0.308 * np.cos(gamma * 0.988 - (gamma**2 / 50000))
+    i_star = xr.where(csza_s > 0.001, rsds_direct / csza_s, 0)
+
+    tmrt = np.power(
+        (
+            (1 / 0.0000000567)
```
aulemahal, line 1609:
```suggestion
            (1 / 5.67e-8)  # Stefan-Boltzmann constant
```

xclim/indices/helpers.py
```diff
@@ -0,0 +1,358 @@
...
+
+import cftime
+import numpy as np
+import xarray as xr
+
+from xclim.core.calendar import datetime_to_decimal_year
+from xclim.core.units import convert_units_to
+
+
+def distance_from_sun(dates: xr.DaraArray) -> xr.DataArray:
```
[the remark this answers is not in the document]
ludwiglierhammer: Unfortunately, the time axis is emtpy too if I calcualte utci directly from the index function.
I added this as a reference for function `distance_from_sun`: U.S. Naval Observatory:Astronomical Almanac. Washington, D.C.: U.S.
Government Printing Office, 1985.
I took teh information from Liljegren et. al (2008).

xclim/indices/helpers.py
ludwiglierhammer: I added this code from [here](https://github.com/QINQINKONG/PyWBGT/blob/v1.0.0/src/coszenith.pyx#L100-L101)
csza = xr.where(((h_s > h_ss)
                         & (h_e < h_sr))
                        |
                        ((h_s < h_sr)
                         & (h_e < h_sr))
                        |
                        ((h_s > h_ss)
                         & (h_e > h_ss)), 0, csza)
I think this should fix it, isn't it?

xclim/indices/_conversion.py
ludwiglierhammer: Is this still necessary with my above fix?

ludwiglierhammer:
Hi @aulemahal, can you please add the radiation variables (`rsds`, `rsus`, `rlds` and `rlus`) to "ERA5/daily_surface_cancities_1990-1993.nc" or "atmosds". I need those variables for the `mrt` and `utci` testings. Thanks.

aulemahal:
I'm working on it!  They should be accessible later today.

aulemahal:
Done! For the development of the branch you can use:
```
ds = open_dataset('ERA5/daily_surface_cancities_1990-1993.nc', branch='add-radiation')
```
Which contains rsds, rlds which are copies of ssr and str from ERA5. Also, rlus = strd - str and rsus = ssrd - ssr. I think I got the signs ok.

tlogan2000:
@ludwiglierhammer  is this ready for a final review?   We are planning a new release (0.37) for the end of the week.

aulemahal:
I believe the warning is not related to the error. The CF checks do not change the data, they only raise a warning. 
When you say "empty", do you mean that it's all NaNs or that the time dimension collapsed to 0?

Is the data accessible in some way? Or maybe you could show us a printout of the datarrays?

ludwiglierhammer:
The data are accessible on [ESGF](https://esgf-data.dkrz.de/search/cordex-dkrz/):
**Domain:** EUR-11
**Institute:** CLMcom-BTU
**Driving Model:** MPI-M-MPI-ESM-LR
**Experiment:** rcp26
**Ensemble:** r1i1p1
**RCM Model:** CCLM4-8-17
**Time Frequency:** 3hr

And here is the time axis DataArray from utci (it is the same for all three stats):

xarray.DataArray'time'time: 0
array([], dtype=object)
Coordinates:
time
(time)
object
height
()
float64
10.0
Attributes:
standard_name :
time
long_name :
time
axis :
T

Hopefully, this is helpful or do you need more information.

ludwiglierhammer:
[This](https://nbviewer.org/github/ludwiglierhammer/Notebooks/blob/main/calculate_utci.ipynb) is the test notebook.

aulemahal:
I believe the issue comes from your data. At least, I wasn't able to reproduce it exactly with my tests on my side. I did not have access to your data so I used CORDEX data from our own model (MRCC5) instead. Instead of loading the data, I used dask and I got a problem where everything was NaN. I think the source of my problem is the same as for yours : radiation values are given at times in the middle of the simulation interval while tas, sfcWind and hurs are given at the beginning of the interval.

Look at the  `time` axis of rlds, rlus, rsus and rsds, t goes 1h30, 4h30, ... On the other hand, the other three variables go : 0h00, 3h00, ... I'd recommend doing something like:
```
rlds['time']  = rlds.time.dt.floor('3H')
# or
rlds['time'] = tas.time
```
Hope this clears out your problem!

While testing, I stumbled on a few problems:
- The `cosine_of_solar_zenith_angle` function is not lazy. The calls to `np.clip` force the loading of dask data, we should replace them by `da.clip(...)` (the xarray method).
- I had weird glitches when computing  Tmrt over NAM. Like impossible values at night and near the polar circle... I'll look into this a bit more later. Here are two examples : 

Noon UTC, in june
image

Midnight UTC, in june
image

ludwiglierhammer:
**To the time axis issue:**
If I implement this 
`rlds['time'] = tas.time`
the probelm is solved.
But then this is only valid for `mean_radiant_temperature`.
Does it make sense to implement a generel check, whether all input time axes are equal. Something like this:

```
def check_time_axis(data_arrays, equalize=False):
    if "time_axes are different":
        if equalize is False:
            raise ValueError("The time axes of your input data do not match.")
        elif equalize is True:
            warnings.warn("The time axes of your input data do not match. Equalize the time axes.")
            time_axis = data_arrays[0].time
            for da in data_arrays[1:]:
                da["time"] = time_axis
        else:
            raise ValueError("equalize must be one of [True, False]")
```
What do you think about this?

**To the `TMRT` issue:**
In `PyWBGT` there are many if cases while calculating some variables:
- [solar hour angle](https://github.com/QINQINKONG/PyWBGT/blob/v1.0.0/src/coszenith.pyx#L50-L55): 
- [hour angle of the starting point](https://github.com/QINQINKONG/PyWBGT/blob/v1.0.0/src/coszenith.pyx#L62-L67)
- The same for `hour angle of end point`
- [cosine zenith angle during only the sunlit period](https://github.com/QINQINKONG/PyWBGT/blob/v1.0.0/src/coszenith.pyx#L94-L119)
- [cosine zenith angle during each interval](https://github.com/QINQINKONG/PyWBGT/blob/v1.0.0/src/coszenith.pyx#L131-L140)
- Our cases are much simpler.

Do you think we should consider this too? 

I get similar weird glitches when computing `MRT` over NAM (I took CORDEX REMO20015_MPI-M-MPI-ESM-LR data) but only for stat is `instant` and not for stat is `sunlit`.

aulemahal:
Quickly, I don't think a function to fix the correct timeseries is necessary : I personally think it's out of scope for xclim. The same way as if you're data is all NaNs, it's not xclim's job to tell you.  Moreover, an expert user could try to use xarray's automatic broadcasting knowingly by passing variables with different time axes.

I'm also not motivated to implement a check (one that raises a warning, not an error), but we do have a mechanism for this (`datachecks`). If you feel it would be very useful, we could add such a check that looks at all input variables of an _indicator_ and warns the user on non-corresponding axes

For me this is a xarray-level user error and I don't want to fill up xclim with too many checks.

Zeitsperre:
In the interest of getting this into the next version (aiming for a 3PM EST release), perhaps adding a `Warning` section to the indice doc string concerning this potential use-case trip fall would be a good idea? I agree with @aulemahal that this looks like an xarray-specific "gotcha". If we were to implement a check for this, it would be much better to propose that in another PR.

@ludwiglierhammer If you feel that this PR is ready, please mark it ready for review and, assuming everything checks out, it will be in the next version. Thanks!

ludwiglierhammer:
@aulemahal: I can fully understand your argumentation and agree with this. Nevertheless, I think a warning if non-corresponding axes exist would be very helpful for the users. 

From my side this PR is ready. Unfortunately, the issue with the weird glitches is still remaining. Honestly, I am not sure how to fix this. 

I've really enjoyed working with you guys so far.

aulemahal (approved):
So I did a last cleanup for docstrings and the unused dep. 

I also renamed `tmrt` to `mrt`. I don't remember why we chose the first one? On wikipedia and elsewhere, the `mrt` acronym seems to be the convention. 

I added a small warning in the docstring of mrt for the polar regions, I'll open a new issue anyway.
Thanks a lot @ludwiglierhammer !

PR #1071, opened 2022-05-06
MAINT: Update typing thanks to `__future__`

### LINKED ISSUE #1065: Proposal: migrate xclim typing to python 3.10 style
### Generic Issue

* xclim version: 0.35.x
* Python version: 3.8
* Operating System: all

### Description

Follow up to this discussion: https://github.com/Ouranosinc/xclim/pull/1050#discussion_r859897719

The [__future__ module](https://docs.python.org/3/library/__future__.html#module-__future__) allows to access some of the latest features of python without needing to increment the minimal python requirement of xclim.

In [PEP604](https://peps.python.org/pep-0604/) for python 3.10, a new type hint syntax has been added. It is made available on older python version by starting modules with the statement `from __future__ import annotations` .
Basically, it replaces `Union` and `Optional` keywords by the pipe operator such: `data: str | int | None`
I believe this makes the typing much more readable and pleasant to use.

It also allows some forward referencing such as:
```

class Cacahuete()

   @classmethod
   def from_generic_food(food) -> Cacahuete: # Without annotations this would need to be a `-> Any` or a string `-> "Cacahuete"`
      return Cacahuete(food)
```

And it removes the need of most imports from `typing` module by allowing to use python types directly:
`def my_func(a: List[str], b: Optional[Tuple[int]]) -> Dict[str]:    pass `
becomes
`def my_func(a: list[str], b: tuple[int] | None) -> dict[str]:    pass `

### Code of Conduct

bzah:
### Pull Request Checklist:
    - This PR fixes #1065

Update typing to the newer and less verbose style.
I used `reorder_python_imports` to add the import, `pyupgrade` to update the typing and pycharm analysis tool to removed unused imports.

No.

### Other information:
I'm not sure if the pre-commit addition is necessary. `reorder_python_imports` is a tool which is supposed to replace isort but it also provides a feature to add `from __future__ import annotations`. It could be convenient to keep this when a new file is created and the developer forgot to add this import manually.
However, as is, the reordering is done by `reorder_python_imports` then is supersede by `isort` when pre-commit runs.


bzah:
I'm bored in my train so I had a look at these errors.
I didn't thought about it beforehand but this feature breaks how InputKind are created:
```
    if Optional[param.annotation] in [
        Optional[DataArray],
        Optional[Union[DataArray, str]],
    ]:
        return InputKind.OPTIONAL_VARIABLE
```
The above code cannot work now because `param.annotation` is now a string instead of being a type.

In the PEP it says:
> This PEP proposes changing function annotations and variable annotations so that they are no longer evaluated at function definition time. Instead, they are preserved in __annotations__ in string form.

Zeitsperre:
@bzah

Interesting! There are so many moving parts in xclim that I didn't even realize we used annotations as construction pieces. This is superb to know this early on that our approach will break in the upcoming changes.

I'll tag @aulemahal as a reviewer here, since he'll have a much better grasp of the amount of work needed to change this behaviour.

xclim/__init__.py
```diff
@@ -1,11 +1,12 @@
 """Climate indices computation package based on Xarray."""
+from __future__ import annotations
+
 from importlib.resources import contents, path
 
 from xclim.core import units  # noqa
 from xclim.core.indicator import build_indicator_module_from_yaml
 from xclim.core.locales import load_locale
 from xclim.core.options import set_options  # noqa
-from xclim.indicators import atmos, land, seaIce  # noqa
```
aulemahal, line 8: haha, what happened here? I think the unused imports remover was a bit insolent and forgot to listen to the `#noqa`? This one of the reason tests are failing. We need to import this to trigger the construction of all indices before being able to build the virtual ones (lower in this file).

Zeitsperre:
Disabling `isort` makes sense, but if possible, I'd want to avoid it. `pre-commit` will need to be re-run over the code in its entirety, as this is a hook that we don't typically run in our `tox` checks. These need to be synchronized better.

aulemahal:
The amount of work wasn't that bad! The new annotations _can_ be parsed and the old code could have worked as-is, but that's only with python 3.10. Even with the `__future__` trick, there's still some features of that PEP that aren't available. Thus, I completely modified the inputkind parser to use the string annotations. It seems to work, but it means we can _only_ use that way and not the old one.

I still need to update some documentation.

aulemahal (commented):
@Zeitsperre @bzah This seems all good for my part! The documentation of InputKind is fixed.

It seems the new hook duplicated the imports instead of grouping them into lists? I personally find it a bit to verbose, but I'll let you decide how to handle these pre-commit issues! 

Thanks @bzah !

Zeitsperre (approved):

Zeitsperre:
Thanks again @bzah and @aulemahal. Excited to use the new syntax!

bzah:
Thanks guys ! 
About the verbosity, in its documentation reorder_python_imports says it's to reduce the amount of git conflicts.
However, I don't think it happens often enough to really care about this.

Zeitsperre:
@aulemahal unless you have other comments. I think we can merge today. All good for you?

PR #1076, opened 2022-05-10
Keep attrs

### LINKED ISSUE #1026: Implement a "keep_attrs" argument when calculating variables
**Is your feature request related to a problem? Please describe.**
When using xclim to calculate a new variable (e.g. `tasmin/tasmax --> tas`), some essential attributes such as `grid_mapping` are lost.

**Describe the solution you'd like**
A "keep_attrs" functionnality could be added, which would compare the attributes of the old and new variables and re-add any missing one. My main issue is with `grid_mapping`, but other similar attributes might be lost too.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1026

* Adds a `keep_attrs` option.
When attributes are "kept" it means that attributes from all inputs (DataArrays) are merged according to strategy "drop_conflicts" and then updated with those provided by the indicator before being put on the output(s).

Behaviour of the option is as follows:

- if xclim's `keep_attrs` is True, attrs are kept.
- if xclim's `keep_attrs` is False, attrs are not kept (current behaviour).
- if xclim's `keep_attrs` is "xarray" and xarray's `keep_attrs` is True, attrs are kept.

xclim's `keep_attrs` defaults to  "xarray" and xarray's `keep_attrs` defaults to "default". Thus, the default behaviour is to NOT keep attributes.

Not really. I modified how attributes are added to the outputs of indicators : instead of _updating_ the dictionary, we now _override_ it. Before, it could happen that for some indicators, attributes were kept, especially if xarray's `keep_attrs` was True. This is not the case anymore, attributes are handled explicitly.


xclim/core/options.py
```diff
@@ -168,6 +172,12 @@ class set_options:
     - ``sdba_encode_cf``:
       Whether to encode cf coordinates in the ``map_blocks`` optimization that most adjustment methods are based on.
       This should have no impact on the results, but should run much faster in the graph creation phase.
+    - ``keep_attrs``:
+      Controls attributes handling in indicators. If True, attributes from all inputs are merged
+      using the `drop_conflicts` strategy and then update with xclim-provided attributes.
```
huard, line 177:
```suggestion
      using the `drop_conflicts` strategy and then updated with xclim-provided attributes.
```

xclim/core/options.py
```diff
@@ -168,6 +172,12 @@ class set_options:
...
+      If False, attributes from the inputs are ignored. If "xarray", xclim will use the option
+      as set in xarray's options. Note that xarray's "default" is equivalent to False.
```
huard, line 179:
```suggestion
      If False, attributes from the inputs are ignored. If "xarray", xclim will use xarray's `keep_attrs` option. 
      Note that xarray's "default" is equivalent to False.
```

huard (approved):
All good.

aulemahal:
To workaround this, I forced xarray's `keep_attrs` to False for the indice computation. This means that attrs are always lost at the first operation and thus the indice fully controls the attributes it outputs. The option is set through a context, so if the user already chose a value it is not modified outside the indice computation.

I also removed all `keep_attrs=True` I found in the indices. It was a trick we often use instead of explicitely setting units that are not modified. But I think it's better to handle units explicitly.

PR #1079, opened 2022-05-12
Move sun-related helpers to their own file

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

So I wanted to help out on `utci` and make all those solar angles and radiation calculations more organized and accessible and the change grew so large that I decided to make a PR so we can discuss this separately from the UTCI stuff.

* I reorganized all solar angle and radiation functions into `_helpers.py`. This file is meant of helper functions that manipulate real data and perform some physical operations. (In opposition to `generic` which is more "meta").
* I used citable equations. This meant modifying some calculations, at least the default version. This was easy going for the extraterrestrial radiation of the potential evapotranspiration, but not for BEDD and Huglin. I thus had to go back an reimplement the "old" ways and add a "method" arg to choose between equations.
* As `latitude` is needed for all algorithms of PET, I added it as an explicit input. We could use `cf_xarray`, but I'm not sure how generic this is.

`lat` is now needed for `potential_evapotranspiration` and `water_budget`.

### Other information:
@ludwiglierhammer When this is ready, you will be able to use `cosine_of_solar_zenith_angle` in the `Tmrt` function.

The issues with precision are because of two functions:
1. Solar declination angle
    The new method is closest to reality, it uses a few terms of the fourier series of the observations. The old method assumed that the orbit is a circle, which is good enough, but you know.
2. Eccentricity correction factor
    Similar: the new method uses the first terms of the fourier series of the obs. The old only used the 2 first terms.

@Zeitsperre  How strict do we need to be with the tests of BEDD and Huglin? The problem is that the very small difference in day length that the different solar declination equation yields adds up over the aggregation period and we get relative errors of around 0.01, which is larger than the current tolerance.

@raquel-ucl Flagging you so you are aware of the changes proposed here. They will slightly affect results of the MB05 methods you guys added in xclim.

Over all (@huard @tlogan2000 ), am I being too verbose here? I thought this would be simple and now I added multiple lines of code for functions that could have been implicit (but duplicated).


aulemahal:
Damned I wanted to merge into UTCI. Closing and opening elsewhere.

aulemahal:
Reopened here : https://github.com/ludwiglierhammer/xclim/pull/1

Issue #131, opened 2018-12-10
Drought indices


Balinus:
Drought indices from the source @huard pointed me is now developing an xarray interface. Might be useful to keep an eye on it.

See https://github.com/monocongo/climate_indices/issues/191

monocongo:
This effort is somewhat stalled, as I discovered limitations when using dask that seem to prevent its use internally (i.e. I'd need to pull out numpy arrays in order to do several things that you can't do using dask, such as indexing individual array elements using integer array indices, etc.). See [this](https://stackoverflow.com/questions/52933553/dask-assignment-error-when-updating-a-value-in-a-dask-array-using-typical-numpy) and [this](https://stackoverflow.com/questions/53292792/dask-equivalent-of-numpy-convolve-hstack) for some of the details.

I'll be happy to pick this up again, as I'd like to be able to leverage dask for larger input datasets, as well as making the library easier to integrate into xclim (if that's the intention here). It's just not clear how I can get there from here -- apparently my xarray/dask-fu is still weak.

BTW I ended up getting much better performance using vanilla multiprocessing mapping of `numpy.apply_along_axis` rather than `xarray.apply_ufunc`, but there may be something obvious I'm missing there. See monocongo/climate_indices#210 for details.

aulemahal:
@huard  @tlogan2000  Is this a library that would be interesting to have as a dependency and wrap into indices/indicators? Or should we reimplement the indices so that they make use of xarray/dask and follow our standards (instead of numpy/numba)?

monocongo:
If this ball gets rolling I'll be happy to help where I can. My current processing scripts utilize a complicated shared memory solution which I'd love to replace with something more elegant by leveraging xarray and/or dask, so if you guys have good ideas I'd love to piggyback on it (my xarray/dask-fu is still weak).

huard:
I think this would certainly be a valuable addition, but looking at the original code, probably not a trivial task. Before we jump into this, I suggest we make sure that
1. We have identified potential users
2. James has some free time to help us out
3. We have enough free cycles to do a good job.

tlogan2000:
In terms of projects / users asking for indices to my knowledge we have only SPEI so far (I believe PET would needed as a sub calculation though).  

So we could maybe look at implementing this single calculation to start with?  FYI We are using environment canada SPEI datasets (already calculated) for current needs so there is no real urgency to have this done in the short-term

monocongo:
[...]

PR #1096, opened 2022-05-31
Spi spei indices

coxipi:
### Pull Request Checklist:
    - This PR fixes (partially) #131

Adds functions to compute new indices
* Standardized precipitation index (SPI)
* Standardized precipitation evapotranspiration index (SPEI)

No.


xclim/indices/_agro.py
```diff
@@ -580,6 +584,167 @@ def water_budget(
...
+        [
+            params.isel(month=i)
+            .drop_vars("month")
+            .assign_coords(time=grouped_time[i][1][0].values)
+            .expand_dims("time")
+            .reindex_like(grouped_time[i][1], method="ffill")
+            for i in range(0, params.sizes["month"])
+        ],
+        dim="time",
+    )
```
aulemahal, line 668:
```suggestion
    params = params.rename(month='time').reindex(time=pr.time.dt.month)
```
I believe this will have the same effect, with the bonus of not relying on an explicit loop.

xclim/indices/_agro.py
```diff
@@ -580,6 +584,167 @@ def water_budget(
...
+
+    # Invert to normal distribution with ppf and obtain SPI
+    params_norm = xarray.DataArray(
+        [0, 1],
+        dims=["dparams"],
+        coords=dict(dparams=(["loc", "scale"])),
+        attrs=dict(scipy_dist="norm"),
+    )
+    params_norm.attrs = ""
+    params_norm.attrs["scipy_dist"] = "norm"
```
aulemahal, line 687:
```suggestion
```
Hmmm. You're setting the attributes, dropping them and rewriting them?

xclim/indices/stats.py
```diff
@@ -153,6 +155,62 @@ def fit(
...
+    Notes
+    -----
+    The resampling frequency time delta must be an integer number of months. It determines time segments used to fit
+    the array group by group. For example, choosing `freq = 'MS` will give 12 time segment, i.e. months.
+    A time series for each month is used to obtain a set of fitting parameters.
+    """
+    # Freq must have a time delta which is an integer number of month
+    dts = pd.date_range("1950-01-01", periods=2, freq=freq).to_pydatetime()
+    if dts[-1].day - dts[0].day != 0 or dts[-1].second - dts[0].second != 0:
+        raise NotImplementedError(f"Resampling frequency `{freq}` not supported.")
```
aulemahal, line 194: The frequency strings are limited, I believe you can hardcode this:
```suggestion
    _, base, _,  _ = parse_offset(freq)
    if base not in ['Y', 'A', 'Q', 'M']:
        raise NotImplementedError(f"Resampling frequency `{freq}` not supported.")
```
This needs `from xclim.core.calendar import parse_offset`

xclim/indices/stats.py
```diff
@@ -153,6 +155,62 @@ def fit(
...
+
+    # groupby argument, depends on the smallest time period allowed with freq
+    group = "time.month"
+
+    # Resampling and rolling
+    da = da.resample(time=freq).mean()
+    if window > 1:
+        attrs = da.attrs.copy()
+        da = da.rolling(time=window).mean(skipna=False)
+        da.attrs.update(attrs)
```
aulemahal, line 204:
```suggestion
        da = da.rolling(time=window).mean(skipna=False, keep_attrs=True)
```

xclim/indices/stats.py
```diff
@@ -153,6 +155,62 @@ def fit(
...
+
+    Parameters
+    ----------
+    da : xarray.DataArray
+      Input data.
+    dist : str
+      Name of the univariate distribution, such as `beta`, `expon`, `genextreme`, `gamma`, `gumbel_r`, `lognorm`, `norm`
+      (see scipy.stats).
+    freq : str
+      Resampling frequency.
```
aulemahal, line 174:
```suggestion
    freq : str
      Resampling frequency. This function only makes sense for monthly or seasonal frequencies.
```

xclim/indices/stats.py
```diff
@@ -153,6 +155,62 @@ def fit(
...
+    # Freq must have a time delta which is an integer number of month
+    dts = pd.date_range("1950-01-01", periods=2, freq=freq).to_pydatetime()
+    if dts[-1].day - dts[0].day != 0 or dts[-1].second - dts[0].second != 0:
+        raise NotImplementedError(f"Resampling frequency `{freq}` not supported.")
+
+    # groupby argument, depends on the smallest time period allowed with freq
+    group = "time.month"
+
+    # Resampling and rolling
+    da = da.resample(time=freq).mean()
```
aulemahal, line 200: I understand that this is what SPI does. but I think I wouldn't expect a function named `fit_group` to average over the group.  When I first suggested this, I thought the SPI fit was taken over _all_ available (daily) values. Seems strange to me to actively remove information before fitting, but that's ok for SPI if that's how they do it.

But, I think a `fit_group` function should fit on the group, not average the group, so the mean could be taken in the `spi` function, before calling this. However, this will break your "window" logic...

xclim/indices/_agro.py
```diff
@@ -11,6 +12,7 @@
 from xclim.core.utils import DayOfYearStr
```
coxipi, line 12:
```suggestion
from xclim.core.utils import DayOfYearStr
from xclim.core.calendar import parse_offset
```

coxipi, line 12:
```suggestion
from xclim.core.utils import DayOfYearStr
from xclim.core.calendar import parse_offset
```

coxipi (commented):
Good suggestions, commented some of them

xclim/indices/stats.py
aulemahal: Rather, my question is on why we couldn't have groups like:
group1 = {Jan 1 yr1, Jan 2 yr 1, Jan 31 yr1, Jan 1 yr2,... Jan 31 yr2, ....}
and so one.

Here the function "fit over the averaged group" instead of "fit over the group". I guess these indices have strict definitions, but wouldn't an SPI that was calibrated on all values actually yield a better results? since, there were more points involved in the fitting?

All that said, if we put the resampling outside this function, the rolling will break, unless we express the window in terms of timesteps of the input data instead of groups.

I think we need to discuss this further, as I may not understand the needs of SPI...

xclim/indices/stats.py
aulemahal: I think my comment here relates to the one below about the resampling. But mainly, if we hardcode `time.month`, then frequencies other than `M' or 'Q' won't do what your are describing, no?  But, I did not think about the year+rolling case you mention, that would make sense here.

En fait, this function is only missing the capacity to fit over the `time.dayofyear` group. (And `time.week` but who does that anyway.)

xclim/indices/stats.py
aulemahal: For the behaviour of the `window` argument, I initially thought it would look like what `xclim.sdba.Grouper` does : if the data is daily, window=31 and group='month', then we have:
group1 = {Dec 15 yr0, ... Feb 15 yr1, Dec 15 yr1, ..., Feb 15 yr2}
group2 = {Jan 15 yr1, ..., Mar 15 yr1, Jan15 yr2, ..., Mar 15 yr2}
and so on

This is different from what is done here and thus from what SPI wants. But I believe it is what would make sense in a `fit_group` function. Meaning that we could simply downstream your resampling and rolling to the SPI indice?

coxipi: Reply to first comment
------------------------------
I'm not sure what is better from a statistical point of view. Fitting:
 group1 = {Jan 1 yr1, Jan 2 yr 1, Jan 31 yr1, Jan 1 yr2,... Jan 31 yr2, ....}
 would indeed give more data to work with, but you would also work with more variability in the dataset, no?
 
 As for definition of SPI, I think indeed it is often applied on a "monthly" dataset, which must mean you don't have information about specific days. For instance: 
 https://www.in.gov/dnr/water/water-availability-use-rights/water-resource-updates/monthly-water-resource-summary/explanation-of-standard-precipitation-index-spi/
 "A 1-month SPI typically compares well to the percent of normal precipitation for the 
  
 Reply to second comment
 -----------------------------------
 Ok, I see the difference. However, in what you describe "data is daily, window=31 and group='month'," I don't see how you would know that DEC,JAN,FEB are grouped together, then JAN,FEB,MAR, then ...  etc

Yes, we can put resampling and rolling directly in the SPI indice, and then simply call `fit` from within that call, removing the need for fit_group.

aulemahal: Ok, having read your link and some other description, I really think  the resampling and rolling should be specific to SPI. 
If you can adapt `fit` so that `fit(da.groupby('time.month'), ...)` is possible, then maybe that's the simplest solution! Otherwise, we can keep a very simple `fit_group`.

coxipi: Yes `fit_group` can be avoided very easily. Only a few lines are necessary to add in the SPI code

```python
pr_cal = pr_cal.resample(time=freq).
pr_cal = pr_cal.rolling(time=window).mean(skipna=False, keep_attrs=True)
group = "time.month"
params = pr_cal.groupby(group).map(lambda x: fit(x, dist, method, dim))
```

By the way, we could give the possibility of a daily SPI. We would need to allow Higher resampling frequency (lower period, for instance `freq="D"`) and choose an appropriate `group = "time.dayofyear"` or something like this given this frequency.

xclim/data/fr.json
```diff
@@ -222,6 +222,18 @@
...
+    "long_name": "Indice de précipitation standardisé (SPI)",
+    "description": "Précipitations sur une fenêtre roullante {window}-X, normalisées tel que la moyenne de SPI est 0. L'unité `X` de la fenêtre est l'unité de temps minimale  déterminée par la fréquence de rééchantillonage {freq}",
+    "title": "Indice de précipitation standardisé (SPI)",
+    "abstract": "Précipitation accumulée sur une fenêtre roullante puis normalisée autour de zéro."
+  },
+  "SPEI": {
+    "long_name": "Indice de précipitation évapotranspiration standardisé (SPEI)",
+    "description": "Budget d'eau (précipitation - évapotranspiration) sur une fenêtre roullante {window}-X, normalisées tel que la moyenne de SPEI est 0. L'unité `X` de la fenêtre est l'unité de temps minimale  déterminée par la fréquence de rééchantillonage {freq}",
+    "title": "Indice de précipitation évapotranspiration standardisé (SPEI)",
+    "abstract": "Budget d'eau accumulé sur une fenêtre roullante puis normalisé autour de zéro."
```
aulemahal, line 235:
```suggestion
    "long_name": "Indice de précipitation standardisé (SPI)",
    "description": "Précipitations sur une fenêtre mobile de {window} {freq:nom}, normalisées tel que la moyenne de SPI est 0.",
    "title": "Indice de précipitation standardisé (SPI)",
    "abstract": "Précipitations sur une fenêtre mobile, normalisées tel que la moyenne de SPI est 0. L'unité de la fenêtre est l'unité de temps déterminée par la fréquence de rééchantillonage",
  },
  "SPEI": {
    "long_name": "Indice de précipitation évapotranspiration standardisé (SPEI)",
    "description": "Budget d'eau (précipitation - évapotranspiration) sur une fenêtre mobile de {window} {freq:nom}, normalisé tel que la moyenne de SPEI est 0.",
    "title": "Indice de précipitation évapotranspiration standardisé (SPEI)",
    "abstract": ""Budget d'eau (précipitation - évapotranspiration) sur une fenêtre mobile, normalisé tel que la moyenne de SPEI est 0. L'unité de la fenêtre est l'unité de temps déterminée par la fréquence de rééchantillonage.",
```
J'avoue que la passe-passe `{freq:nom}' est pas clairement expliquée dans la doc. Je pense que ça va être plus joli. Mais il faut aussi ajouter:
```
"D": [
    "quotidien",
    "quotidienne",
    "quotidiens",
    "quotidiennes",
    "jours
]
```
dans le "attrs_mapping", plus haut dans ce même fichier.

xclim/indicators/atmos/_precip.py
```diff
@@ -247,6 +250,31 @@ class HrPrecip(Hourly):
     parameters={"tas": {"kind": InputKind.VARIABLE}, "phase": "solid"},
 )
 
+standardized_precipitation_index = PrecipWithIndexing(
+    title="Standardized Precipitation Index (SPI)",
+    identifier="spi",
+    units="",
+    standard_name="spi",
+    long_name="Standardized Precipitation Index (SPI)",
+    description="Precipitations over rolling window {window}-X window, normalized such that SPI averages to 0. The window unit `X` is the minimal time period defined by resampling frequency {freq}",
```
aulemahal, line 259:
```suggestion
    description="Precipitations over a rolling window of {window} {freq:noun}, normalized such that SPI averages to 0.",
```
Pareil. Ici il faudra ajouter `"D": ["daily", "days"]` dans `xclim.formatting.default_formatter`.

xclim/indicators/atmos/_precip.py
```diff
@@ -247,6 +250,31 @@ class HrPrecip(Hourly):
...
+    parameters=dict(window=1, dist="gamma", method="ML"),
+)
+
+standardized_precipitation_evapotranspiration_index = PrecipWithIndexing(
+    title="Standardized Precipitation Evapotranspiration Index (SPEI)",
+    identifier="spei",
+    units="",
+    standard_name="spei",
+    long_name="Standardized Precipitation Evapotranspiration Index (SPEI)",
+    description="Water budget (precipitation - evapotranspiration) over rolling window {window}-X window, normalized such that SPEI averages to 0. The window unit `X` is the minimal time period defined by resampling frequency {freq}",
```
aulemahal, line 272:
```suggestion
    description="Water budget (precipitation - evapotranspiration) over a rolling window of {window} {freq:noun}, normalized such that SPEI averages to 0.",
```

xclim/indicators/atmos/_precip.py
```diff
@@ -247,6 +250,31 @@ class HrPrecip(Hourly):
...
+    title="Standardized Precipitation Index (SPI)",
+    identifier="spi",
+    units="",
+    standard_name="spi",
+    long_name="Standardized Precipitation Index (SPI)",
+    description="Precipitations over rolling window {window}-X window, normalized such that SPI averages to 0. The window unit `X` is the minimal time period defined by resampling frequency {freq}",
+    cell_methods="",
+    compute=indices.standardized_precipitation_index,
+    # the freq parameter seems to cause errors when doing the tests
+    parameters=dict(window=1, dist="gamma", method="ML"),
```
aulemahal, line 263: Doesn't this injects the parameters? Don't we want these params to be accessible to the user?

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+    params = pr_cal.groupby(group).map(lambda x: fit(x, dist, method))
+    params = resample_to_time(params, pr)
+
+    # ppf to cdf
+    # zero-bounded distributions;  'pearson3' will also go in this group once it's implemented
+    if dist in ["gamma", "pearson3"]:
+        prob_pos = dist_method("cdf", params, pr.where(pr > 0))
+        prob_zero = resample_to_time(
+            pr.groupby(group).map(
+                lambda x: (x == 0).sum() / x.dropna(dim="time").shape[0]
```
aulemahal, line 728:
```suggestion
                lambda x: (x == 0).sum('time') / x.notnull().sum('time')
```
We can't assume dimension order!

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+    params_norm = xarray.DataArray(
+        [0, 1],
+        dims=["dparams"],
+        coords=dict(dparams=(["loc", "scale"])),
+        attrs=dict(scipy_dist="norm"),
+    )
+    spi = dist_method("ppf", params_norm, prob)
+    spi.attrs["units"] = ""
+    spi.attrs["standard name"] = "SPI"
+    spi.attrs["long name"] = "Standardized Precipitation Index"
```
aulemahal, line 744:
```suggestion
```
That's the indicator's job.

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+        # Distributions bounded by zero: Water budget must be shifted, only positive values
+        # are allowed. The offset choice is arbitrary and needs to be revisited.
+        # In monocongo, the offset would be 1000/(60*60*24) in [kg m-2 s-1]
+        # The choice can lead to differences as big as +/-0.2 in the SPEI.
+        # If taken too big, there are problems with the "ML" method  (this should be an
+        # issue with the fitting procedure that also needs attention)
+        offset = xarray.DataArray([1e-04])
+        offset.attrs["units"] = "kg m-2 s-1"
+        convert_units_to(offset, wb.units)
+        offset = offset.values.item()
```
aulemahal, line 803:
```suggestion
        offset = convert_units_to("1e-4 kg m-2 s-1", wb.units)
```

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+        # Increase offset if negative values remain
+        if wb.min() + offset < 0 or wb_cal.min() + offset < 0:
+            offset = offset - 2 * min(wb.min(), wb_cal.min())
```
aulemahal, line 806: You can't test a DataArray within a `if` because it might be using dask, which means it hasn't been computed yet and the user doesn't expect a computation right now.

Is this related to our previous discussion? I mean, can't you simply always  add the full offset? It could then look like:
```suggestion
         offset = offset - 2 * min(wb.min(), wb_cal.min(), 0)
```

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+        with xarray.set_options(keep_attrs=True):
+            wb, wb_cal = wb + offset, wb_cal + offset
+
+    spei = standardized_precipitation_index(wb, wb_cal, freq, window, dist, method)
+    spei.attrs["standard name"] = "SPEI"
+    spei.attrs["long name"] = "Standardized Precipitation Evapotranspiration Index"
```
aulemahal, line 812:
```suggestion
```

xclim/indices/stats.py
```diff
@@ -121,7 +131,8 @@ def fit(
         output_dtypes=[float],
         keep_attrs=True,
         kwargs=dict(
-            dist=dc if method == "ML" else lm3dc,
+            # Don't know how APP should be included, this works for now
```
aulemahal, line 134: A bit overkill for small data, but it works and might even be more performant than calling the `_fit_start` method directly for data with many chunks.

xclim/testing/tests/test_indices.py
```diff
@@ -413,6 +413,112 @@ def test_effective_growing_degree_days(
...
+            ("MS", 12, "gamma", "ML", [0.5982, 1.5598, 1.6931, 0.9964, 0.7028], 0.034),
+        ],
+    )
+    def test_standardized_precipitation_index(
+        self, freq, window, dist, method, values, diff_tol
+    ):
+
+        # Only location=1 is tested
+        ds = open_dataset("sdba/CanESM2_1950-2100.nc").isel(location=1)
+        pr = ds.pr.sel(time=slice("2000"))
```
aulemahal, line 455:
```suggestion
        pr = ds.pr.sel(time="2000")
```

xclim/testing/tests/test_indices.py
```diff
@@ -413,6 +413,112 @@ def test_effective_growing_degree_days(
...
+            ("MS", 1, "gamma", "APP", [1.505, 1.753, 2.334, -2.484, 0.879], 0.2),
+            ("MS", 12, "gamma", "APP", [0.6012, 1.6908, 1.8678, 0.9801, 0.6566], 0.024),
+            ("MS", 1, "gamma", "ML", [1.505, 1.753, 2.334, -2.484, 0.879], 0.76),
+            ("MS", 12, "gamma", "ML", [0.6012, 1.6908, 1.8678, 0.9801, 0.6566], 0.038),
+        ],
+    )
+    def test_standardized_precipitation_evapotranspiration_index(
+        self, freq, window, dist, method, values, diff_tol
+    ):
+        print(values)
```
aulemahal, line 488:
```suggestion
```
héhé some leftovers?

xclim/testing/tests/test_indices.py
```diff
@@ -413,6 +413,112 @@ def test_effective_growing_degree_days(
...
+        # Only location=1 is tested
+        ds = (
+            open_dataset("sdba/CanESM2_1950-2100.nc")
+            .isel(location=1)
+            .sel(time=slice("2000"))
```
aulemahal, line 493:
```suggestion
            .sel(time="2000")
```

xclim/testing/tests/test_indices.py
```diff
@@ -413,6 +413,112 @@ def test_effective_growing_degree_days(
...
+        )
+        pr = ds.pr
+        # generate water budget
+        with xr.set_options(keep_attrs=True):
+            tasmax = ds.tasmax
+            tas = tasmax - [2 + 1 * 0.5 for i in range(ds.tasmax.shape[0])]
+            tasmin = tasmax - [4 + 2 * 0.5 for i in range(ds.tasmax.shape[0])]
```
aulemahal, line 500:
```suggestion
            tas = tasmax - 3.5
            tasmin = tasmax - 7
```
Or did I miss something?

xclim/testing/tests/test_precip.py
```diff
@@ -92,6 +93,68 @@ def test_with_different_phases(self):
...
+class TestStandardizedPrecip:
+    nc_ds = os.path.join("sdba", "CanESM2_1950-2100.nc")
+
+    def test_3d_data_with_nans(self):
+        # test with data
+        ds = open_dataset(self.nc_ds).isel(location=1)
+        pr = ds.pr.sel(time=slice("2000"))  # kg m-2 s-1
+        prMM = pr
+        prMM *= 86400 * 1000
+        prMM.attrs["units"] = "mm/day"
```
aulemahal, line 105:
```suggestion
        prMM = convert_units_to(pr, "mm/day")
```
with `from xclim.core.units import convert_units_to` at the top of the file.

xclim/testing/tests/test_precip.py
```diff
@@ -92,6 +93,68 @@ def test_with_different_phases(self):
...
+            dist="gamma",
+            method="ML",
+        )
+        np.testing.assert_array_almost_equal(out1, out2, 3)
+
+        # preparing water_budget for SPEI test
+        with xr.set_options(keep_attrs=True):
+            tasmax = ds.tasmax
+            tas = tasmax - [2 + 1 * 0.5 for i in range(ds.tasmax.shape[0])]
+            tasmin = tasmax - [4 + 2 * 0.5 for i in range(ds.tasmax.shape[0])]
```
aulemahal, line 132:
```suggestion
            tas = tasmax - 3.5
            tasmin = tasmax - 7
```

xclim/testing/tests/test_precip.py
```diff
@@ -92,6 +93,68 @@ def test_with_different_phases(self):
...
+            wb = xci.water_budget(pr, None, tasmin, tasmax, tas)
+            wbMM = wb
+            wbMM *= 86400 * 1000
+            wbMM.attrs["units"] = "mm/day"
```
aulemahal, line 136:
```suggestion
            wbMM = convert_units_to(wb, "mm/day")
```

xclim/testing/tests/test_precip.py
```diff
@@ -92,6 +93,68 @@ def test_with_different_phases(self):
...
+            method="ML",
+        )
+        np.testing.assert_array_almost_equal(out1, out2, 3)
+
+        # preparing water_budget for SPEI test
+        with xr.set_options(keep_attrs=True):
+            tasmax = ds.tasmax
+            tas = tasmax - [2 + 1 * 0.5 for i in range(ds.tasmax.shape[0])]
+            tasmin = tasmax - [4 + 2 * 0.5 for i in range(ds.tasmax.shape[0])]
+            wb = xci.water_budget(pr, None, tasmin, tasmax, tas)
```
aulemahal, line 133:
```suggestion
            wb = atmos.water_budget(pr=pr, tasmin=tasmin, tasmax=tasmax)
```
Also, you don't need `tas` if you're using the "BR65" method, which is default.

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+      Daily precipitation.
+    pr_cal : xarray.DataArray
+      Daily precipitation used for calibration. Usually this is a temporal subset of `pr` over some reference period.
+    freq : str
+      Resampling frequency. A monthly or daily frequency is expected.
+    window : int
+      Averaging window length relative to the resampling frequency. For example, if `freq="MS"`, i.e. a monthly resampling, the window
+      is an integer number of months.
+    dist : str
+      Name of the univariate distribution, such as `beta`, `expon`, `genextreme`, `gamma`, `gumbel_r`, `lognorm`, `norm`
```
aulemahal, line 624:
```suggestion
    dist : {'gamma'}
      Name of the univariate distribution, only `gamma` is currently implemented
```
We shouldn't list methods we don't support I think.

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+      (see :py:mod:`scipy.stats`).
+    method : str
+      Name of the fitting method, such as `ML` (maximum likelihood), `APP` (approximate). The approximate method
+      simply uses the parameters output of `_fit_start`.
```
aulemahal, line 628:
```suggestion
      uses a deterministic function that doesn't involve any optimization.
```
Not sure how to phrase it, but I don't think most users will be familiar with `_fit_start`.

aulemahal (commented):
Première passe de cette version finale!
Demain je fais quelques tests.

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+        group = "time.dayofyear"
+    else:
+        _, base, _, _ = parse_offset(freq)
+        if base in ["Y", "A", "Q", "M"]:
+            group = "time.month"
+        else:
+            raise NotImplementedError(f"Resampling frequency `{freq}` not supported.")
+
+    # Resampling precipitations
+    if freq != "D":
```
tlogan2000, line 683: Question of someone who is maybe confused : Does it even make sense to calculate SPEI on ['Y', 'A', 'Q'] ?   Or should we only support daily and monthly frequency?

xclim/indices/_agro.py
```diff
@@ -589,6 +595,225 @@ def water_budget(
...
+    method: str = "ML",
+) -> xarray.DataArray:
+    r"""Standardized Precipitation Evapotranspiration Index (SPEI).
+
+    Precipitation minus potential evapotranspiration data (PET) fitted to a statistical distribution (dist), transformed to a cdf,  and inverted back to a gaussian normal pdf. The potential evapotranspiration is calculated with a given method (method).
+
+    Parameters
+    ----------
+    wb : xarray.DataArray
+      Daily water budget (pr - pet).
```
tlogan2000, line 768: Same comment as before (Daily and monthly I think)

xclim/indices/_agro.py
coxipi: I have seen some definitions of SPI as "annual-SPI-1" which is distinguished from "SPI-12". See for instance [1]. 
* "SPI-12" is resampled over month (`freq=MS`), with a window of 12 months; 
* "annual-SPI-1" is resampled over year (`freq=YS`) with a window of 1 year.

This definition doesn't seem to be used in many research articles though. The expression "annual SPI" is sometimes used as a synonym for "SPI-12" too.  The `monocongo` library doesn't support this definition of SPI either. 

All in all, including this possibility doesn't cost much, but if you feel it could bring confusion, we could decide to forbid this.

[1] Aryal, 2022: https://mdpi-res.com/d_attachment/earth/earth-03-00025/article_deploy/earth-03-00025-v2.pdf

xclim/indicators/atmos/_precip.py
coxipi: The SPI index is guaranteed to be centered on 0 by definition: The CDF (from whatever distribution is chosen) is inverted back to a normal distribution. So I don't think we need to ask for a constraint on the calibration data. Or did I misunderstand your point?

xclim/indices/_agro.py
aulemahal: If I understand correctly, SPI-12 on january is the same as annual-SPI-1 ? 
If so, 'Y', and 'Q' are just options to drop the overlapped members of the result, we could _not_ implement them and show in the doc how to retrieve them from the monthly version?
Ex to be clear:
```
spiM12 = spi(pr, pr_cal, freq='MS', window=12)
spiY1 = spiM12.sel(time=(spiM12.time.dt.month == 1))
```

coxipi: Right, what you describe is pretty much the same I agree. The only difference I see: The annual-SPI-1 is like having the Januaries of SPI-12 *with a window rolling forward in time* if `freq=YS` is used.

xclim/testing/tests/test_indices.py
coxipi: Sounds good, thanks.

xclim/testing/tests/test_indices.py
coxipi: Just saw it too oops

xclim/testing/tests/test_indices.py
coxipi: Earlier writing had an index dependent random number, but not anymore. Thanks

xclim/testing/tests/test_precip.py
```diff
@@ -7,6 +7,7 @@
 import pytest
 import xarray as xr
 
+import xclim.indices as xci
 from xclim import atmos, core, set_options
 from xclim.core.calendar import build_climatology_bounds, percentile_doy
 from xclim.core.utils import PercentileDataArray
```
coxipi, line 13:
```suggestion
from xclim.core.units import convert_units_to
from xclim.core.utils import PercentileDataArray
```

xclim/indicators/atmos/_precip.py
```diff
@@ -247,6 +250,28 @@ class HrPrecip(Hourly):
     parameters={"tas": {"kind": InputKind.VARIABLE}, "phase": "solid"},
 )
 
+standardized_precipitation_index = PrecipWithIndexing(
```
aulemahal, line 253: I think this is not the best class (`PrecipWithIndexing`). Because it 1 - expects only daily inputs, 2- adds the possibility of subsetting the data temporally _before_ computing the indicator, which would mess with this calculation.

I suggest creating a new class (at the top of the file):
```
class StandardizedIndexes(ResamplingIndicator):
    """Resampling but flexible inputs indicators."""
    src_freq = ["D", "M"]
    context = "hydro"
```
This needs the import of `ResamplingIndicator` from `xclim.core.indicator`.

xclim/indices/stats.py
coxipi: Ok, sounds good.

xclim/indicators/atmos/_precip.py
tlogan2000: ok. I might be confused about what we're discussing but looking at SPEI data calculated for the climatedata.ca project. I see that output spei values are centred more or less on 0 for 'calibration' period (historical grey data in this case) 
https://climatedata.ca/explore/variable/?coords=64.69793160441417,-103.88397216796876,8&delta=&geo-select=&var=spei_3m&var-group=precipitation&mora=jul&rcp=rcp85&decade=2070s&sector=

image

I still think having the calibration period in the metadata somewhere would be an important piece of info to retain
(From what I see in the signature the `history` attr will only tell us that there was an `xr.DataArray` used)

xclim/indices/_agro.py
```diff
@@ -589,6 +595,221 @@ def water_budget(
...
+    wb_cal : xarray.DataArray
+      Daily water budget used for calibration.
+    freq : str
+      Resampling frequency. A monthly or daily frequency is expected.
+    window : int
+      Averaging window length relative to the resampling frequency. For example, if `freq="MS"`, i.e. a monthly resampling, the window
+      is an integer number of months.
+    dist : str
+      Name of the univariate distribution, such as `beta`, `expon`, `genextreme`, `gamma`, `gumbel_r`, `lognorm`, `norm`
+      (see :py:mod:`scipy.stats`).
```
aulemahal, line 780:
```suggestion
    dist : {'gamma'}
      Name of the univariate distribution. Only "gamma" is currently implemented. (see :py:mod:`scipy.stats`).
```

xclim/indices/_agro.py
```diff
@@ -589,6 +595,221 @@ def water_budget(
...
+    dist_and_methods = {"gamma": ["ML", "APP"]}
+    if dist not in dist_and_methods.keys():
+        raise NotImplementedError(f"The distribution `{dist}` is not supported.")
+    elif method not in dist_and_methods[dist]:
+        raise NotImplementedError(
+            f"The method `{method}` is not supported for distribution `{dist}`."
+        )
+
+    # calibration period
+    cal_period = [pr_cal.time.values[0], pr_cal.time.values[-1]]
```
aulemahal, line 672:
```suggestion
    cal_period = pr_cal.time[[0, -1]].dt.strftime('%Y-%m-%dT%H:%M:%S').values.tolist()
```
List indexing to be fancy, but most importantly, this formats the dates into a string. NetCDFs do not support datetime objects in their attributes, and this string format is very common (ISO 8601, without the timezone info).

xclim/indicators/atmos/_precip.py
aulemahal: I checked and the french translations are working well but I think you forgot to change the english version as well.

aulemahal (approved):
I ran some tests and this seems to work as intended!

With dask, the daily version is quite slow and heavy. That's sad but the easiest solution I see would be to import sdba's `map_groups` monstruosity into the core xclim, which is not to be done here. It seems ok for the monthly cases, except for an issue with non-dimension coordinates, but that's to be fixed upstream : pydata/xarray#4205. (I tested with ERA5 and the `lon` and `lat` coords are triggering dask computations unexpectedly because they get chunked for no reason)

Nice!

tlogan2000 (approved):

PR #1114, opened 2022-06-20
Fix upstream installation recipe for PyPI-compliance

### LINKED ISSUE #1113: Move installation process of VCS-controlled libraries out of setup.py
For security reasons, VCS-based dynamic links to development version of dependencies is not allowed on PyPI (see discussion: https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package).

This means that the recipe for installation of `upstream` dependencies needs to move elsewhere, most likely into the `makefile` and/or `tox.ini`.

Will be resolved ASAP.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1113 

* Move upstream dependencies to its own requirements_upstream.txt
* Add requirements_upstream.txt to MANIFEST.in
* Set tox to use upstream dependencies
* Add a Makefile recipe for installing upstream dependencies and xclim
* Update history and documentation

No.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -13,7 +13,9 @@ on:
       - setup.py
       - tox.ini
       - xclim/__init__.py
-      - docs/**.rst
+      - docs/**.ipynb
+      - docs/*.py
+      - docs/*.rst
       - .github/**.yml
       - .pre-commit-config.yaml
 
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -6,10 +6,14 @@ on:
       - master
     paths-ignore:
       - HISTORY.rst
+      - Makefile
+      - MANIFEST.in
+      - requirements_upstream.txt
       - setup.cfg
       - setup.py
       - tox.ini
       - xclim/__init__.py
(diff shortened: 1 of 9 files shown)


.github/workflows/bump-version.yml
```diff
@@ -6,10 +6,14 @@ on:
...
     paths-ignore:
       - HISTORY.rst
+      - Makefile
+      - MANIFEST.in
+      - requirements_upstream.txt
       - setup.cfg
       - setup.py
       - tox.ini
       - xclim/__init__.py
+      - docs/**.rst
```
aulemahal, line 16: What does this fix?
So, if I understand correctly, this means we bump the version only when the code changes, but not for a doc PR.

aulemahal (approved):

.github/workflows/bump-version.yml
Zeitsperre: Precisely. This will trigger if a notebook changes, though. Should I exclude both?

aulemahal: Ok! Bah, I don't think I care that much about the difference, but indeed excluding the notebooks would fit the idea.

PR #1115, opened 2022-06-21
Wrap SBCK

### LINKED ISSUE #1109: Add experimental wrapping of SBCK to sdba
### Generic Issue

Yoan Robin's [SBCK](https://github.com/yrobink/SBCK) implements many bias adjustment methods, including most interesting multivariate ones. It follows a fit-predict syntax quite similar to xclim's sdba. It uses numpy, but only acts (AFAIU) on timeseries (by that I mean that is doesn't try to vectorize or loop the adjustment, there's no parallelism).

Which means, it could be quite easy (I usually underestimate things) to wrap it up into xclim. I'm talking of automatically generating sdba Adjustment classes from SBCK ones, and using `apply_ufunc` to loop on elements of the xarray inputs (like gridpoints. This would allow usage of dask with the same constraints as the other methods.

However, the license (GPL3) doesn't allow sub-licensing, which means we can't copy the code as-is into xclim. Let's just hope there's no bug...

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1109 

* All adjustment methods of [SBCK](https://github.com/yrobink/SBCK) are wrapped in `sdba.adjustment`, using `Adjust` classes (single step mode). This allow the use of dask and xarray, but is quite basic otherwise.

No.

### Other information:
SBCK depends on `Eigen` a C++ library. It is available on conda, but the CI uses pip, so I can't install it to run the new test. Or am I wrong?

### DIFF
--- a/xclim/sdba/__init__.py
+++ b/xclim/sdba/__init__.py
@@ -96,6 +96,7 @@ from .processing import (
 # TODO: ISIMIP ? Used for precip freq adjustment in biasCorrection.R
 # Hempel, S., Frieler, K., Warszawski, L., Schewe, J., & Piontek, F. (2013). A trend-preserving bias correction &ndash;
 # The ISI-MIP approach. Earth System Dynamics, 4(2), 219–236. https://doi.org/10.5194/esd-4-219-2013
+# If SBCK is installed, create adjustment classes wrapping SBCK's algorithms.
 if hasattr(adjustment, "_generate_SBCK_classes"):
     for cls in adjustment._generate_SBCK_classes():
         adjustment.__dict__[cls.__name__] = cls
--- a/xclim/sdba/__init__.py
+++ b/xclim/sdba/__init__.py
@@ -57,6 +57,19 @@ Most methods acting on grouped data also accept a `window` int argument to pad t
 Units of `window` are the sampling frequency of the main grouping dimension (usually `time`). For more complex grouping,
 one can pass an instance of :py:class:`xclim.sdba.base.Grouper` directly.
 
+Experimental wrap of SBCK
+=========================
+The `SBCK`_ python package implements various bias-adjustment methods, with an emphasis on multivariate
+methods and with a care for performance. If the package is correctly installed alongside xclim, the methods will be wrapped into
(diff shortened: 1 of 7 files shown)


Zeitsperre:
@aulemahal Not sure why, but the test is being skipped.
```
xclim/testing/tests/test_sdba/test_adjustment.py::test_SBCK[True-method0] SKIPPED [ 94%]
```
I'll leave you to do the configuring there.

aulemahal:
haha, sorry to tell you that, @Zeitsperre , but the test is skipped because SBCK isn't being installed ;)
The last commit should have fixed the issue, it seems related with tox's naming tricks.

Zeitsperre:
That's interesting. You're right. I feel like I've been able to throw keywords at the end of builds (e.g. `-lm3`) in order to trigger specific action in the main `testenv` block, but adding `-slow` broke that one. I wonder if it was a 'bug'/'feature'?

aulemahal:
My guess is that the env name needs to be a bit more specific. All other keyword-based variation we have happen in the generic unnamed env. May be `sbck-{,slow}` ? The doc has examples with choices (like `django{15,16}`) but not with an optional flag.

xclim/sdba/__init__.py
```diff
@@ -70,18 +83,19 @@
...
     stack_variables,
     unpack_moving_yearly_window,
     unstack_variables,
 )
-from .properties import *
 
 # TODO: ISIMIP ? Used for precip freq adjustment in biasCorrection.R
 # Hempel, S., Frieler, K., Warszawski, L., Schewe, J., & Piontek, F. (2013). A trend-preserving bias correction &ndash;
 # The ISI-MIP approach. Earth System Dynamics, 4(2), 219–236. https://doi.org/10.5194/esd-4-219-2013
+if hasattr(adjustment, "_generate_SBCK_classes"):
```
huard, line 99:
```suggestion
# If SBCK is installed, create adjustment classes wrapping SBCK's algorithms.
if hasattr(adjustment, "_generate_SBCK_classes"):
```

PR #1134, opened 2022-07-20
Optimized run length

### LINKED ISSUE #956: Bad performance of run length algorithms
As flagged in #912, an important weakness of xclim in term of performance is the huge number of tasks it generates when using dask. That issue discusses improvements that could be made at the indicator-level, but one major functionality of xclim that generates heavy load is the run-length algorithm(s),

As issue #916 raises, implementation choices have been made to ensure reasonable dask loads even though they were not scientifically correct. Specifically, we decided to resample and then count events, which incorrectly handles events spanning multiples periods but reduces the number of tasks a lot by running `rle` on smaller datasets.

Moreover, the `xclim.indices.run_length` module is somewhat messy because of the existence of 3 separate algorithms for most base functions : the `window=1` version,  the "ufunc" version (`run_length.rle_1d`) and the "normal" verson (`run_length.rle`).

The "ufunc" version requires a single chunk along  `time` and is said to be faster with smaller datasets, which was not verified extensively. The "normal" version is generic and works with any chunking or data size, but it generated an awful lot of dask tasks.

Before rewriting indices to their scientifically correct version, we must rewrite `rle` in order for it to work on large datasets.

EDIT: Just to be clear, the current `rle` does the following:
Takes in a boolean array and flags the runs of true values. The output is a float array where the first day of each run is the length of the run, all other days of the run are NaN and days outside of the runs are 0. An option exists to put the run-length at the end of the run instead. A new implementation could modify how the output is given, if that new output allows the computation of the same statistics.

coxipi:
### Pull Request Checklist:
    - This PR fixes #956

* Implements a new version of the internal run length algorithm. The run lengths are computed in a different way (using cumulative sums), and the use of memory in the computation of big datasets is more reasonable without relying on resampling.


tlogan2000:
This is some really interesting work by @coxipi  ... I thought it might want a good look by a few people so have tagged multiple reviewers.  @coxipi  not 100% sure if this is ready for review or not?  Note @aulemahal  and I will be on vacation soon... No real rush to merge so I suggest we take our time

coxipi:
> This is some really interesting work by @coxipi ... I thought it might want a good look by a few people so have tagged multiple reviewers. @coxipi not 100% sure if this is ready for review or not? Note @aulemahal and I will be on vacation soon... No real rush to merge so I suggest we take our time

The replacement of the `rle` function is done, yes. I was keeping the old version as `rle_old` for testing, so that's why there is a coverage error I think, I will simply remove `rle_old`. I was still comparing `rle` and `rle_1d`, but maybe this could be in a separate PR.

huard:
Et en terme de performance, qu'est-ce que ça dit?

coxipi:
> Et en terme de performance, qu'est-ce que ça dit?

C'est surtout l'utilisation de la mémoire qui est beaucoup plus efficace.  Le nouvel algorithme ne nécessite pas un seul chunk sur la dimension choisie, donc on évite le rechunking.  C'était l'étape la plus coûteuse.
Par exemple, voici un dataset de `espo-R`, slicé tel que time x lat x lon ~ 50 000 x 200 x 200. J'ai créé un `da` avec un treshold sur les données de précipitations, `pr > pr_T`. 

[memory200_old.txt](https://github.com/Ouranosinc/xclim/files/9168268/memory200_old.txt)
[memory200_new.txt](https://github.com/Ouranosinc/xclim/files/9168269/memory200_new.txt)

Avec plus de points spatiaux, par exemple 400 x 400, j'atteins la limite admise par Doris avec l'ancienne méthode et le calcul ne se complète, mais la nouvelle se comporte bien. 
mem_old_400
mem_new_400

Voici aussi des données avant de loader le output de `rle`. Il y a un gain important dans les tasks obtenues (`out.__dask_graph__().keys()`)
|                        | old     | new    |
| ---------------------- | ------- | ------ |
| memory [MiB]     | 59.6    | 26.1   |
| time   [s]             | 16.29   | 5.99   |
| tasks (raw number)     | 331 896 | 166784 |
| tasks gain (1-new/old) | N/A     | 49.75% |

huard:
Impressionnant !

xclim/indices/run_length.py
```diff
@@ -88,83 +88,30 @@ def rle(
...
+    # "first" case: Algorithm is applied on inverted array and output is inverted back
+    if index == "first":
+        da = da[{dim: slice(None, None, -1)}]
+
+    # Get cumulative sum for each series of 1, e.g. da == 100110111 -> cs_s == 100120123
+    cs = da.cumsum(dim=dim)
+    cs2 = cs.where(da == 0)
+    cs2[{dim: 0}] = 0
+    cs2 = cs2.ffill(dim=dim)
+    cs_s = cs - cs2
```
bzah, line 102: I feel that these few lines of code are really great !
They should probably be in a separate function because this is basically a cumsum with a reset mechanism when encountering a zero so I think it can be useful in a few other cases.
Also it would help for readability.

bzah (commented):
LGTM, nice work !

Maybe this rle and the `cumsum with reset on 0` could be useful in xarray directly.

xclim/indices/run_length.py
coxipi: Ok sounds good! I guess `core/utils` would be the right place?  Let me know if you think there's a better place.

bzah: Yes, it's just an intuition because I would expect other cases where a `cumsum_with_reset` would be useful.

I would make it a private function in run_length module, mainly for readability, until it needs to be reused elsewhere. But that's just my way of doing things.

tlogan2000 (approved):

Issue #744, opened 2021-06-11
Reduce duplicated indicator source code


Zeitsperre:
With the push to synchronize our code base with the `clix-meta` project, it may be worthwhile to start deprecating some indicators we offer currently. One example of duplicated indicators is:

- `xclim.indices.daily_temperature_range_variability` and 
  - https://github.com/Ouranosinc/xclim/blob/e425aac1ecf56f6624610657ad1946ea82e1802c/xclim/indices/_multivariate.py#L357 
- `xclim.indices.generic.interday_diurnal_temperature_range` 
  - https://github.com/Ouranosinc/xclim/blob/e425aac1ecf56f6624610657ad1946ea82e1802c/xclim/indices/generic.py#L602

This obviously entails breaking changes, so identifying all affected indicators and gradually deprecating them is the safest way forward.

Zeitsperre:
Here's another such example:

* https://github.com/Ouranosinc/xclim/blob/798c036a1022e8e1ad70cc98bba3a69144719aa4/xclim/indices/_multivariate.py#L564
* https://github.com/Ouranosinc/xclim/blob/798c036a1022e8e1ad70cc98bba3a69144719aa4/xclim/indices/generic.py#L657

I suppose in this case, we could offer both, seeing as the `_multrivariate` version has units checking, but we could also just port this units checking directly to `generic`, no?

aulemahal:
Indeed. This example exists because it's part of the clix-meta functions, and for this reason the `generic` one is not decorated with `declare_units`. We _could_ port the unit checking to  `generic` but that might break the automatic build of the `cf`module and it would be incoherent with the other clix-meta functions (which are not units-ready)...
I can accept this duplication, but we could also make the `_multivariate` version directly call the `generic` one, with added unit checking and a better signature.

Zeitsperre:
Interesting. This is certainly a bit confusing (I wasn't sure what the difference was at first glance). It might be good to add a `Warning` to these, or maybe better flesh out their methodologies? Perhaps this is a case for a `method` switch?

aulemahal:
Well, the one in `generic` is defined by clix-meta, so I'd be reluctant to change its name or signature. BUT, the docstring needs a fix, because it's currently wrong.

However, implementing this other version in the `_multivariate` one could be interesting, indeed using a `method` switch.

aulemahal:
[...]

PR #1157, opened 2022-08-22
Allow binary operations in generic indice call signatures

### LINKED ISSUE #389: Operators in indicator call signatures
Follow-up on a discussion point raised in https://github.com/Ouranosinc/xclim/pull/379#discussion_r386490720

When creating new indices based on the existing set, there are instances where it would be useful to be able to set a threshold with a different operator. What this means is that for indices that use `> | <` in some definitions and `>= | <=` in others, it would make for a more versatile system if when wrapping an existing base indicator, we were able to specify the type of operation we can use to trigger the conditions.

In other words, the proposal is to add the following kwargs to all indices based on conditional thresholds:
```
f = indices.tropical_nights(tas, thresh="20 degC", operator="gteq | gt | > | >=")
```
This would result in wrapped indices that look more like this:
```
R75p = partial(indices.days_over_precip_thresh, thresh="1 mm/day", operator=">")
```

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #389 and addresses some issues raised in #744 

* Adds the ability to explicitly set a binary operator for some existing indices.
* Adds the ability to constrain the allowed operators for a given indice (e.g. `xyz_days_above` → (">" | ">=") only).
* `op` is now more universally applied as the comparison operation for many generic indices.
* Adds tests for several previously untested generic indices (`count_level_crossings`, `count_occurrences`, `first_occurrence`, `last_occurrence`)
* Test `degree_days` with more operators.
* Removes `select_time` from `xclim.core.generic` (This has been deprecated for some time)
* Removed `daily_freezthaw_cycles` from `xclim.indices._multivariate.py` (This has been deprecated for some time).
* Various housekeeping fixes (Additions of `noqa` where needed, removal of useless imports).
* Revised the clix-meta parsing tools to support the new "op" operator handler, and added a smoke test to ensure is working.

Yes. 

 - Call signatures for generic indices have been modified to accept `op` in many cases, and `condition`/`conditional`/`operation` has been renamed to `op` for consistency.
 - Some generic indices call signatures have been reordered.
 - `select_time` has been removed from `generic.py` (was deprecated).
 - `daily_freezthaw_cycles` has been removed from `_multivariate.py` (was deprecated).

### DIFF
--- a/xclim/indices/generic.py
+++ b/xclim/indices/generic.py
@@ -7,6 +7,7 @@ Helper functions for common generic actions done in the computation of indices.
 """
 from __future__ import annotations
 
+import warnings
 from typing import Sequence
 
 import cftime
@@ -133,10 +134,17 @@ def get_op(op: str, constrain: Sequence[str] | None = None):
     constrain : sequence of str, optional
       A tuple of allowed operators.
     """
+    if op == "gteq":
+        warnings.warn(f"`{op}` is being renamed `ge` for compatibility.")
+        op = "ge"
+    if op == "lteq":
+        warnings.warn(f"`{op}` is being renamed `le` for compatibility.")
+        op = "le"
+
     if op in binary_ops.keys():
-        op = binary_ops[op]
+        binary_op = binary_ops[op]
     elif op in binary_ops.values():
-        pass
+        binary_op = op
     else:
         raise ValueError(f"Operation `{op}` not recognized.")
 
@@ -149,9 +157,9 @@ def get_op(op: str, constrain: Sequence[str] | None = None):
 
     if constrain:
         if op not in constraints:
-            raise ValueError("Operation `{op}` not permitted for indice.")
+            raise ValueError(f"Operation `{op}` not permitted for indice.")
 
-    return xr.core.ops.get_op(op)  # noqa
+    return xr.core.ops.get_op(binary_op)  # noqa
 
 
 def compare(
@@ -504,7 +512,7 @@ def statistics(data: xr.DataArray, reducer: str, freq: str) -> xr.DataArray:
     return out
 
 
-def compare_operation(
+def compare_arrays(
     left: xr.DataArray, op: str, right: xr.DataArray, constrain: Sequence[str] | None
 ) -> xr.DataArray:
     """Compare two variables using a binary logical operation.
@@ -524,9 +532,7 @@ def compare_operation(
     -------
     xr.DataArray
     """
-    op = get_op(op, constrain)
-
-    func = getattr(left, "_binary_op")(get_op(op))
+    func = getattr(left, "_binary_op")(get_op(op, constrain))
     return func(left, right)
 
 
--- a/xclim/indices/generic.py
+++ b/xclim/indices/generic.py
@@ -12,7 +12,7 @@ from typing import Sequence
 import cftime
 import numpy as np
 import xarray as xr
-from xarray.coding.cftime_offsets import _MONTH_ABBREVIATIONS
+from xarray.coding.cftime_offsets import _MONTH_ABBREVIATIONS  # noqa
 
 from xclim.core.calendar import (
     DayOfYearStr,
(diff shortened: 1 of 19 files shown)


xclim/indices/generic.py
```diff
@@ -496,8 +504,39 @@ def statistics(data: xr.DataArray, reducer: str, freq: str) -> xr.DataArray:
     return out
 
 
+def compare_operation(
+    left: xr.DataArray, op: str, right: xr.DataArray, constrain: Sequence[str] | None
+) -> xr.DataArray:
```
aulemahal, line 509: This function seems a good idea indeed. I only have a suggestion for its name.
```suggestion
def compare_arrays(
    left: xr.DataArray, op: str, right: xr.DataArray, constrain: Sequence[str] | None
) -> xr.DataArray:
```

bzah (approved):
LGTM!

bzah:
For the record, I'm implementing something quite similar in icc with a Threshold class that wrap the operator, the threshold value and the unit. You can then do stuff like:
`count_occurrences(data, ">= 25 degC")`
 or
`count_occurrences(data, ">= 75 doy_per")`.
Then metadata are generated with jinja.
If it makes sense, it should not be too hard to move that to xclim once it's finished.

xclim/indices/generic.py
```diff
@@ -147,42 +123,79 @@ def default_freq(**indexer) -> str:
...
-        pass
+        binary_op = op
     else:
         raise ValueError(f"Operation `{op}` not recognized.")
-    return xr.core.ops.get_op(op)  # noqa
+
+    constraints = list()
+    if isinstance(constrain, (list, tuple, set)):
+        constraints.extend([binary_ops[c] for c in constrain])
+        constraints.extend(constrain)
```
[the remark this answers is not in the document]
Zeitsperre: Constrain accepts the symbolic operators (e.g. `"<", "<="`), but if a user inputs the word equivalent notation (`"lt", "le"`), we don't want to raise an error as that usage is allowed.

xclim/indices/generic.py
```diff
@@ -291,15 +327,19 @@ def count_level_crossings(
...
     out = (lower & higher).resample(time=freq).sum()
     return to_agg_units(out, low_data, "count", dim="time")
 
 
 def count_occurrences(
-    data: xr.DataArray, threshold: str, condition: str, freq: str
+    data: xr.DataArray,
+    threshold: str,
+    freq: str,
+    op: str,
```
aulemahal, line 341: I think the change from "condition" to "op" has to be made in the automatic clix-meta parser in `xclim.core.utils.adapt_clix_meta_yaml` too.

xclim/indices/run_length.py
```diff
@@ -20,11 +20,9 @@
 from xclim.core.options import OPTIONS, RUN_LENGTH_UFUNC
 from xclim.core.utils import DateStr, DayOfYearStr, uses_dask
 
+# Arrays with less than this number of data points per slice will trigger
+# the use of the ufunc version of run lengths algorithms.
```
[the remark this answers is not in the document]
Zeitsperre: Oh! I had no idea we could introduce descriptions for constant variables! Will revert.

xclim/indices/generic.py
Zeitsperre: I'm trying to generate a YAML file via clix-meta so that I can test these changes, but it's not very clear to me how that's done. Can you explain how, or pass me a YAML file example?

aulemahal: You need to install their python package which is in the "python" folder of the directory: https://github.com/clix-meta/clix-meta/tree/main/python
and then run `ci-meta-exporter path/to/the/excel/file.xls`.

Zeitsperre: Fixed. Test for this is now added.

aulemahal (approved):
Looks good!
I think the history was messed up by some merges, but other than that, this is good to go!

PR #1159, opened 2022-08-23
Add McArthur Forest Fire Danger Index

### LINKED ISSUE #1152: Add McArthur Forest Fire Danger Index?
### Addressing a Problem?

Thanks for this terrific package!

Is there interest in adding the McArthur Forest Fire Danger Index (FFDI) to `xclim`?

The FFDI is a fire weather index that is heavily used in operation and research, particularly in Australia, e.g.

- http://www.bom.gov.au/vic/forecasts/fire-map.shtml
- https://www.nature.com/articles/s41467-021-27225-4

The calculation of the FFDI requires the calculation of a Drought Factor, commonly the Griffiths Drought Factor (DF) using the Keech-Byram Drought Index (KBDI), so functions for calculating these (at least) would also need to be included.

### Potential Solution

If there’s interest in including the FFDI, DF and KBDI in `xclim`, I’d be happy to have a stab at implementation. I have some code already, though it’s currently just some functions in a notebook [here](https://notebooksharing.space/view/a38d8bc9093ab0b612f8aaab8767a48bad686944ff0774cc6af6af49da6f2e8c#displayOptions=). My code could probably benefit from some refactoring when slotting into xclim (efficient implementation of DF, in particular, is a little sticky), but the gist of it is there.

### Additional context

_No response_

### Contribution

### Code of Conduct

dougiesquire:
### Pull Request Checklist:
    - This PR fixes #1152 

Introducing new indices related to and including the McArthur Forest Fire Danger Index:
* The Keetch-Byram Drought Index (commonly KBDI)
* The Griffiths Drought Factor (commonly DF)
* The McArthur Forest Fire Danger Index (commonly FFDI)

A new `xclim.indices.fire` module has been introduced for housing fire related indices. The existing Canadian fire indices have been moved into here along with the new FFDI indices. As part of this, the exisiting function `fire_weather_indexes` has been renamed to `cffwis_indices`. Depreciation warnings indicating that `fire_weather_indexes` will be removed in future releases have been added.


Zeitsperre:
@dougiesquire Thanks so much for this PR! If you'd like to give me write access to your branch, I can add a few fixes to better integrate the documentation into the streamlined API section.

The CI is failing because Python3.8/3.9 need to be using Python3.10's annotation system. The following should fix it:
```
$ pre-commit install
$ pre-commit run --all
```

dougiesquire:
Thanks @zeitsperre. It looks like you have access to my branch already.

I still have to add tests, indicators and get the units stuff working. I’ll try to get to these in the next few days.

I have a question about where the McArthur FFDI and related indices should sit in `xclim`. You’ll see that, for now, I’ve added them to a new submodule `xclim.indices.ffdi`. However, the FFDI is a fire weather index, so it seems to make more sense for these indices to live in the existing submodule `xclim.indices.fwi`.

Currently, `xclim.indices.fwi` is very much framed around the Canadian Forest Fire Weather Index System (CFFWIS). Unfortunately some of the names of the components of the CFFWIS are a little ambiguous when talking about multiple fire rating systems. For example, the CFFWIS fire weather index is simply called “Fire Weather Index”. Currently, `xclim.indices.fwi` includes the following public functions:
- `fire_weather_indexes` : This name seems a little ambiguous. This function calculates the 6 component indices of the CFFWIS, one of which is called the Fire Weather Index. There are other fire weather indices, e.g. the McArthur Forest Fire Danger Index implemented in this PR. (Also it's not clear to me why "indexes" is used here but "indices" is used elsewhere).
- `drought_code` : This is probably unambiguous. Drought indicators used by other systems seem to have distinct names, e.g. the Keetch-Byram Drought Index implemented in this PR.
- `overwintering_drought_code` : See previous point
- `fire_season` : This is probably unambiguous and other methods for defining the fire season could always be added as options to this function.

Some suggested options for implementation of this PR:
1. If we’re okay with **breaking changes** we could rename some of the functions in `xclim.indices.fwi` so that they are less ambiguous. For example, `xclim.indices.fwi` could expose the following public functions:
    - `cffwis_indices` : renamed from `fire_weather_indexes`
    - `cffwis_drought_code` : renamed from `drought_code` (maybe not necessary)
    - `cffwis_overwintering_drought_code` : renamed from `overwintering_drought_code` (maybe not necessary)
    - `fire_season`
    - `mcarthur_forest_fire_danger_index` : added in this PR
    - `griffiths_drought_factor` : added in this PR
    - `keetch_byram_drought_index` : added in this PR
3. Leave as is, with McArthur FFDI indices in `xclim.indices.ffdi` and CFFWIS indices in `xclim.indices.fwi`. This has the potential to be confusing to users who go looking for all fire weather indices in `xclim.indices.fwi`.
4. Simply add the McArthur FFDI to `xclim.indices.fwi` and rely on the documentation to clarify any ambiguity.
5. Other suggestions?

FWIW, point 1. is my preferred option since I think it's most user-friendly, but a breaking change may be unacceptable.

Zeitsperre:
Hi @dougiesquire and thank you again for your contribution! 

I'm in agreement that your first option would be better if we ever want to support more approaches/methodologies in the future. This layout proposal would be easier to document and manage as well. There are likely going to be a few breaking changes in the upcoming release (see #1157), so this might be a good opportunity to introduce changes to the FWI module at the same time. These changes will require some adjustments in a few projects that serve fire-based indices, but this is reasonable, IMO.

@aulemahal, what are your thoughts on this?

aulemahal:
Thanks for this nice addition!

I think I would prefer option 1, but without renaming the drought code indices. I feel "drought code", even if not totally unambiguous, is unique enough. The documentation page, however, should be clearly divided into two sections, one for the Canadian system and the other for this Australian one.

I'll divert the question anyway to @tlogan2000, who may be a better judge on how this will impact our current users.

tlogan2000:
Hello @dougiesquire thanks for the addition! This is really cool.  
 
For the code structure I would like to avoid a breaking change as much as possible (not a deal breaker) but I think I would initially see if we can't go with @aulemahal's suggestion of a combined `fwi` but with very clear separation / explanation in the docs?   In this case nothing would stop us from eventually moving to the full option 1 if we decide to go that way down the road.   I agree that separate `fwi` and `ffdi` is not the best solution as fire related algorithms should be grouped together in my mind.

aulemahal:
I suggest we do some magic for the first few versions after this change: `fire_weather_indexes` will still work but throw a deprecation warning that it has been renamed `cffwis_indices`. And similarly if we decide to rename other functions.

tlogan2000:
> I suggest we do some magic for the first few versions after this change: `fire_weather_indexes` will still work but throw a deprecation warning that it has been renamed `cffwis_indices`. And similarly if we decide to rename other functions.

Ok yes even better!

Zeitsperre:
I can add some changes based on this discussion if we'd like. Should have some time this afternoon!

Zeitsperre:
I've added some changes. See: https://github.com/dougiesquire/xclim/pull/1

dougiesquire:
As expected, there're a few minor issues from the merge causing some pre-commit checks to fail. To avoid treading on toes, I'll leave them for you to fix, @Zeitsperre, when you get a chance - thanks!

dougiesquire:
I've added a few basic tests and added indicators for all the indices. Jotting down a few things on my mind while I remember:

- There is currently no direct test of the indicators. The documentation recommends writing these using real data, but I'm not aware of any official FFDI data we could use. Recommendations for what to do here would be welcome.
- I'm not totally confident about my implementation of the indicators and units stuff. I mostly just copied the existing fire indicators. It would be great if someone could cast a close eye on this stuff during review.
- @Zeitsperre's restructure still needs to be added to HISTORY.rst. @Zeitsperre, would you mind doing this? I think you're most across the changes you made.
- I still need to check how all the documentation comes together following @Zeitsperre's restructure.

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+        if dfw > 10.0:
+            dfw = 10.0
+
+        df[d] = dfw
+
+
+# SECTION 2 - Public methods and indices
+
+
+def _keetch_byram_drought_index_pass(pr, tasmax, pr_annual, kbdi0):
```
aulemahal, line 200: I would declare this function inside the main one (`keetch_byram_drought_index`) to "remove" it from the module and avoid any confusion with the vectorized one. Same thing with the drought factor one.

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+    return _griffiths_drought_factor(pr, smd, lim)
+
+
+@declare_units(
+    pr="[precipitation]",
+    smd="[precipitation]",
+)
+def griffiths_drought_factor(
+    pr: xr.DataArray,
+    smd: xr.DataArray,
```
aulemahal, line 286: Please add an entry to `xclim/data/variables.yml` for this new variable.

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+        input_core_dims=[["time"], ["time"]],
+        output_core_dims=[["time"]],
+        dask="parallelized",
+        output_dtypes=[pr.dtype],
+    )
+    df.attrs["units"] = ""
+
+    # First non-zero entry is at the 19th time point since df is calculated
+    # from a 20-day rolling window
+    return df.isel(time=slice(19, None))
```
aulemahal, line 344: Would returning NaN for the 19 first days make sense? I would prefer that because xclim doesn't usually _crop_ the outputs. One could normally expect to receive the same data structure as the input for indicator that do not perform resampling.

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+):
+    """Calculate the McArthur forest fire danger index (FFDI) Mark 5.
+
+    This method calculates the FFDI using the formula in :cite:t:`ffdi-noble_1980`.
+
+    Parameters
+    ----------
+    D : xr.DataArray
+        The drought factor to use in the FFDI calculation. Often the daily Griffiths
+        drought factor (see `griffiths_drought_factor`).
```
aulemahal, line 367:
```suggestion
        drought factor (see :py:func:`griffiths_drought_factor`).
```

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+    H="[]",
+    V="[speed]",
+)
+def mcarthur_forest_fire_danger_index(
+    D: xr.DataArray,
+    T: xr.DataArray,
+    H: xr.DataArray,
+    V: xr.DataArray,
+):
+    """Calculate the McArthur forest fire danger index (FFDI) Mark 5.
```
aulemahal, line 359:
```suggestion
    """McArthur forest fire danger index (FFDI) Mark 5.
```

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+@declare_units(
+    pr="[precipitation]",
+    smd="[precipitation]",
+)
+def griffiths_drought_factor(
+    pr: xr.DataArray,
+    smd: xr.DataArray,
+    limiting_func: str = "xlim",
+) -> xr.DataArray:
+    """Calculate the Griffiths drought factor based on the soil moisture deficit.
```
aulemahal, line 289:
```suggestion
    """Griffiths drought factor based on the soil moisture deficit.
```

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,393 @@
...
+    pr_annual="[precipitation]",
+    kbdi0="[precipitation]",
+)
+def keetch_byram_drought_index(
+    pr: xr.DataArray,
+    tasmax: xr.DataArray,
+    pr_annual: xr.DataArray,
+    kbdi0: xr.DataArray | None = None,
+) -> xr.DataArray:
+    """Calculate the Keetch-Byram drought index (KBDI) for soil moisture deficit.
```
aulemahal, line 219:
```suggestion
    """Keetch-Byram drought index (KBDI) for soil moisture deficit.
```

aulemahal (commented):
First pass of comments.

For the "real data", one alternative would be to compute the indicators on one of the xclim test datasets using a reference implementation. (Matlab? R? idl?)
Test datasets are here : https://github.com/Ouranosinc/xclim-testdata/ and can be opened in the code with  `xclim.testing.open_dataset`. I suggest using `ERA5/daily_surface_cancities_1990-1993.nc` if possible.

Then, the "expected" values can either be saved into a dataset we add to `xclim-testdata` or simply copy-pasted in the tests (with a comment on their provenance).

xclim/indices/fire/_ffdi.py
aulemahal: I note that "soil moisture deficit" doesn't seem to exist in the CMIP6 models and thus there is no official variable name to follow. From my quick search, `smd` seems the most commonly used, even though CMIP6 uses `mrso` for soil moisture.

Zeitsperre:
> * I still need to check how all the documentation comes together following @Zeitsperre's restructure.

I don't think I can share access to the ReadTheDocs builds easily since you aren't a maintainer, but if you run the following from the top-level:
```
$ make docs
```
You should be able to build them yourself, provided you're running in the python environment you're developing from. Usually takes a few minutes to run and should open a browser window automatically.

xclim/indices/fire/_ffdi.py
dougiesquire: Added. I went through the same process as you looking for a CMIP name and found nothing. In `xlim/data/variables.yml` I've given `smd` canonical units of mm since that seemed like the most obvious choice.

dougiesquire:
> For the "real data", one alternative would be to compute the indicators on one of the xclim test datasets using a reference implementation. (Matlab? R? idl?)

Thanks @aulemahal, I'll have a dig around for a reference implementation. I'm aware of a couple of existing implementations but a few have bugs and the others calculate the drought factor using an old approach.

xclim/data/variables.yml
```diff
@@ -113,6 +113,11 @@ variables:
     standard_name: sea_ice_area_fraction
     data_flags:
       - percentage_values_outside_of_bounds:
+  smd:
+    canonical_units: mm
```
aulemahal, line 117: In the function, you put "[precipitation]" as the dimensions of `smd` and then you convert it to `mm/d`. Shouldn't then that be the canonical units?

Zeitsperre:
@dougiesquire

There was a lot of movement today, but the code base should now be up-to-date. Our next release is slated for Friday, so we're trying to close a bunch of issues. 

The PR is looking good for far. If you think this would be ready for final review for sometime during Thursday morning (America time), we can likely get this merged for v 0.38. Let us know!

xclim/data/variables.yml
dougiesquire: Would it be appropriate to add something like:
```
units.define("[soil_moisture_deficit] = [length] / [time]")
```
to `xclim/core/units.py` and then use that to specify the dimensions? Or is it fine to use "[precipitation]" here given that `smd` and `pr` should always have the same units?

dougiesquire:
> The PR is looking good for far. If you think this would be ready for final review for sometime during Thursday morning (America time), we can likely get this merged for v 0.38. Let us know!

The only thing holding this up is finding a good way to test the indicators against real data. 

I thought maybe the [CEMS GEFF dataset](https://cds.climate.copernicus.eu/cdsapp#!/dataset/cems-fire-historical?tab=overview) could be useful, but it's been difficult to interpret as there isn't great documentation. Looking at the source code, they use old formulations of the KBDI and Drought Factor so we can't expect our results to match theirs.

I've also looked for other implementations of the FFDI indices to test against. I've found two R packages ([ClimInd](https://www.rdocumentation.org/packages/ClimInd/versions/0.1-3) and [ecbtools](https://rdrr.io/github/ozjimbob/ecbtools/)), but both have a number of differences from the "modern" way to calculate the FFDI (i.e. following [Finkele et al. 2006](https://webarchive.nla.gov.au/awa/20060903105143/http://www.bom.gov.au/bmrc/pubs/researchreports/RR119.pdf)). I also found documentation for a Python package called [ffdi-kbdi](https://kbdi-ffdi.readthedocs.io/en/latest/index.html), but the source code doesn't appear to exist anymore, so it's not installable.

I could add a test that simply calculates the FFDI indicators using the `ERA5/daily_surface_cancities_1990-1993.nc`. It wouldn’t check the values against anything, but it would at least check that the indicators run on real data (we could also check other basic things like the shape of the output etc). Is this a sufficient test? If yes, then I should have time to get this in and ready for review before Thursday morn.

Zeitsperre:
> I could add a test that simply calculates the FFDI indicators using the `ERA5/daily_surface_cancities_1990-1993.nc`. It wouldn’t check the values against anything, but it would at least check that the indicators run on real data (we could also check other basic things like the shape of the output etc). Is this a sufficient test? If yes, then I should have time to get this in and ready for review before Thursday morn.

It looks as though there really isn't much available for us to compare our results to as yet, so I think having a test that runs the FFDI indices based off of an ERA5 file would be sufficient. So long as expected results are reasonable on inspection, we can then test for expected values at a given coordinate, output shape, etc.

If you'd like to add a small (like a handful of pixels across, maybe a few seasons long time series with the variables needed to run FFDI; no more than 500 KB) of ERA5/Land to the [xclim testing data](https://github.com/Ouranosinc/xclim-testdata) for a given region in Australia, feel free to submit a PR.

From a fork of `xclim-testdata` containing your addition (under `ERA5/daily_australia_2020.nc`, let's say). You could call it within xclim with the following:

```
from xclim.testing.utils import open_dataset

ds = open_dataset("ERA5/daily_australia_2020.nc",
             github_url="https://github.com/dougiesquire/xclim-testdata",
             branch="my_branch")
```
We can handle the data integration from there.

Hope this helps!

dougiesquire:
Thanks @Zeitsperre. I think this is now ready for review. I ended up just testing the indicators on the `ERA5/daily_surface_cancities_1990-1993.nc` data as I don't think having an Australia-specific region would add anything. I'll have a little time on my Friday to resolve reviews before your Friday morning - hopefully we can get this into v0.38!

xclim/indices/fire/_ffdi.py
```diff
@@ -0,0 +1,402 @@
...
+    T="[temperature]",
+    H="[]",
+    V="[speed]",
+)
+def mcarthur_forest_fire_danger_index(
+    D: xr.DataArray,
+    T: xr.DataArray,
+    H: xr.DataArray,
+    V: xr.DataArray,
+):  # noqa: D403
```
aulemahal, line 367: Would it be possible to use more common variable names? I understand the definitions used here are a bit different, but they seem close enough?
Like: D->  drought_factor, or df
T -> tasmax
H -> hurs or hursmax
V -> sfcWind
?

aulemahal (changes requested):
This looks quite good! I have only a small request.

Also, the  file  `docs/xclim.indices.fire.rst` shouldn't be present in the repo. It is automatically generated by sphinx, but as it is new, it is absent from the `.gitignore`. Could you remove it from the repo and add it to the gitignore?

Zeitsperre:
@dougiesquire

Just an FYI, but there are still a few more adjustments we want to make before v0.38, so we're going to push the release to Tuesday, Sept 6. Feel free to make tweaks/changes over the weekend!

dougiesquire:
I'm happy for this to be merged if everyone else is happy

Zeitsperre:
Congrats @dougiesquire !

PR #1161, opened 2022-08-25
Resampling after run length

### LINKED ISSUE #505: Consecutive threshold spanning across intervals

coxipi:
### Pull Request Checklist:
    - This PR fixes #505, #916, #917

An option `resample_before_rl` allows choosing between two behaviours for certain indices:
1. `resample_before_rl == False`: Apply `run_length` algorithms first, then resample the results in given periods and obtain various statistics
2. `resample_before_rl == True`:  Separate the data in given periods, then apply `run_length` algorithms and find statistics.

Parameters `reducer` and `window` in function `rle_statistics` are now positional


xclim/testing/tests/test_run_length.py
```diff
@@ -200,6 +267,7 @@ def test_real_simple(self):
         assert 10 == i
 
     def test_real_data(self):
+        # !!! No test here?  !!!
```
Zeitsperre, line 270:
```suggestion
        # FIXME: No test here?!
```
Good catch!

xclim/indices/run_length.py
```diff
@@ -58,7 +58,7 @@ def use_ufunc(
         If ufunc_1dim is "auto", returns True if the array is on dask or too large.
         Otherwise, returns ufunc_1dim.
     """
-    if ufunc_1dim is False and freq is not None:
+    if ufunc_1dim is True and freq is not None:
```
Zeitsperre, line 61: Just a heads-up here, but evaluating against `True` here will result in `False` if `ufunc_1dim` is a valid string. Otherwise, evaluating `ufunc_1dim` for "truthiness" (ie: ` if ufunc_1dim and freq is not None`) will accept `True` or a valid string.

coxipi: Thanks, that does what I want then. I want a warning only if `ufunc_1dim` was asked specifically, i.e. set to `True` as opposed to `"auto"` or `"from_context"`. I want the user to be notified this is not a valid choice.

coxipi:
@tlogan2000 @RondeauG: I was discussing with @aulemahal about the implementation we would for the new parameter in this PR: `resample_before_rl` (see above "What kind of change does this PR introduce?" for more info on this parameter). It was implemented in a similar fashion to `run_length_ufunc`, but it is also used as an argument in indices (not only a global parameter affecting computation). But this results in tests failures as the "kind" `str | bool` is not recognized. What do you think should be the implementation:
A) Use `resample_before_rl` as an argument in indices, and add a new kind
B) Do not use  `resample_before_rl` as an argument in indices

I did the comparison with `run_length_ufunc`, but this parameter has consequences for what output is obtained (and not only internal computations like `run_length_ufunc`). This might incline us to write things as in A). However, it might be more simple to do B), and in the future maybe we simply don't want this as an option. Any thoughts?

RondeauG:
I'm not really too familiar with the internal workings of xclim, so I can't offer any kind of precise answer. What I can say, is that there might be cases where we want a mix of `resample_before_rl = True/False`, so a purely global parameter might not be the most appropriate? That would point to A being the better option?

xclim/core/options.py
```diff
@@ -163,21 +163,22 @@ class set_options:
...
+      algorithm. Default: ``False``.
+    - ``sdba_encode_cf``:
+      Whether to encode cf coordinates in the ``map_blocks`` optimization that most adjustment methods are based on.
+      This should have no impact on the results, but should run much faster in the graph creation phase.
+    - ``keep_attrs``:
+      Controls attributes handling in indicators. If True, attributes from all inputs are merged
+      using the `drop_conflicts` strategy and then updated with xclim-provided attributes.
+      If False, attributes from the inputs are ignored. If "xarray", xclim will use xarray's `keep_attrs` option.
+      Note that xarray's "default" is equivalent to False.
+      Default: ``"xarray"``.
```
coxipi, line 181: I know this change should be removed, trying to find how within github

xclim/indices/_agro.py
```diff
@@ -765,7 +765,7 @@ def needs_rechunking(da):
                     stacklevel=2,
                 )
                 return True
-            return False
+                return False
```
coxipi, line 768: Indendation error to be corrected

xclim/testing/tests/test_formatting.py
```diff
@@ -32,11 +32,10 @@ def test_indicator_docstring():
...
     assert doc[12] == "  Default : `ds.tasmin`. [Required units : [temperature]]"
-    assert doc[38] == (
-        "  Number of heat wave events (Tmin > {thresh_tasmin} and Tmax > {thresh_tasmax} for >= {window} days) "
-        "(heat_wave_events), with additional attributes: **description**: {freq} number of heat wave events "
-        "over a given period. An event occurs when the minimum and maximum daily temperature both exceeds specific "
-        "thresholds : (Tmin > {thresh_tasmin} and Tmax > {thresh_tasmax}) over a minimum number of days ({window})."
+    assert doc[2] == (
+        "Number of heat waves over a given period. A heat wave is defined as an event "
+        "where the minimum and maximum daily temperature both exceed specific thresholds "
+        "over a minimum number of days."
```
coxipi, line 38: Needs to be rechecked. Currenty, "doc[38] == Returns"

xclim/indices/run_length.py
```diff
@@ -56,13 +58,63 @@ def use_ufunc(
...
+      Keyword arguments needed in `compute`
+
+    Returns
+    -------
+    xr.DataArray
+      Output of compute resampled according to frequency {freq}.
+    """
+    if resample_before_rl and freq is not None:
+        out = da.resample({dim: freq}).map(compute, args=args, dim=dim, **kwargs)
+    else:
```
aulemahal, line 115:
```suggestion
    if resample_before_rl:
        out = da.resample({dim: freq}).map(compute, args=args, dim=dim, **kwargs)
    else:
```
If someone forgets `freq` but does give `resample_before_rl=True,` I believe it should fail. Here, the error will come from xarray's resample, we could also raise an explicit error by testing `if freq is None` _within_ the `if`.  The current code silently switches to resampling after if one forgets `freq`.

xclim/indices/_multivariate.py
```diff
@@ -733,7 +770,14 @@ def heat_wave_max_length(
...
-    max_l = cond.resample(time=freq).map(rl.longest_run, dim="time")
+    max_l = rl.resample_and_rl(
+        cond,
+        resample_before_rl,
+        rl.longest_run,
+        freq=freq,
+    )
+    # TODO: Should this step be integrated as an option in longest_run?
+
     out = max_l.where(max_l >= window, 0)
```
aulemahal, line 781:
```suggestion
    out = rl.resample_and_rl(
        cond,
        resample_before_rl,
        rl.rle_statistics,
        reducer='max',
        window=window,
        freq=freq,
    )
```
Not sure why the code was done this way. `rle.longest_run` is only a specific call to  `rle_statistics`which already exposes the needed functionality. Your comment made me notice it!

aulemahal (approved):
This looks good!

I believe you forgot to change `xclim.indices.generic.spell_length`, which is making your tests fail.

xclim/indices/run_length.py
coxipi: I think I would agree if `resample_before_rl = True` was not the default option. I think the user should be able to specify  `freq is None` without having to care about `resample_before_rl`, because with `freq is None`,  resampling does not occur.

coxipi: Should we add an option `None` for resample_before_rl` and check that the user makes a coherent choice?

|`resample_before_rl` | `freq` | Result? |
|--|--|--|
| `None` | `None` | Success |
| `None` | "YS", "MS",... | Fail |
| `True`, `False` | `None` | Fail |
| `True`, `False` | "YS", "MS",... | Success |

aulemahal: I didn't think of the case where one would want NOT to resample. However, wouldn't it be better for such a user to call the  `compute` function directly instead? I'm not sure I see a case where someone would use `resample_before_rl` in a way that needs to be "generic" to both resampling and non-resampling.

If such a case does exist, your idea is the best way to go!

coxipi: True, I'm not sure if that functionality would be used. 

Say you have a dataset over 30 years. Using `heat_wave_max_length(da,...,freq=None,...)` would allow you to find the maximal heat wave of your whole dataset directly. You could also use `heat_wave_max_length(da,...,freq="YS",..., resample_before_rl=False)` and  simply find the maximum in the set of 30 `heat_wave_max_length` sub-output for each year

coxipi: On a side note, I realized that the signature in the indices is:
``freq: str = "YS"` 
as opposed to letting  `None`  as a possibility
``freq: str | None = "YS"` .

I will wait to see what we decide to do, but if we keep `freq=None`, I'll need to fix this.

tlogan2000:
> I'm not really too familiar with the internal workings of xclim, so I can't offer any kind of precise answer. What I can say, is that there might be cases where we want a mix of `resample_before_rl = True/False`, so a purely global parameter might not be the most appropriate? That would point to A being the better option?

I agree that I think we should allow for both

aulemahal (approved):
I like the added example!

PR #1163, opened 2022-08-29
Pylint adjustments

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Addresses the pylint stack overflow crashes that have been happening since the release of pylint v2.15
* Moves the `pylint` configuration to its own `.pylintrc` based on the examples offered at pylint
* Updates makefile to use `.pylintrc`
* Removed pylint from tox (issues with virtualenv and pylint)
* Addresses some easily-addressed pylint warnings

No.

### Other information:

I've removed the Codefactor.io check, as the configuration for it was invalid this entire time (hence why so few errors were ever raised). `pylint` is much more synchronized with our code style, and we should aim to address/mark-for-exception the issues that we deem to be errors. Once that work is done, we can remove the `--exit-zero` flag and allow it to fail builds.


xclim/core/indicator.py
```diff
@@ -1684,6 +1686,7 @@ def _merge_attrs(dbase, dextra, attr, sep):
                 data, identifier=identifier, module=module_name
             )
 
+        # FIXME: Handle the potential exceptions explicitly
```
aulemahal, line 1689: Well I know this is what pylint whines about, but the point of this construct is that we actually don't know what could cause the function to fail. BUT, we don't want to stop the execution of the program if it does.
I don't think the construction makes any sense if we narrow it down to a list of exceptions.

xclim/sdba/adjustment.py
```diff
@@ -134,10 +134,10 @@ def _harmonize_units(cls, *inputs, target: str | None = None):
         return (convert_units_to(inda, target) for inda in inputs), target
 
     @classmethod
-    def _train(ref, hist, **kwargs):
+    def _train(cls, hist, **kwargs):
```
aulemahal, line 137:
```suggestion
    def _train(cls, ref, hist, **kwargs):
```
My bad for this signature. The goal was to show what it should look like.

xclim/sdba/adjustment.py
```diff
@@ -134,10 +134,10 @@ def _harmonize_units(cls, *inputs, target: str | None = None):
...
         raise NotImplementedError()
 
-    def _adjust(sim, *args, **kwargs):
+    def _adjust(self, *args, **kwargs):
```
aulemahal, line 140:
```suggestion
    def _adjust(self, sim, *args, **kwargs):
```

aulemahal (approved):
Good with me!

xclim/core/indicator.py
Zeitsperre: Understood, I don't think there's any issue with marking "exceptions" to the pylint rules, so long as we're explicit about them.

PR #1183, opened 2022-09-12
Log-logistic distribution for SPEI

### LINKED ISSUE #1141: Log-logistic Distribution for SPEI
### Addressing a Problem?

According to the [docs](https://xclim.readthedocs.io/en/latest/xclim.indices.html?highlight=spei#xclim.indices._agro.standardized_precipitation_evapotranspiration_index) `Gamma` is the only supported distribution for the SP(E)I calculation. AFAIK the Log-logistic distribution can be used for the SPEI as well, which is not supported by xclim yet.

### Potential Solution

According to the [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.fisk.html) there is a log-logistic (aka fisk) dist available in `scipy.stats.fisk`. From a quick view at the code, it seems that only the name of the distribution needs to be added in some lines. If it is that easy to implement I'd try to do it in a PR.

### Additional context

_No response_

### Contribution

### Code of Conduct

coxipi:
### Pull Request Checklist:
    - This PR fixes #1141 

Adds the possibility to use a log-logistic distribution with SPI and SPEI indices

Minor: The definition of SPEI function is slightly changed to better reproduce results of the `monocongo/climate_indices` library (reference for the gamma distribution)


Zeitsperre:
@coxipi I think 1aa9ad330238370e562c48ab81f56814592f5a67 removed a lot of tests and fixes that were added to the code base in a previous PR. Things that checked for operator behaviour and implicit lat have been removed. Can you rollback those changes?

coxipi:
> @coxipi I think [1aa9ad3](https://github.com/Ouranosinc/xclim/commit/1aa9ad330238370e562c48ab81f56814592f5a67) removed a lot of tests and fixes that were added to the code base in a previous PR. Things that checked for operator behaviour and implicit lat have been removed. Can you rollback those changes?

Thanks for catching this!

coxipi:
One thing that remains: Computing SPI or SPEI using the ML method from Scipy leads to important variations when using tox in the github tests (errors of order 0.01). So I had to increase the tolerance on SPI/ SPEI tests up to 0.02. That's not specific to the `log-logistic` distribution, so maybe opening a separate issue would be better? I think that this is more an internal issue and giving access to log-logistic now could be good.

xclim/indices/stats.py
```diff
@@ -485,16 +486,29 @@ def _fit_start(x, dist, **fitkwargs) -> tuple[tuple, dict]:
...
+        x_pos = x[x > 0]
+        m = x_pos.mean()
+        v = x_pos.var()
+        # pdf =  (beta/alpha) (x/alpha)^{beta-1}/ (1+(x/alpha)^{beta})^2
+        # Compute f_1 and f_2 which only depend on beta:
+        # f_1 := mean/alpha = <x>/alpha
+        # f_2 := variance/alpha^2 = (<x^2> - <x>^2)/alpha^2
+        # In the large beta limit, f_1 -> 1 and f_1/sqrt(f_2) -> 0.56*beta - 0.25
+        # Solve for alpha and beta below:
+        alpha_0, beta_0 = m, (1 / 0.56) * (m / np.sqrt(v) + 1 / 4)
```
coxipi, line 509: I didn't find any parameter estimation in the literature, so I computed my own. I described the process, but should I include more details / include a notebook somewhere?

For reference, here are the functions described in line 507 
[f1.pdf](https://github.com/Ouranosinc/xclim/files/9619010/f1.pdf)
[ratio.pdf](https://github.com/Ouranosinc/xclim/files/9619009/ratio.pdf)

HISTORY.rst
```diff
@@ -4,17 +4,19 @@ History
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * Documentation now supports intersphinx mapping references within code examples via `sphinx-codeautolink` and copying of code blocks via `sphinx-copybutton`. (:pull:`1182`).
+* Log-logistic distribution added to `stats.py` for use with `standardized_precipitation_index` and `standardized_precipitation_evapotranspiration_index` (:issue:`1141`, :pull:`1183`).
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
 * Indices that accept `lat` or `lon` coordinates in their call signatures will now use `cf-xarray` accessors to gather these variables in the event that they are not explicitly supplied. (:pull:`1180`). This affects the following
     - ``huglin_index``, ``biologically_effective_degree_days``, ``cool_night_index``, ``latitude_temperature_index``, ``water_budget``, ``potential_evapotranspiration``
 * ``cool_night_index`` now accepts ``lat: str = "north" | "south"`` for calculating CNI over DataArrays lacking a latitude coordinate. (:pull:`1180`).
+* The offset value in `standardized_precipitation_evapotranspiration_index` is changed to better reproduce results in the reference library `monocongo/climate_indices`.
```
aulemahal, line 19:
```suggestion
* The offset value in ``standardized_precipitation_evapotranspiration_index`` is changed to better reproduce results in the reference library ``monocongo/climate_indices``.
```

xclim/indices/_agro.py
```diff
@@ -864,16 +864,11 @@ def standardized_precipitation_evapotranspiration_index(
...
-        # In monocongo, the offset would be 1000/(60*60*24) in [kg m-2 s-1]
-        # The choice can lead to differences as big as +/-0.2 in the SPEI.
-        # If taken too big, there are problems with the "ML" method  (this should be an
-        # issue with the fitting procedure that also needs attention)
-        offset = convert_units_to("1e-4 kg m-2 s-1", wb.units)
-        # Increase offset if negative values remain
-        offset = offset - 2 * min(wb.min(), wb_cal.min(), 0)
+        # are allowed. The offset choice is arbitrary and the same offset as the monocongo
+        # library is taken
+        offset = convert_units_to(f"{1e3 / (24*60*60*1000)} kg m-2 s-1", wb.units)
```
aulemahal, line 871:
```suggestion
        offset = convert_units_to("1 mm/d", wb.units)
```

xclim/indices/_agro.py
```diff
@@ -728,7 +728,7 @@ def standardized_precipitation_index(
 
     """
     # "WPM" method doesn't seem to work for gamma or pearson3
-    dist_and_methods = {"gamma": ["ML", "APP"]}
+    dist_and_methods = {"gamma": ["ML", "APP"], "fisk": ["ML"]}
```
aulemahal, line 731: If you implemented a `_fit_start` for `fisk`, couldn't we allow "APP" here?  Sorry if I forgot a previous discussion about this...

aulemahal (approved):
This looks good! 

A detail, but the `dist` argument currently does not appear in the attributes of SPI or SPEI. Does @RondeauG think it should appear at least in the `description`?

xclim/indices/_agro.py
coxipi: Yes, sounds good. The previous idea we discussed was not to use "APP" with `fisk`. But this was back when I was solving saddle point equations with `fsolve` directly in `_fit_start` to get parameters. Now,  there is a simple approximation in terms of mean and variance to get parameters. I agree "APP" should be available for `fisk` now.

RondeauG:
> This looks good!
> 
> A detail, but the `dist` argument currently does not appear in the attributes of SPI or SPEI. Does @RondeauG think it should appear at least in the `description`?

Probably. More metadata is never bad.

xclim/indices/_agro.py
```diff
@@ -694,8 +694,8 @@ def standardized_precipitation_index(
     window : int
         Averaging window length relative to the resampling frequency. For example, if `freq="MS"`,
         i.e. a monthly resampling, the window is an integer number of months.
-    dist : {'gamma'}
-         Name of the univariate distribution, only `gamma` is currently implemented
+    dist : {"gamma", "fisk"}
+        Name of the univariate distribution.
```
huard, line 698: Should probably mention somewhere that fisk is another name for log-logistic.

xclim/indices/stats.py
```diff
@@ -90,8 +91,8 @@ def fit(
         Name of the univariate distribution, such as beta, expon, genextreme, gamma, gumbel_r, lognorm, norm
         (see :py:mod:scipy.stats for full list). If the PWM method is used, only the following distributions are
         currently supported: 'expon', 'gamma', 'genextreme', 'genpareto', 'gumbel_r', 'pearson3', 'weibull_min'.
-    method : {"ML", "PWM"}
-        Fitting method, either maximum likelihood (ML) or probability weighted moments (PWM), also called L-Moments.
+    method : {"ML", "PWM", "APP"}
+        Fitting method, either maximum likelihood (ML), probability weighted moments (PWM), also called L-Moments, or approximate method (APP)
```
huard, line 95: I'm a bit confused by the APP method. It essentially returns the value of _fit_start, which is just a rough estimate of initial values.

huard: What I mean is that I'm not sure it's a good idea to expose those initial values as a legitimate fit methods in the public interface.

coxipi: This was introduced in a previous PR for the SPI/SPEI using only the gamma distribution and we were comparing with the library `monocongo/climate_indices` (used by `drought.gov`) which only uses the approximate method as in  `_fit_start`.  Is the issue here that we label this as a "gamma fit" even though only approximate methods are used? Or is the concern about the usefulness of such results? When I was developing this, I saw some rare instances where the approximate method seem more robust than the full method using ``scipy``, but I'm not sure I noted these values: I can try to find them back.

huard: Yes, my concern is that we're exposing an approximate fitting method, appropriate for this case, inside a generic function.

coxipi: Ok, I see. Do you think that calling `_fit_start`  directly within `standardardized_precipitation_index` and `standardardized_precipitation_evapotranspiration_index` if the method "APP" is given as input could be an acceptable alternative?

huard: Hum, this is going to add complexity. For me, the best option would be to remove the APP method from the public interface, and make sure the scipy fit function works well, but that might prove tricky or require upstream changes. I'll keep thinking about this.

PR #1186, opened 2022-09-19
Refactor first_day indices into generics

### LINKED ISSUE #1175: The `first_day_below` and `first_day_above` indices should be made into generics
I think it would be a good idea to move the two `first_day` indices under `_temperature.py` to `generic.py` so that we can construct better indicators out of them.

Despite their docstrings being appropriate for general temperature data, they currently use `tasmin` in their call signature (since our regional focus largely concerns winter phenomena). For the wrapped indicators of these, it means that a dataset with `tasmin` among its data_vars will have that variable used by default. This can be misleading when using the indicator and indice forms.

If we were to move these over, I would elect to deprecate `first_day_above` and `first_day_below` and propose the following replacement, under `generic.py`:

 * first_day_above: `def first_day_below(da: xarray.DataArray, thresh: str = None, before_date: DayOfYearStr = "07-01", window: int = 1, freq: str = "YS",) -> xarray.DataArray`
 * first_day_below: `def first_day_below(da: xarray.DataArray, thresh: str = None, after_date: DayOfYearStr = "07-01", window: int = 1, freq: str = "YS",) -> xarray.DataArray`

under `_temperature.py`:

* first_day_{tn | tg | tx}_above: `def first_day_above({tasmin | tas | tasmax}: xarray.DataArray, thresh: str = None, before_date: DayOfYearStr = "07-01", window: int = 1, freq: str = "YS",) -> xarray.DataArray`
* first_day_{tn | tg | tx}_below: `def first_day_below({tasmin | tas | tasmax}: xarray.DataArray, thresh: str = None, after_date: DayOfYearStr = "07-01", window: int = 1, freq: str = "YS",) -> xarray.DataArray`

This would mean making some deprecation warnings for the existing indices/indicators, with breaking changes coming later.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1175 

* Uses `DayOfYearStr` from same file source
* Refactors indices into generics that compose temperature-based indices and indicators unique to variables
* Adds deprecation notices where needed

Yes. `first_day_below` and `first_day_above` are now generics, and users must transition to using `first_day_tn_below` and `first_day_tn_above` for previous behaviour. Warnings have been added to their docstrings. For previous versions, they remain available, but will show `DeprecationWarnings` when called by user.

`xclim.indices._threshold.first_day_above|below` will be removed in xclim>=0.40.0

### Other information:

The `first_day_XX_above|below` indices all perform local imports of generic `first_day_above` and `first_day_below`. This was needed to ensure that the deprecated versionsof these indices remain available without overwriting the newer generics. This will no longer be needed in xclim>=0.40.0

### DIFF
--- a/xclim/indicators/atmos/_temperature.py
+++ b/xclim/indicators/atmos/_temperature.py
@@ -626,8 +626,42 @@ first_day_below = Temp(
     standard_name="day_of_year",
     long_name="First day of year with temperature below {thresh}",
     description="First day of year with temperature below {thresh} for at least {window} days.",
-    compute=indices.first_day_below,
+    compute=indices.first_day_tn_below,
+    _version_deprecated="0.39.0",
 )
+
+first_day_tn_below = Temp(
+    identifier="first_day_below",
+    units="",
+    standard_name="day_of_year",
+    long_name="First day of year with minimum temperature below {thresh}",
+    description="First day of year with minimum temperature below {thresh} for at least {window} days.",
+    compute=indices.first_day_tn_below,
+    parameters=dict(thresh={"default": "0 degC"}, after_date={"default": "07-01"}),
+)
+
+
+first_day_tg_below = Temp(
+    identifier="first_day_below",
+    units="",
+    standard_name="day_of_year",
+    long_name="First day of year with mean temperature below {thresh}",
+    description="First day of year with mean temperature below {thresh} for at least {window} days.",
+    compute=indices.first_day_tg_below,
+    parameters=dict(thresh={"default": "0 degC"}, after_date={"default": "07-01"}),
+)
+
+
+first_day_tx_below = Temp(
+    identifier="first_day_below",
+    units="",
+    standard_name="day_of_year",
+    long_name="First day of year with maximum temperature below {thresh}",
+    description="First day of year with maximum temperature below {thresh} for at least {window} days.",
+    compute=indices.first_day_tx_below,
+    parameters=dict(thresh={"default": "0 degC"}, after_date={"default": "07-01"}),
+)
+
 
 first_day_above = Temp(
     identifier="first_day_above",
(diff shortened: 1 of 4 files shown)


xclim/indicators/atmos/_temperature.py
```diff
@@ -626,18 +626,83 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
+    standard_name="day_of_year",
+    long_name="First day of year with minimum temperature below {thresh}",
+    description="First day of year with minimum temperature below {thresh} for at least {window} days.",
+    compute=indices.first_day_tn_below,
+    parameters=dict(thresh={"default": "0 degC"}, after_date={"default": "07-01"}),
+)
+
+
+first_day_tg_below = Temp(
+    identifier="first_day_below",
```
aulemahal, line 645: All your indicators have the same identifier! It should be the same as the python name.

aulemahal (commented):
Suggestion, if you like it : To further generalize this, we could use the indicator-creation "input" arg to transform the generic function directly into an indicator. This way, you wouldn't need all the indices.

Example:
```python3
first_day_tn_below = Temp(
    identifier="first_day_tn_below",
    units="",
    standard_name="day_of_year",
    long_name="First day of year with minimum temperature below {threshold}",
    description="First day of year with minimum temperature below {threshold} for at least {window} days.",
    title="First day of year with minimum temperature below a threshold.",
    abstract="Returns first day of period where minimum temperature is inferior to a threshold over a given number of days, limited to a starting calendar date.",
    compute=indices.generic.first_day_below,
    input={'data': 'tasmin'},
    parameters=dict(threshold={"default": "0 degC"}, after_date={"default": "07-01"}),
)
```


Also. We already have  "tn_days_below",  wouldn't it be coherent to name the new indicators like: "tn_first_day_below" ?

xclim/indicators/atmos/_temperature.py
Zeitsperre: Ah! Yes, absolutely. Good eye!

xclim/indices/generic.py
```diff
@@ -808,3 +813,107 @@ def degree_days(tas: xr.DataArray, threshold: str, op: str) -> xr.DataArray:
         raise NotImplementedError(f"Condition not supported: '{op}'.")
 
     return to_agg_units(out, tas, op="delta_prod")
+
+
+def first_day_above(
```
bzah, line 818: To go one step further into generification you could have instead a `first_day_threshold_reached` with an operator parameter  taking either `>` or `<` as a valid value.
It would reduce the duplication of code and doc.
This would also ease the adding of new operators such as `>=`, `<=`, `==`
It similar to what is done on `degree_days` for example.

Zeitsperre: Absolutely! Will make changes. Thanks!

Zeitsperre:
This is now ready for final review

HISTORY.rst
```diff
@@ -26,7 +36,8 @@ Internal changes
 * The documentation build now relies on `sphinx-codeautolink` and `sphinx-copybutton`. (:pull:`1182`).
 * Many docstrings did not fully adhere to the `numpy docstring format <https://numpydoc.readthedocs.io/en/latest/format.html>`_. Fields and entries for many classes and functions have been adjusted to adhere better. (:pull:`1182`).
 * The xdoctest namespace now provides access to session-scoped ``{variable}_dataset`` accessors, as well as a ``path_to_atmos_file`` object. These can be used for running doctests on all variables made in the pytest ``atmosds()`` fixture. (:pull:`1882`).
-* Upgrade CodeQL github action to v2 (:issue:`1188`, :pull:`1189`).
+* Upgrade CodeQL GitHub Action to v2. (:issue:`1188`, :pull:`1189`).
+* New generic indice ``first_day_threshold_reached`` is now used to compose all ``first_day_XYZ`` indices. (:issue:`1175`, :pull:`1186`).
```
bzah, line 40:
```suggestion
* New generic index ``first_day_threshold_reached`` is now used to compose all ``first_day_XYZ`` indices. (:issue:`1175`, :pull:`1186`).
```

bzah (approved):
lgtm!

aulemahal (commented):
I believe the french translations are missing ;)!

Otherwise, this looks good!

PR #1197, opened 2022-10-07
Adopt operators in warm/cold spell, dry, and wet days indices

### LINKED ISSUE #1194: Adopt operators in  `hot_spell_frequency` and `hot_spell_max_length`
### Generic Issue

* xclim version: 0.38.0
* Python version: All 
* Operating System: Any

### Description

`hot_spell_frequency` and `hot_spell_max_length` use ">" directly in their calculations. These indices should adopt the new "operator" feature introduced in v0.38.0, with accompanying changes available for the wrapped indicators. Default values should still be ">" for both.

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1194 

* Integrate operators into the indices and indicators of:
  * `hot_spell_max_length`
  * `hot_spell_frequency`
  * `cold_spell_days`
  * `cold_spell_frequency`
  * `heat_wave_index`
  * `warm_day_frequency (indice only)`
  * `warm_night_frequency (indice only)`
  * `dry_days`
  * `wetdays`
  * `wetdays_prop`

Not really. The same behaviour is used by default for all affected indices/indicators.

### DIFF
--- a/xclim/indicators/atmos/_temperature.py
+++ b/xclim/indicators/atmos/_temperature.py
@@ -247,7 +247,6 @@ heat_wave_index = Temp(
     "temperatures exceed given thresholds for a number of days.",
     cell_methods="",
     compute=indices.heat_wave_index,
-    parameters={"op": {"default": ">"}},
 )
 
 
@@ -264,7 +263,6 @@ hot_spell_frequency = Temp(
     cell_methods="",
     keywords="health,",
     compute=indices.hot_spell_frequency,
-    parameters={"op": {"default": ">"}},
 )
 
 hot_spell_max_length = Temp(
@@ -280,7 +278,6 @@ hot_spell_max_length = Temp(
     cell_methods="",
     keywords="health,",
     compute=indices.hot_spell_max_length,
-    parameters={"op": {"default": ">"}},
 )
 
 tg_mean = TempWithIndexing(
@@ -470,7 +467,6 @@ cold_spell_days = Temp(
     "consecutive days with mean daily temperature below a given threshold.",
     cell_methods="",
     compute=indices.cold_spell_days,
-    parameters={"op": {"default": "<"}},
 )
 
 cold_spell_frequency = Temp(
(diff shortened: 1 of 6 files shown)


xclim/indicators/atmos/_temperature.py
```diff
@@ -467,6 +470,7 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
     "consecutive days with mean daily temperature below a given threshold.",
     cell_methods="",
     compute=indices.cold_spell_days,
+    parameters={"op": {"default": "<"}},
```
aulemahal, line 473: I haven't checked them all, but you don't need to pass a "parameter" arg here if you are not changing anything. For  `cold_spell_days`, the default already is "<", so we usually do not repeat the information here.

aulemahal (approved):
Looks good! I think you may have gone too verbose in the indicator part.

xclim/indicators/atmos/_temperature.py
Zeitsperre: If `parameters` is not passed to the indicator wrapper, is it still possible for a user to change/control these values? I figured why not offer this as a possible option, but if it's too much in your opinion, I don't mind reverting some of these changes.

aulemahal: When `parameters` are passed as dict, it simply overrides (updates) the configuration parsed from the indice's docstring and signature. So yes, even if it is not passed as you did, the argument still appears in the indicator's signature, and it is modifiable.

That provision is meant mostly to be used when we want to _inject_ a parameter and remove it from the indicator's signature. For example, in the same file you can see `daily_freezethaw_cycles`, which is a stricter version of `multiday_temperature_swing` : 
```
daily_freezethaw_cycles = TempWithIndexing(
   [...]
    parameters={
        "op": "sum",
        "window": 1,
        "thresh_tasmax": {"default": "0 degC"},
        "thresh_tasmin": {"default": "0 degC"},
        "op_tasmax": {"default": ">"},
        "op_tasmin": {"default": "<="},
    },
)
```
Here the `op` and `window` parameters are no longer controllable, but the others have only seen their default value change.

Zeitsperre: I see! This demystifies the behaviour. Thanks for the explanation!

PR #1198, opened 2022-10-07
Spatial sdba properties - rewrite annual cycle properties - StatistialPropertyMeasure

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* New base indicator class for sdba: ``StatisticalPropertyMeasure``, those measures that also reduce the time (as a property does). This has no effect real on xclim yet, but should open a way to use those in xscen.
* New spatial sdba properties and measures : ``spatial_correlogram``, ``scorr`` and ``first_eof``. The later needs the optional dependency "eofs". The first two are quite heavy to compute as they perform pairwise  spearman correlations...
    - It may help to add a "max_distance" arg to both, to limit the number of points. 
    - We could add a "spatial correlation length" by making use of `spatial_correlogram`.
* The ``relative_annual_cycle_amplitude`` and ``annual_cycle_amplitude`` have been rewritten to match the version defined in the VALUE project, outputs will change drastically (for the better). See below.
* The `test-py38-sbck-slow-conda`  github build has been renamed to `test-py38-opt-slow-conda`. It also installs Eofs, the new optional dependency.

Yes for ``relative_annual_cycle_amplitude`` and ``annual_cycle_amplitude``.

The previous code computing the annual cycle amplitude for each year individually and then averaged them together. This mean we had the mean of (max - min). The new version compute the averaged annual cycle and then takes the amplitude (max(mean) - min(mean)). This is how VALUE does it. The former version depends heavily on the interannual variability of the dataset, while the latter less so. The former version was not uninteresting, it just didn't seem to be the annual cycle amplitude I was expecting. 

While fixing this, I also integrated the `window` argument, as VALUE has it. This further hammers the nails of smoothing the annual cycle first and taking the statistic in a second step.
I rewrote all "annual cycle" properties as cases of a single function.

### Other information:
Diagnostics of ESPO-R5 were computed with this branch.


juliettelavoie:
For the annual cycle, I have a vague memory of having a discussion about this were we decided to not follow VALUE. I thought you were involved in it, but it might have been in a weekly meeting with LP and Travis. I remember seeing that VALUE was doing the mean first and that didn't make much sense to me. If the phase changes, the mean annual cycle might be flatten and then its amplitude will not be representative of the amplitude of each annual cycles. Hence, I coded amplitude first, then mean. Could you explain why doing the mean first is better ?

aulemahal:
Oh! I do not remember that conversation, but I may have simply forgotten...

I think the noise in the previous version is due to the use of `max` and `min` on raw daily data. Indeed, may be we could smooth the data before hand and get something more interesting?

When I talk about what "I expected", I mean that without reading the property's code. I thought of the "annual cycle"  of the data as this "daily climatology", thus the `tas.groupby('time.dayfoyear').mean()`. As in : the data for a single year is not an "annual cycle", that concept is on the climatology side for me. Also, I expected the bias-adjustment to be almost perfect on this matter, which is not what I got with the previous version.

To me, the previous version was kinda looking at the average annual range more than its "cycle amplitude". And it is interesting to see that the DQM used in ESPO-R5 did not adjust that, at all. However, this might only be noise, because when adding a smoothing to the previous version I get more coherent results of an improved property.

I think I would keep my changes to ensure we have a correct mapping between xclim and VALUE? But I will reimplement the previous version, with an added window parameter. I'm not sure on the name to use though...

juliettelavoie:
Okay, let's change it to fit VALUE. I did not realise the association between annual cycle and climatology, but if this is more common in the litterature and people's mind, it is better to do it that way.  I was imagining a sin wave with given frequency and amplitude that repeats for many cycles.

It think `mean_annual_range` is fine for the name of the old version.  or maybe `timeseries_amplitude` ?

tox.ini
```diff
@@ -58,11 +58,12 @@ commands =
...
     pytest --rootdir xclim/testing/tests/ --xdoctest xclim --ignore=xclim/testing/tests/ --durations=10
 
-[testenv:sbck-slow]
-description = Run tests with SBCK (experimental) and pytest under {basepython} (Anaconda distribution)
+[testenv:opt-slow]
+description = Run tests with optional requirements (SBCK (experimental), eofs) and pytest under {basepython} (Anaconda distribution)
 extras =
 conda_channels = conda-forge
 conda_deps =
+    eofs
```
Zeitsperre, line 66: I would also move the `pybind11` dependency into here (remove from line 86).

xclim/sdba/measures.py
```diff
@@ -197,20 +300,31 @@ def _rmse(sim, ref):
...
 
     Parameters
     ----------
     sim : xr.DataArray
-      data from the simulation (a time-series for each grid-point)
+        data from the simulation (a time-series for each grid-point)
     ref : xr.DataArray
-      data from the reference (observations) (a time-series for each grid-point)
+        data from the reference (observations) (a time-series for each grid-point)
+    group: str
```
Zeitsperre, line 325:
```suggestion
    group : str
```

xclim/sdba/measures.py
```diff
@@ -246,20 +369,23 @@ def _annual_cycle_correlation(
     Parameters
     ----------
     sim : xr.DataArray
-      data from the simulation (a time-series for each grid-point)
+        data from the simulation (a time-series for each grid-point)
     ref : xr.DataArray
-      data from the reference (observations) (a time-series for each grid-point)
+        data from the reference (observations) (a time-series for each grid-point)
     window: int
```
Zeitsperre, line 375:
```suggestion
    window : int
```

xclim/sdba/measures.py
```diff
@@ -246,20 +369,23 @@ def _annual_cycle_correlation(
...
-      Size of window around each day of year around which to take the mean.
-      E.g. If window=31, Jan 1st is averaged over from December 17th to January 16th.
+        Size of window around each day of year around which to take the mean.
+        E.g. If window=31, Jan 1st is averaged over from December 17th to January 16th.
+    group: str
```
Zeitsperre, line 378:
```suggestion
    group : str
```

xclim/sdba/measures.py
```diff
@@ -268,5 +394,51 @@ def _annual_cycle_correlation(
...
+    Compute the inter-site correlations of each array, compute the difference in correlations and sum.
+    Taken from Vrac (2018). The spatial and temporal dimensions are reduced.
+
+    Parameters
+    ----------
+    sim : xr.DataArray
+        data from the simulation (a time-series for each grid-point)
+    ref : xr.DataArray
+        data from the reference (observations) (a time-series for each grid-point)
+    dims: sequence of strings, optional
```
Zeitsperre, line 422:
```suggestion
    dims : sequence of strings, optional
```

xclim/sdba/measures.py
```diff
@@ -268,5 +394,51 @@ def _annual_cycle_correlation(
...
+        Name of the spatial dimensions. If None (default), all dimensions except 'time' are used.
+    group: str
```
Zeitsperre, line 424:
```suggestion
    group : str
```

Zeitsperre (approved):

xclim/sdba/measures.py
```diff
@@ -231,11 +345,20 @@ def _mae(sim, ref):
     return out.assign_attrs(units=ensure_delta(ref.units))
 
 
-mae = StatisticalMeasure(identifier="mae", compute=_mae)
+mae = StatisticalMeasure(
```
juliettelavoie, line 348:
```suggestion
mae = StatisticalPropertyMeasure(
```

xclim/sdba/measures.py
```diff
@@ -268,5 +394,51 @@ def _annual_cycle_correlation(
 
 
 annual_cycle_correlation = StatisticalMeasure(
```
juliettelavoie, line 396:
```suggestion
annual_cycle_correlation = StatisticalPropertyMeasure(
```

xclim/sdba/properties.py
```diff
@@ -403,112 +406,186 @@ def _acf(ds, *, dim, lag, freq):
...
+)
+
+annual_cycle_asymmetry = StatisticalProperty(
+    identifier="annual_cycle_asymmetry",
+    aspect="temporal",
+    compute=_annual_cycle,
+    parameters={"stat": "asymmetry"},
+    allowed_groups=["group"],
+    units="yr",
+)
```
juliettelavoie, line 514: min and max exist in stat, but they are not create as indicators. Is this on purpose ?

juliettelavoie (approved):
looks good! Il y a deux mesures que je pense devait être des StatisticalPropertyMeasure parce que leur input sont des timeseries. Je pense que c'est peut-être juste un oubli.

xclim/sdba/properties.py
aulemahal: More like laziness. I'll add them :).

Zeitsperre (approved):

aulemahal:
@Zeitsperre If you have time, could you check the errors? I get ` ERROR: unknown environment 'py39-opt-slow'`, so I guess I didn't do something, but I can't see what! Thanks!

juliettelavoie:
`spatial_correlogram` is not working for me... When I try to run exactly the same code as the test, the first test works but for 
```
sim = open_dataset("NRCANdaily/nrcan_canada_daily_tasmax_1990.nc").tasmax
out = sdba.properties.spatial_correlogram(
    sim.isel(lon=slice(0, 50)), dims=["lon", "lat"], bins=20
)
```
I get an error: `KeyError: "DataArray.cf does not understand the key 'longitude'.Use 'repr(DataArray.cf)' (or 'DataArray.cf' in a Jupyter environment) to see a list of key names that can be interpreted."`

If I try to do it mannually, `corr.cf` doesn't have the information linking longitudes to 'lon'.

aulemahal:
Hum, I tried with xarray 2022.06.0 and 2022.10.0 and it worked both times. This really looks like an xarray issue, which version do you have?

Zeitsperre:
@aulemahal The tox configuration is a convoluted, but I'm going to revise and clean this up in another PR at some point. Not blocking the release.

aulemahal:
Ah fiou! Super alors. On fusionne quand les tests réussissent!

PR #1206, opened 2022-10-18
ENH: Update convert_units_to function

### LINKED ISSUE #1205: Convert [length] into [length]/[time] when cell_methods and/or standard_name allows it
### Generic Issue

* xclim version: 0.38

### Description

There are precipitation datasets, such as E-OBS, that bear a `[length]` units (usually "mm") that could be interpreted as a `[length]/[time]` (usually "mm/day") when computing climate indices.

I think `xclim.units.convert_units_to` could handle such cases when the input is a DataArray. 

Two parameters would help us find if this conversion would be safe:
- if cell_methods contains "time: sum", it means it's an accumulation over time.
- if the variable standard_name is the CF `thickness_of_rainfall_amount`:
> "Amount" means mass per unit area. The construction thickness_of_[X_]rainfall_amount means the accumulated "depth" of rainfall i.e. the thickness of a layer of liquid water having the same mass per unit area as the rainfall amount.

Note that "cell_methods" seems to be missing from e-obs datasets, so it might not be the best idea to use that.

----

What do you think about that ?
If you are okay with adding this feature, I can implement it.

ref:
- icclim issue: https://github.com/cerfacs-globc/icclim/issues/195#issuecomment-1193951194
- e-obs: https://surfobs.climate.copernicus.eu/dataaccess/access_eobs_months.php
- cf: https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

### Code of Conduct

bzah:
### Pull Request Checklist:
    - This PR fixes #1205

This PR adds standard-name-aware unit conversions. Those are listed in `data/variables.yml` in a new  "conversions" section. For now, the `amount2rate` and the new `amount2lwethickness` are available (and reverse operations).

`convert_units_to` will also perform those conversion automatically if the change in dimensionality requires it and the input has a valid standard name. 

The new functions are often the equivalent of the "hydro" context. Both are complementary: the hydro context is still to be used in indicators for which the transformation is implicitly valid, no matter the CF attributes. 

This is necessary to be able to compute precipitation indices on datasets that express RR as "thickness_of_rainfall_amount" instead of the usual "precipitation_flux" such as the e-obs datasets.

I can't make these transformations available automatically in the indices with the current structure state of  `declare_units`, I'll look into this in a future PR. For now, as shown in the edited `units.ipynb` these "CF conversions" must be down explicitly beforehand.

Yes, the error `pint.errors.DimensionalityError` is no longer raised when attempting to convert this kind of datasets described above into a `[length] / [time]` compatible unit.


bzah:
A more generalist approach could be to do something like (pseudo-code):
```
if  is_amount(source) and is_rate(target):
  src_rate = amount2rate(source)
  return convert(src_rate, target)
elif is_rate(source) and is_amount(target):
   src_amount rate2amount(source)
   return convert(src_amount, target)
```
That would work for "thickness_of_rainfall_amount" case, but also for any attempted conversion between an amount and a rate.

WDYT ?

xclim/core/units.py
```diff
@@ -42,6 +41,7 @@
 
 
 units = pint.UnitRegistry(autoconvert_offset_to_baseunit=True, on_redefinition="ignore")
+Convertible = TypeVar("Convertible", str, xr.DataArray, units.Unit, units.Quantity)
```
Zeitsperre, line 44: Neat! Can you briefly explain this syntax? How will this render in the docs?

xclim/core/units.py
```diff
@@ -285,60 +287,79 @@ def convert_units_to(
...
 
 
+def _is_precipitation_amount(source: xr.DataArray, source_unit: units.Unit) -> bool:
+    standard_name = source.attrs.get("standard_name", None)
+    return standard_name == PR_AMOUNT_STANDARD_NAME and _is_amount(source_unit)
+
+
+def _is_amount(source_unit: units.Unit) -> bool:
+    quantity = units.Quantity(1, source_unit)
+    return quantity.check("[length]") or quantity.check("[mass] / [length]**2")
```
aulemahal, line 360: Oh I didn't know about this "check", nice!

xclim/core/units.py
```diff
@@ -285,60 +287,79 @@ def convert_units_to(
...
+        if source_unit == target_unit:
             # The units are the same, but the symbol may not be.
-            source.attrs["units"] = tu_u
+            source.attrs["units"] = target_cf_unit
             return source
 
         with units.context(context or "none"):
+            if (
+                _is_precipitation_amount(source, source_unit)
+                and target_unit.dimensionality.get("[time]") == -1
```
aulemahal, line 324: Would it be too strict to do:
`_is_precipitation(source) and _is_amount(source_unit) and _is_rate(target_unit)`
?

This would indeed allow, to have a reverse operation, if ever needed.

xclim/core/units.py
```diff
@@ -285,60 +287,79 @@ def convert_units_to(
...
+            ):
+                source = amount2rate(source)
```
aulemahal, line 326: I realized a potential bug in `amount2rate` yesterday.
If the time axis is not continuous, the sampling rate is determined by the time diff. This means that if you have a large gap in your data, the data point at the start of that gap gets divided by a very large time period. This is silent.
Thus, doing this conversion automatically seems dangerous to me! The solution could be that `amount2rate` raises if `xr.infer_freq` fails on the data, unless the behaviour is deactivate by some argument to be added.
What do you think?

huard: I agree with this. I was surprised to realize the function was so liberal.

xclim/core/units.py
huard: I suggest we keep the reverse operation for another PR if needed.

xclim/core/units.py
huard: Fixed in [d1fea25](https://github.com/Ouranosinc/xclim/pull/1206/commits/d1fea25b72dc1dbf715e5b39db0ff88971cdb937)

huard:
@aulemahal  Ready for final review. Should we add a test for irregular time series to confirm it fails ?

xclim/core/units.py
```diff
@@ -277,56 +283,91 @@ def convert_units_to(
...
     if isinstance(source, units.Quantity):
-        return source.to(tu).m
+        return source.to(target_unit).m
 
     if isinstance(source, xr.DataArray):
-        fu = units2pint(source)
-        tu_u = pint2cfunits(tu)
-
-        if fu == tu:
+        source_unit = units2pint(source)
```
huard, line 309: I have the feeling some of the logic here would fit better within `_rate_and_amount_converter`.

xclim/core/units.py
aulemahal: The annotation shows `"Convertible"`, so it seems equivalent to 
```
Convertible = str | xr.DataArray | units.Unit | units.Quantity 
```

aulemahal:
Woupsi sorry.

aulemahal:
@huard  and the others, I believe this is ready!

I edited the top post to explain what was done here. Grosso modo : Instead of abel's units-based hook into `converts_units_to`, I chose to go with a metadata-based conversion with it's own functions. `convert_units_to` can also make use of these. Unit conversion that changes the dimensionality is now possible, but requires valid standard names.

docs/notebooks/units.ipynb
```diff
@@ -18,27 +18,28 @@
...
     "\n",
-    "For precipitation data, `xclim` expects precipitation fluxes. This could be units of `length/time`, such as mm/d, or units of `mass / area / time`, such as kg/m²/s. Units of `length` only, such as mm, are not supported, because the interpretation depends on the frequency of the data, which cannot always be inferred explicitly from the data. For example, if a daily precipitation series records total daily precipitation and has units of `mm`, change the units attribute to `mm/d` before computing indicators. Note that `xclim` will automatically convert between `mass / area / time` and `length/time` using a water density of 1000 kg/m³ when the context is hydrology.\n",
+    "The main unit handling method is [`xclim.core.units.convert_units_to`](../xclim.core.rst#xclim.core.units.convert_units_to) which can also be useful on its own. `xclim` builts upon [pint](https://pint.readthedocs.io/) to perform the unit handling.\n",
```
huard, line 40:
```suggestion
    "The main unit handling method is [`xclim.core.units.convert_units_to`](../xclim.core.rst#xclim.core.units.convert_units_to) which can also be useful on its own. `xclim` relies on [pint](https://pint.readthedocs.io/) for unit handling.\n",
```

docs/notebooks/units.ipynb
```diff
@@ -78,7 +79,69 @@
...
+    "\n",
+    "We can also leverage the CF conventions to perform some other \"smart\" conversions. For example, if the CF standard name of an input refers to liquid water, the flux <-> rate and amount <-> thickness conversions explained above will be automatic in `xc.core.units.convert_units_to`, whether or not the \"hydro\" context is activated. Another CF-driven conversion is between amount and flux or thickness and reate. Here again, `convert_units_to` will see if the  `standard_name` attribute allors and will need to infer the frequency of the data. For example, if a daily precipitation series records total daily precipitation and has units of `mm`, it should use the `lwe_thickness_of_precipitation_amount` standard name and have a regular time coordinate, so that we can understand and convert it to a precipitation flux. These conversions are however not automatically done in the indicator and  `convert_units_to` should be called beforehand. \n",
```
huard, line 86: reate -> rate
allors ?

The beginning of the paragraph is clear, but the end gets confusing.

xclim/core/units.py
```diff
@@ -277,60 +304,117 @@ def convert_units_to(
...
+        # Automatic pre-conversions based on the dimensionalities and CF standard names
+        standard_name = source.attrs.get("standard_name")
+        if (
+            standard_name is not None
+            and source_unit.dimensionality != target_unit.dimensionality
+        ):
+            dim_order_diff = source_unit.dimensionality / target_unit.dimensionality
+            for convname, convconf in CF_CONVERSIONS.items():
+                for direction, sign in [("to", 1), ("from", -1)]:
+                    # If the dimensionality diff compatible with this conversion
```
huard, line 344:
```suggestion
                    # If the dimensionality diff is compatible with this conversion
```

xclim/core/units.py
```diff
@@ -528,45 +622,65 @@ def _rate_and_amount_converter(
...
     Parameters
     ----------
     rate : xr.DataArray
         "Rate" variable, with units of "amount" per time. Ex: Precipitation in "mm / d".
     dim : str
         The time dimension.
+    sampling_rate_from_coord : boolean
+        For data with irregular time coordinates. If True, the diff of the time coordinate will be used as the sampling rate,
+        meaning each data point will be assumed to apply for the interval ending at the next point. See notes.
+        Defaults to False, which raises an error if the time coordiante is irregular.
```
huard, line 675:
```suggestion
        Defaults to False, which raises an error if the time coordinate is irregular.
```

docs/notebooks/units.ipynb
aulemahal: Woups indeed, I committed a partial edit. It is updated.

bzah:
Nice to see this being merge 🎉
Thanks for taking over on this PR!

PR #1217, opened 2022-10-31
Time and calendar improvements

### LINKED ISSUE #1111: Check compatibility of time coordinates of inputs
When calling an indicator with multiple inputs, the user has to make sure all those inputs have compatible time coordinates. For example, if one variable has daily values on midnight and the other at noon, xarray will try to align the two and the result will either have no time steps (the intersection) or interspersed ones (the union). In the second case, everything will usually be NaN.

Q1 : Does it fall in xclim's scope to detect this problem?

Q2 : If so, to what extent would xclim intervene? 
If we implement something, I would suggest a `datacheck`, so the behaviour is controllable from the options.

What do @huard and @tlogan2000  think?

### Code of Conduct

### LINKED ISSUE #1190: Better calendar choice in ensembles - allow to pass conversion options
### Addressing a Problem?

In  `create_ensemble` the `calendar` argument defaults to "default", which means, for example, that if only "noleap" data is passed, the data is converted anyway with the default arguments of `convert_calendar`. This results in a daily dataset that fails the `check_daily` test (and also the `missing_any` on leap years).

### Potential Solution

`calendar` could default to the "smallest common calendar" ? For example, a mixed input of "noleap" and "360_day" would default to "noleap" (because 360_day is specifically converted with `align_on='year'` already). A single "proleptic_gregorian" entry would trigger a `default` conversion.

In my example, some realization would still fail `missing_any` , but that's on the user's at this point. We could add a notice in the doc.

### Additional context

One might still want to have an ensemble with a "larger" calendar than the inputs but without having it failing `check_daily`. The solution is then to pass  `missing=np.nan` to `convert_calendar`, and thus we should make this option accessible in  `create_ensemble`.

### Contribution

### Code of Conduct

### LINKED ISSUE #1213: parse_offset doesn't fail on invalid offsets
### Generic Issue

 `xclim.core.calendar.parse_offset` is meant to divide a pandas' time offset (frequency) into its 4 parts. Currently, it doesn't fail for malformed offset, I think it should!

For example:
```python3
from xclim.core.calendar import parse_offset

parse_offset('day')
# (1, 'd', False, None)
```
But "day" isn't a valid offset (here it should be "D").

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1213, fixes #1190, fixes #1111

* Cleaner `xclim.core.calendar.parse_offset` : fails on invalid frequencies, return implicit anchors (YS -> JAN) and implicit `is_start_anchored` (D -> True)
* `xclim.core.calendar.common_calendar` to find the best calendar to use when uniformizing an heteroclite collection of data.
* `xclim.ensembles.create_ensemble` now accepts `calendar=None`, and uses the above function to guess the best one. It also now accepts `cal_kwargs` to fine tune the calendar conversion.
* New data check : `xclim.core.datachecks.check_common_time` that ensures all inputs of multivariate indicators have the same frequency (and the same time anchoring for daily and hourly data).

`parse_offset` has changed. The results are better and more consistent, but could break previous uses.

`construct_offset(*parse_offset(freq))` will fail in more cases than before because we now return implicit anchors.

### Other information
If your time is limited I suggest the following review dispatch:
@huard  : `datachecks.check_common_time`
@tlogan2000  : `ensembles.create_ensemble`
@Zeitsperre : `calendar.parse_offset`.


xclim/core/calendar.py
```diff
@@ -613,19 +650,28 @@ def parse_offset(freq: str) -> Sequence[str]:
...
     Returns
     -------
-    multiplier (int), offset base (str), is start anchored (bool), anchor (str or None)
-      "[n]W" is always replaced with "[7n]D", as xarray doesn't support "W" for cftime indexes.
+    multiplier : int
+       "[n]W" is always replaced with "[7n]D", as xarray doesn't support "W" for cftime indexes.
+    offset_base : str
       "Y" is always replaced with "A".
+    is_start_anchored : bool
+      Whether coordinates of this frequency should correspond to the beginning of the period (True) or it's end (False).
```
Zeitsperre, line 658:
```suggestion
      Whether coordinates of this frequency should correspond to the beginning of the period (`True`) or its end (`False`).
```

xclim/core/calendar.py
```diff
@@ -638,9 +684,9 @@ def construct_offset(mult: int, base: str, start_anchored: bool, anchor: str | N
     base : str
       The base period string (one char).
     start_anchored: bool
-      If True, adds the "S" flag.
+      If True and base in [Y, A, Q, M], adds the "S" flag.
     anchor: str, optional
-      The month anchor of the offset.
+      The month anchor of the offset. Defaults to JAN for bases AS, Y and QS and to DEC for bases A and Q.
```
Zeitsperre, line 689: Very nice! `A` starting in `DEC` needs to be made very clear in the history of breaking changes.

xclim/core/datachecks.py
```diff
@@ -59,3 +59,40 @@ def check_daily(var: xr.DataArray):
     This does not check for gaps in series.
     """
     return check_freq(var, "D")
+
+
+@datacheck
+def check_common_time(*inputs):
+    """Raise an error if the list of inputs doesn't have a single common frequency.
+
+    This raises if:
```
Zeitsperre, line 68: I would use the numpy docstring `raises` convention here: https://numpydoc.readthedocs.io/en/v1.5.0/format.html#raises

Zeitsperre (approved):
Very nice! Will need to adapt changes to the changes in Miranda, but much cleaner!

tlogan2000 (approved):
This works for me.  Something about the term `smallest common calendar` seems strange to me but after reading the description in `common calendar` I understand better.   I don't think we need to block merging but we could maybe modify the description at some point.

xclim/core/calendar.py
```diff
@@ -146,6 +145,44 @@ def get_calendar(obj: Any, dim: str = "time") -> str:
     raise ValueError(f"Calendar could not be inferred from object of type {type(obj)}.")
 
 
+def common_calendar(*calendars, join="outer") -> str:
+    """Return a calendar common to all calendars from a list.
+
```
huard, line 150:
```suggestion

Parameters
----------
calendars: Sequence
  List of calendar names.
join : {'inner', 'outer'}
  ... 
   

```

xclim/core/calendar.py
```diff
@@ -146,6 +145,44 @@ def get_calendar(obj: Any, dim: str = "time") -> str:
...
+        "366_day": "all_leap",
+        "365_day": "noleap",
+        "julian": "standard",
+    }
+    ranks = {"360_day": 0, "noleap": 1, "standard": 2, "all_leap": 3}
+    calendars = sorted([trans.get(cal, cal) for cal in calendars], key=ranks.get)
+
+    if join == "outer":
+        return calendars[-1]
+    return calendars[0]
```
huard, line 183:
```suggestion
  elif join == "inner":
    return calendars[0]
  else:
    raise ValueError(join)
```

xclim/core/calendar.py
```diff
@@ -613,19 +650,28 @@ def parse_offset(freq: str) -> Sequence[str]:
 
     Returns
     -------
-    multiplier (int), offset base (str), is start anchored (bool), anchor (str or None)
-      "[n]W" is always replaced with "[7n]D", as xarray doesn't support "W" for cftime indexes.
+    multiplier : int
+       "[n]W" is always replaced with "[7n]D", as xarray doesn't support "W" for cftime indexes.
```
huard, line 654: Start with definition of output

xclim/core/datachecks.py
```diff
@@ -59,3 +65,42 @@ def check_daily(var: xr.DataArray):
     This does not check for gaps in series.
     """
     return check_freq(var, "D")
+
+
+@datacheck
+def check_common_time(*inputs):
```
huard, line 71: Why `*inputs` and not `inputs` ?

xclim/core/datachecks.py
```diff
@@ -59,3 +65,42 @@ def check_daily(var: xr.DataArray):
...
+    """Raise an error if the list of inputs doesn't have a single common frequency.
+
+    Raises
+    ------
+    ValidationError
+        - if the frequency of any input can't be inferred
+        - if inputs have differing frequencies
```
huard, line 78:
```suggestion
        - if inputs have different frequencies
```

xclim/core/datachecks.py
```diff
@@ -59,3 +65,42 @@ def check_daily(var: xr.DataArray):
...
+    ----------
+    inputs : xr.DataArray
+        Input arrays.
+    """
+    # Check all have the same freq
+    freqs = {xr.infer_freq(da.time) for da in inputs}
+    if None in freqs:
+        raise ValidationError(
+            "Unable to infer the frequency of the time series. "
+            "To mute this, set xclim's option data_validation='log'."
```
huard, line 91: Should this message appear in the other errors as well ?

xclim/core/datachecks.py
```diff
@@ -59,3 +65,42 @@ def check_daily(var: xr.DataArray):
...
+        )
+    if len(freqs) != 1:
+        raise ValidationError(f"Inputs have different frequencies. Got : {freqs}.")
+
+    # Check if anchor is commont
```
huard, line 96:
```suggestion
    # Check if anchor is the same
```

xclim/core/datachecks.py
```diff
@@ -59,3 +65,42 @@ def check_daily(var: xr.DataArray):
...
+        - if inputs have differing frequencies
+        - if inputs have a daily or hourly frequency, but they are not given at the same time of day.
+
+    Parameters
+    ----------
+    inputs : xr.DataArray
+        Input arrays.
+    """
+    # Check all have the same freq
+    freqs = {xr.infer_freq(da.time) for da in inputs}
```
huard, line 87: Suggest using a list comprehension, then convert to a set to check for unicity. This way, you can print the list in the error message.

huard (approved):

xclim/core/calendar.py
```diff
@@ -146,6 +145,44 @@ def get_calendar(obj: Any, dim: str = "time") -> str:
     raise ValueError(f"Calendar could not be inferred from object of type {type(obj)}.")
 
 
+def common_calendar(*calendars, join="outer") -> str:
```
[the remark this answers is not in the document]
aulemahal: No reasons other than "I can". ;)
I'll change it.

Issue #1129, opened 2022-07-14
Modules should tolerate non-xclim variable names


aulemahal:
### Setup Information

- Xclim version: master
- Python version: all
- Operating System: all

### Description

When creating an indicator through a Yaml file and using the `input: {old: new}` construct to rename an input, xclim currently need a name (`new`) that exists in `variables.yml`. It fails otherwise. This was done because this construct activates the CF checks, so standard CF attributes must be available and the only way currently is through that internal file (or adding them dynamically to `xclim.core.utils.VARIABLES`).

### Steps To Reproduce

```yaml
realm: land
indicators:
  freq_analysis:
    base: freq_analysis
    input: {da: Dis}
    parameters:
      mode: max
      t: [2, 20, 100, 350]
      dist: genextreme
```
When loading this as an xclim module, it will fail with :
```
ValueError: Compute argument da was mapped to variable Dis which is not understood by xclim or CMIP6. Please use names listed in `xclim.core.utils.VARIABLES`.
```

### Additional context

I think we could simply downgrade the error to a warning and mention that the unknown name means no CF-checks will be performed.

A fancier solution would be to provide tools to add variables to the known list. If it could be made through the yaml file and be module-dependent, that seems to me to be elegant enough?

@RondeauG 

### Contribution

- [X] I would be willing/able to open a Pull Request to address this bug.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

aulemahal:
Idea after discussion:
Add an optional section in the virtual module's yaml : "variables" where one declares those new variables that xclim doesn't know. It must contain at least the "canonical_units" entry and optionally all other CF fields, so that xclim is able to build input units and CF checks for custom variables.

The potentially confusing thing here is that the `xclim.core.utils.VARIABLES` dict is global. There is not module-dependent version and thus when adding (or _overriding_) variables through a module, they get added to the global dict, for as long as the session runs.

RondeauG:
This sounds like a very good compromise! As for the point in the last paragraph, this implementation could maybe be purely additive (aka, no overriding accepted)?

PR #1225, opened 2022-11-08
Check for discharge unit compliance in streamflow statistical indicators

### LINKED ISSUE #1130: `frequency_analysis` indicator weirdly declared
### Setup Information

- Xclim version: master
- Python version: all
- Operating System: all

### Description

https://github.com/Ouranosinc/xclim/blob/1505aacbd9c5fd86885ee202aed99012e45c5abd/xclim/indicators/land/_streamflow.py#L69-L78

Here you see that the output of `frequency_analysis` is a discharge (var name `q`, units m³/s). However, the input variable has a generic name `da` and no prescribed dimensionality.  This would mean that if you pass anything else then a discharge, the unit check will not fail, but the indicator will fail when attempting to convert the output units.

Indeed, the next example fails. But even weirder,  is that we have a special CF check that looks for the "water_volume_transport_in_river_channel" standard name. Thus, we already are expecting a discharge.

### Steps To Reproduce

```python3
import xclim as xc

ds = open_dataset('ERA5/daily_surface_cancities_1990-1993.nc')
xc.land.freq_analysis(da=ds.dtas, t=100, dist='norm', mode='max')
```
warns:
```
UserWarning: Variable has a non-conforming standard_name: Got `dew_point_temperature`, expected `['water_volume_transport_in_river_channel']`
```
and fails:
```
DimensionalityError: Cannot convert from 'kelvin' ([temperature]) to 'meter ** 3 / second' ([length] ** 3 / [time])
```
The expect error would be:
```
ValidationError: Data units K are not compatible with requested [discharge].
```

### Additional context

@RondeauG 

### Contribution

### Code of Conduct

huard:
### Pull Request Checklist:
    - This PR fixes #1130 

* Check input data compliance with discharge units in streamflow indicators.
* Add discharge variable to variables.yml 
* Add test confirming failure and informative error message if units do not comply. 
* Deprecate generic streamflow indicators (fit, stats, freq_analysis)
* Create `generic` indicator realm with the indicators fit, stats, and return_level. 

Three land indicators for streamflow analysis have been deprecated.


RondeauG (approved):
Looks good to me! 

One remaining issue is that `da` is very nondescript and doesn't make is obvious that a discharge is required. If we followed conventions established for other functions in `xclim`, I'd expect `q` or something similar. However, I understand that changing it at this point would likely be a breaking change...

huard:
It's because the stats indices are generic, and are meant to be applicable to any kind of variable. 

Unless @aulemahal has magic trick up his sleeve ?

xclim/indicators/land/_streamflow.py
```diff
@@ -79,7 +79,7 @@ def cfcheck(self, **das):
     "distribution.",
     abstract="Streamflow frequency analysis on the basis of a given mode and distribution.",
     units="m^3 s-1",
-    compute=declare_units(da=None)(frequency_analysis),
+    compute=declare_units(da="[discharge]")(frequency_analysis),
```
aulemahal, line 82:
```suggestion
    compute=frequency_analysis,
    inputs={'da': 'q'}
```
This is my up-sleeve trick! It sets the correct dimensionality (and units check), expected standard name and expected cell_methods.

The trick is currently limited by #1129 as the new name (`q`) must be documented in `data/variables.yml`. And it is currently not there.

Originally meant to allow changing the variable name of a virtual-module indicator to make the use of `ind(ds=ds)` work, we realized this technique could be useful to wrap generic indices in a meaningful but easy way.  The code that manages it is here : https://github.com/Ouranosinc/xclim/blob/d565f5c0985a3eda09c395b0b7dfc4ffdb764393/xclim/core/indicator.py#L563

huard: Ok, so I understand we're OK with breaking backward compatibility in an effort to have consistent input variable naming patterns across indicators?

Are we good with the name `discharge`?

RondeauG: The decision for backwards compatibility is more so in your hands (aka, the xclim group), but `discharge` sounds good to me.

Zeitsperre: If it's possible to keep both (ie: their names are different), I'm a big fan of indicating that an indicator will be deprecated. There's a special field in `Indicator` classes for marking an indicator as being deprecated: https://github.com/Ouranosinc/xclim/blob/d565f5c0985a3eda09c395b0b7dfc4ffdb764393/xclim/core/indicator.py#L360

huard: It's the same indicator, but with a different signature, so I don't think it applies here.

aulemahal: I am ok with backward compatibility for this small set of indicators.
On the other hand, now that the indicators are not _generic_ anymore, maybe the name should reflect that?
Ex:
`freq_analysis` -> `discharge_return_period_amount`
`fit` -> `discharge_distrtibution`
?
The three others already have discharge-related names. Quoique, `doy_qmax` is a bit confusing now that we are using the `discharge` variable name.

RondeauG: I agree that `freq_analysis/fit` should probably be generic (for use with something like precipitation, too), while discharge-specific indicators should have a more explicit name.

huard:
This PR took an unexpected turn. It now includes a new `generic` realm to hold `stats`, `fit` and `return_level` (previously `freq_analysis`.

RondeauG:
I'll let the others approve the more internal changes, but it looks good to me.

xclim/indicators/land/_streamflow.py
```diff
@@ -94,17 +67,22 @@ def cfcheck(self, **das):
...
+    missing="any",
+    input={"da": "discharge"},
+    _version_deprecated="0.40",
 )
 
-fit = Fit(
+
+fit = Indicator(
     title="Distribution parameters fitted over the time dimension.",
     identifier="fit",
```
aulemahal, line 87:
```suggestion
    identifier="discharge_fit",
```
Or `discharge_distribution` ? We currently can't have two indicators with the same identifier, even from different realms.

huard: discharge_distribution_fit ?

aulemahal: Also good with me. 
This will make the identifier different from the name in the `land` namespace. This is not uncommon in xclim, but annoying. Changing it would however be a worst breaking change than only changing the identifier. I'll leave the decision to you.

PR #1231, opened 2022-11-09
New variables section in yaml module

### LINKED ISSUE #1129: Modules should tolerate non-xclim variable names

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1129

* Virtual modules can add variables to ``xclim.core.utils.VARIABLES`` through the new `variables` section of the yaml files.

No.


xclim/core/indicator.py
```diff
@@ -56,11 +63,11 @@
...
         compute: <function name>  # Referring to a function in the supplied `Indices` module, xclim.indices.generic or xclim.indices
-        input:  # When "compute" is a generic function this is a mapping from argument
-                # name to what CMIP6/xclim variable is expected. This will allow for
-                # declaring expected input units and have a CF metadata check on the inputs.
-                # Can also be used to modify the expected variable, as long as it has
-                # the same units. Ex: tas instead of tasmin.
+        input:  # When "compute" is a generic function this is a mapping from argument name to the expected variable.
+                # This will allow the input units and CF metadata checks to run on the inputs.
+                # Can also be used to modify the expected variable, as long as it has the same dimensionality
+                # Ex: tas instead of tasmin.
```
RondeauG, line 69: How will this affect metadata and the `history` field? Will it correctly keep track of the changed variable?

aulemahal: I'll add a check for this. But it should keep track : the history string is built using the Indicator's signature which includes the mapped variable.

RondeauG (approved):
Looks good!

PR #1235, opened 2022-11-10
Add Dryness Index

Zeitsperre:
### Pull Request Checklist:
    - This PR advances progress on #355 

* Adds the agroclimate metric Dryness Index from Riou et al. 1994 (https://hal.inrae.fr/hal-02843898)
* Adds the changes to the atmosds dataset creation #1269 to the main development branch
* Updates `pytest-xdist` to use the `--dist=worksteal` scheduler

**Yes.** Testing datasets have been modified, so changes to expected results apply for all tests based on the ERA5 Canadian cities' dataset.

### Other information:

This PR lead to the discovery of aberrant values for evspsblpot in our testing data (https://github.com/Ouranosinc/xclim-testdata/issues/20)


xclim/indices/_agro.py
```diff
@@ -507,6 +508,133 @@ def cool_night_index(
...
+def dryness_index(
+    pr: xarray.DataArray,
+    evspsblpot: xarray.DataArray,
+    lat: xarray.DataArray | str | None = None,
+    wo: str = "200 mm",
+    freq: str = "YS",
+) -> xarray.DataArray:
+    """Dryness Index.
+
+    Mean minimum temperature for September (northern hemisphere) or March (Southern hemisphere).
```
coxipi, line 521: This description is for `cool_night_index`,  it doesn't seem related to `dryness_index`, right?

xclim/indices/_agro.py
```diff
@@ -507,6 +508,133 @@ def cool_night_index(
...
+    Warnings
+    --------
+    Dryness Index expects CF-Convention conformant potential evapotranspiration (positive up).
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+        Precipitation.
+    evspsblpot : xarray.DataArray
+        Potential evapotranspiration.
```
coxipi, line 533: Probably precising `Daily XYZ` could be good? Monthly could also work it seems, although you do a `.resample(freq='MS')`, so probably assuming it's daily?

xclim/indices/_agro.py
```diff
@@ -507,6 +508,133 @@ def cool_night_index(
...
+    adjustment_array_north = xarray.DataArray(
+        [0, 0, 0, 0.1, 0.3, 0.5, 0.5, 0.5, 0.5, 0, 0, 0],
+        dims="month",
+        coords=dict(month=np.arange(1, 13)),
+    )
+    adjustment_array_south = xarray.DataArray(
+        [0.5, 0.5, 0.5, 0.5, 0, 0, 0, 0, 0, 0, 0.1, 0.3],
+        dims="month",
+        coords=dict(month=np.arange(1, 13)),
+    )
```
coxipi, line 575: The 2004 paper says: 
```
In the Southern Hemisphere, k = 0.1 for October,
0.3 for November and 0.5 for the months from De-
cember to March.
```
whereas your array implies this starts in November. Is this just a typo, or maybe I don't understand correctly what `di_south` does below?

xclim/indices/_agro.py
```diff
@@ -507,6 +508,133 @@ def cool_night_index(
...
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+        Precipitation.
+    evspsblpot : xarray.DataArray
+        Potential evapotranspiration.
+    lat : xarray.DataArray or {"north", "south"}, optional
+        Latitude coordinate as an array, float or string.
+        If None, a CF-conformant "latitude" field must be available within the passed DataArray.
```
coxipi, line 536: In any case, method `.replace` will fail if the passed DataArrays don't have cftimes, correct? I think this is unsual for the data we work with? Maybe it should be said somewhere in the docstring too? Or add a conversion step if needed?

aulemahal: Woups sorry @Zeitsperre, I didn't know about that we I suggested it! 

There are hundred of ways to do this, I just came up with:
```
di_south['time'] = di_south.indexes['time'].shift(-6, 'MS')
```

xclim/indices/_agro.py
Zeitsperre: Woops, yes. That was a copy-paste. Will address.

xclim/indices/_agro.py
Zeitsperre: I think we expect all datasets passed into `indices` and `indicators` to be day-length. I can adjust the docstring though.

xclim/indices/_agro.py
Zeitsperre: It's a typo, for sure. I likely read the paper wrong. Will revise.

xclim/indices/_agro.py
Zeitsperre: I'll _replace_ the `replace` line with Pascal's suggestion!

xclim/indices/_agro.py
aulemahal: We do have options for other frequencies! See `StandardizedIndexes` in `atmos/precip.py`. It simply needs passing `src_freq=['D', 'MS']` in the base class.

Zeitsperre: Should I change the `Precip` instantiation to `StandardizedIndexes` then?

aulemahal: Haha, well it semantically makes no sense, but they are similar enough (same variables, context and input frequencies), so I'm ok with it!

coxipi: > I think we expect all datasets passed into `indices` and `indicators` to be day-length. I can adjust the docstring though.

Yes, at lot of indices have `Daily precipitation` in their docstring, therefore I was wondering if this is something we want to have uniformized. I was not 100% sure, because for some indices it would be ok to pass a monthly precipitation, but I guess this is more the exception than the rule.

coxipi: Regarding `Cool night index`, I see two things:
1. It's based on the following class "tree" (I mean a chain of inheritances, not sure what is the word): `Temp <- Daily <- ResamplingIndicator <- ...`. The class `Daily` defines `src_freq = "D"`, so this is fixed by the class used for this index.  
2. The `allowed_periods = "A"` concerns the resampling frequency, correct? I think this is fine for your needs here. 

Regarding point 1, I guess you can stick with the specific classes, and specify that the source variables are `daily` in the docstring.

Zeitsperre: I could also perform a resampling to daily before running through the calculation? This would allow for finer/smaller time steps.

coxipi: I know that hourly variables are sometimes used for potential_evapotranspiration computations  where a specific method uses hourly data. In this case, I think it would make sense to allow for hourly data, if there's a method related to it. Is this case here, could you use a different method with hourly data? If it's just a matter of convenience, I would suggest to let the users do the transformation `hourly -> daily` themselves. 

The point I was trying to make with monthly data is kind in the opposite direction. In the end, you ask for daily precipitations, but the function is only computing things with the monthly precipitations, correct? So in a sense, if you only have monthly precipitations at hand, you could still compute the `dryness_index` in principle, you just need the function to allow you to do that. Then again, maybe it's a very specific corner case, and users will generally have access daily precips anyways.

coxipi: I have the impression that the approach that reflects more broadly what is done in XClim in general is: Just assume users will input daily precips.

Then, for corner-cases, users have to think about:
* Users with hourly precips should resample before-hand, the docstring  would instruct them to input daily precips.
* Users with monthly precips could use the index anyways, as the monthly resampling `dryness_index` will do nothing. You just have to be aware of this.

Zeitsperre:
There are currently no tests for this indice due to issues with the testing dataset (https://github.com/Ouranosinc/xclim-testdata/issues/20). I'll be updating the affected netcdf file before finalizing this branch.

Thanks for the review and suggestions @coxipi and @aulemahal !

xclim/indices/_agro.py
```diff
@@ -506,6 +519,197 @@ def cool_night_index(
...
+
+    """
+    if parse_offset(freq) != (1, "A", True, "JAN"):
+        raise ValueError(f"Freq not allowed: {freq}. Must be `YS` or `AS-JAN`")
+
+    # Resample all variables to monthly totals in mm units.
+    evspsblpot = (
+        convert_units_to(evspsblpot, "mm", context="hydro").resample(time="MS").sum()
+    )
+    pr = convert_units_to(pr, "mm", context="hydro").resample(time="MS").sum()
```
aulemahal, line 635: This will only work if the variable has a correct `standard_name` (`water_potential_evaporation_flux`), otherwise `convert_units_to` will fail as the dimensionalities are wrong ([mass]/([time][length]^2)  vs [length]). In either case, `context='hydro'` does nothing as it doesn't implement this transformation.

As you can already assume correct inputs, but not the presence of a `standard_name` I suggest to perform the transformation explicitely:
```suggestion
    evspsblpot =  rate2amount(evspsblpot, out_units="mm").resample(time="MS").sum()
    pr = rate2amount(pr, out_units="mm").resample(time="MS").sum()
```

aulemahal (commented):
This looks quite good! Here are a few comments about units and long names, mainly. 
Also, I think this is missing a test ?

xclim/indices/_agro.py
Zeitsperre: Thanks!

xclim/indicators/atmos/_precip.py
```diff
@@ -173,6 +174,24 @@ class HrPrecip(Hourly):
     compute=indices.dry_days,
 )
 
+dryness_index = Precip(
+    title="Dryness index",
+    identifier="dryness_index",
+    units="mm",
+    long_name="Estimation of growing season humidity (precipitation minus adjusted evapotranspiration) for the period "
+    "of April to September (Northern Hemisphere) or October to March (Southern Hemisphere), "
+    "with initial soil moisture content set to {wo}",
```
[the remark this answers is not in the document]
Zeitsperre: I think the title is safe with "Dryness index", but the long name should be closer to the description, aiming to describe its contents. Thanks for the suggestion!

xclim/indicators/atmos/_precip.py
```diff
@@ -173,6 +174,24 @@ class HrPrecip(Hourly):
...
+    description="Estimation of growing season humidity (precipitation minus adjusted evapotranspiration) for the "
+    "period of April to September (Northern Hemisphere) or October to March (Southern Hemisphere), with initial soil "
+    "moisture content set to {wo} and an adjustment based on monthly precipitation and evapotranspiration limits.",
+    abstract="The dryness index is a characterization of the water component in winegrowing regions which considers "
+    "the precipitation and evapotranspiration factors without deduction for surface runoff or drainage. "
+    "Metric originally published in Riou et al. (1994). ",
```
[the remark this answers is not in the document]
Zeitsperre: I think "Growing season humidity" makes the most sense. I'm not sure why I added "Absolute" in the Returns there, but it doesn't belong. Will update.

xclim/indices/_agro.py
Zeitsperre: I'm running into a strange issue here. It appears that evspsblpot's units are not being recognized properly. From the ncdump:
```
        float evspsblpot(location, time) ;
                evspsblpot:_FillValue = NaNf ;
                evspsblpot:cell_methods = "time: mean within days" ;
                string evspsblpot:description = "Total potential evaporation thickness converted to mass flux using a water density of 1000 kg/m³." ;
                evspsblpot:long_name = "Mean daily potential evaporation flux" ;
                evspsblpot:original_long_name = "Potential evaporation" ;
                evspsblpot:original_variable = "pev" ;
                evspsblpot:standard_name = "water_potential_evaporation_flux" ;
                evspsblpot:units = "kg m-2 s-1" ;
                evspsblpot:coordinates = "lat lon"
```

Trace back:
```
../../core/indicator.py:845: in __call__
    outs = self.compute(**compute_das, **kwargs, **var_kwargs)
<boltons.funcutils.FunctionBuilder-54>:2: in dryness_index
    ???
../../core/units.py:976: in wrapper
    out = func(*args, **kwargs)
../../indices/_agro.py:632: in dryness_index
    evspsblpot = rate2amount(evspsblpot, out_units="mm").resample(time="MS").sum()
../../core/units.py:753: in rate2amount
    return _rate_and_amount_converter(
../../core/units.py:676: in _rate_and_amount_converter
    out = convert_units_to(out, out_units)
../../core/units.py:398: in convert_units_to
    out = source.copy(data=units.convert(source.data, source_unit, target_unit))
/home/tjs/mambaforge/envs/xclim/lib/python3.10/site-packages/pint/facets/plain/registry.py:911: in convert
    return self._convert(value, src, dst, inplace)
/home/tjs/mambaforge/envs/xclim/lib/python3.10/site-packages/pint/facets/context/registry.py:393: in _convert
    return super()._convert(value, src, dst, inplace)
/home/tjs/mambaforge/envs/xclim/lib/python3.10/site-packages/pint/facets/nonmultiplicative/registry.py:206: in _convert
    return super()._convert(value, src, dst, inplace)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pint.registry.UnitRegistry object at 0x7fa7aebca290>
value = array([[-0.13347669,  0.17290459,  0.16369012, ...,  0.07921789,
         0.03846444,  0.09547401],
       [ 0.3923424...789,  0.22684312],
       [ 0.02936513,  0.0195477 ,  0.01270458, ...,  0.00563113,
         0.00070897, -0.00178045]])
src = <UnitsContainer({'kilogram': 1, 'meter': -2})>
dst = <UnitsContainer({'millimeter': 1})>, inplace = False
check_dimensionality = True

~~~

>               raise DimensionalityError(src, dst, src_dim, dst_dim)
E               pint.errors.DimensionalityError: Cannot convert from 'kilogram / meter ** 2' ([mass] / [length] ** 2) to 'millimeter' ([length])
```

Why isn't `s-1` being found?

aulemahal: oups, my bad here. `rate2amount` is converting the kg m-2 s-1 into kg m-2, and then trying to convert to mm. This is a transformation that needs the `hydro` context turned on, but the  `rate2amount` function doesn't pass it.

```python3
evspsblpot =  convert_units_to(rate2amount(evspsblpot), "mm", context="hydro").resample(time="MS").sum()
pr = convert_units_to(rate2amount(pr), "mm", context="hydro").resample(time="MS").sum()
```
or (but I'm less sure here):
```python3
with units.context("hydro"):
    evspsblpot =  rate2amount(evspsblpot, out_units="mm").resample(time="MS").sum()
    pr = rate2amount(pr, out_units="mm").resample(time="MS").sum()
```

aulemahal: Or, maybe even better as it is completely explicit:
```python3
evspsblpot =  amount2lwethickness(rate2amount(evspsblpot), out_units="mm").resample(time="MS").sum()
pr =   amount2lwethickness(rate2amount(pr), out_units="mm").resample(time="MS").sum()
```

Zeitsperre: I'd opt for the first option (`Explicit is better than implicit.`). Will push changes to the function itself.

aulemahal (approved):

xclim/indices/_agro.py
```diff
@@ -506,6 +525,199 @@ def cool_night_index(
...
+    >>> from xclim.indices import dryness_index
+    >>> dryi = dryness_index(pr_dataset, evspsblpot_dataset, wo="200 mm")
+
+    References
+    ----------
+    :cite:cts:`tonietto_multicriteria_2004,riou_determinisme_1994`
+
+    """
+    if parse_offset(freq) != (1, "A", True, "JAN"):
+        raise ValueError(f"Freq not allowed: {freq}. Must be `YS` or `AS-JAN`")
```
coxipi, line 635: `freq` is not used in the index apart from this line which just checks that is has a correct value, correct? Was there plans to allow a different resampling frequency that those below? If not, is there a reason to keep `freq` as an argument? It's to be used like other functions that do allow a greater flexibility for resampling?

Zeitsperre: I think this has more to do with compatibility for `Indicators` that supply a `freq` argument than anything else. The value for this indicator is only valid for annual estimates, so I don't see any other values being allowed. There's already some handling for the Southern Hemisphere growing season, so that would provide even more reason to constrict this to `AS-JAN`.

I don't see any other `freq` values possible.

xclim/testing/tests/test_indicators.py
```diff
@@ -461,7 +461,11 @@ def test_all_parameters_understood(official_indicators):
             if param["kind"] == InputKind.OTHER_PARAMETER:
                 problems.add((identifier, name))
     # this one we are ok with.
-    if problems - {("COOL_NIGHT_INDEX", "lat")}:
+    if problems - {
+        ("COOL_NIGHT_INDEX", "lat"),
+        ("DRYNESS_INDEX", "lat"),
+        ("DRYNESS_INDEX", "wo"),
+    }:
```
coxipi, line 468: `wo`  is a `Quantified`, so should it not avoid the `InputKind` error? 

Also, this is less important and I'm just thinking outloud:  `lat` is not a `Quantified` so that it can be a `None` type?  Perhaps  `Quantified` ought to allow `None`? Another strategy could be to replace the use of `None` with an empty string for `lat`, but perhaps it's not  very clean and might as well keep this exception above.

xclim/indices/_agro.py
```diff
@@ -506,6 +525,199 @@ def cool_night_index(
...
+    ----------
+    pr : xarray.DataArray
+        Precipitation.
+    evspsblpot : xarray.DataArray
+        Potential evapotranspiration.
+    lat : xarray.DataArray or {"north", "south"}, optional
+        Latitude coordinate as an array, float or string.
+        If None, a CF-conformant "latitude" field must be available within the passed DataArray.
+    wo : str
+        The initial soil water reserve accessible to root systems [length]. Default: 200 mm.
```
coxipi, line 559:
```suggestion
    wo : Quantified
        The initial soil water reserve accessible to root systems [length]. Default: 200 mm.
```

xclim/testing/tests/test_indicators.py
Zeitsperre: I think what complicates things for `lat` is that it can optionally be a `str` instance as well (in the case where we supply a point with no `lat` information easily found, like station obs data). An empty string might raise an exception with the way CNI and DI are currently written.

`wo` here is not `optional` and must have a `units` value. It's hardcoded in the indice. Not sure if I follow why we would allow `None` here, unless there are other cases ?

coxipi: Ok, thanks, clear up things I was confused with in another PR!

coxipi:
Can't find the approve button (and I assume 1 approval is enough), but looks good to me!

coxipi (approved):

xclim/testing/tests/test_indicators.py
aulemahal: `Quantified` inputs must all be polymorphic so that in the example below `A_conv` can be used for arithmetic with xarray DataArrays. Which  means is must either be a string (quantity with units, `A_conv` will be a float) or a `DataArray` with units (`A_conv` will be a DataArray).

```python3

A : Quantified = <something>

A_conv = convert_units_to(A, 'units')
```

Sadly this does not allow `None`, we are left with this test-skipping for `lat`...

But indeed, `wo` shouldn't have to appear here!

Zeitsperre: > But indeed, `wo` shouldn't have to appear here!

Interesting. I have test failures otherwise. Bug?

coxipi: For the empty string for lat, I was indeed thinking to replace the condition that checks if `lat` is `None` and check if `lat` is an empty string instead, so it would mean some rewriting. I don't think this approach is used elsewhere in XClim though, that was just to explore possibilities to avoid the exception. But perhaps the exception is a lesser evil

PR #1252, opened 2022-12-05
More diags

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add 2 properties:`decorrelation_length` and `transition_probability`

no 

### Other information:
`decorrelation_length` was used on the small performance zones for the analysis of ESPO-G, but it does not work on info-crue data (ERA5-Land resolution over Québec).


xclim/sdba/properties.py
```diff
@@ -947,6 +1009,147 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+    group: str
+      Useless for now.
+
+    Returns
+    -------
+    xr.DataArray, [dimensionless]
+      Decorrelation length.
+
+    Notes
+    _____
```
Zeitsperre, line 1045: The checker here is suggesting that to underline the section, it requires "dashed lines". Very confusing error message. 
```suggestion
    Notes
    -----
```

juliettelavoie: Thanks!

xclim/sdba/properties.py
```diff
@@ -753,6 +753,68 @@ def _relative_frequency(
...
+    """
+    # mask of the ocean with NaNs
+    mask = ~(da.isel({group.dim: 0}).isnull()).drop_vars(group.dim)
+
+    today = da.isel(time=slice(0, -1))
+    tomorrow = da.shift(time=-1).isel(time=slice(0, -1))
+
+    ops = {">": np.greater, "<": np.less, ">=": np.greater_equal, "<=": np.less_equal}
+    t = convert_units_to(thresh, da)
+    cond = ops[initial_op](today, t) * ops[final_op](tomorrow, t)
```
aulemahal, line 801: For the comparing thing, I suggest using `xclim.indices.generic.compare`.

```suggestion
    # To be transported at the top
    from xclim.indices.generic import compare

    t = convert_units_to(thresh, da)
    cond = compare(today, initial_op, t) * compare(tomorrow, final_op, t)
```

xclim/sdba/properties.py
```diff
@@ -753,6 +753,68 @@ def _relative_frequency(
...
+    if group.prop != "group":
+        cond = cond.groupby(group.name)
+    out = cond.mean(dim=group.dim)
```
aulemahal, line 804: It is not often possible to use it normally in the `properties`, but this case is the point of the `Grouper` object:
```suggestion
    out = group('mean', cond)
```

xclim/sdba/properties.py
```diff
@@ -753,6 +753,68 @@ def _relative_frequency(
...
+    out = out.where(mask, np.nan)
+    out.attrs["units"] = ""
+    print("test")
```
aulemahal, line 807: `test`

aulemahal (commented):
Partial review for now: `decorrelation_length` is not made for a 17h50 task.  :sweat_smile:

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,147 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+      Decorrelation length.
+
+    Notes
+    -----
+    Calculating this property requires a lot of memory. It will not work with large datasets.
+    """
+    # save attrs
+    coords_attrs = {}
+    for c in da.coords:
+        coords_attrs[c] = da[c].attrs
```
aulemahal, line 1046: Could `xclim.sdba.utils.copy_all_attrs` be used for the same goal?

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,147 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+
+    if dims is None:
+        dims = [d for d in da.dims if d != "time"]
```
aulemahal, line 1049:
```suggestion
        dims = [d for d in da.dims if d != group.dim]
```

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,147 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+
+    dists, mn, mx = _pairwise_haversine_and_bins(
+        corr.cf["longitude"].values, corr.cf["latitude"].values
+    )
+
+    # fill bottom triangle
+    np.nan_to_num(dists, 0)
+    # TODO: do it in 2 shots instead?
+    dists = dists + dists.T  # - np.diag(np.diag(dists))
+    dists = dists.astype(int)
```
aulemahal, line 1061: Integer distances?

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,147 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+    corr = _pairwise_spearman(da, dims)
+
+    dists, mn, mx = _pairwise_haversine_and_bins(
+        corr.cf["longitude"].values, corr.cf["latitude"].values
+    )
+
+    # fill bottom triangle
+    np.nan_to_num(dists, 0)
+    # TODO: do it in 2 shots instead?
+    dists = dists + dists.T  # - np.diag(np.diag(dists))
```
aulemahal, line 1060: I think this could be more efficient if done in `sdba.nbutils._pairwise_haverinse_and_bins`?

```python3
@njit
def _pairwise_haversine_and_bins2(lond, latd, transpose=False):
    """Inter-site distances with the haversine approximation."""
    N = lond.shape[0]
    lon = np.deg2rad(lond)
    lat = np.deg2rad(latd)
    dists = np.full((N, N), np.nan)
    for i in range(N - 1):
        for j in range(i + 1, N):
            dlon = lon[j] - lon[i]
            dists[i, j] = 6367 * np.arctan2(
                np.sqrt(
                    (np.cos(lat[j]) * np.sin(dlon)) ** 2
                    + (
                        np.cos(lat[i]) * np.sin(lat[j])
                        - np.sin(lat[i]) * np.cos(lat[j]) * np.cos(dlon)
                    )
                    ** 2
                ),
                np.sin(lat[i]) * np.sin(lat[j])
                + np.cos(lat[i]) * np.cos(lat[j]) * np.cos(dlon),
            )
            if transpose:
                dists[j, i] = dists[i, j]
    mn = np.nanmin(dists)
    mx = np.nanmax(dists)
    if transpose:
        np.fill_diagonal(dists, 0)
    return dists, mn, mx
```

With 1000 points, I get 50 ms with this code, no matter if `transpose` is True or False, which is the same as with the previous version. And 65 ms with your version. I thought the difference would be larger, but here is my code now that I did the test...

xclim/sdba/properties.py
juliettelavoie: yes `copy_all_attrs(out,da)` would do the trick here.

but I find the doctring of `copy_all_attrs` a bit confusing... | read "Copy all attributes of ds to ref," as take attrs from ds and put them in ref, but the fonction does it the other way around.

aulemahal: Indeed! I can fix this in this PR.

aulemahal (commented):
Another run of comments.
I am going to play with it today, but it looks good! Thanks!

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,138 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+    # only keep points inside the radius
+    ds = ds.where(ds.distance < radius)
+
+    ds = ds.where(ds.distance2 < radius)
+
+    def _bin_corr(corr, distance):
+        """Bin and mean."""
+        return stats.binned_statistic(
+            distance, corr, statistic="mean", bins=bins
+        ).statistic
```
aulemahal, line 1086: I just tried the function and it works well, and faster than I expected! `binned_statistic` is a good find! Adapt my `spatial_correlogram` to use this, on the Haudenosaunee region, I went from 1m25 to 25s!

There is an issue with it however: it doesn't perform the nan-aware statistic. Points where `corr` is NaN will yield NaN bins. I think the issue gets hidden here with the use of `idxmin`, so you skip those bins already. However, in `spatial_correlogram`, it is obvious. In my temporary tests I did:
```python3
dists = dists.where(corr.notnull())
```
I'm not sure if this could be more efficient if done in numpy within `_bin_corr`.

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,138 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+            ds.corr,
+            ds.distance,
+            input_core_dims=[["_spatial2"], ["_spatial2"]],
+            output_core_dims=[["distance_bins"]],
+            dask="parallelized",
+            vectorize=True,
+            output_dtypes=[float],
+            dask_gufunc_kwargs={
+                "allow_rechunk": True,
+                "output_sizes": {"distance_bins": 100},
```
aulemahal, line 1101:
```suggestion
                "output_sizes": {"distance_bins": len(bins)},
```

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,138 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+
+    Distance from a grid cell where the correlation with its neighbours goes below the threshold.
+    A correlogram is calculated for each grid cell following the method from ``xclim.sdba.properties.spatial_correlogram``.
+    Then, we find the first bin closest to the correlation threshold.
+
+    Parameters
+    ----------
+    da: xr.DataArray
+      Data.
+    radius: int
```
aulemahal, line 1020:
```suggestion
    radius: float
```

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,138 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+        Radius (in km) defining the region where correlations will be calculate between a point and its neighbours.
```
aulemahal, line 1021:
```suggestion
        Radius (in km) defining the region where correlations will be calculated between a point and its neighbours.
```

xclim/sdba/properties.py
```diff
@@ -947,6 +1004,138 @@ def _spatial_correlogram(da: xr.DataArray, *, dims=None, bins=100, group="time")
...
+      Same as argument `bins` from :py:meth:`scipy.stats.binned_statistic`.
+      If given as a scalar, the equal-width bin limits from 0 to radius are generated here
+      (instead of letting scipy do it) to improve performance.
+    group: str
+      Useless for now.
+
+    Returns
+    -------
+    xr.DataArray, [dimensionless]
+      Decorrelation length.
```
aulemahal, line 1037:
```suggestion
    xr.DataArray, [km]
      Decorrelation length.
```

xclim/sdba/properties.py
```diff
@@ -753,6 +753,63 @@ def _relative_frequency(
...
+        Operation to verify the condition for the initial state.
+        The condition is variable {op} threshold.
+    final_op: {">", "gt", "<", "lt", ">=", "ge", "<=", "le", "==", "eq", "!=", "ne"}
+        Operation to verify the condition for the final state.
+        The condition is variable {op} threshold.
+    thresh : str
+        Threshold on which to evaluate the condition.
+    group : {"time", "time.season", "time.month"}
+        Grouping on the output.
+        e.g. For "time.month", the relative frequency would be calculated on each month, with all years included.
```
aulemahal, line 785:
```suggestion
        e.g. For "time.month", the transition probability would be calculated on each month, with all years included.
```

xclim/sdba/properties.py
juliettelavoie: oh you're right, I missed that!

aulemahal (approved):
This looks good!

PR #1256, opened 2022-12-13
Rain season index

### LINKED ISSUE #842: Proposal of a rain season group of climate indices
* xclim version: 0.25.0
* Python version: 3.7.6
* Operating System: LinuxMint 20.1

### Description
A group of climate indices related to rain season has been developed (currently available in /scen_workflow_afr/indices.py). The algorithm is currently being tested (using climate data from West Africa) and documented (link methods to scientific literature). These indices can be used to identify the impacts of climate change in countries where a monsoon is present. It could potentially be used with a snow `xarray.Dataset`.

### What I Did
A call to the developed function returns 4 `xarray.DataArray` instances : 'rainstart' (beginning of season), 'rainend' (end of season), 'raindur' (duration of season) and 'rainqty' (amount of precipitation received during season). Parameters allow adjusting detection rules to different countries or regions. 3 methods were implemented for the end of the season. The detection rules are summarized below:

rainstart : The season begins when the total amount of precipitation received (_P_total_) over a period of _D_start_ consecutive days is greater than a threshold value (_P_threshold_), and if there is no pause longer than _D_dry_ days within a period of _D_wet_ days since the beginning of the rain season. Optionally, the user can specify the search range for the first day of the season (between day _DOY_min_ and _DOY_max_).

rainend ('depletion' method): The season ends after a water column of height _P_threshold_ has completed evaporated, considering that evapotranspiration (a constant daily evaporation rate of _ETP_threshold_ or a `xarray.DataArray` of projected values can be provided) and precipitation both affect the height of the water column.

rainend ('event' method) : The season ends on the last day of a period of _D_threshold_ consecutive days during which the daily precipitation amount (_P_day_) is lower than a threshold value (_P_threshold_).

rainend ('total' method) : The season ends on the last day of a period of _D_threshold_ consecutive days during which the total amount of precipitation received (_P_total_) is lower than a threshold value (_P_threshold_).

rainend (any method) : Optionally, the user can specify the search range for the last day of the season (between day _DOY_min_ and _DOY_max_). If there are multiple rain seasons in a year (which means several 'rainseason' groups of indices), it is possible to impose an upper limit, which is based on the beginning of the following rain season. This allows to avoid an overlap between two seasons.

raindur: It is the difference between 'rainend' and 'rainstart', plus one.

rainqty: It is the total amount of precipitation received between 'rainstart' and 'rainend'.

It is possible that no value is found locally (a few cells) in the resulting `xarray.DataArrays`. That's not necessarily a problem, since interpolation can be used to replace 'nan' values (strategy used in scen_workflow_afr) or simply ignored when calculating statistics. Too many missing values could mean that the detection rules are not compatible with the climate in the studied region.

### What I Received
This section is not required.

coxipi:
### Pull Request Checklist:
    - This PR fixes  #842

* Adds a rain season index (useful notably in West Africa)

No


xclim/indices/run_length.py
```diff
@@ -638,6 +638,80 @@ def get_out(rls):
     return da.copy(data=out.transpose(*da.dims).data)
 
 
+def rle_with_holes(
+    da_start: xr.DataArray,
+    window_start: int,
+    index_start: str,
+    da_stop: xr.DataArray | None = None,
+    window_stop: int | None = None,
+    index_stop: str | None = None,
```
coxipi, line 647: Not sure if this a good idea to allow `None` here.  Maybe only having `dim` as a kwarg would be better.

xclim/indices/run_length.py
```diff
@@ -638,6 +638,80 @@ def get_out(rls):
...
+    dim: str = "time",
+) -> xr.DataArray:
```
coxipi, line 649: Not sure if this a good idea to allow `None` here.  Maybe only having `dim` as a kwarg would be better.

xclim/indices/_agro.py
```diff
@@ -922,7 +1240,7 @@ def dry_spell_frequency(
     Examples
     --------
     >>> from xclim.indices import dry_spell_frequency
-    >>> pr = xr.open_dataset(path_to_pr_file).pr
```
Zeitsperre, line 925: see above

xclim/indices/_agro.py
```diff
@@ -672,6 +673,159 @@ def water_budget(
...
+        First day of year when season can end ("mm-dd").
+    end_date_max: DayOfYearStr
+        Last day of year when season can end ("mm-dd").
+    freq : str
+      Resampling frequency.
+    coord : {None, "default", name of a DateTimeAccessor}
+        If `coord == "default"`, the function returns values along `dim` instead of indexes (`coord is None`).
+        If `dim` has a datetime dtype, `coord` can also be a str of the name of the
+        DateTimeAccessor object to use (ex: 'dayofyear').
+
```
coxipi, line 746: `coord`  can't be a `str | bool | None` like in `rl.season` (unless I add a new input kind.

In `rl.season`, there are 3 possibilities:
* `coord==False` or `coord is None` : output index, an integer
* `coord==True` : output the related time
* or, for example, more specific: `coord=="dayofyear"`

With a `str | bool`, I achieve the 3 cases above with:
* `coord is None` : output index, an integer
* `coord=="default"` : output the related time
* or, for example, more specific: `coord=="dayofyear"`

Is that an acceptable/clear workaround?

xclim/indices/_agro.py
```diff
@@ -672,6 +673,159 @@ def water_budget(
...
+        end = _get_end_first_run(pram.where(mask))
+        length = xarray.where(end.notnull(), end - start, pram[dim].size - start)
+
+        # changing index to time if requested
+        if coord:
+            crd = pram[dim]
+            if coord != "default":
+                crd = getattr(crd.dt, coord)
+            start = rl.lazy_indexing(crd, start)
+            end = rl.lazy_indexing(crd, end)
```
coxipi, line 817: See comment about `coord` above

xclim/indices/_agro.py
```diff
@@ -884,6 +885,159 @@ def water_budget(
     return out
 
 
+@declare_units(
+    pr="[precipitation]",
+    s_thresh_wet="[length]",
+    s_thresh_dry="[length]",
+    e_thresh_dry="[length]",
+)
+def rain_season(
```
Zeitsperre, line 894: This might be a big ask, but it would be nice to have a `note` with the breakdown of the formulae involved here.

Would `thresh_wet_start` be better than `s_thresh_wet`?

Zeitsperre (approved):

xclim/data/fr.json
```diff
@@ -1140,16 +1140,33 @@
     "title": "Température radiative moyenne",
     "abstract": "La température moyenne des rayonnements solaires et thermiques incidents sur un corps à l'extérieur."
   },
+  "RAIN_SEASON": {
+    "description": "Saison des pluies",
```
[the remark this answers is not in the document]
coxipi: Hmm, no idea. You mean there are already 3 other entries doing the job? I'll try without

xclim/indices/_agro.py
Zeitsperre: I think this is fine. You agree @aulemahal ?

Zeitsperre:
Rain season is just missing a translation!

coxipi:
> Rain season is just missing a translation!

That's the one I just removed, I've put it back

xclim/indices/_agro.py
aulemahal: Well, I agree that your annotation and "default" makes sense. However, I'm not sure if we really need to give this option to the user?
What are the use cases for something else that a `doy` ?
Disclaimer : I haven't looked at this PR more than this comment and section for now.

coxipi: Not really a big motivation, I just instinctively wanted to give as much possibility as in `rl.season`.  

I see by comparison that `frost_free_season_end` returns only a `doy`, same for `growing_season_end`, all the snow_indices, etc. So might as well follow your suggestion and make this uniform.

aulemahal: Well, I don't have such strong opinions about what the indice does, but for consistency, xclim indicators always return the same units (i.e. they don't depend on parameters). So if `coord` stays in the function, the indicator should at least inject "dayofyear".

coxipi: I changed it as you suggested, the function is already complicated enough, I am gladly removing some lines.

xclim/indices/run_length.py
```diff
@@ -639,6 +651,85 @@ def get_out(rls):
...
 
+def rle_with_holes(
+    da_start: xr.DataArray,
+    window_start: int,
+    da_stop: xr.DataArray,
+    window_stop: int,
+    index: str = "first",
+    dim: str = "time",
+) -> xr.DataArray:
+    """Generate modified run length function with run sequences that can contain holes.
```
coxipi, line 662: What the function does is actually more generic. What would correspond IMO to `rle_with_holes` would be with the additional constraint `da_stop = 1 - da_start`. In this way, it would be very similar to a season, except you allow to set the window of holes (instead of having the same window as the start condition).

Maybe finding a better name and maybe adding a function `rle_with_holes` simply setting `da_stop = 1 - da_start` would be interesting

xclim/indices/_agro.py
```diff
@@ -884,6 +885,171 @@ def water_budget(
...
+    """
+    # Unit conversion.
+    pram = rate2amount(pr, out_units="mm")
+    thresh_wet_start = convert_units_to(thresh_wet_start, pram)
+    thresh_dry_start = convert_units_to(thresh_dry_start, pram)
+    thresh_dry_end = convert_units_to(thresh_dry_end, pram)
+
+    # won't work for cftime?
+    last_doy = pram.isel(time=-1).time.dt.date.values.item().strftime("%m-%d")
+
```
coxipi, line 981: This only works for datetime? I need to verify this

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,31 @@ class HrPrecip(Hourly):
     cell_methods="time: sum over days",
     compute=indices.cold_and_wet_days,
 )
+
+rain_season = PrecipWithIndexing(
+    title="Rain season",
+    identifier="rain_season",
+    realm="atmos",
+    var_name=["start", "end", "length"],
```
aulemahal, line 628:
```suggestion
    var_name=["rain_season_start", "rain_season_end", "rain_season_length"],
```
I think these are less ambiguous.

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,31 @@ class HrPrecip(Hourly):
...
+    ],
+    description=[
+        "First step of a run where i) a sequence of {window_wet_start} days accumulated {thresh_wet_start}"
+        "of precipitations ii) followed by a sequence of {window_not_dry_start} days with no dry sequence, i.e. a sequence of {window_dry_start} days"
+        "with at least {thresh_dry_start} {method_dry_start}. The start of the season is on the last day of the first sequence i) and must be between {date_min_start} and {date_max_start}",
+        "Last day in a dry sequence after the start of the season, i.e.  a sequence of {window_dry_end} days"
+        "with at least {thresh_dry_end} {method_dry_end}. It must be between {date_min_end} and {date_max_end}",
+        "Number of steps of the original series in the season, between 'start' and 'end'.",
+    ],
+    units="",
```
aulemahal, line 643: `PrecipWithIndexing` has `src_freq='D'`, so we can safely attribute "days" as the units of `rain_season_length` :
```suggestion
    units=["", "", "days"],
```

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,31 @@ class HrPrecip(Hourly):
...
+    abstract="The rain season begins when two conditions are met: 1) There must be a number of wet days with"
+    "precipitations above or equal to a given threshold; 2) There must be another sequence following, where, for a given period in time, there are no"
+    "dry sequence (i.e. a certain number of days where precipitations are below or equal to a certain threshold). The rain season ends"
+    "when there is a dry sequence.",
```
aulemahal, line 647:
```suggestion
```
Better put it in the compute function so the latter is better documented.

xclim/indices/_agro.py
```diff
@@ -884,6 +885,171 @@ def water_budget(
...
+      Resampling frequency.
+
+    Returns
+    -------
+    rain_season_start: xr.DataArray, [dimensionless]
+    rain_season_end: xr.DataArray, [dimensionless]
+    rain_season_length: xr.DataArray, [dimensionless]
+
+    Summary
+    -------
```
aulemahal, line 961: "Summary" is not a valid numpy docstring section (https://numpydoc.readthedocs.io/en/latest/format.html#sections)

You could use :
```suggestion
    Notes
    -----
```

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,31 @@ class HrPrecip(Hourly):
     cell_methods="time: sum over days",
     compute=indices.cold_and_wet_days,
 )
+
+rain_season = PrecipWithIndexing(
+    title="Rain season",
+    identifier="rain_season",
+    realm="atmos",
+    var_name=["start", "end", "length"],
+    standard_name=["rain_season_start", "rain_season_end", "rain_season_length"],
```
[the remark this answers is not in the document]
coxipi: Ok, just to confirm, I don't _need_ to include a `standard_name`, right?

xclim/indicators/atmos/_precip.py
coxipi: Should I simply remove the current `Notes` with this?

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,30 @@ class HrPrecip(Hourly):
...
+        "with at least {thresh_dry_start} {method_dry_start}. The start of the season is on the last day of the first sequence i) and must be"
+        "between {date_min_start} and {date_max_start}",
+        "Last day in a dry sequence after the start of the season, i.e.  a sequence of {window_dry_end} days"
+        "with at least {thresh_dry_end} {method_dry_end}. It must be between {date_min_end} and {date_max_end}",
+        "Number of steps of the original series in the season, between 'start' and 'end'.",
+    ],
+    units="",
+    abstract="Start time, end time and length of the rain season, notably useful for West Africa (sivakumar, 1998). The rain season starts with "
+    "a period of abundant rainfall, followed by a period without prolonged dry sequences, which must happen before a given date."
+    " The rain season stops during a dry period happening after a given date",
```
coxipi, line 646: Simplified abstract. Should I translate this shorter abstract in french, or keep the longer one for the translation?

xclim/indicators/atmos/_precip.py
aulemahal: Good news, `Precip` also has `src_freq='D'` (through  `Daily`)!

xclim/indices/run_length.py
coxipi: Some input on whether this function does what we want would be good.

Usually, a season is something like:
da = [**1,1,1**, 0,1,**0,0,0**,1...]
We want a given window (say 3) to be true, then, once the opposite condition is found again the season ends.  The season starts at **1 1 1 ** and ends at **0 0 0**. 

You can see this as two run lengths being computed, one that find the sequence **1 1 1**, another that finds **0 0 0**.

I generalized this idea by decoupling these two run lengths searches by defining 
`da_start` and `da_stop`. In the simple example above, `da_stop = 1- da_start`

This can lead to weird situations:
`da_start = [1,1,1,1,1,1,1,1,1,0,0,0,0,0,0,0, .........], window_start = 1`
`da_stop = [0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0, ....], window_stop = 1`
will give: `run_lengths = [2,NaN, 0, ~ len(da_start)]`

The reason is that 
[1,1,1,**1**,1,1,1,1,1,0,0,0,0,0,0,0, .........]
has nothing to stop it once we pass the first and only stop. In this sense, it works much more like a season than a run_length, so maybe a change of name would be warranted.

This weird behaviour will likely not happen when `da_stop` and `da_start`  are related like `da_stop  = 1 - da_start` like in the simpler case.

Another choice I had made above: The run starts backs _after_ we have passed the stop, i.e. we obtain 
`[2,NaN, 0, ~ len(da_start)]`
and not
`[2,NaN, 1, ~ len(da_start) + 1]`

Again, this is a very specific corner case that may not play an important role in many cases, but if you know of any situation where this is important, we actually have an idea of what we want, that would be great.

For "verglas" for instance, the only generalization needed is really having `window_start` and `window_stop` I believe. The fact we have `da_stop = 1 - da_start` will make it so that there is no overlapping ambiguity such as above.

xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +620,31 @@ class HrPrecip(Hourly):
     cell_methods="time: sum over days",
     compute=indices.cold_and_wet_days,
 )
+
+rain_season = PrecipWithIndexing(
```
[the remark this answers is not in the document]
coxipi: I'm confused:
 *  `cold_and_wet_days` is  a  `PrecipWithIndexing`, but has no `indexer` injection
 *  `dry_spell_total_length` is a `Precip`, but has an `indexer` injection. 
 Am I right in thinking these two examples contradict what you just said? Or I'm missing something?

xclim/data/fr.json
```diff
@@ -1176,11 +1176,6 @@
     "title": "Température radiative moyenne",
     "abstract": "La température moyenne des rayonnements solaires et thermiques incidents sur un corps à l'extérieur."
   },
-  "RAIN_SEASON": {
-    "description": "Saison des pluies",
-    "title": "Saison des pluies",
-    "abstract": "La saison des pluies commence lorsque deux conditions sont réunies : 1) Il doit y avoir un certain nombre de jours humides avec précipitations supérieures ou égales à un seuil donné ; 2) Il doit y avoir une autre séquence suivante, où, pendant une période de temps donnée, il n'y a pas de séquence sèche (c'est-à-dire un certain nombre de jours où les précipitations sont inférieures ou égales à un certain seuil). La saison des pluies se termine lorsqu'il y a une séquence sèche"
-  },
```
Zeitsperre: Is this not needed as a base for the other `RAIN_SEASON.*` indicators?

coxipi: Yes, I was trying some things to make it work. I don't know how to test this locally?

coxipi: I've already added it back, but I still have the issue with ["", "", "day"]

coxipi: Hmm I still get incomplete translation:
```
FAILED xclim/testing/tests/test_locales.py::test_xclim_translations[fr] - Failed: 0 indicators are missing translations and RAIN_SEASON have incomplete translations for official locale `fr`.
```

xclim/testing/tests/test_locales.py
```diff
@@ -147,10 +147,13 @@ def test_xclim_translations(locale, official_indicators):
         for attrs, transattrs in zip(indcls.cf_attrs, trans["cf_attrs"]):
             if {"long_name", "description"} - set(transattrs.keys()):
                 is_complete = False
+                print({"long_name", "description"} - set(transattrs.keys()))
 
         if not is_complete:
             incomplete.append(indname)
+    import xarray as xr
 
+    # xr.DataArray(tempo).to_netcdf("/home/eridup1/test.nc")
```
Zeitsperre, line 156: I think this was committed in error.

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+        elif method_dry_start == "total":
+            da_stop = pram.rolling({"time": window_dry}).sum() <= thresh_dry_start
+            # equivalent to rolling forward in time instead, i.e. end date will be at beginning of dry run
+            da_stop = da_stop.shift({"time": -(window_dry - 1)}, fill_value=False)
+            window_dry = 1
+
+        # First and second condition combined in a run length
+        run_positions = (
+            rl.rle_events(da_start, 1, da_stop, window_dry, "first")
+            >= window_not_dry_start + window_wet_start
```
aulemahal, line 1016: I found it weird that `window_dry` was given in the signature.
```suggestion
    def _get_first_run_start(pram):
        pram = select_time(pram, date_bounds=(date_min_start, last_doy))

        # First condition: Start with enough precipitation
        da_start = pram.rolling({"time": window_wet_start}).sum() >= thresh_wet_start

        # Second condition: No dry period after
        if method_dry_start == "per_day":
            da_stop = pram <= thresh_dry_start
            window_dry = window_dry_start
        elif method_dry_start == "total":
            da_stop = pram.rolling({"time": window_dry_start}).sum() <= thresh_dry_start
            # equivalent to rolling forward in time instead, i.e. end date will be at beginning of dry run
            da_stop = da_stop.shift({"time": -(window_dry_start - 1)}, fill_value=False)
            window_dry = 1

        # First and second condition combined in a run length
        run_positions = (
            rl.rle_events(da_start, 1, da_stop, window_dry, "first")
            >= window_not_dry_start + window_wet_start
```

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+    def _get_first_run(run_positions, start_date, end_date):
+        run_positions = select_time(run_positions, date_bounds=(start_date, end_date))
+        first_start = run_positions.argmax("time")
+        return xarray.where(
+            first_start != run_positions.argmin("time"), first_start, np.NaN
+        )
+
+    # Find the start of the rain season
+    def _get_first_run_start(pram, window_dry=window_dry_start):
+        pram = select_time(pram, date_bounds=(date_min_start, last_doy))
```
aulemahal, line 999: Why `last_doy` ? Like you could have an incomplete daily series finishing in july, but still having a few full years.

2 ideas:
- Adapt `select_time` so it can accept incomplete tuples.
- Recompute `last_doy` each time.
```suggestion
        last_doy = pram.indexes["time"][-1].strftime("%m-%d")
        pram = select_time(pram, date_bounds=(date_min_start, last_doy))
```

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+    def _get_rain_season(pram):
+        start = _get_first_run_start(pram)
+
+        # masking value before  start of the season (end of season should be after)
+        start_ind = xarray.where(start.notnull(), start.astype(int), -1)
+        mask = xarray.where(
+            pram.time == pram.isel(time=start_ind).time, 1, np.NaN
+        ).ffill("time")
+        mask[{"time": start_ind}] = np.NaN
+        mask = mask.notnull()
```
aulemahal, line 1041:
```suggestion
    # Get valid integer indexer of the day after the first run starts
    start_ind = (start + 1).fillna(-1).astype(int)
    mask = (pram * np.NaN) 
    # Put "True" on the day of run start
    mask[{'time': start_ind}] = 1
    # Mask back points without runs, propagate the True
    mask = mask.where(start.notnull()).ffill('time')
    mask = mask.notnull()
```

I tried something to make this code clearer and shorter....

I assumed  `start + 1` is safe : the need for a stop condition makes it so that `start` can't be the last index. no ?

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+        return _get_first_run(run_positions, date_min_start, date_max_start)
+
+    # Find the end of the rain season
+    def _get_first_run_end(pram):
+        if method_dry_end == "per_day":
+            da_stop = pram <= thresh_dry_end
+            run_positions = rl.rle(da_stop, index="first") >= window_dry_end
+        elif method_dry_end == "total":
+            da_stop = pram.rolling({"time": window_dry_end}).sum() <= thresh_dry_end
+            run_positions = da_stop
```
aulemahal, line 1028:
```suggestion
            run_positions = pram.rolling({"time": window_dry_end}).sum() <= thresh_dry_end
```

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+        ).ffill("time")
+        mask[{"time": start_ind}] = np.NaN
+        mask = mask.notnull()
+        end = _get_first_run_end(pram.where(mask))
+
+        length = xarray.where(end.notnull(), end - start, pram["time"].size - start)
+
+        # converting to doy
+        crd = pram["time"]
+        crd = getattr(crd.dt, "dayofyear")
```
aulemahal, line 1048: Héhé copied from `run_length.py` :) ?
```suggestion
        crd = pram.time.dt.dayofyear
```

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+                "rain_season_length": length,
+            }
+        )
+        return out
+
+    out = pram.resample(time=freq).map(_get_rain_season)
+    for outd in out.values():
+        outd.attrs["units"] = ""
+    out["rain_season_length"].attrs["units"] = "days"
+    return out["rain_season_start"], out["rain_season_end"], out["rain_season_length"]
```
aulemahal, line 1065:
```suggestion
    out["rain_season_start"].attrs["is_dayofyear"] = np.int32(1)
    out["rain_season_end"].attrs["is_dayofyear"] = np.int32(1)
    return out["rain_season_start"], out["rain_season_end"], out["rain_season_length"]
```
This is a xclim-specific fix because there is no CF convention to specify the output is a doy. The `np.int32` is because some netCDF tools do not support int64 attributes.

aulemahal (approved):
Mostly comments on some details.
This is nice! I'm sorry I took so long to review it.

xclim/indices/_agro.py
```diff
@@ -882,6 +883,188 @@ def water_budget(
...
+                "rain_season_start": start,
+                "rain_season_end": end,
+                "rain_season_length": length,
+            }
+        )
+        return out
+
+    out = pram.resample(time=freq).map(_get_rain_season)
+    for outd in out.values():
+        outd.attrs["units"] = ""
```
aulemahal, line 1063:
```suggestion
    out["rain_season_start"].attrs["units"] = ""
    out["rain_season_end"].attrs["units"] = ""
```

xclim/indices/_agro.py
coxipi: Just a side comment:

>    # Eliminate negative values.
    pram = xarray.where(pram < 0, 0, pram)
    pram.attrs["units"] = "mm"
    
    
    This is legacy from Wf_africa, I should just remove this

coxipi: > Adapt select_time so it can accept incomplete tuples

This would be nice, I will take a look

coxipi: > Like you could have an incomplete daily series finishing in july, but still having a few full years.

This is meant to be used in a resampled computation, so I'm not sure if that concern will be important? In each resampled group, the last_doy could be different

aulemahal: Indeed! But the current code takes `last_doy` from the full time series, not the resampled group. (Line 983)

xclim/indices/_agro.py
coxipi: mayyybe

tests/test_indices.py
```diff
@@ -2720,6 +2720,56 @@ def test_snow_season_end(self, snd_series, snw_series):
...
+    elif result_type == "start_cond2_fails":
+        pr[{"time": slice(10, 10 + 7)}] = 0
+        out_exp = [np.NaN, np.NaN, np.NaN]
+    elif result_type == "end_cond_fails":
+        pr[{"time": 99 + 20 - 1}] = 5
+        out_exp = [3, np.NaN, 363]
+
+    # convert mm -> kg m-2 s-1
+    with xr.set_options(keep_attrs=True):
+        pr = pr / (60 * 60 * 24)
```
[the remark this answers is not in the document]
coxipi: I'm confused, why is this working in the first place? I will recheck / refactor

xclim/indices/_agro.py
coxipi: I think it's fine yes. Maybe we should add:
```
start_ind = xr.where(start_ind < len(pram.time), start_ind,  -1)
```
to be on the safe side. 

Also, I think: 
```
mask = mask.where(start.notnull()).ffill('time')
```
could just be 
```
mask = mask.ffill('time')
```
at this point, it is also non-NaN only at places where start is non-NaN because we used : `start_ind = (start + 1).fillna(-1).astype(int)`

tests/test_run_length.py
```diff
@@ -123,6 +123,52 @@ def test_rle(ufunc, use_dask, index):
...
+        da = da.chunk({"a": 1, "b": 2})
+
+    out = rl.rle_events(da != 0, 1, da == 0, 1, index=index).mean(["a", "b", "c"])
+
+    # Same output as rle when da_stop is the opposite of da_start and window_start == window_stop = 1
+    # But:
+    # - `rle_events` gives either length or NaN.
+    # - `rle` gives either length, NaN, _or_ 0
+    if index == "last":
+        expected = np.zeros(365) * np.nan
```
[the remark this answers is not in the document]
coxipi: Done

coxipi:
By the way, `rle` has either 0's or NaN's, and `rle_events` has only NaN's. I remember we discussed at some points that it doesn't matter, but it might be nice to uniformize things. In `rle_events` it made sense not to add back 0's as it would be an extra useless step. It would be nice to avoid this if possible with `rle` too, but many tests (symbolically since we do not differentiate 0 and NaN in this context) really will start to fail as a result if it is all NaNs

coxipi:
Ok, I tried to get a closer notation to what is already used in xclim. `events` is a form of generalized `season` (without constraint on time with date). 

e.g. imagine a season with window=3
[1,1,1,0,1,1,0,0,1,0,0,0]
The season would be the 9 first entries. It begins at element 0 because of the three 1's and ends at element 9 because of the three 0's. An `event==season` would be:
[1,1,1,1,1,1,1,1,1, 0,0,0]
i.e. 1's in the season sequence. 

With this in mind, should I change the name of `rle_events`? We already have `season_lengths`, so this could also `event_lengths`? Or does `rle_events` seems fine?

Zeitsperre:
I'm much more a fan of function names that clearly signal what it is that they do. `event_lengths` is generic yet clear and concise.

coxipi:
> I'm much more a fan of function names that clearly signal what it is that they do. `event_lengths` is generic yet clear and concise.

Ok sounds good. I just realize, one difference compared with `season_length` gives the length of the first run, just one number per time series.

Here, `event_lengths` would give something like:

[5,NaN,NaN,NaN,NaN,0,0,0, 3,NaN,NaN, ...]
the first event has length 5, the second event has length 3, etc.

Is this type of output ok for something we would call `event_lengths`? I initially called it `rle_events` because it is the same kind of output as `rle`. I see the advantage of clearer function names, I just want to make sure it's ok we call this `event_lengths` given the form of the output.

coxipi:
Yes. In the end, `rle_events` is almost useless. It's just a shortcut for:

```
events = extract_events(...)
out = rle(events)

# same as 
out = rle_events(...)
```

I thought of `rle_events` as accomplishing a similar objective as `season`, but as underlined above, it's a bit different

coxipi:
Sounds good, I'm dropping it for now and referring this PR in a new issue

Zeitsperre:
@coxipi @aulemahal Thanks so much for the work here!

PR #1258, opened 2022-12-13
Pin Tox and re-enable conda-tox in workflows

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Pins tox at v3 due to some really breaking bugs in the current version (released last week).
* Re-enables the conda-tox build using tox at v3
* fixes a broken link to a PDF in our references
* updates a notebook that made mention of an obsolete module

No. It fixes something that was breaking.


Zeitsperre:
This good for review!

aulemahal (approved):
Looks like a good (temporary ?) fix to me! Thanks!

aulemahal:
As I see the HDF Error is back, I suggest we remove the `git clone` line in `tox.ini`, as it didn't really solve it. We were simply lucky I guess.

aulemahal:
I have no idea if this is related, but I noticed that we download the MD5 file even if the netCDF file already exists:
https://github.com/Ouranosinc/xclim/blob/939adcf46df944dd16d601b10cd3825cddccdf1c/xclim/testing/utils.py#L189-L196

This means we are still writing to disk each time `open_dataset` is called. Maybe there some non-thread-safe things happenning? (I have no idea how it could resolve in a HDF error though).

We could try to:
- Download the MD5 to memory instead than to a file?
- Use locks to prevent multiple write attemps.
?

Zeitsperre:
The thread problem is still there, but at least we have something that errors out LESS. I fear that the only way to fix this is to completely throw out most of the file gathering tools for `pooch`, and I don't want to do that just yet.

Will update history and merge.

PR #1278, opened 2023-01-20
Migrate from setup.py to pyproject.toml (PEP 517 and PEP 621)

Zeitsperre:
### Pull Request Checklist:
    - This PR addresses comments raised in https://github.com/pyOpenSci/software-review/issues/73

* `setup.py` is no longer used as the packaging system, `pyproject.toml` with [flit](https://flit.pypa.io/en/stable/) is now the packaging engine.
* several configurations for linting tools (including Manifest.in) have either been rendered obsolete or been migrated into the `pyproject.toml`
* TOML checkers are now present in pre-commit

#### What other changes need to happen?

- [x] package requirements need to be migrated into the package
- [x] user-end installation tests need to be performed
-  ~README and HISTORY require a pre-build step in order to properly be packaged and uploaded to PyPI. The ways this was done to date involved a hacky string replace step in `setup.py`. This might now need to be handled with a build-helper script~
    - [x] Drop the changelog from the PyPI index readme

Yes. Very much. Installation of the package should be the same on the user end, but development tooling has changed significantly. Documentation will need to be updated accordingly.

### Other information:
https://chadsmith.dev/python-packaging/


pyproject.toml
```diff
@@ -0,0 +1,191 @@
...
+append_only = true
+add_imports = "from __future__ import annotations"
+
+[tool.mypy]
+python_version = 3.8
+show_error_codes = true
+warn_return_any = true
+warn_unused_configs = true
+
+[[tool.mypy.overrides]]
```
bzah, line 129: According to [mipy doc  ](https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml), you can group the `ignore_missing_imports` in a single `tool.mypy.overrides` entry: 
```
[[tool.mypy.overrides]]
module = [
    "xarray.*",
    "numpy.*"
]
ignore_missing_imports = true
```

xclim/indices/_synoptic.py
```diff
@@ -1,6 +1,8 @@
 # noqa: D100
 from __future__ import annotations
 
+from typing import Tuple
```
bzah, line 4: T|his import is probably as you can use `tuple` (lowercase) when defining types.

Zeitsperre: You're right. The new annotations are only enabled on xclim and I forget that we can use them here. Will modify.

pyproject.toml
Zeitsperre: Right. I read somewhere about how splitting them up is preferable? At this point in time, we aren't using mypy, so let's make this change.

docs/installation.rst
```diff
@@ -44,26 +44,29 @@ We also recommend using the subsetting tools in `clisops`_ (see: :doc:`clisops.c
...
 It can't be installed directly from pip or conda and has one complex dependency : `Eigen`_.
-Please refer to Eigen's and SBCK's docs for the recommended installation instructions.
-However, Eigen is available on conda, so one can do:
+Please refer to Eigen's and SBCK's docs for the recommended installation instructions. However, Eigen is available on conda, so one can do:
 
-.. code-block::
+.. code-block:: shell
 
    $ conda install -c conda-forge eigen pybind11
    $ pip install "git+https://github.com/Ouranosinc/SBCK.git@easier-install#egg=sbck&subdirectory=python"
```
aulemahal, line 67:
```suggestion
   $ pip install "git+https://github.com/yrobink/SBCK.git@master#egg=sbck&subdirectory=python"
```
The branch has been merged some time ago, we can use this PR to update the doc too!

docs/installation.rst
```diff
@@ -44,26 +44,29 @@ We also recommend using the subsetting tools in `clisops`_ (see: :doc:`clisops.c
...
 
-The last line will install SBCK>=v0.4.0 from Ouranos' fork which implements a small fix that allows this
+The last line will install ``SBCK>=v0.4.0`` from Ouranos' fork which implements a small fix that allows this
 direct installation from pip within a virtual environment.
```
aulemahal, line 70:
```suggestion
```

xclim/ensembles/_reduce.py
```diff
@@ -100,7 +100,7 @@ def kkz_reduce_ensemble(
 def kmeans_reduce_ensemble(
     data: xarray.DataArray,
     *,
-    method: dict = None,
+    method: dict,
```
aulemahal, line 103: Somewhere lower, there's a `if method is None` check with a default value.
Do I understand this change was to fix a typing error? If so, I suggest:
```suggestion
    method: dict | None = None,
```

aulemahal (approved):
Looks good to me! 
I have an unrelated suggestion and an alternative fix.

docs/installation.rst
aulemahal: This would require a change to the `opt-slow` workflow too, I couldn't suggest it as it was out of reach.

xclim/ensembles/_reduce.py
Zeitsperre: Ah, I was basing this off the numpy docstring. Will revert, thanks!

Zeitsperre:
@aulemahal test-py39-opt-slow is consistently giving me the following:
```
FAILED xclim/testing/tests/test_sdba/test_properties.py::TestProperties::test_spatial_correlogram - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 5 / 5 (100%)
Max absolute difference: 7.40815654e-05
Max relative difference: 4.549936e-07
 x: array([ 26.543211,  67.716254, 108.889298, 150.062341, 191.235384])
 y: array([ 26.543199,  67.716227, 108.889254, 150.062282, 191.23531 ])
```

Is this just a matter of adjusting tolerances, or should I be investigating further?

aulemahal:
argh.
Yes easiest would be to adjust the tolerances.

PR #1285, opened 2023-01-30
Mannwhitneyu

ludwiglierhammer:
Hi guys,

I  added the Mann-Whitney U-test to ``ensembles.change_significance`` and implemented ``realization`` as an argument.
See https://github.com/euro-cordex/py-cordex/issues/33

Greetings from Hamburg, Germany,
Ludwig


Zeitsperre:
@ludwiglierhammer

Hi Ludwig! Thanks for the contribution. The CI tests should be fixed momentarily, and we'll be reviewing your PR soon!

xclim/ensembles/_robustness.py
```diff
@@ -40,11 +41,13 @@ def change_significance(
...
         `fut` and `ref` must be of the same type (Dataset or DataArray). If they are
         Dataset, they must have the same variables (name and coords).
-    test : {'ttest', 'welch-ttest', 'threshold', None}
+    test : {'ttest', 'welch-ttest', 'mannwhitney-utest', 'threshold', None}
         Name of the statistical test used to determine if there was significant change. See notes.
     weights : xr.DataArray
         Weights to apply along the 'realization' dimension. This array cannot contain missing values.
         Note: 'ttest' and 'welch-ttest' are not currently supported with weighted arrays.
+    realization: str
+        Name of the realization dimension. If you don't have multiple cliamte realizations or models, chosse None.
```
huard, line 50:
```suggestion
        Name of the realization dimension. If you don't have multiple climate realizations or models, choose None.
```

xclim/ensembles/_robustness.py
```diff
@@ -152,12 +164,16 @@ def change_significance(
         p_change = kwargs.setdefault("p_change", 0.05)
 
         # Test hypothesis of no significant change
+        input_core_dims = [[realization, "time"], [realization]]
+        input_core_dims = remove_emptys_from_list(input_core_dims)
+        output_core_dims = [[realization]]
+        output_core_dims = remove_emptys_from_list(output_core_dims)
```
huard, line 170: I think the code would be easier to read with something like
```
if realization is None:
  input_core_dims=[['time'], []]
  output_core_dims=[[]]
else:
  ...
```

huard: or I wonder if it would make sense to create a dummy `realization` dimension in case it doesn't exist, so the internal code doesn't have to handle this special case.

ludwiglierhammer:
What do you think about returning 'changed'? 

We could add some robustness checks, too.
Something like [this]( https://github.com/IPCC-WG1/Atlas/blob/main/datasets-interactive-atlas/hatching-functions/hatching-functions.R) from IPCC.

ludwiglierhammer:
Or should we implement a new function to calculate robustness from `changed`. We can also implement more robustness methods in `robustness_coefficient` and make it more flexible.

xclim/ensembles/_robustness.py
```diff
@@ -228,6 +265,15 @@ def change_significance(
         f"Significant change was tested with test {test} with parameters {kwargs_str}."
     )
     das = {"fut": fut} if ref is None else {"fut": fut, "ref": ref}
+    changed.attrs.update(
+        description="Change significance. " + test_str,
+        units="",
+        test=str(test),
+        histroy=update_history(
+            f"change_significance(fut=fut, ref=ref, test={test}, {kwargs_str})",
```
aulemahal, line 273: You could use `xclim.core.formatting.gen_call_string` here I think!

xclim/ensembles/_robustness.py
```diff
@@ -228,6 +265,15 @@ def change_significance(
         f"Significant change was tested with test {test} with parameters {kwargs_str}."
     )
     das = {"fut": fut} if ref is None else {"fut": fut, "ref": ref}
+    changed.attrs.update(
+        description="Change significance. " + test_str,
+        units="",
+        test=str(test),
+        histroy=update_history(
```
aulemahal, line 272:
```suggestion
        history=update_history(
```

xclim/ensembles/_robustness.py
aulemahal: Haha, I just realized we didn't use it further down... Woups, I can change that.

xclim/ensembles/_robustness.py
```diff
@@ -247,7 +293,7 @@ def change_significance(
             **das,
         ),
     )
-    return change_frac, pos_frac
+    return changed, change_frac, pos_frac
```
aulemahal, line 296:
```suggestion
    return change_frac, pos_frac, changed
```
This is a breaking change, but I think the impact is lower if the order is unchanged. 
Also could you add something to the docstring about this new return value?

xclim/ensembles/_robustness.py
```diff
@@ -121,28 +126,37 @@ def change_significance(
     ...     delta, test="threshold", abs_thresh=2
     ... )
     """
+    if realization is None:
```
aulemahal, line 129: I agree with the dummy `realization` dimension method.

However, I think I would prefer to stick to xclim's current specification of  `realization` as the ensemble dimension. All functions in `xclim.ensembles` depend on this syntax. I feel introducing a `realization: str` argument to every one of them is overkill when one only needs to do: `ds.rename(my_ensemble='realization')`. The same way we always expect `time` as the time coordinate.

However, this is only my opinion and I am open to discussion!

aulemahal (commented):
Thanks @ludwiglierhammer !
This looks good.

I have a few suggestions and comments.

huard:
Hi Ludwig, I've made some modifications on my end, will clean and test and push them here for your consideration. In particular, I've removed the "realization" as suggested by Pascal, and suggest to return the pvals instead of changed, as it carries more information. A keyword would trigger returning pvals to avoir breaking the API.

xclim/ensembles/_robustness.py
huard: Agree to stick to the ensemble convention.

ludwiglierhammer:
@huard: thanks for your work. It makes sence to me.
I saw in your modifications that SciPy >= 1.8.0 is required when using the Mann-Whitney U-test. 
I changed the scipy version request from `string` to `parse_version` due to the new Scipy 1.1.0. release.

ludwiglierhammer:
Are you interested that I implement the Brown-Forsythe test too. This test is helpful if all input samples are from populations with equal variances e.g. such as annual temperature and/or precipitation variations.

huard:
Hi Ludwig, 
Thanks for the version fix ! 
I'm not familiar with it, but if your team uses the Brown-Forsythe test in climate analytics workflows, then yes ! Let's open another PR for it though. I'll merge this one right away.

huard (approved):

PR #1297, opened 2023-02-14
Upgrade to tox v4.0 and adjust local testing default configurations

### LINKED ISSUE #1295: Streamline local and remote development testing with tox/ GitHub CI
### Addressing a Problem?

There are three issues at hand here:

 - Presently, local development and CI tests are largely identical workflows thanks to tools like `tox`. While this reduces the need to update our development steps considerably (documentation and CI stages), there are instances where this is not well suited to local development. One such issue is the use of `pytest-cov`, which is helpful for CI and noisy for local developers.

- In order to speed up testing, in a recent release we adopted `pytest-xdist` as a means of distributing tests between processes. GitHub CI offers Linux VMs with two cores, so this effectively halves testing time needed (which is great!). This becomes an issue when running tests for individual tests locally, as regardless of the number of tests being run, workers need to be spun-up to meet this configuration (automatic distribution; 6 workers maximum). To add insult to injury, importing `xclim` remains something of a lengthy process (several seconds), despite advancements that have sped this up considerably.

- For us to maintain and ensure that the test are being run with the most recent testing data, the test data is synchronized with the main branch of `xclim-testdata` at every call of `pytest` by default. This adds a lot of delay to the running of tests as these files are always verified at every run.

### Potential Solution

Most of these configurations work very well for our CI configurations. With some tweaking, there are some ways to further streamline this but we seem to have a workable configuration for our test ensemble.

We need to define a set of configurations for local development, and this will neccessitate making some decisions:
- How do we want to enable distributed testing (`pytest-xdist`) when running tests locally?
  - Where should distributed testing configurations be housed?
- What logic/checks should we use to determine when we should perform a synchronization of the local testdata with remote?
  - Can we think of a better approach to gathering the testdata across workers?

Some approaches could be:
- Only run distributed testing configuration when the entirety of the test ensemble is run (i.e. when running tests with `tox`, or when running on CI).
- Using a different approach to copying over test data between workers (see: https://pytest-xdist.readthedocs.io/en/latest/how-to.html#specifying-rsync-dirs-in-an-ini-file)
- Only synchronizing the testing data with remote if the folder is not present.

Perhaps it would be good to define a an ideal set of configs for each use-case and seeing how best we can accomodate them?
 - CI Testing
 - Local Testing (full suite of tests)
 - Local Testing (individual tests)

### Additional context

_No response_

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1295 

* Revises two environment variables (`XCLIM_TESTDATA_BRANCH` and `XCLIM_PREFETCH_TESTING_DATA`) to better limit their scope and clarify their purposes.
* Upgrades `tox` configuration to use version 4.0 by default
* Moves some configurations out of  `pyproject.toml` to better facilitate local development
* Removes several geojson files that are no longer used (previously used for subset tools).

Yes. Environment variables and testing data have been modified.


HISTORY.rst
```diff
@@ -27,7 +39,8 @@ Internal changes
 * Coveralls GitHub Action removed as it did not support ``pyproject.toml``-based configurations. (:pull:`1278`).
 * Add a remark about how `xclim`'s CFFWIS is different from the original 1982 implementation. (:issue:`1104`, :pull:`1284`).
 * Update CI runs to use Python3.9 when examining upstream dependencies. Replace `setup-conda` action with `provision-with-micromamba` action. (:pull:`1286`).
-* Update CI runs to always use `tox~=3.0` and the `latest` Ubuntu images (now `v22.04`). (:pull:`1288`).
+* ~Update CI runs to always use `tox~=3.0` and the `latest` Ubuntu images (now `v22.04`). (:pull:`1288`).~
+   * CI and tox defaults now use `tox~4.0` and the `latest` Ubuntu images. (:pull:`1297`).
```
huard, line 43: This is a bit confusing, with versions 3 and 4.

huard (approved):
Thanks !

PR #1302, opened 2023-02-22
Fix: pre-commit conda-forge package

bzah:
Boring fix of pre-commit conda-forge package used in environment.yml.

There are two conda-forge packages for `pre-commit` I was afraid `pre_commit` (underscored) might be a fraud (that's a very common way of distributing malware, at least on npm), but it seems to just be the former version of pre-commit.

See: https://github.com/search?q=org%3Aconda-forge+pre_commit

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -43,7 +43,7 @@ dependencies:
     - nc-time-axis
     - netCDF4>=1.4
     - pooch
-    - pre_commit
+    - pre-commit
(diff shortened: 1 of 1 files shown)


Zeitsperre (approved):
Oh, I had no idea! We have a dependency scanner on xclim at the very least, but let's deal with this. Thanks!

PR #1311, opened 2023-02-28
Improve SPI performance

### LINKED ISSUE #1270: Xclim Standardised Precipitation Index always throws error
### Setup Information

- Xclim version: 0.39.0
- Python version:3.8.10
- Operating System:Linux

### Description

I'm attempting to calculate Standardised Preciptation Index using Xclim with two data arrays (reference period and projection). Even when specifying the following chunks: 

`hist = xr.open_dataarray("./historic.nc",chunks={"time":-1,"lat":50,"lon":50})`

The SPI indicator function always throws the following error: 
```
"The input data is chunked on time dimension and must be fully rechunked to"
                    " run `fit` on groups ."
                    " Beware, this operation can significantly increase the number of tasks dask"
                    " has to handle.",
```

From a quick look in the source code, this is due to the resample function in xarray not preserving the time based chunking when resampling i.e. it is chunked time:1, lat:XX, Lon:XX after resampling. Therefore the following code (from the source code) always throws an error and is extremely confusing. 

```python3
   if freq != "D":
        pr = pr.resample(time=freq).mean(keep_attrs=True)
        pr_cal = pr_cal.resample(time=freq).mean(keep_attrs=True)

        def needs_rechunking(da):
            if uses_dask(da) and len(da.chunks[da.get_axis_num("time")]) > 1:
                warnings.warn(
                    "The input data is chunked on time dimension and must be fully rechunked to"
                    " run `fit` on groups ."
                    " Beware, this operation can significantly increase the number of tasks dask"
                    " has to handle.",
                    stacklevel=2,
                )
                return True
            return False

        if needs_rechunking(pr):
            pr = pr.chunk({"time": -1})
        if needs_rechunking(pr_cal):
            pr_cal = pr_cal.chunk({"time": -1})
```

This also leads to the following Xarray warning here: 

`/home/ubuntu/.local/lib/python3.8/site-packages/xarray/core/indexing.py:1374: PerformanceWarning: Slicing with an out-of-order index is generating 95 times more chunks
  return self.array[key]`

This could also explain the increasing number of dask tasks generated by the SPI indicator function here as well as I think multiple rounds of chunking (one done by the user and one internally) can drastically increase the number of tasks and the movement of data around in RAM. 

Something to look out for. 

Let me know if there's something wrong with what I'm saying here, happy to discuss. 

Thanks!

### Steps To Reproduce

_No response_

### Additional context

_No response_

### Contribution

### Code of Conduct

### LINKED ISSUE #1416: xclim.indices.standardized_precipitation_index() return infinite values
### Setup Information

- Xclim version: 0.39.0
- Python version: 3.8.13 
- Operating System: Ubuntu 22.04.02

### Context

The standardized precipitation index function returns infinite values (np.inf) for some time periods in which the precipitation is maximum or minimum (see Figure). The input data (dataset_pr.pr) does not have any nan or fill/missing value. 

It seams that the function can not deal with some precipitation thresholds (maximum and minimum).

I am using the following code:

```
dataset = standardized_precipitation_index(
        pr=dataset_pr.pr,
        pr_cal=dataset_pr.sel(time=slice(datetime(1971, 1, 1), datetime(2010, 12, 31))).pr,
        freq="MS",
        window=6,
        dist="gamma",
        method="ML",
    )
```
Thank you very much in advance for the help.

Screenshot from 2023-07-07 13-11-54

### Code of Conduct

### LINKED ISSUE #1474: _agro.standardized_precipitation_evapotranspiration_index creates NaNs
### Generic Issue

* xclim version: 0.44.0
* Python version: 3.10.12
* Operating System: Windows 10 Enterprise 22H2

### Description
I am using ISIMIP chelsa-w5e5v1.0 30 arcsec resolution daily data from 1979 to 2016 (dimensions, time: 13880, y: 30, x: 57, bbox: [73.533, 18.967, 74.008, 19.217]) pr, tasmin and taxmax data to first calculate the pet (_xci.potential_evapotranspiration_) and the water budget (xci._agro.water_budget_). No values are lost in these calculations and there are no NaNs present. However, when I then use xci._agro.standardized_precipitation_evapotranspiration_index_ to calculate the SPEI, two things happen: 

1. Along the x axis, after a certain moment the SPEI values/pixels always become NaNs. These values do generally respond to the lowest SPEI values alongside the x-axis (in the west there is most rainfall due to orographic uplift, the eastern areas are substantially drier), however, there are lower water_budget values with other y coordinates that are not NaNs during other timesteps. Is it possible that at these pixels the algorithm is not able to fit distributions? If so, what would be a fix? 
2. When I change the window below 12 periodically (corresponding to roughly 12 months) 3-5 timesteps (months) become NaNs. And with a window of 12 the first 10 timesteps become NaNs. 

The settings for the function are as follows: 
```
wb_cal = water_budget.sel(time=slice('1981-01-01', '2010-01-01'))

xci._agro.standardized_precipitation_evapotranspiration_index(wb = water_budget, wb_cal = wb_cal, freq = "MS", window = 12, dist = 'gamma', method = 'APP')
```

### What I Received
For example, timestep 54 and 94:
image
image

And with a window of 6: 

image

Comparison of water budget and SPEI at the same timestep:

image
image

Does anyone know what could be going wrong?

### Code of Conduct

coxipi:
### Pull Request Checklist:
    - This PR fixes #1270 and fixes #1416 and fixes #1474

* Make SPI/SPEI faster
* fit params are now modular, can be computed before computing SPI/SPEI. This allows more options to segment computations and allow to obtain the fitting params if troubleshooting is needed.
* time indexing now possible
* `dist_method` now avoids `vectorize=True` in its `xr.apply_ufunc`. This is the main  improvement in SPI/SPEI. 
* Better document the limits of usage of standardized indices. Now standardized indices are capped at extreme values ±8.21. The upper bound is a limit resulting of the use of float64.

Yes.

* `pr_cal` or `wb_cal` will not be input options in the future: 

> Inputing `pr_cal` will be deprecated in xclim==0.46.0. If `pr_cal` is a subset of `pr`, then instead of:
`standardized_precipitation_index(pr=pr,pr_cal=pr.sel(time=slice(t0,t1)),...)`, one can call:
`standardized_precipitation_index(pr=pr,cal_range=(t0,t1),...)`.
If for some reason `pr_cal` is not a subset of `pr`, then the following approach will still be possible:
`params = standardized_index_fit_params(da=pr_cal, freq=freq, window=window, dist=dist, method=method)`.
`spi = standardized_precipitation_index(pr=pr, params=params)`.
This approach can be used in both scenarios to break up the computations in two, i.e. get params, then compute
standardized indice

I could revert this breaking change if we prefer. This was a first attempt to make the computation faster, but the improvements are now independent of this change. We could also keep the modular structure for params, but revert to `pr_cal` instead of `cal_range`. It's a bit less efficient when `pr_cal` is simply a subset of `pr`, because you end up doing resampling/rolling two times on the calibration range for nothing. When first computing `params`, then obtaining `spi` in two steps, then it makes no difference


xclim/indices/_agro.py
```diff
@@ -951,80 +953,63 @@ def standardized_precipitation_index(
             f"The method `{method}` is not supported for distribution `{dist}`."
         )
 
-    # calibration period
-    cal_period = pr_cal.time[[0, -1]].dt.strftime("%Y-%m-%dT%H:%M:%S").values.tolist()
+    # freq is None allows to use monthly precips
+    if freq is None:
+        group = "time.month"
```
coxipi, line 958: Probably it would be best to allow for any grouper type. group should be inferred from data in this case

xclim/indices/_agro.py
```diff
@@ -886,12 +886,11 @@ def water_budget(
 
 @declare_units(
     pr="[precipitation]",
-    pr_cal="[precipitation]",
 )
 def standardized_precipitation_index(
     pr: xarray.DataArray,
-    pr_cal: Quantified,
-    freq: str = "MS",
```
coxipi, line 894: We could add `params` as a possible input if keeping the possibility of user-inputted calibration is still wanted

xclim/indices/_agro.py
```diff
@@ -944,6 +950,8 @@ def standardized_precipitation_index(
     ----------
     :cite:cts:`mckee_relationship_1993`
     """
+    uses_params = params is not None
+    uses_range = cal_range is not None
```
Zeitsperre, line 967: Is there a possibility that a user could send in a variable that isn't "Truthy"? In all honesty, I kind of like this pattern, but you could probably get away with `if params:` below.

coxipi: Weird, `if params.any()` takes between 10 and 20 sec. It might be better if I stick to `if params is not None`

coxipi: I've removed `uses_range` for now, if you change idea on `uses_param` later just let me know!

xclim/indices/_agro.py
```diff
@@ -884,14 +885,63 @@ def water_budget(
...
+    params.attrs.update(
+        {
+            "Calibration period": str(cal_range),
+            "Distribution": dist,
+            "Method": method,
+        }
+    )
+    return params
+
+
```
coxipi, line 935: I'm still unsure about this structure. I wanted to get the fitting function out of `spi`. Before, `spi` had 2 output types, either I asked for `params`, or for actual `spi` values. Mixing output types (both unitless, but still conceptually different with different shapes for the output DataArray) is not ideal.

However, the functions that then need to get out are quite large. I'm not sure how I could clean this up, make it cleaner.

xclim/indices/_agro.py
```diff
@@ -923,123 +977,98 @@ def standardized_precipitation_index(
...
-                    " run `fit` on groups ."
-                    " Beware, this operation can significantly increase the number of tasks dask"
-                    " has to handle.",
-                    stacklevel=2,
-                )
-                return True
-            return False
-
-        if needs_rechunking(pr):
-            pr = pr.chunk({"time": -1})
```
coxipi, line 986: Before, I would rechunk everything, then compute parameters and compute spi values.

Now, I compute parameters, rechunk only the concerned subset, and then move on and compute spi values with the original chunked `pr`.

This removes warnings about out-of-index slices or something like this.  However, it greatly increases the number of dask tasks (from about ~50% of original SPI tasks to 150% of original SPI tasks)

coxipi:
I think the computation is still quite heavy. A few things should help. There is less redundant computation (we can either reuse params, or at least we don't resample/roll more or less twice (like it used to do for `pr` and `pr_cal`)).

The user can now dissect the computation more easily:
* Produce params before hand, reuse them for many computations
* Use of indexer to obtain SPI only for certain periods. This not like selecting, say month of June, from the onset. The indexing must be done  _after_ the rolling. Said another way, you still need months "Jan to June" even if you only want  SPI-6 for June. So the indexer must be used mid-computation, this is what is done.

xclim/indices/_agro.py
```diff
@@ -884,28 +885,97 @@ def water_budget(
...
 @declare_units(
     pr="[precipitation]",
-    pr_cal="[precipitation]",
+    params="[]",
 )
 def standardized_precipitation_index(
     pr: xarray.DataArray,
-    pr_cal: Quantified,
-    freq: str = "MS",
+    cal_range: tuple[DateStr, DateStr] | None = None,
```
coxipi, line 957: This could be two DateStr, `date1` and `date2`, if we want to avoid creating a new type.

xclim/indices/_agro.py
```diff
@@ -884,28 +885,97 @@ def water_budget(
...
+        if indexer != {}:
+            if da.isnull.all():
+                select_dims = {d: 0 for d in da.dims if d != "time"}
+                with xarray.set_options(keep_attrs=True):
+                    params = (
+                        fit(da.isel(time=slice(0, 2))[select_dims], dist, method)
+                        * da.isel(time=0, drop=True)
+                        * np.NaN
+                    )
+                return params
```
coxipi, line 942: I thought that for periods of time that have been cut by indexing, we should not really compute fitting parameters (or spi values later) we just fill in with NaN. I still run it through the fitting pipeline so that I have appropriate dimensions. I broadcast the result to the full `da`. 

Perhaps this logic could be used more widely or better formulated. I wonder if a trick like this could speed-up the computation for all-NaN slices, or if scipy already makes such checks.

xclim/indices/_agro.py
```diff
@@ -1178,10 +1069,118 @@ def _get_rain_season(pram):
...
+        "freq": freq,
+        "window": window,
+        "dist": dist,
+        "method": method,
+    }
+    return params
+
+
+# TODO : ADD
+"""
```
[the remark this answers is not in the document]
coxipi: Oh I see. Honestly, this particular note  doesn't really belong to this PR, but it's good to know in general!

coxipi:
Benchmarking already resampled data (1812 months x 30 lat x 60 lon): 
```
spi = xclim.indices.standardized_precipitation_index(
    da,
    da.sel(time=slice(datetime(1971, 1, 1), datetime(2010, 12, 31))),
    freq=None, 
    window=6, 
    dist="gamma", 
    method="APP",
    )
```
New : 14 s
Old : 4 min and still going

xclim/indices/_agro.py
```diff
@@ -1067,17 +1069,186 @@ def _get_rain_season(pram):
...
+NEW dist: GEV, log-normal
+https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.genextreme.html#scipy.stats.genextreme
+https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.lognorm.html
+"""
+
+
 @declare_units(
     pr="[precipitation]",
     pr_cal="[precipitation]",
+    params="[]",
```
coxipi, line 1239: Is this OK? 

`pr_cal` is maintained temporary as a xr.DataArray | None
`params` is Quantified without units, does this need a declare_units?

xclim/indices/_agro.py
```diff
@@ -1295,6 +1290,8 @@ def standardized_precipitation_index(
     * Supported statistical distributions are: ["gamma", "fisk"], where "fisk" is scipy's implementation of
        a log-logistic distribution
     * If `params` is given as input, it overrides the `cal_start`, `cal_end`, `freq` and `window`, `dist` and `method` options.
+    * The standardized index is bounded by ±8.21. 8.21 is the largest standardized index as constrained by the float64 precision in \
+      the inversion to the normal distribution.
```
Zeitsperre, line 1294:
```suggestion
    * The standardized index is bounded by ±8.21. 8.21 is the largest standardized index as constrained by the float64 precision in
      the inversion to the normal distribution.
```

CHANGES.rst
```diff
@@ -24,8 +24,10 @@ Internal changes
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-* Standardized indices (``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index``) are faster. (:issue:`1270` ,:pull:`1311`).
-* Standardized indicators can be computed in two steps: First compute fit parameters with ``xclim.indices.standardized_index_fit_params``, then use the output in the standardized indices functions.  (:issue:`1270` ,:pull:`1311`).
+* Standardized indices (``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index``)  (:issue:`1270`, :issue:`1416`, ,:pull:`1311`) were changed:
+    * More optimized
+    * Can be computed in two steps: First compute fit parameters with ``xclim.indices.standardized_index_fit_params``, then use the output in the standardized indices functions.
+    * The standardized index values are now clipped by ± 8.21.  This reflects the float64 precision of the computation when cdf values are inverted to a normal distribution.
```
Zeitsperre, line 30:
```suggestion
* Standardized indices (``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index``)  (:issue:`1270`, :issue:`1416`, :pull:`1311`) were changed:
    * More optimized and noticeably faster calculation.
    * Can be computed in two steps: First compute fit parameters with ``xclim.indices.standardized_index_fit_params``, then use the output in the standardized indices functions.
    * The standardized index values are now clipped to ± 8.21. This reflects the ``float64`` precision of the computation when cumulative distributed function values are inverted to a normal distribution.
```

xclim/indices/_agro.py
```diff
@@ -1221,16 +1221,11 @@ def wrap_cdf_ppf(da, params, group_idxs):
...
     )
-
-
-# TODO : ADD
-"""
-streamflow index (SSI)     GEV, log-logistic (Vincente-Serrano et al., 2012) Tweedie (Barker et al., 2016)    (Station de suivi)
-groundwater index (SGI)   log-normal, gamma, GEV (Bloomfield et Marchant, 2013)	                            (Station de suivi)
-NEW dist: GEV, log-normal
-https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.genextreme.html#scipy.stats.genextreme
-https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.lognorm.html
```
Zeitsperre: Should this information be placed elsewhere? Moved into an issue maybe?

xclim/indices/_agro.py
```diff
@@ -1247,6 +1420,21 @@ def standardized_precipitation_evapotranspiration_index(
...
+        Start date of the calibration period. A `DateStr` is expected, that is a `str` in format `"YYYY-MM-DD"`. Default option `None`
+        means that the calibration period begins at the start of the input dataset.
+    cal_end: DateStr | None
+        End date of the calibration period. A `DateStr` is expected, that is a `str` in format `"YYYY-MM-DD"`. Default option `None`
+        means that the calibration period finishes at the end of the input dataset.
+    params: xarray.DataArray
+        Fit parameters. The `params` can be computed using ``xclim.indices.standardized_index_fit_params`` in advance.
+        The ouput can be given here as input, and it overrides other options.
+    offset: Quantified
+        For distributions bounded by zero (e.g. "gamma", "fisk"): Water budget must be shifted, only positive values are allowed.
```
huard, line 1433:
```suggestion
        For distributions bounded by zero (e.g. "gamma", "fisk"), an offset must be added to the water budget to make sure there are no negative values. Keep the offset as small as possible to minimize its influence on the results. 
```

xclim/indices/_agro.py
```diff
@@ -1261,16 +1449,38 @@ def standardized_precipitation_evapotranspiration_index(
     -----
     See Standardized Precipitation Index (SPI) for more details on usage.
```
huard, line 1450: If results include NaNs, check that the offset parameter is larger than the minimum water budget values.

xclim/indices/_agro.py
```diff
@@ -1259,18 +1447,42 @@ def standardized_precipitation_evapotranspiration_index(
...
+        )
+        if params_offset != offset:
+            extra_msg = (
+                " (using default SPEI value 1 mm/d)" if uses_default_offset else ""
+            )
+            warnings.warn(
+                f"The offset in `params` differs from the input `offset`{extra_msg}."
+                "Procceding with the value given in `params`."
+            )
+            offset = params_offset
```
huard, line 1470: Why this, and not the offset explicitly given ?

xclim/indices/_agro.py
```diff
@@ -1067,17 +1069,185 @@ def _get_rain_season(pram):
...
+            vals[vals == 0] = np.NaN
+            paramsd = {param_names[jj]: pars[jj] for jj in range(len(pars))}
+            probs_of_zero = zeros / vals.shape[0]
+            dist_probs = scipy_dist.cdf(vals, **paramsd)
+            probs = probs_of_zero + ((1 - probs_of_zero) * dist_probs)
+            spi[indices] = norm.ppf(probs)
+        return spi
+
+    group_idxs = da.time.dt.month if group_name == "month" else da.time.dt.dayofyear
+    std_index = xarray.apply_ufunc(
```
huard, line 1212: Why can't this be a groupby.map ?

xclim/indices/_agro.py
```diff
@@ -1067,17 +1069,185 @@ def _get_rain_season(pram):
...
+        "group": group,
+        "time_indexer": indexer,
+        "units": "",
+        "offset": offset,
+    }
+    return params
+
+
+def _get_standardized_index(da, params, **indexer):
+    scipy_dist = get_dist(params.attrs["scipy_dist"])
```
huard, line 1188: Add docstring. It took me a while to understand what the function was doing.

huard (approved):
I have the feeling we could add features to some of the functions in stats.py that would reduce the complexity of the code here.

CHANGES.rst
```diff
@@ -6,10 +6,22 @@ v0.46.0
 --------------------
 Contributors to this version: Éric Dupuis (:user:`coxipi`).
 
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* Standardized indices (``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index``)  (:issue:`1270`, :issue:`1416`, :issue:`1474`, :pull:`1311`) were changed:
+    * Optimized and noticeably faster calculation.
+    * Can be computed in two steps: First compute fit parameters with ``xclim.indices.standardized_index_fit_params``, then use the output in the standardized indices functions.
+    * The standardized index values are now clipped to ±8.21. This reflects the ``float64`` precision of the computation when cumulative distributed function values are inverted to a normal distribution.
```
coxipi, line 14: There were two suggestions on the same line, I think the second one was dropped here, I will try to find it back

coxipi:
```suggestion
    * The standardized index values are now clipped to ±8.21. This reflects the ``float64`` precision of the computation when cumulative distributed function values are inverted to a normal distribution and avoids returning infinite values.
```

xclim/indices/_agro.py
coxipi: This is indeed a bit more complicated than it need to be since I give "1 mm/d" as a default value. This was done to avoid doing  a breaking change. I think it might be good to have a breaking change so that users are aware of this, but if this is not the final version of SPEI, maybe we should reserve a breaking change for the final formulation?

xclim/indices/_agro.py
coxipi: The previous formulation of those indices did rely on a `groupby.map`, but it resulted in a very slow computation. This convoluted writing is what allowed the great speed up observed.

coxipi: There is a similar procedure ``standardized_index_fit_params``, but there, I found no improvement in using this complicated formulation over a `groupby.map`, so I kept the simpler `groupby.map` there.

xclim/indices/_agro.py
```diff
@@ -1067,17 +1069,185 @@ def _get_rain_season(pram):
...
+    group_name = params.attrs["group"].rsplit(".")[-1]
+    param_names = params.dparams.values
+
+    def wrap_cdf_ppf(da, params, group_idxs):
+        if indexer != {} and np.isnan(da).all():
+            return np.zeros_like(da) * np.NaN
+
+        spi = np.zeros_like(da)
+        # loop over groups (similar to np.groupies idea, but np.groupies can't be used with custom functions)
+        for ii, group_idx in enumerate(list(dict.fromkeys(group_idxs).keys())):
```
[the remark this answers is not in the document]
coxipi: I think this can be done, yes. It would just require a tad bit more attention, to make sure that arrays are well sorted. But I think this was already a pre-requesite, I will double check, then 

```python
 for group_idx in (list(dict.fromkeys(group_idxs).keys()):
            pars = params[group_idx - 1]
```
or something even simpler for the loop should do the trick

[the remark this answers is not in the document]
huard: I thought params` keys where the group index.

xclim/indices/_agro.py
huard: Your call. My suggestion is really to say that a parameter in the function call is more explicit of user intent than a attribute in another parameter, so should override the later.

xclim/indices/_agro.py
huard: Ok. Just out of curiosity, do you understand where the speed up comes from?

xclim/indices/_agro.py
coxipi: Yes, I would agree. The problem is that to avoid a breaking change, I was using a default value for `offset`. In this situation, the parameter in the function call can either be something put in explicitly by the user. Given that may be replaced in the future, a strategy could be to use a dummy value by default (e.g. "UNSET") to track if a value was indeed given.

xclim/indices/_agro.py
coxipi: It might be for the same reason that. I had the intuition to do this because : 
1. I saw this part was the bottleneck ; 
2. I knew that `numpy.groupies` or `flox` can sometimes be faster than `groupby.map`. It's not possible to use the functions with my complex implementation, so I had to reproduce approximately what is done in these functions. 

In the end, I have indeed seen a good speedup. But why is this a faster way? Maybe this way of working avoids a costly regrouping step in the `groupby.map`, but I'm not sure. I think the answer is the same as why  `numpy.groupies` or `flox` can sometimes be faster than `groupby.map`

coxipi:
> I have the feeling we could add features to some of the functions in stats.py that would reduce the complexity of the code here.

I've thinking about this. One change could be to modify `fit` to allow this code block:
```
    def wrap_fit(da):
        if indexer != {} and da.isnull().all():
            select_dims = {d: 0 if d != "time" else [0, 1] for d in da.dims}
            fitted = fit(da[select_dims], dist, method)
            return fitted.broadcast_like(da.isel(time=0, drop=True))
        return fit(da, dist, method)

    params = da.groupby(group).map(wrap_fit)
 ```  
 in a single call. So `fit` would need to accept `group` and `indexer` arguments.
 
 Other than that, I find it hard to see how to avoid creating new functions "stats.py" to reduce the complexity.   The function `_get_standardized_index` really achieves: pdf -> cdf -> ppf using a given scipy dist and parameters for this function. This procedure should be agnostic to the previous processing (rolling and resampling) that usually takes in computing indices used in climate service.   This could be put directly in stats.py?

File: xclim/indices/_agro.py [hunk 20]
```diff
@@ -1265,6 +1467,21 @@ def standardized_precipitation_evapotranspiration_index(
...
+    cal_start: DateStr | None
+        Start date of the calibration period. A `DateStr` is expected, that is a `str` in format `"YYYY-MM-DD"`. Default option `None`
+        means that the calibration period begins at the start of the input dataset.
+    cal_end: DateStr | None
+        End date of the calibration period. A `DateStr` is expected, that is a `str` in format `"YYYY-MM-DD"`. Default option `None`
+        means that the calibration period finishes at the end of the input dataset.
+    params: xarray.DataArray
+        Fit parameters. The `params` can be computed using ``xclim.indices.standardized_index_fit_params`` in advance.
+        The ouput can be given here as input, and it overrides other options.
+    offset: Quantified
```
aulemahal, line 1479: Suggestion:
```suggestion
    offset: Quantified | None = None,
```

```python3
# pseudo-code
if offset is None:
  if params is not None:
      if dist in bounded_distributions:
          offset = 1000 mm / d
      else: 
        offset = 0
    else:
        offset = params.offset
else:
    if params is not None and params.offset != offset:
        warning
```

C'est à dire : put a default of None and clearly explain in the docstring what the default behaviour is. I know I just suggested something different for the snow thing, but that was in a "temporary" perspective!

aulemahal: Also, could there some option where `offset = -wb.min()` ?

xclim/indices/_agro.py
```diff
@@ -1079,17 +1081,214 @@ def _get_rain_season(pram):
...
+    window : int
+        Averaging window length relative to the resampling frequency. For example, if `freq="MS"`,
+        i.e. a monthly resampling, the window is an integer number of months.
+    indexer
+        Indexing parameters to compute the indicator on a temporal subset of the data.
+        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
+    """
+    _, base, _, _ = parse_offset(freq or xarray.infer_freq(da.time))
+    try:
+        group = {"D": "time.dayofyear", "M": "time.month"}[base]
```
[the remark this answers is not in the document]
coxipi: I think I wanted to allow "M" or "MS" when I first wrote the index, not "4MS" or similar. I thought "M" could still work, but indeed working with "MS" is cleaner / less confusion. You could imagine a standardized index for many time steps as "4MS". But indeed, SPI / SPEI and probably other definitions used in climate services use either daily ("1D") or monthly ("1MS") data as input, so we might as well restrict frequencies to that.

xclim/indices/_agro.py
```diff
@@ -1079,17 +1081,214 @@ def _get_rain_season(pram):
...
+    param_names = params.dparams.values
+
+    # transform `da` values to a cdf distribution with fitted parameters `params`
+    # then invert back to a normal distrubtion (ppf)
+    def wrap_cdf_ppf(da, params, group_idxs):
+        if indexer != {} and np.isnan(da).all():
+            return np.zeros_like(da) * np.NaN
+
+        spi = np.zeros_like(da)
+        # loop over groups (similar to np.groupies idea, but np.groupies can't be used with custom functions)
```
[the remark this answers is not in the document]
coxipi: I had tried but got a confusing error message leading me to believe this claim was maybe false. But I will retry, it would be worth it to clean the function.

xclim/indices/_agro.py
```diff
@@ -1118,123 +1331,108 @@ def standardized_precipitation_index(
...
+            method=method,
+        )
 
-    # Rolling precipitations
-    if window > 1:
-        pr = pr.rolling(time=window).mean(skipna=False, keep_attrs=True)
-        pr_cal = pr_cal.rolling(time=window).mean(skipna=False, keep_attrs=True)
+    # If params only contains a subset of main dataset time grouping
+    # (e.g. 8/12 months, etc.), it needs to be broadcasted
+    template = pr.groupby(params.attrs["group"]).apply(lambda da: da.isel(time=0))
```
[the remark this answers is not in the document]
coxipi: nice

xclim/indices/_agro.py
```diff
@@ -1079,17 +1081,214 @@ def _get_rain_season(pram):
...
+
+    if offset:
+        with xarray.set_options(keep_attrs=True):
+            da = da + convert_units_to(offset, da, context="hydro")
+
+    da, group = _preprocess_standardized_index(da, freq, window, **indexer)
+
+    def wrap_fit(da):
+        # if all values are null, obtain a template of the fit and broadcast
+        if indexer != {} and da.isnull().all():
```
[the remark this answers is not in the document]
coxipi: Good point about Dask. There is no way then to achieve a fastpath with lazy computing, no?

For the other question: Yes, indeed, spatial regions with only NaNs independently of time selecting would also benefit from this speedup (ignoring the Dask issue). I wanted to reserve this check for cases when I was sure there was a potential of all-NaN slices, e.g. when there was time-selecting. I'm not sure how costly it is to check .isnull.all, but probably very small in comparison to the whole algorithm.

xclim/indices/_agro.py
coxipi: Currently, we have a legacy default "1 mm/d", I was trying not to breaking things for now. Ideally:
* By default, offset = None
* If a dist bounded by 0 I chosen but not offset is given, raise an error

I will check, but I think actually, there is not a restriction in scipy to use gamma/fisk distrubutions with 0 bounded data. A parameter can also be fitted for this shift. This is was an idea for another PR though.

the problem with `offset = -wb.min()` is that you can become sensitive to what data is included. Imagine your reference data is 1980-2010 and you fit two cases:
1. 1980-2020
2. 1980-2050

You could have a different minimum in cases 1 &  2, so different offset. Ideally, since this is a trick, we would like that our computation are not too sensitive on this. But in any, it would be nice that the results of 1. are en exact subset of 2. if you use the same calibration data / methods for both computations.

coxipi: Ideally, I would eventually like to get rid of the default 1 mm/day default. It was just copied on monocongo's climate indices. It's a good rule of thumb, but users have found cases where a bigger offset was needed. Maybe we can keep it in but warn users about having NaNs?  Maybe just in the docstring. Not sure. I could check the R implementation to see if I can dig ideas.

coxipi:
I think it may be worthwhile to benchmark a bit the previous implementation, where is the bottleneck. The main culprit I believe is `dist_method`. For instance, let us first obtain some parameters.

```python
import xclim
import scipy

pr = open_dataset("sdba/CanESM2_1950-2100.nc").pr
freq, window, dist, method = "MS", 1, "gamma", "APP"
params = xclim.indices.stats.standardized_index_fit_params(
    pr.sel(time=slice("1950", "1980")), freq=freq, window=window, dist=dist, method=method
)
# broadcast params like pr
params = params.rename(month="time").reindex(time=pr.time.dt.month)
params["time"] = pr.time
```

Now, we either use a numpy/ scipy approach (fast): 

```python
# ~0.5 seconds
import xarray as xr
from xclim.indices.stats import get_dist


dist = get_dist(params.attrs["scipy_dist"])
def wrap_cdf(da, pars):
    return dist.cdf(da[:], *pars)

out1 = xr.apply_ufunc(
    wrap_cdf,
    pr.where(pr>0), 
    params,
    input_core_dims=[["time"], ["dparams","time"]],
    output_core_dims=[["time"]],
    vectorize=True,
)
out1.values
```

or we use `dist_method` (50 times slower)
```python
# about 25 seconds
from xclim.indices.stats import dist_method
out2 = dist_method("cdf", params, pr.where(pr > 0))
out2.values
```

Both methods give the same results. It remains to be seen if the speed-up persists as we scale things up, but this is already a weird result.

xclim/indices/stats.py
```diff
@@ -1,14 +1,19 @@
 """Statistic-related functions. See the `frequency_analysis` notebook for examples."""
 from __future__ import annotations
 
+import warnings
 from typing import Any, Sequence
 
+import flox.xarray as floxx
```
Zeitsperre, line 7: `flox` is currently listed as an optional dependency (it's not even listed as a dev dependency). Is there a way to make this import optional as well or are we officially adopting this here?

coxipi:
# Some more info on `dist_method` slowness

I think the performance issues really stem from the `input_core_dims` used in the `xr.apply_ufunc` call of `dist_method`. With my little example, I can reproduce the slowness of `dist_method`:

## like `dist_method`, about 25 s
```python
    def wrap_cdf(da, pars):
        return dist.cdf(da, *pars)

    out = xr.apply_ufunc(
        wrap_cdf,
        pr.where(pr>0), 
        params,
        input_core_dims=[[], ["dparams"]],
        output_core_dims=[[]],
        vectorize=True,
    )
```

while using "time" in the core dims goes much faster. This remains true if I try using dask etc., but maybe I'm not using this well. At this point, we could chunk the time dimension, but earlier steps in `stats` usually require unchunked time dimension. If using "time" as a core dimension is always appropriate in this scenario, this could be way to improve `dist_method` easily

xclim/indices/stats.py
coxipi: I didn't use flox in the end, didn't seem worth it

Zeitsperre: If there is a significant speedup that you think we could benefit from, feel free to add it in. Otherwise, what I would propose is opening an issue to discuss adopting `flox` as a "true" optional dependency (like `lmoments3`) and seeing where we could easily integrate it for analysis speedups.

Zeitsperre (approved):
Well done!

xclim/indices/stats.py
```diff
@@ -585,20 +587,204 @@ def dist_method(
...
+
+    def reindex_time(da, da_ref):
+        if group == "time.dayofyear":
+            da = da.rename(day="time").reindex(time=da_ref.time.dt.dayofyear)
+            da["time"] = da_ref.time
+            da = resample_doy(da, da_ref)
+        elif group == "time.month":
+            da = da.rename(month="time").reindex(time=da_ref.time.dt.month)
+            da["time"] = da_ref.time
+        return da.chunk({"time": -1})
```
aulemahal, line 766: Hum, are we rechunking even if dask was never used ?

```suggestion
        return da if not uses_dask(da) else da.chunk({"time": -1})
```

aulemahal (approved):
Two small comments, but it looks good!

xclim/indices/stats.py
coxipi: oups, I had those checks earlier, must have dropped them by accident

Zeitsperre:
Ignore the cancelled builds, merging when docs clear.

PR #1318, opened 2023-03-10
Remove deprecated indicators/indices, handle warnings 

### LINKED ISSUE #1281: `discharge_distribution_fit` has wrong  `_registry_id`
### Generic Issue

* xclim version: 0.40
* Python version: all
* Operating System: all

### Description

`xclim.land.fit` has the wrong `_registry_id`. It shows :  `xclim.core.indicator.DISCHARGE_DISTRIBUTION_FIT`. The `xclim.core.indicator` part should not be there.

This make finch skip it when building its processes. Easily fixable, but not so clean.

This is because `fit` is a subclass of the base Indicator class and thus is assigned to the module of that class. We usually create a dummy class in the correct file so that it gets assigned to the correct module.

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1281

* Deprecated Indicators have been removed:
  * first_day_below, first_day_above, snow_cover_duration, continuous_snow_cover_start, continuous_snow_cover_end
* Deprecated indices have been removed:
  * snow_cover_duration, continuous_snow_cover_start, continuous_snow_cover_end
* Most remaining deprecated calls to tests have been updated to new standards
* ~Indices that emit FutureWarnings through regular use have been identified:~
  * ~`snowfall_approximation` (`method='brown'`)~
  * ~`effective_growing_degree_days`~
  * See #1319 

Yes. This removes all currently-marked deprecated indices and indicators.

### Other information:

The volume of warnings emitted from `pytest` is significantly reduced. The only warnings now emitted are those coming from lower-level libraries, like `numpy` and `xarray`. This should make it easier to identify and handle these instances moving forward.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,12 +19,15 @@ Breaking changes
 * Many previously deprecated indices and indicators have been removed from `xclim` (:pull:`1318`), with replacement indicators suggested as follows:
     - ``xclim.indicators.atmos.first_day_above`` ->  ``xclim.indicators.atmos.first_day_{tn | tg | tx}_above``
     - ``xclim.indicators.atmos.first_day_below`` -> ``xclim.indicators.atmos.first_day_{tn | tg | tx}_below``
-    - ``xclim.indicators.land.snow_cover_duration`` -> ``xclim.indicators.land.snd_season_length``
-    - ``xclim.indices.snow_cover_duration`` -> ``xclim.indices.snd_season_length``
-    - ``xclim.indicators.land.continuous_snow_cover_start`` -> ``xclim.indicators.land.snd_season_start``
-    - ``xclim.indices.continuous_snow_cover_start`` -> ``xclim.indices.snd_season_start``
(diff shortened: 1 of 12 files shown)


aulemahal (changes requested):
I realized that we are emitting `DeprecationWarning` instead of `FutureWarning`. As we discovered elsewhere,  the former is not visible to the user by default... Which means, that all the deprecation we are removing here have not been properly announced to the users...

Is the HISTORY enough to justify removing them? Or do we want to wait one more version ?

At least, I think we should fix the issue here : https://github.com/Ouranosinc/xclim/blob/e0f67df7e1e07087999910b50aabc96a094cee09/xclim/core/indicator.py#L1355

Zeitsperre:
@aulemahal 

I think we would be fine effecting these changes. In some cases, the history has indicated that we are removing these functions as far back as v0.39, however, if we opt to wait until v0.43, I could remove all remaining deprecated indicators/indices as well.

If we want to mark this PR for v0.43, I'd open another PR to deal with the `FutureWarning` issue, and maybe address the usage warnings I found in the testing suite. What do you think?

huard (approved):

PR #1323, opened 2023-03-15
Fix last run resampling

### LINKED ISSUE #1321: Run length & resample not working for `last_run`
### Setup Information

- Xclim version: 0.40
- Python version: 3.10
- Operating System: Linux

### Description

@RondeauG was trying to compute the last run of "low-water" (étiage) in a given year and got an error mesage:

`ValueError: index must be monotonic for resampling`

### Steps To Reproduce

Running:
```python
rl.last_run(da, window, freq)
```
with any `freq` other than `None` yields the error.

### Additional context

If `da` is reversed in time (as it is when using `last_run`), then we can't use a resampling. 

A minimal example to reproduce the error could be:

```python
reversed_da = da.sortby(dim, ascending=False)
reversed_da.resample({"time": "YS"}).map(lambda da:da)
```

Note: This only affects the *new* behaviour allowed, 1) run_length ; 2) resample. This problem does not affect current indices with the given default `resample_before_rl = True`.

### Contribution

### Code of Conduct

coxipi:
### Pull Request Checklist:
    - This PR fixes #1321 

* `freq` other than `None` now works with `last_run` 

No


xclim/indices/run_length.py
```diff
@@ -451,11 +452,15 @@ def coord_transform(out, da):
...
+            da = da[{dim: slice(None, None, -1)}]
+        dmax_ind = da.argmax(dim=dim)
         # If `d` has no runs, dmax_ind will be 0: We must replace this with NaN
-        out = dmax_ind.where(dmax_ind != d.argmin(dim=dim))
-        out = coord_transform(out, d)
+        out = dmax_ind.where(dmax_ind != da.argmin(dim=dim))
+        if first is False:
+            out = da[dim].size - out - 2
+        out = coord_transform(out, da)
         return out
```
coxipi, line 464: I'm not saying this should be implemented as it is. Having a `first` variable used in `first_run` to switch between `first_run` and `last_run` modes is not ideal. I just wanted to show how this can be done, and get some ideas on how to organize things. Should I make a parent function to `first_run` and `last_run`, e.g. `first_or_last_run` ?

xclim/indices/run_length.py
```diff
@@ -465,16 +470,20 @@ def get_out(d):
...
             out = coord_transform(out, da)
 
+    # Last_run not working in this state
     elif ufunc_1dim:
+        if first is False:
+            da = da[{dim: slice(None, None, -1)}]
         out = first_run_ufunc(x=da, window=window, dim=dim)
+        if first is False:
+            out = da[dim].size - out - 2
         out = coord_transform(out, da)
```
coxipi, line 483: I just realized passing `index="last"` to `first_run_ufunc` should do the trick to make it work too, will test this soon

xclim/testing/tests/test_run_length.py
```diff
@@ -558,7 +558,7 @@ def test_run_with_dates_different_calendars(self, calendar, expected):
             .resample(time="AS-MAR")
             .map(rl.last_run_before_date, date="03-02", window=2)
         )
-        np.testing.assert_array_equal(out.values[1:], np.array(expected) + 1)
+        np.testing.assert_array_equal(out.values[1:], np.array(expected))
```
coxipi, line 561: The previous output of `last_run_before_date` was not what I would expect, and the modified implementation I have gives something else. So I modified the implementation of the test, but let me explain and we can decide.

The example has a run starting on march 1, which means doy=[61,60] on a [leap year, non-leap year]. The call to `first_run_after_"03-01"`  and `last_run_before_"03-02"`^[both boundary dates are included given how these functions are constructed] are meant to isolate this run. 
        - `First_run_after "03-01"` gives [61,60], which is the `expected` array
        - Now, I find that intuitively, `Last_run_before_"03-02"` should give the same answer. This was not like this before, because this output was compared to  `expected+1` instead.
        - With my current version of `Last run before "03-02"`, I also get [61,60], which is I think 
        what we should expect from the function last run. For a *same run* selected by `first_run_after_date` and  `last_run_before_date`, I think it would make sense to get the same answer.  Of course, there might be multiple runs possibles, so that the two algorithm will give different output.

coxipi: ... should have read the doc...
`last_run` : `Return the index of the **last item** of the last run of at least a given length.`
The function does what it claims to do. I could include an argument `index` ("first" or "last") in `_boundary_run` that lets the choice of first or last element.

Zeitsperre: That would be good, IMO. Breaking changes are fine so long as we can replicate past behaviour and results.

Zeitsperre: In order to be less confusing, you could try going with `position` or `attribute/value_position`, maybe?

I agree that the level of control is starting to make it a bit confusing.

Zeitsperre: That already sounds better!

CHANGES.rst
```diff
@@ -2,9 +2,9 @@
 Changelog
 =========
 
-v0.42.0 (unreleased)
---------------------
-Contributors to this version: Trevor James Smith (:user:`Zeitsperre`).
+0.42.0 (unreleased)
```
Zeitsperre: The `v` is needed!
```suggestion
v0.42.0 (unreleased)
```

CHANGES.rst
```diff
@@ -25,6 +25,10 @@ Breaking changes
 * following indices now accept the `op` argument for modifying the threshold comparison operator (:pull:`1325`):
     * ``snw_season_length``, ``snd_season_length``, ``growing_season_length``, ``frost_season_length``, ``frost_free_season_length``, ``rprcptot``, ``daily_pr_intensity``
 
+Bug fixes
+^^^^^^^^^
+* `last_run` nows works when `freq` is not `None` (:issue:`1321`, :pull:`1323`).
```
Zeitsperre, line 33:
```suggestion
* ``xclim.indices.run_length.last_run`` nows works when ``freq`` is not ``None``. (:issue:`1321`, :pull:`1323`).
```

CHANGES.rst
coxipi: Already fixed!

xclim/indices/run_length.py
```diff
@@ -399,13 +399,14 @@ def windowed_run_count(
...
 
-def first_run(
+def _boundary_run(
     da: xr.DataArray,
     window: int,
-    dim: str = "time",
-    freq: str | None = None,
-    coord: str | bool | None = False,
-    ufunc_1dim: str | bool = "from_context",
+    dim: str,
```
aulemahal, line 405: The docstring should be updated to mirror this change to no defaults!

aulemahal (approved):
This looks good! I agree with the choice not to open the options for now. If the need comes up, we could reconsider it.

PR #1356, opened 2023-04-28
Add test_timeseries

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add a function `test_timeseries`. The function takes in an array and the name of a variable (not indicator) and creates a DataArray or Dataset ready to be used for tests.
* It will also be useful for xscen tests.

no

### Other information:
I don't really know how pytest and fixtures work, but I tried something...
1. @RondeauG @aulemahal Is this what you wanted?
2. ~I put it in xclim.testing.helpers as discussed in the meetingthursday, but it seems like it is not found as a fixture. If I try to use it in `test_atmos.py`, it fails. If I put it in `conftest.py`, then I am able to use it in `test_atmos.py`. Should I move it or do something else to make it accessible ?
Should I do something like line 14 of test_sdba/conftest.py: `cannon_2015_dist = pytest.fixture(tu.cannon_2015_dist)`?~ -> this is fixed.
3. If the variable is not recognized, should I put "unknown" in the attrs or just put no attrs ?


xclim/testing/helpers.py
```diff
@@ -195,3 +199,42 @@ def add_example_file_paths(cache_dir: Path) -> dict[str]:
             ns[f"{variable}_dataset"] = ds.get(variable)
 
     return ns
+
+
+@pytest.fixture
+def test_timeseries():
+    def _test_timeseries(
+        values, variable, start="7/1/2000", units=None, freq="D", as_dataset=False
+    ):
```
aulemahal, line 208: I suggest `test_timeseries` as a simple function (your `_test_timeseries`) in this module.

Then we could either have a fixture in `conftest`:
```
from xclim.testing.helpers import test_timeseries

@pytest.fixture
def timeseries():
    return test_timeseries
```
(The "test_" prefix has a meaning for pytest, I wouldn't use it in a fixture)

OR

we simply import test_timeseries everywhere.

juliettelavoie: that works! thanks!

xclim/testing/helpers.py
```diff
@@ -195,3 +200,36 @@ def add_example_file_paths(cache_dir: Path) -> dict[str]:
...
+
+    else:
+        warnings.warn(
+            f'Variable {variable} not recognised. Attrs will filled with "unknown".'
+        )
+        # TODO: is it better to put this or just no attrs ?
+        attrs = {
+            a: "unknown"
+            for a in ["description", "standard_name", "cell_methods", "units"]
+        }
```
Zeitsperre, line 225: It depends on whether this is used in Indicators. I would opt to remove it for now and see what breaks.

xclim/testing/helpers.py
```diff
@@ -2,10 +2,15 @@
 from __future__ import annotations
 
 import os
+import warnings
+from importlib.resources import open_text
 from pathlib import Path
 
 import numpy as np
+import pandas as pd
+import pytest
```
Zeitsperre, line 11: Needed?

Zeitsperre (approved):
LGTM. Were you planning on replacing the existing timeseries fixtures with this? Might be a lengthy process but I can give you a hand.

juliettelavoie:
> LGTM. Were you planning on replacing the existing timeseries fixtures with this? Might be a lengthy process but I can give you a hand.

The initial goal of this PR was to create something that xscen could use for tests. It was not my plan to replace existing fixtures, but I/we can do it if it's something that people want/need.

juliettelavoie:
what about just having the `tas_series` fixture (and others) call `timeseries ?

Zeitsperre:
@aulemahal Just mentioned that as a much nicer approach to me. Let's go with that. I can effect that changes.

Zeitsperre:
This PR has the @RondeauG of approval

Issue #1352, opened 2023-04-18
some more climate indicators


ludwiglierhammer:
### Addressing a Problem?

Hi guys,

I would like to implement some more climate indicators:

- mean_precipitation: `mean(dim="time")` instead of `sum(dim="time")`; see [precip_accumulation](https://xclim.readthedocs.io/en/stable/_modules/xclim/indices/_multivariate.html#precip_accumulation)
- snowfall_frequency: see `Sfreq` in [Frei P., 2016](https://www.meteoswiss.admin.ch/dam/jcr:2695424d-351a-47ab-8da3-696a6870225b/SR101_Frei.pdf)
- snowfall_intensity: see `Sint` in [Frei P., 2016](https://www.meteoswiss.admin.ch/dam/jcr:2695424d-351a-47ab-8da3-696a6870225b/SR101_Frei.pdf)
- dry_spell_max_length: according to [hot_spell_max_length](https://xclim.readthedocs.io/en/stable/_modules/xclim/indices/_threshold.html#hot_spell_max_length)
- mean_daily_wind_speed: see `FG` from [ECA&D indices](https://www.ecad.eu/indicesextremes/indicesdictionary.php#9)
- maximum_daily_maximum_wind_speed: see `FXx` from [ECA&D indices](https://www.ecad.eu/indicesextremes/indicesdictionary.php#9)
- delayed_frost_days: we can use something like [frost_days](https://xclim.readthedocs.io/en/stable/indices.html#xclim.indices.frost_days) but with a `before_date` as in [last_spring_frost](https://xclim.readthedocs.io/en/stable/api.html#xclim.indicators.atmos.last_spring_frost). Is there already a function implemented in `xclim` whihc differentiate between the Northern and the Southern hemisphere? Let's say a `delayed_frost_day` at the Northern hemisphere is a frost day between the 1st of April and the 31st of June and a `delayed_frost_day` at the Southern is a frost day between the 1st of October and the 31st of December (random dates; just to illustrate)
- precpitation_amount_from_wet_days_above_threshold: Total precipitation amount with precipitation > {perc}th percentile on wet days (precipitation >= 1 mm)
- precpitation_fraction_from_wet_days_above_threshold: Precipitation fraction with precipitation > {perc}th percentile on wet days (precipitation >= 1 mm)

[...]

PR #1358, opened 2023-05-02
Simple indicators

ludwiglierhammer:
Hi guys,
this is my first PR regarding issue #1352.
It contains the implementation of the following indices:
- `snowfall_frequency`
- `snowfall_intensity`
- `mean_daily_windspeed` named `sfcWind_mean` according to `tg_mean`
- `maximum_daily_maximum_wind_speed` named `sfcWindmax_max` according to `tx_max`
- `mean_precipitation` named `precip_average` according to `precip_accumulation`
- `solid_precip_average` and `liquid_precip_average`

I also implemented a conversion from a snowfall flux (`prsn`) in `kg/m**2*s` to a snowfall flux in `mm/day`. I use this conversion to calculate `snowfall_frequency` and `snowfall_intensity`. In additon you you can find this conversion in `first_snowfall`, `last_snowfall` and `days_with_snow`. This allows to give thresholds in  both `kg/m**2*s`  and `mm/day`.

Unfortunately, I can't test the new code snippets in my local repository. I get this error message:

    fixture 'xdoctest_namespace' not found

Can you help me with this issue?

I'll add some more tests and adjust `CHANGES.rst`

Cheers, Ludwig


docs/references.bib
```diff
@@ -1894,3 +1894,19 @@ @article {sturm_swe_2010
       pages = "1380 - 1394",
       url = "https://journals.ametsoc.org/view/journals/hydr/11/6/2010jhm1202_1.xml"
 }
+
+@article {frei_snowfall_2018,
+    author =
+}
+
```
coxipi, line 1901:
```suggestion
```

xclim/indices/_threshold.py
```diff
@@ -1331,6 +1342,16 @@ def first_snowfall(
...
         Threshold precipitation flux on which to base evaluation.
+    snr : xarray.DataArray, optional
+        Snow density [kg/m^3].
+        Use this option if unit of low and/or high is mm/day
+        to convert prsn [kg m-2 s-1] to [mm day-1].
+        If `snr` is None `const` will be used for convertion.
+    const: Quantified
+        Constant snow density [kg/m^3]
+        `const` is only used if `snr` is None.
+        Use this option if unit of low and/or high is mm/day
```
coxipi, line 1353: "low and/or high" -> "thresh"

and similarly below

xclim/indices/_threshold.py
```diff
@@ -1340,10 +1361,18 @@ def first_snowfall(
...
+
     References
     ----------
     :cite:cts:`cbcl_climate_2020`.
     """
+    unit = str2pint("mm/day")
+    unit_thresh = str2pint(thresh)
+    if unit_thresh._units == unit._units:
+        prsn = prsn_to_mm_per_day(prsn, snr=snr, const=const)
     thresh = convert_units_to(thresh, prsn, context="hydro")
```
coxipi, line 1376: Similar problem here with `context="hydro"` and having `prsn` with units `mm day-1`. 

If we have `prsn [mm day-1]` and `thresh [kg m-2 day-1]`, then `thresh = convert_units_to(thresh, prsn, context="hydro")` will do a conversion that is not consistent with the choice of `const=312 kg m-3`. 

On the other hand, having both `prsn [mm day-1]` and `thresh [mm day-1]` means that we would use `prsn [mm day-1]` as an input in `prsn_to_mm_per_day`, so this is a concrete occurence of the problem outlined in my comment above.

coxipi: We could 
1) Restrict the use of the indicator `first_snowfall` to `prsn` with  [mass]/[area]/[time] units. (also, in this case,   `context="hydro" would not be necessary).
2) Allow `prsn` with  [length]/[time] units.  We should then figure out what we do when we have  `prsn [mm day-1]` and `thresh [kg m-2 day-1]`. We could convert  `prsn` with a similar function `prsn_to_kg_per_m2_per_day` (with some better  name)

Maybe taking `thresh` with units ` [kg m-2 day-1]`, so maybe it just complicates things for nothing. Or maybe it's really unusual to have `prsn` with `[mm day-1]` to start with, and sticking with option 1) would make sense. 

Anyways, I think we need to restrict use cases (1) or make sure every case is covered correctly (2).

xclim/indices/_threshold.py
```diff
@@ -1431,16 +1503,143 @@ def days_with_snow(
...
+    :cite:cts:`frei_snowfall_2018`
+    """
+    unit = str2pint("mm/day")
+    unit_thresh = str2pint(thresh)
+    if unit_thresh._units == unit._units:
+        prsn = prsn_to_mm_per_day(prsn, snr=snr, const=const)
+    thresh = convert_units_to(thresh, prsn, context="hydro")
+    cond = prsn >= thresh
+    mean = cond.resample(time=freq).mean(dim="time")
+    return xarray.where(mean == np.nan, 0, mean)
```
coxipi, line 1640: I think that this current function `snowfall_intensity` achieves what is expected from `snowfall_frequency` (you would only need an extra factor of 100), since you sum on the condition.
 
To get the intensity, you could mask  `prsn` and then sum on it
```suggestion
    cond = prsn >= thresh
    mean = prsn.where(cond).resample(time=freq).mean(dim="time")
    return mean.where(mean.notnull(), 0)
```

ludwiglierhammer:
> Hi Ludwig, thanks for the nice work.
> 
> I'm not sure what can cause the fixture problem, I cloned your PR and don't see a problem.
> 
> There has been some reorganization, the tests in xclim are not where they used to be. Maybe that's the root of the problem? To test a new feature, I activate xclim with `conda activate xclim` then update with
> 
> ```
> conda env update xclim && pip install ".[dev]"
> ```
> 
> Maybe this way of proceeding would solve your fixtures problem?

Many thanks! 
```
pip install ".[dev]"
```
solved my problem.

xclim/indices/_conversion.py
```diff
@@ -991,6 +992,45 @@ def snd_to_snw(
...
+
+    Notes
+    -----
+    The estimated mean snow density value of 312 kg m-3 is taken from :cite:t:`sturm_swe_2010`.
+
+    References
+    ----------
+    :cite:cts:`sturm_swe_2010`
+    """
+    prsn = convert_units_to(prsn, "kg/(m**2*day)", context="hydro")
```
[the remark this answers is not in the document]
ludwiglierhammer: Sound good to me. I tried your second approach.

coxipi:
In #1359, we discussed the `maximum_consecutive_*_days` is just a special case of `*_spell_max_length`. 

The same can be said of `sfcWindmax_max` and similar functions here (although it's a bit more abstract). The function `xclim.indices.generic.statistics` covers those cases. For instance, the indicator of `sfcWindmax_max` could be written as: 
```python
sfcWindmax_max = Wind(
...
    # compute=indices.sfcWindmax_max,
    compute=indices.generic.statistics,
    parameters={"reducer": "max"},
)
```
@Zeitsperre , what is your take on this: Should we indeed reduce the amount of needed indices to a minimum like this?

xclim/indices/_threshold.py
```diff
@@ -1311,10 +1316,17 @@ def first_day_temperature_above(
...
+    const="[mass]/[volume]",
+)
 def first_snowfall(
     prsn: xarray.DataArray,
     thresh: Quantified = "0.5 mm/day",
+    snr: xarray.DataArray | None = None,
+    const: Quantified = "312 kg m-3",
     freq: str = "AS-JUL",
 ) -> xarray.DataArray:
     r"""First day with solid precipitation above a threshold.
```
coxipi, line 1332: @Zeitsperre Could you give your insight on this. I realize that this index (and the next one) differs from how `pr` indices usually work, and I want to know if we want to allow this extra layer of complexity.

Indices with `pr` usually simply keep the units of `pr` and simply adapt the threshold if needed. This is OK since the conversion is always uniform, same density 1000 kg m-3 everywhere regardless. The case of `prsn` with possibly space-dependent density (`prsn`) is more complex. 

Should we:
1. Allow the conversion of `prsn` inside the index as things are now? 
2. Only let the index convert `thresh` with a constant conversion factor if needed as in `pr` indices. In this case, one would need to convert `prsn` if [length]/[time] units are desired.

Maybe the confusion also comes from the fact that [precipitation] can mean both [mass]/[area]/[time] and [length]/[time]. In case of actual `pr`, this is not so ambiguous, since the conversion factor is uniform. But for `prsn`, things are more complex. We were discussing how `prsn` should be reserved for [mass]/[area]/[time], and there should be another variable name when units are [length]/[time].

Zeitsperre: I can see how quickly this case is becoming complicated; I think we need to agree on some approaches for handling precipitation. The CF Conventions are pretty specific on what counts as Precipitation Flux (`pr`; [mass]/[area]/[time]) and Precipitation Amount (no official acronym; [mass]/[area]).

I don't like the idea of hiding conversions within an indice, especially if this conversion step might be used in multiple indices. We don't currently log these conversion steps into the metadata history either, making matters worse. I do think we should be able to support amounts and fluxes within an indicator/index, but how we handle the conversion needs to be clearly communicated (via warnings?) to the user.

The current behaviour to only use a rule-of-thumb liquid water equivalent conversion only if `snr` is not provided feels like a good approach, so I wouldn't change that. 

The more I think about it, the more I agree that we need to define two variables for snow amount and flux. Or at the very least have some code to handle the case of prsn as an amount.

Zeitsperre:
> @Zeitsperre , what is your take on this (and the similar discussion in #1359): Should we indeed reduce the amount of needed indices to a minimum like this or keep more specific indices? Also, we could have a generic function that cover indices like `XYZ_days`, but maybe too much abstraction is not what we aim, I'm not sure.

There are definitely a few key indices for temperature that we opted to not remove when we were doing our last major refactor/simplification, simply because they're _always_ needed (e.g. `t{n|g|x}_{min|mean|max}`). One reason to keep these around is that indicators can intelligently select the proper variable from a dataset if needed. @aulemahal showed a small application of this off yesterday.

Striking a balance between what is popular and what reduces redundancy is hard, but we should opt for more succinct indices that can be built from (generics) and more offerings of customized indicators.

coxipi:
@ludwiglierhammer Do you agree with the equivalences I'm making:

| xclim name | `snw`   (kg m-2 s-1)  | `swe`    (m)                           | `snd`    (m)             |
| ---------- | --------------------- | -------------------------------------- | ------------------------ |
| cfname     | `surface_snow_amount` | `lwe_thickness_of_surface_snow_amount` | `surface_snow_thickness` |


| xclim name | `prsn`   (kg m-2 s-1) | `lwe_prsn` (tentative)    (m s-1) | `pr_solid` (tentative)    (m s-1) |
| ---------- | --------------------- | --------------------------------- | --------------------------------- |
| cfname     | `snowfall_rate`       | `lwe_snowfall_rate`               | ???                               |

Do you know if `pr_solid` has cf_name?

coxipi:
> If I may : `snowfall_rate` doesn't exist I think. A [mass]/([area]*[time]) is a flux in CF's vocabulary, thus `prsn` would be `snowfall_flux`.
> 
> Indeed, I think a `snowfall_rate` would be [length]/[time], the `pr_solid` that doesn't exist, your `thickness_of_snowfall_rate` (I think here "thickness of rate" is the part that doesn't make sense in CF).

Oh thanks, got mixed up.  
`snowfall_flux` is the right expression with units kg m-2 s-1
When discussing the lwe, in this case we have:
`lwe_snowfall_rate` with units m s-1

coxipi:
@ludwiglierhammer After discussion with @Zeitsperre we thought it was better to keep conversions out of indices. Sorry for the extra work, but it was useful to show me how to write the flux2rate converter.

I defined `pr_solid` as `prsnd` for now, which I also call  "snowfall_rate" in the comment above. Would you agree this is the quantity that is computed with your converter (what you originally called `prsn_to_mm_per_day`)?

Since `prsn` and `prsnd` are non-trivially related, just like `snw` and `snd`, we thought it would be better to have well separated variable. It is still possible to use either `prsn` or its converted variable `prsnd` in indices because I kept [precipitation] as units and not something more specific, but you need to use an appropriated threshold with it. 

I phrased docstrings and variables names accordingly to the default thresholds you gave (some indices work with `prsn` and "snowfall_flux", some work with `prsnd` and "snowfall_rate".

ludwiglierhammer:
@coxipi:  Many thanks for your flux2rate converter. I think this is very helpful and can simplify the code.

> @ludwiglierhammer I would say: snw is to snd what prsn is to pr_solid, do you agree?

Yes, I'd agree with this.

> Do you know if pr_solid has cf_name?

Unfortunately, I couldn't find a cf_name too :disappointed:.

> @ludwiglierhammer After discussion with @Zeitsperre we thought it was better to keep conversions out of indices. Sorry for the extra work, but it was useful to show me how to write the flux2rate converter.

This is pretty understandable to me.  Nevertheless I have some comments on this:
- For all those snow indices (`first_snowfall`, `last_snowfall`, `snowfall_frequency` and `snow_intensity` ) I'd recommend to add a warning for the users if the unit dimensionalities of `prsn` and `thresh` do not match. Something like this:
```
thresh_dim = str2pint(thresh).dimensionality
thresh_units = str2pint(thresh)._units
prsnd_dim = unit2pint(prsn).dimensionalty
prsnd_units = unit2pint(prsn)._units

if  thresh_dim != prsnd_dim:
    warnings.warn(f"Dimension of prsnd {prsnd_dim} and thresh {thresh_dim} do not match.")
    warnings.warn(f"Note: thresh will be converted from {thresh_units} to {prsn_units}")
    warnings.warn(f"using the liquid water density of 1000 kg m-3.")
    warnings.warn(f"Before calling this function you can use the function xclim.indices.prsn_to_prsnd")
    warnings.warn(f"to convert snowfall flux ([mass]/[area]/[time]) to snowfall rate ([length]/[time])")
    warnings.warn(f"using snow density ([mass]/[volume]) or simply set thresh unit to prsn unit.")
```
- `snowfall_frequency` calls `days_with_snow`. `snowfall_frequency` uses `prsnd` as input but `days_with_snow` uses `prsn`. This yields some conversion inconsistencies. Could we change `days_with_snow` to something like this? Then all the snowflux indices would depend on `prsnd` (3e11 mm/day as high limit? converted 1e6 kg m-2 s-1 with 312 kg m-3):

```
def days_with_snow(
    prsnd: xarray.DataArray,
    low: Quantified = "0 mm/day",
    high: Quantified = "? mm/day",
    freq: str = "AS-JUL",
) -> xarray.DataArray:
```

coxipi:
>  warnings.warn(f"Note: thresh will be converted from {thresh_units} to {prsnd_units}")
    warnings.warn(f"using the liquid water density of 1000 kg m-3.")

This *won't* happen because we use `thresh = convert_units_to(thresh, prsnd)`, i.e. without the option `context="Hydro"`, no conversion with 1000 kg m-3. This can only change units (e.g. mm/day -> mm/s) but not dimensionality (e.g. kg m-2 s-1 -> mm/day). Therefore, if dimensionality of `thresh` and `prsnd` don't match, this will give an error. I implemented an additional warning about `prsn_to_prsnd`, thanks.

> Then all the snowflux indices would depend on prsnd (3e11 mm/day as high limit? converted 1e6 kg m-2 s-1 with 312 kg m-3)

I didn't see any mention of 1e6 kg m-2 s-1 in matthews_planning_2017, all thresholds seem already in cm.  In any case, 1e6 kg m-2 s-1 is just a very large number, basically the same as infinity in this context, yes? We could simply use 1e12 mm/day.

xclim/core/units.py
```diff
@@ -30,13 +30,14 @@
...
     "declare_units",
-    "flux_and_rate_converter",
+    "flux2rate",
     "infer_context",
     "infer_sampling_units",
     "lwethickness2amount",
     "pint_multiply",
     "pint2cfunits",
     "rate2amount",
+    "rare2flux",
```
Zeitsperre:
```suggestion
    "rate2flux",
```

ludwiglierhammer:
> > warnings.warn(f"Note: thresh will be converted from {thresh_units} to {prsnd_units}")
warnings.warn(f"using the liquid water density of 1000 kg m-3.")

> This won't happen because we use thresh = convert_units_to(thresh, prsnd), i.e. without the option context="Hydro", no conversion with 1000 kg m-3. This can only change units (e.g. mm/day -> mm/s) but not dimensionality (e.g. kg m-2 s-1 -> mm/day). Therefore, if dimensionality of thresh and prsnd don't match, this will give an error. I implemented an additional warning about prsn_to_prsnd, thanks.

Ah, ok, now I fully understand the option context="hydro". Honestly, sometimes just reading the code would be pretty helpful :smile:

docs/references.bib
```diff
@@ -1896,7 +1896,7 @@ @article {sturm_swe_2010
 }
 
 @article {frei_snowfall_2018,
-    auhtor = {Frei, P. and Kotlarski, S. and Liniger, M. A. and Sch\"ar, C.},
+    author = {Frei, P. and Kotlarski, S. and Liniger, M. A. and Sch\"ar, C.},
```
Zeitsperre, line 1899:
```suggestion
    author = {Frei, P. and Kotlarski, S. and Liniger, M. A. and Schär, C.},
```

Zeitsperre: AFAIK UTF-8 is allowed in BibTeX

ludwiglierhammer:
According to `tg_*` and `tx_*`, I added `sfcWind_max`, `sfcWind_min`, `sfcWindmax_mean` and `sfcWindmax_min`.

coxipi:
I discussed in #1359 about avoiding breaking changes, I did not realize I made some breaking changes myself when editing this PR, I thought I was working with new indices/indicators.

For instance:
1. Indicators name change: `snw_to_snd` -> `snd_from_snw` and vice-versa
2. Thresholds `low, high` -> `thresh_min, thresh_max` 

These are also non-necessary changes that simply uniformize things a bit. I feel in this case this might pass since the concerned indices are more recent in XClim (I think?), but still same question here @Zeitsperre , are those changes accepted or not?

Zeitsperre:
> 1. Indicators name change: `snw_to_snd` -> `snd_from_snw` and vice-versa

I think if we really want to tackle this properly, we should do something about some of the existing converter names (like `sfcWind_2_uas_vas`). The idea of always going from one thing `to` another thing makes the most sense, but that's just me. I say that with the knowledge that I may have already broken that rule (with `from`). It would be swell to fix it all in one fell swoop (and provide aliases with warnings, as we normally do with major breaking changes).
 
> 2. Thresholds `low, high` -> `thresh_min, thresh_max`

I would absolutely change indices that are tied to a specific variable to use `thresh`, but for those under `generic`, I would keep them as is with `low` and `high` (there are a couple). It further differentiates that the generic indices should not be used directly by end-users. 
 
> These are also non-necessary changes that simply uniformize things a bit. I feel in this case this might pass since the concerned indices are more recent in XClim (I think?), but still same question here @Zeitsperre , are those changes accepted or not?

See my comments about breaking changes in the other PR. I like the idea of uniformity, and we've been working towards that for quite some time.

coxipi:
> I would absolutely change indices that are tied to a specific variable to use thresh, but for those under generic, I would keep them as is with low and high (there are a couple). It further differentiates that the generic indices should not be used directly by end-users.

That's what I had in mind, good.

> I think if we really want to tackle this properly, we should do something about some of the existing converter names (like sfcWind_2_uas_vas). 

Ok. On top of my head, here are a few different conventions I remember
* in low level files like "units.py" : `rate2amount`
* More high level : `snw_to_snd`
* the weird in-between I introduced : `sfcWind_2_uas_vas`

Do we want eventually to use `_to_` everywhere?  

> The idea of always going from one thing to another thing makes the most sense, but that's just me. I say that with the knowledge that I may have already broken that rule (with from). It would be swell to fix it all in one fell swoop (and provide aliases with warnings, as we normally do with major breaking changes).

Got it, yes I was following the names of some indicators with `_from_`, but I'll revert to `_to_` notation.

xclim/indices/_threshold.py
```diff
@@ -1399,48 +1430,141 @@ def last_snowfall(
...
+    low="[precipitation]",
+    high="[precipitation]",
+)
 def days_with_snow(
-    prsn: xarray.DataArray,  # noqa
-    low: Quantified = "0 kg m-2 s-1",
-    high: Quantified = "1E6 kg m-2 s-1",
+    prsn: xarray.DataArray,
+    low: Quantified = "0 mm/day",
+    high: Quantified = "1e12 mm/day",
```
coxipi, line 1445:
```suggestion
    low: Quantified = "0 kg m-2 s-1",
    high: Quantified = "1E6 kg m-2 s-1",
```

coxipi:
Old snow indices now target primarily `prsn` as they used to. Default behaviour now uses context="hydro" if conversion is needed, but I added a warning. The new indices are written in terms of `prsnd` as it was the motivation to include these indices. We could also write more similarly to other snowfall indices. Anyways, currently, `prsnd` can be given as input, you just need to give an appropriate `thresh`. I think it's a good balance between avoiding breaking changes, but allowing new functionality.

ludwiglierhammer:
> I reverted many changes I had made without realizing they were breaking changes. Old snow indices now target primarily `prsn` as they used to. Default behaviour now uses context="hydro" if conversion is needed, but I added a warning. The new indices are written in terms of `prsnd` as it was the motivation to include these indices. We could also write more similarly to other snowfall indices. Anyways, currently, `prsnd` can be given as input, you just need to give an appropriate `thresh`. I think it's a good balance between avoiding breaking changes, but allowing new functionality.

I think it's a good balance too. Can you imagine to add those breaking changes in further releases?

ludwiglierhammer:
The unit of default threshold values for prsn depending indices (`first_snowfall`, `last_snowfall` and `days_with_snow`) is `mm/day`. Since the CF unit of `prsn` is `kg m-2 s-2` I would recommend to set the thresholds' units to `kg m-2 s-2` as well. Do you agree?

coxipi:
> The unit of default threshold values for prsn depending indices (`first_snowfall`, `last_snowfall` and `days_with_snow`) is `mm/day`. Since the CF unit of `prsn` is `kg m-2 s-2` I would recommend to set the thresholds' units to `kg m-2 s-2` as well. Do you agree?

Hmm that's the thing. Many indices assume we are interested in discussing solid precipitations through its liquid water equivalent (convert with 1000 kg m-3), so they implicitly do the conversion ("hydro") if needed. That's why the threshold is in `mm/day` and not `kg m-2 s-1`.  It is not the only index that has a threshold that can involve the implicit transformation to a water equivalent. So I think having the possibility to input  a threshold in `mm/day` rather than `kg m-2 s-1` is regarded as a desired feature.  I kept this behaviour in the concerned indices and added a warning to make this more explicit.  So to answer your other question as well:

> Can you imagine to add those breaking changes in further releases?

I would keep these indices as-is. It might be clearer if `prsnd` had its own index? But It might be redundant. However, I  understand that the implicit conversion allowed with 1000 kg m-3 in the indices is responsible for this confusion.

ludwiglierhammer:
@coxipi Thanks for your explicit explanation! Yes, we should stick to the threshold's unit of the reference. I try to untagle the knot in my brain and clarify my confusion :dizzy_face:.

CHANGES.rst
```diff
@@ -18,6 +29,7 @@ Internal changes
 ^^^^^^^^^^^^^^^^
 * In order to ensure documentation can be rebuilt at a later time, errors raised by `sphinx` linkcheck are now set to be ignored when building the documentation. (:pull:`1375`).
 * With the publication of `xclim`, the code repository now offers a `CITATION.cff` configuration for users to properly cite the software (APA formatted and raw BibTeX) for academic purposes. (:issue:`95`, :pull:`250`).
+* Added `prsnd`, i.e. snowfall rate ([length]/[time]) to list of defined variables. (:pull:`1271`).
```
coxipi, line 32:
```suggestion
* Added `prsnd`, i.e. snowfall rate ([length]/[time]) to list of defined variables. (:pull:`1358`).
```

coxipi:
@ludwiglierhammer In frei_snowfall_2018 cited in the `intensity` and `frequency`indices, it seems that everything is formulated in terms of SWE mm

> A set of six different snowfall indices is considered (Table 2).
Mean snowfall (Smean) refers to the (spatio-)temporally aver-
aged snowfall amount in mm snow water equivalent (SWE;
note that from this point on we will use the unit “mm” as
a synonym for “mm SWE” as the unit of several snowfall
indices).

This means that the suggested threshold 1mm/day is in SWE, or I should say `lwe_prsn` and *not* `prsnd`. Using the 1000 kg/m3 / 100 kg/m3 (water density / snowfall density) , I recover the same threshold 1cm/day as in cbcl_climate_2020.  Do you agree with this?

ludwiglierhammer:
@coxipi: Thanks for your carefull reading. Here are my suggestions/comments (I hope I got it right this time :wink:):
- I like your idea of using `lwe_prsn` thresholds in all snow indices (for consistency reasons). For `{first/last}_snowfall`, this means we could change the 0.5 mm/day threshold to 1 mm/day. I also would add a note in the docstrings and expand the warning (am I right with the assumption in the second box?):

```
Notes
---------
The 1 mm/day liquid-water-equivalent threshold corresponds to the 1cm/day snow thickness threshold 
in :cite:cts:`cbcl_climate_2020` using a snow denstiy of 100 kg/m**3.
``` 
```
warnings.warn(
    f"..."
    ...
    "We recommend to use the snow density of 100 kg/m**3 from :cite:cts:`cbcl_climate_2020` for convertion."
)
```
- frei_snowfall_2018 used CORDEX EUR-11 `prsn` data in there analysis. Maybe we should use `prsn` too instead of `prsnd` for calculating `snowfall_{frequency/intensity}`.  If so we can add the warning from `{first/last}_snowfall` to the code.

I hope, this brings us a little closer to the solution.

coxipi:
That sounds good. About the second box, since we are assuming a liquid water equivalent snowfall rate, you just need to change the threshold you choose. Suppose you want a fixed snowfall rate threshold 1 cm/day, then you can choose:
* 1 mm/day lwe_prsn by supposing a 100 kg m-3  density
* 2.5 mm/day lwe_prsn by supposing a 250 kg m-3 density
etc.

Of course, 1 cm/day with a 250 kg m-3 density will produce the same threshold as 2.5 cm/day with a 100 kg m-3 because it corresponds to the same amount of snow. 

For now, I have kept the precipitation types for `prsn` and `thresh`, meaning that different types of output are possible. But as soon as `prsn` and `thresh` have different units, it is assumed that one of them is a `lwe_prsn` and the other is `prsn`. 

If you work with `prsnd` and `thresh` both in [length/time] units, then no conversion will be done. So the threshold would really be a `prsnd` type threshold in this case. I'm wondering if we should forbid this behaviour and simply write two indicators, one that works with `prsn` and the other with `prsnd`, or if we can simply drop the idea of working directly with `prsnd`.  I have yet to see a study that uses `prsnd` the same way that `snd` is sometimes used. The closest I saw is cbcl_climate_2020  which really formulates things in terms of `prsnd`, but it uses a constant density, so effectively we can use the `prsn` + a `lwe_prsn` threshold approach.

tests/test_precip.py
```diff
@@ -416,16 +478,17 @@ class TestSnowfallDate:
...
+        return xci.prsn_to_prsnd(prsn)
 
     def test_first_snowfall(self, open_dataset):
         with set_options(check_missing="skip"):
             fs = atmos.first_snowfall(
-                prsn=self.get_snowfall(open_dataset), thresh="0.5 mm/day"
+                prsn=self.get_snowfall_rate(open_dataset), thresh="0.5 mm/day"
             )
 
         np.testing.assert_array_equal(
```
coxipi, line 494: We could revert this?

tests/test_precip.py
```diff
@@ -441,7 +504,7 @@ def test_first_snowfall(self, open_dataset):
     def test_last_snowfall(self, open_dataset):
         with set_options(check_missing="skip"):
             ls = atmos.last_snowfall(
-                prsn=self.get_snowfall(open_dataset), thresh="0.5 mm/day"
+                prsn=self.get_snowfall_rate(open_dataset), thresh="0.5 mm/day"
             )
```
coxipi, line 508: We could revert this?

tests/test_precip.py
ludwiglierhammer: Yes, we could.

tests/test_precip.py
ludwiglierhammer: yes, we could.

ludwiglierhammer:
@coxipi: Many thanks for your work. 

Now it is a bit clearer to me. Since the thresholds are liquid water equivalent thresholds  we don't need those warnings anymore if we use `prsn` as input variables because we have to convert with the density of water, isn't it? 

I like the new flexibility of input. 

In the metadata of the snow indicators I added a note that we use liquid water equivalent thresholds. I think it is important for the users to mention that somehow in the metadata. But I am not sure if my idea is useful. Note that the unit of indicator `snowfall_intensity` is `mm` (in liquid water equivalent).

We should rename this PR into `Not realy simple indicators`  :smile:

ludwiglierhammer:
> Currently the threshold is "[precipitation]", so could be a `prsn` threshold too. I've changed the documentation to specify we either have a thresh which as "snowfall flux" or "lwe snowfall rate". Is that all right? We could also force the users to input a "lwe snowfall rate" as threshold by requiring thesh=[length]/[time].

I would stick with this flexible solution. The problem is that the users have to think while they work ... :smile:

> Currently, the `snowfall_intensity` was giving the same units as the input. I've changed this so that the output is forced to be in `lwe snowfall rate` [mm/day] (as it is in the reference paper). Is that all right? We can also revert this and ask to kg m-2 s-1 which would match the units of the input, as you prefer.

I absolutely agree with a `lwe_snowfall_rate` output as it is in the reference paper.

Zeitsperre:
@ludwiglierhammer @coxipi This PR is looking very good! 

I can explain why some of the doctests are failing; The examples are running pieces of code against actual objects, so if you refer to `path_to_sfcWind_file` and `path_to_sfcWindmax_file`, they need to be defined within the doctest setup here: https://github.com/Ouranosinc/xclim/blob/e7664ee72ee12336fb8d5fd518025e19bd4e88d8/xclim/testing/helpers.py#L152

The file `ERA5/daily_surface_cancities_1990-1993.nc` contains a few variables that could be used: `sfcWind` and `wsgsmax` (standard name: "wind_speed_of_gust").

If you'd like to rename `wsgsmax` to `sfcWindmax` (not necessary), this can be done by modifying the file and opening a pull request here: https://github.com/Ouranosinc/xclim-testdata

xclim/indicators/atmos/_precip.py
```diff
@@ -251,12 +256,27 @@ class HrPrecip(Hourly):
     description="{freq} total precipitation.",
     abstract="Total accumulated precipitation. If the average daily temperature is given, the phase parameter can be "
     "used to restrict the calculation to precipitation of only one phase (liquid or solid). Precipitation is "
-    "considered solid if the average daily temperature is below 0°C (and vice versa).",
+    "considered solid if the average daily temperature is below {thresh} (and vice versa).",
```
Zeitsperre, line 259:
```suggestion
    "considered solid if the average daily temperature is below 0°C (and vice versa).",
```

The convention we landed on for these templates is to only have fillable fields in the `description` and `long_name`

xclim/indicators/atmos/_precip.py
```diff
@@ -251,12 +256,27 @@ class HrPrecip(Hourly):
...
 )
 
+precip_average = PrecipWithIndexing(
+    title="Averaged precipitation (solid and liquid)",
+    identifier="prcpavg",
+    units="mm",
+    standard_name="lwe_average_of_precipitation_amount",
+    long_name="Averaged precipitation",
+    description="{freq} mean precipitation.",
+    abstract="Averaged precipitation. If the average daily temperature is given, the phase parameter can be "
```
Zeitsperre, line 272: Same here. Mention of `{thresh}` shouldn't be here.

xclim/indicators/atmos/_precip.py
```diff
@@ -280,12 +299,26 @@ class HrPrecip(Hourly):
     long_name="Total accumulated precipitation when temperature is above {thresh}",
     description="{freq} total {phase} precipitation, estimated as precipitation when temperature is above {thresh}.",
     abstract="Total accumulated liquid precipitation. "
-    "Precipitation is considered liquid when the average daily temperature is above 0°C.",
+    "Precipitation is considered liquid when the average daily temperature is above {thresh}.",
```
Zeitsperre, line 302:
```suggestion
    "Precipitation is considered liquid when the average daily temperature is above 0°C.",
```

xclim/indicators/atmos/_precip.py
```diff
@@ -280,12 +299,26 @@ class HrPrecip(Hourly):
...
 
+liquid_precip_average = PrTasxWithIndexing(
+    title="Averaged liquid precipitation.",
+    identifier="liquidprcpavg",
+    units="mm",
+    standard_name="lwe_average_of_liquid_precipitation_amount",
+    long_name="Averaged precipitation when temperature is above {thresh}",
+    description="{freq} mean {phase} precipitation, estimated as precipitation when temperature is above {thresh}.",
+    abstract="Averaged liquid precipitation. "
+    "Precipitation is considered liquid when the average daily temperature is above {thresh}.",
```
Zeitsperre, line 316: Same here.

xclim/indicators/atmos/_precip.py
```diff
@@ -294,12 +327,26 @@ class HrPrecip(Hourly):
     long_name="Total accumulated solid precipitation",
     description="{freq} total solid precipitation, estimated as precipitation when temperature at or below {thresh}.",
     abstract="Total accumulated solid precipitation. "
-    "Precipitation is considered solid when the average daily temperature is at or below 0°C.",
+    "Precipitation is considered solid when the average daily temperature is at or below {thresh}.",
```
Zeitsperre, line 330:
```suggestion
    "Precipitation is considered solid when the average daily temperature is at or below 0°C.",
```

xclim/indicators/atmos/_precip.py
```diff
@@ -294,12 +327,26 @@ class HrPrecip(Hourly):
...
 
+solid_precip_average = PrTasxWithIndexing(
+    title="Averaged solid precipitation.",
+    identifier="solidprcpavg",
+    units="mm",
+    standard_name="lwe_average_of_snowfall_amount",
+    long_name="Averaged solid precipitation",
+    description="{freq} mean solid precipitation, estimated as precipitation when temperature at or below {thresh}.",
+    abstract="Averaged solid precipitation. "
+    "Precipitation is considered solid when the average daily temperature is at or below {thresh}.",
```
Zeitsperre, line 344: Same here.

xclim/indices/_conversion.py
```diff
@@ -936,59 +991,93 @@ def snw_to_snd(
     ----------
     :cite:cts:`sturm_swe_2010`
     """
-    if snr is None:
-        snr = const
+    density = snr if (snr is not None) else const
+    snd = flux2rate(snw, density=density, out_units=out_units).rename("snd")
+    snd.attrs["standard_name"] = "surface_snow_thickness"
+    return
```
Zeitsperre, line 997: Empty return?!

xclim/indices/_conversion.py
```diff
@@ -904,29 +916,72 @@ def rain_approximation(
...
+    # TODO: Leave this operation to rate2flux? Maybe also the variable renaming above?
+    snw.attrs["standard_name"] = "surface_snow_amount"
+    return snw
+
+
 @declare_units(snw="[mass]/[area]", snr="[mass]/[volume]", const="[mass]/[volume]")
 def snw_to_snd(
     snw: xr.DataArray,
-    snr: xr.DataArray | None = None,
+    snr: Quantified | None = None,
```
Zeitsperre, line 963: So to confirm, we only handle `Quantified` changes here from now on?

Zeitsperre (approved):
I think we're about done. A few bugs need to be addressed before merging, though.

xclim/indices/_conversion.py
coxipi, line 997:
```suggestion
    return snd
```

xclim/indices/_conversion.py
Zeitsperre: My mistake! I was confusing this with the `Quantity` type. I didn't realize this includes `xr.DataArray`s. All good!

tests/test_indices.py
```diff
@@ -2440,30 +2491,69 @@ def test_rain_approximation(pr_series, tas_series, method, exp):
...
 
 
-def test_first_snowfall(prsn_series):
-    prsn = prsn_series(30 - abs(np.arange(366) - 180), start="2000-01-01")
-    out = xci.first_snowfall(prsn, thresh="15 kg m-2 s-1", freq="YS")
+def test_first_snowfall(prsn_series, prsnd_series):
+    mmday2ms = 86400000
+    prsnd = prsnd_series(
+        (30 - abs(np.arange(366) - 180)) / mmday2ms, start="2000-01-01"
+    )
```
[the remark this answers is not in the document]
ludwiglierhammer: Unfortunately, I cannot follow your question. Can you give an example how you would do the test? We test whether we get the same results for both `prsn` and `prsnd` input.

ludwiglierhammer:
> @ludwiglierhammer @coxipi This PR is looking very good!
> 
> I can explain why some of the doctests are failing; The examples are running pieces of code against actual objects, so if you refer to `path_to_sfcWind_file` and `path_to_sfcWindmax_file`, they need to be defined within the doctest setup here:
> 
> https://github.com/Ouranosinc/xclim/blob/e7664ee72ee12336fb8d5fd518025e19bd4e88d8/xclim/testing/helpers.py#L152
> 
> The file `ERA5/daily_surface_cancities_1990-1993.nc` contains a few variables that could be used: `sfcWind` and `wsgsmax` (standard name: "wind_speed_of_gust").
> 
> If you'd like to rename `wsgsmax` to `sfcWindmax` (not necessary), this can be done by modifying the file and opening a pull request here: https://github.com/Ouranosinc/xclim-testdata

@Zeitsperre: Thanks for your explanation. I'll try this PR. If this is done are there any parts we should adjust in `xclim`?

xclim/indicators/atmos/_precip.py
Zeitsperre: We can definitely set the defaults at the indicator or use the defaults that are inherited from `indices`. Generally, the description and long_name should have information within it that more specifically details the way that this indicator was called (containing fillable fields, like `{thresh}`), while the abstract would be what the user expects to see when reading the docstring for the function (before having called it).

I think we should explain what is returned when calling it without having modified the call signature.

Zeitsperre:
Feel free to ignore the failures from `Verify Testing Data`; This is simply a workflow to ensure that the modified testdata tag in our CI matches the latest tag (it does).

tests/test_indices.py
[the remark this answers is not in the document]
ludwiglierhammer: Ah, that makes sense to me. I adjusted the tests with `xclim.core.units.convert_units_to` and added some more tests for the snowfall indices:
- test with prsnd [mm day-1]
- test with prsnd [m s-1]
- test with prsn [kg m-2 s-1]

All those test yield the same results.

In `test_indices.py` there is this `K2C`. We could delete it and call the temperature series with: 
```
da = tas_series(a, units="C")
```
Then we were more independent of hard-coded conversion things.

xclim/indicators/atmos/_precip.py
ludwiglierhammer: I think I found the knot in my brain.

The indicators `xclim.indicators.atmos.precip_{accumulation|average}` are explicitly defined as indicators respecting both liquid **and** solid precipitation. If one will compute an indicator for either liquid **or** solid precipitation one can call `xclim.indicators.atmos.{liquid|solid}_precip_{accumulation|average}`.

Thus, we don't need this liquid-solid differentiation in `xclim.indicators.atmos.precip_{accumulation|average}` since we can find it in both `long_name` and `description` of `xclim.indicators.atmos.{liquid|solid}_precip_{accumulation|average}`.

Isn't it?

Zeitsperre:
> @Zeitsperre: Thanks for your explanation. I'll try this PR ([Ouranosinc/xclim-testdata#25](https://github.com/Ouranosinc/xclim-testdata/pull/25)). If this is done are there any further parts we should adjust in `xclim`?

No, I think we're in the home stretch! I'm waiting on a review of #1388 to fix the final CI-related failures, and I think that this PR will follow right afterwards. Thanks so much once again for all your time and effort!

coxipi (approved):

xclim/indicators/atmos/_precip.py
Zeitsperre: Yes, precisely! `Indicators` make it simpler for users to return a specific variable with a different default call signature from the `indice`, which includes all CF-checks and without needing to change any settings.

There are a lot of layers to consider, but the way I envision it is:
- `indices.generic`: Building blocks for creating climate indexes (exposed but no checks for units, nothing specifically tied to a variable/unit).
- `indices.*`: Climate indexes, includes checks for proper units on both input and output datasets, lots of user control and granularity in settings (operators, thresholds, methods, etc.)
- `indicators.*`: Most user-friendly way of calling `indices.*` with all checks integrated. This is where very specific indexes can be defined with fixed default values and lots of finely-tuned metadata.

tests/test_indices.py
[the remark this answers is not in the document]
Zeitsperre: True! I know in some cases we are testing to see whether the `convert_units_to` is working properly, but for examples testing indices specifically, we should migrate, absolutely. Not necessary in this PR!

coxipi:
Thanks for the nice work @ludwiglierhammer  @Zeitsperre

PR #1359, opened 2023-05-02
More spell indicators

ludwiglierhammer:
Hi guys,

this is the second PR regarding issue #1352.
I implemented `cold_spell_max_length`, `dry_spell_frequency` and `dry_spell_max_length` according to `hot_spell_frequency` and `hot_spell_max_length`.

What should we do with all those `maximum_consecutive_*_days`? This is nothing but `*_spell_max_length` without the `window` parameter (as I understand). Should we leave it as it is or replace it with all those `*_spell_max_length` indicators?

Test will follow.

Cheers, Ludwig

### DIFF
--- a/xclim/indicators/atmos/_precip.py
+++ b/xclim/indicators/atmos/_precip.py
@@ -21,6 +21,8 @@ __all__ = [
     "wetdays",
     "wetdays_prop",
     "dry_days",
+    "dry_spell_frequency",
+    "dry_spell_max_length",
     "dryness_index",
     "maximum_consecutive_dry_days",
     "maximum_consecutive_wet_days",
@@ -618,4 +620,34 @@ cold_and_wet_days = PrecipWithIndexing(
     abstract="Number of days with temperature below a given percentile and precipitation above a given percentile.",
     cell_methods="time: sum over days",
     compute=indices.cold_and_wet_days,
+)
+
+dry_spell_frequency = Precip(
+    title="Dry spell frequency",
+    identifier="dry_spell_frequency",
+    units="",
+    standard_name="dry_spell_frequency",
+    long_name="Total number of series of at least {window} consecutive days "
+    "where the mean daily precipitation is below {thresh}",
+    description="{freq} number dry spell events. A dry spell is definedas a minimum number of "
+    "consecutive days with daily precipitation below {thresh}.",
+    abstract="The number of dry spell events. A dry spell is defined as a minimum number of consecutive days with "
+    "mean daily precipitation below a given threshold.",
+    cell_methods="",
+    compute=indices.dry_spell_frequency,
+)
+
+dry_spell_max_length = Precip(
+    title="Dry spell maximum length",
+    identifier="dry_spell_maximum_length",
+    units="days",
+    standard_name="spell_length_of_days_with_precipitation_below_threshold",
+    long_name="Longest series of at least {window} consecutive days "
+    "with daily precipitation below {thresh}",
+    description="{freq} maximum length of dry spell events occurring within a given period. "
+    "A dry spell event occurs when the daily precipitation is below {thresh} over at least {window} days.",
+    abstract="Maximum length of dry spells events within a given period. A dry spell occurs when the daily"
+    "precipitation is below a given threshold for a minimum number of days.",
+    cell_methods="",
+    compute=indices.dry_spell_max_length,
 )
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -37,8 +37,11 @@ __all__ = [
     "calm_days",
     "cold_spell_days",
     "cold_spell_frequency",
+    "cold_spell_max_length",
     "daily_pr_intensity",
     "degree_days_exceedance_date",
+    "dry_spell_frequency",
+    "dry_spell_max_length",
     "cooling_degree_days",
     "snd_season_end",
     "snd_season_start",
@@ -230,6 +233,56 @@ def cold_spell_frequency(
     )
     out.attrs["units"] = ""
     return out
+
+
+@declare_units(tas="[temperature]", thresh="[temperature]")
+def cold_spell_max_length(
+    tas: xarray.DataArray,
+    thresh: Quantified = "-10 degC",
+    window: int = 5,
+    freq: str = "AS-JUL",
+    op: str = "<",
+    resample_before_rl: bool = True,
+) -> xarray.DataArray:
+    r"""Longest cold spell.
+
+    Longest spell of temperatures over a given period.
+    The longest series of consecutive days with tas at or below -10°C. Here, there is no minimum threshold for number
+    of days in a row that must be reached or exceeded to count as a spell. A year with zero -10°C days will return a
+    longest spell value of zero.
+
+    Parameters
+    ----------
+    tas : xarray.DataArray
+        Mean daily temperature.
+    thresh : Quantified
+        The temperature threshold needed to trigger a coldwave event.
+    window : int
+        Minimum number of days with temperatures below thresholds to qualify as a coldwave.
+    freq : str
+        Resampling frequency.
+    op : {"<", "<=", "lt", "le"}
+        Comparison operation. Default: "<".
+    resample_before_rl : bool
+        Determines if the resampling should take place before or after the run
+        length encoding (or a similar algorithm) is applied to runs.
+
+    Returns
+    -------
+    xarray.DataArray, [time]
+        Maximum length of continuous cold days at the wanted frequency.
+    """
+    thresh = convert_units_to(thresh, tas)
+
+    cond = compare(tas, op, thresh, constrain=("<", "<="))
+    max_l = rl.resample_and_rl(
+        cond,
+        resample_before_rl,
+        rl.longest_run,
+        freq=freq,
+    )
+    out = max_l.where(max_l >= window, 0)
+    return to_agg_units(out, tas, "count")
 
 
 @declare_units(snd="[length]", thresh="[length]")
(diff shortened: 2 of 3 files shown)


xclim/indicators/atmos/_precip.py
```diff
@@ -619,3 +621,33 @@ class HrPrecip(Hourly):
...
+
+dry_spell_frequency = Precip(
+    title="Dry spell frequency",
+    identifier="dry_spell_frequency",
+    units="",
+    standard_name="dry_spell_frequency",
+    long_name="Total number of series of at least {window} consecutive days "
+    "where the mean daily precipitation is below {thresh}",
+    description="{freq} number dry spell events. A dry spell is definedas a minimum number of "
+    "consecutive days with daily precipitation below {thresh}.",
```
coxipi, line 633:
```suggestion
    description="{freq} number of dry spell events. A dry spell is defined as a minimum number of "
    "consecutive days with daily precipitation below {thresh}.",
```

xclim/indices/_threshold.py
```diff
@@ -232,6 +235,56 @@ def cold_spell_frequency(
...
+    tas: xarray.DataArray,
+    thresh: Quantified = "-10 degC",
+    window: int = 5,
+    freq: str = "AS-JUL",
+    op: str = "<",
+    resample_before_rl: bool = True,
+) -> xarray.DataArray:
+    r"""Longest cold spell.
+
+    Longest spell of temperatures over a given period.
```
coxipi, line 249:
```suggestion
    Longest spell of low temperatures over a given period.
```

xclim/indices/_threshold.py
```diff
@@ -2635,3 +2688,102 @@ def winter_storm(
...
+    precipitation is below a specific threshold (default: 1 mm/day) over a minimum number of days (default: 3).
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+        Mean daily precipitation flux.
+    thresh : Quantified
+        The precipitation threshold needed to trigger a dry wave event.
+    window : int
+        Minimum number of days with temperatures above thresholds to qualify as a dry wave.
```
coxipi, line 2714:
```suggestion
        The precipitation threshold needed to trigger a dry spell.
    window : int
        Minimum number of days with precipitation above threshold to qualify as a dry spell.
```

xclim/indices/_threshold.py
```diff
@@ -2635,3 +2688,102 @@ def winter_storm(
...
+    op : {"<", "<=", "lt", "le"}
+        Comparison operation. Default: "<".
+    resample_before_rl : bool
+        Determines if the resampling should take place before or after the run
+        length encoding (or a similar algorithm) is applied to runs.
+
+    Returns
+    -------
+    xarray.DataArray, [dimensionless]
+        Number of dry wave at the wanted frequency
```
coxipi, line 2726:
```suggestion
        Number of dry spells at the wanted frequency
```

xclim/indices/_threshold.py
```diff
@@ -2635,3 +2688,102 @@ def winter_storm(
...
+    freq: str = "YS",
+    op: str = "<",
+    resample_before_rl: bool = True,
+) -> xarray.DataArray:
+    r"""Longest dry spell.
+
+    Longest spell of precipitation over a given period.
+    The longest series of consecutive days with precipitation at or below 1 mm/day. Here, there is no minimum threshold for number
+    of days in a row that must be reached or exceeded to count as a spell. A year with zero 1mm/day days will return a
+    longest spell value of zero.
```
coxipi, line 2756:
```suggestion
    Longest spell of precipitation over a given period.
    Longest series of at least {window} consecutive days with precipitations at or below {thresh}.
```

xclim/indices/_threshold.py
```diff
@@ -2635,3 +2688,102 @@ def winter_storm(
...
+
+    Parameters
+    ----------
+    pr : xarray.DataArray
+        Mean daily precipitation flux.
+    thresh : Quantified
+        The precipitation threshold needed to trigger a dry wave event.
+    window : int
+        Minimum number of days with precipitation below thresholds to qualify as a dry wave.
```
coxipi, line 2765:
```suggestion
        The precipitation threshold needed to trigger a dry spell.
    window : int
        Minimum number of days with precipitation below thresholds to qualify as a dry spell.
```

xclim/indices/_threshold.py
ludwiglierhammer: I changed the docstring similar to `hot_spell_max_length`.

coxipi:
> What should we do with all those `maximum_consecutive_*_days`? This is nothing but `*_spell_max_length` without the `window` parameter (as I understand). Should we leave it as it is or replace it with all those `*_spell_max_length` indicators?

Thanks for this useful remark, I agree. I think we already discussed the idea of having less of the specific indices but keep the related indicators. For instance, the indicator `consecutive_frost_days` currently depends on `maximum_consecutive_frost_days`. It could be written with `cold_spell_max_length`
```python
    consecutive_frost_days = Temp(
    ...
    compute=indices.maximum_consecutive_frost_days,
    parameters={"thresh": {"default": "0 degC"}},
    )
    ```

**Comment by ludwiglierhammer** [comment]
Is there any reason why default window is 5 at `cold_spell_frequency`? Can we change it to 3 according to `hot_spell_frequency`?

**Comment by ludwiglierhammer** [comment]
I recognized that there are multiple `dry_spell_*` indices. 
- `dry_spell_frequency` in both `_threshold.py` and `_agro.py` (slightly different codes)
- `dry_spell_total_length` in `_agro_.py`
- `dry_spell_max_length` in `_threshold.py` 

I implemented the ones in `_threshold.py` because the other spell indices are in `_threshold.py` too. In your view, where should we have the `dry_spell_*` indices?

I think the threshold parameters in `_agro,.py` should be `thresh="[precipitation]"` and not `thresh="[length]"`.

**Comment by ludwiglierhammer** [comment]
> You can represent a precipitation as its rate [mm/d], or integrate over time and work with an amount [mm]. The input precipitation here is a rate, and the threshold is an amount. The rate pr is converted to an amount pram before comparing with thresh, so it's OK.

I was just wondering because the test fail:

FAILED tests/test_indices.py::test_dry_spell[pr0-3-3-7-outs0] - xclim.core.utils.ValidationError: Data units 3.0 millimeter are not compatible with requested [precipitation].

**Comment by ludwiglierhammer** [comment]
> It allows a supplementary mode "op == sum" to account for spells where the criterion is cumulative (a sequence where for window days, the total precipitation is under a given threshold). The mode "max" is equivalent to what you coded (a sequence where for window days, each day in the window is under a given threshold).

For comparability resaons, I think we should add this option to the other `*_spell_frequency` indices. What do you think about it?

**Comment by coxipi** [comment]
> > You can represent a precipitation as its rate [mm/d], or integrate over time and work with an amount [mm]. The input precipitation here is a rate, and the threshold is an amount. The rate pr is converted to an amount pram before comparing with thresh, so it's OK.
> 
> I was just wondering because the test failed:
> 
> FAILED tests/test_indices.py::test_dry_spell[pr0-3-3-7-outs0] - xclim.core.utils.ValidationError: Data units 3.0 millimeter are not compatible with requested [precipitation].

There are currently two `dry_spell_frequency` functions, the old one and the new one, which have different expectation for the dimension of `thresh`. What is likely happening is: A `thresh` with dim=[length] is given, but the function expecting dim=[precipitation] is used. I think this should be cleared once we use only one function. 

> For comparability resaons, I think we should add this option to the other *_spell_frequency indices. What do you think about it?

I agree with you. I already started adapting your new functions, I can take care of it.

**Comment by ludwiglierhammer** [comment]
I found this in the docstring of `hot_spell_*` indices:

The thresholds of 22° and 25°C for night temperatures and 30° and 35°C for day temperatures were selected by
Health Canada professionals, following a temperature–mortality analysis. These absolute temperature thresholds
characterize the occurrence of hot weather events that can result in adverse health outcomes for Canadian
communities :cite:p:`casati_regional_2013`.

In :cite:t:`robinson_definition_2001`, the parameters would be
`thresh_tasmin=27.22, thresh_tasmax=39.44, window=2` (81F, 103F).

I think this note has something to do  with `heat_wave_*` indices and we could delete it from `hot_spell_*` indices, couldn't we?

**Comment by coxipi** [comment]
> I think this note has something to do with `heat_wave_*` indices and we could delete it from `hot_spell_*` indices, couldn't we?

I agree that double thresholds concern heat waves. But it seems we just used the  `tasmax` threshold of `heat_wave_*`  in `hot_spell_*` indices, so I would say that these references are still useful. But as it is, I agree it is confusing. I've updated the notes.

**Comment by coxipi** [comment]
I see a few breaking changes in already existing indices:
* `thresh_tasmax` -> `thresh`
* `window=1` -> `window=3`

I agree that uniformizing things in this way would be nice (especially for `thresh`. For the `window` default, there might have been a reason why `window=1`  in `total_length` indices, I don't know). 

But, we tend to avoid breaking changes if they are not necessary.  Just double-checking with you @Zeitsperre , what do you think?

**Comment by Zeitsperre** [comment]
@coxipi 

This is just my opinion, so I'd welcome others to weigh in here, but if we have adequate justification for changing the default settings of an index and indicator (literature, uniformity with other indices, common sense, etc.), our primary responsibility to the users is to firmly inform them of changes. For the interim of these call signature changes, what I've done in the past is something that resembles this:

```python
def my_indicator(tasmax, window="UNSET", thresh_tasmax="UNSET", thresh=1):
    if window == "UNSET":
        warnings.warn("The default value for this indicator will change in xclim>=X.Y.Z,  from `1` to `3`. Using `1` for now.")
        window = 1
    if thresh_tasmax != "UNSET":
        warnings.warn("The call signature for this indicator will change from `thresh_tasmax` to `thresh` in xclim>=X.Y.Z. Passing `thresh_tasmax` value to `thresh`.")
        thresh = thresh_tasmax
    ...
```

Given that these changes affect a much older indice, it might be worthwhile to verify with @tlogan2000 that these indices are not currently being used for any major projects (like climatedata.ca). If so, the maintainers there might also need a heads-up.

We need to give at least a few months of warnings before we make significant changes and ensure that they are both in the code and in the changelog.

ludwiglierhammer:
> You can represent a precipitation as its rate [mm/d], or integrate over time and work with an amount [mm]. The input precipitation here is a rate, and the threshold is an amount. The rate pr is converted to an amount pram before comparing with thresh, so it's OK.

> Without more info, I would also prefer thresh="[precipitation]". It must be a matter of convention where precipitations thresholds are often discussed in terms of total amount in the literature, but I can't say more.

Could we specify both thresh="[precipitation]" and thresh="[length]"?

coxipi:
> Could we specify both thresh="[precipitation]" and thresh="[length]"?

I guess you would actually want: [precipitation], [length], or [mass]/[area], i.e. :
[mass]/[area]/[time], [length]/[time], [length], or [mass]/[area].

I'm not sure what would be the best way to proceed: I suppose a new type could be precised. I tried looking at precipitation 
in `units.py` and I only see `precipitation = [mass]/[length]**2/[time]`, so I'm not sure how the equivalence with [length]/[time] is done. I might also possible to specify multiple types directly in the indicator? @Zeitsperre Do you think we should allow this, and if so, any thoughts on what I just wrote?

CHANGES.rst
```diff
@@ -13,16 +13,19 @@ Announcements
...
+    * ``{cold | dry}_spell_max_length`` for calculating maximum length of {cold | dry} spell events. (:issue:`1352`, :pull:`1359`).
+    * ``dry_spell_frequency`` for calculating total number of dry spells. (:issue:`1352`, :pull:`1359`).
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * ``xclim.sdba.processing.escore`` performance was improved with a faster runtime (:pull:`1360`).
 * New generic function (``flux_and_rate_converter``) converting flux to a rate (and vice-versa) using a density. `snw_to_snd` and `snd_to_snw` were refactored using this function. (:issue:`1352`, :pull:`1358`)
 * New function (``prsn_to_prsnd``) to convert snowfall flux ([mass]/[area]/[time]) to snowfall rate ([length]/[time]) using snow density ([mass]/[volume]). (:issue:`1352`, :pull:`1358`)
-* New variables: Snowfall rate  ``prsnd`` and surface maximum wind speed ``sfcWindmax``.
+* New variables: Snowfall rate ``prsnd`` and surface maximum wind speed ``sfcWindmax``.
```
Zeitsperre, line 28: :issue: and :pull:

tests/test_indices.py
```diff
@@ -3262,6 +3334,136 @@ def test_dryness_index(self, atmosds):
...
+                7.422,
+                14.8854,
+                13.4334,
+                0.0012,
+                0.0782,
+                31.2916,
+                0.0379,
+            ]
+        )
+    )
```
Zeitsperre: This is a bit long. Maybe shorten?

Zeitsperre (approved):
Looks great! Thanks once again @ludwiglierhammer and @coxipi !

coxipi:
Thanks again @ludwiglierhammer  @Zeitsperre  nice work

ludwiglierhammer:
Many thanks to both of you @coxipi @Zeitsperre for your great help/work.

PR #1361, opened 2023-05-03
delayed_frost_days

ludwiglierhammer:
Hi guys,

this is the next PR regarding #1352.
I implemented `delayed_frost_days`. 
I think we can optimize my approach to differentiate between the Northern and the Southern Hemisphere. I copied it almost completely from [`cool_night_index`](https://github.com/Ouranosinc/xclim/blob/1966c0f8bb6bdeceb97bc7a70fc5f4eb8121be1b/xclim/indices/_agro.py#L450) and [`huglin_index`](https://github.com/Ouranosinc/xclim/blob/1966c0f8bb6bdeceb97bc7a70fc5f4eb8121be1b/xclim/indices/_agro.py#L140).

Test will follow.

Cheers, Ludwig


xclim/indices/_threshold.py
```diff
@@ -2635,3 +2638,52 @@ def winter_storm(
...
+    start_date_nh: DayOfYearStr = "04-01",
+    start_date_sh: DayOfYearStr = "10-01",
+    end_date_nh: DayOfYearStr = "06-30",
+    end_date_sh: DayOfYearStr = "12-31",
+    freq: str = "YS",
+) -> xarray.DataArray:
+    r"""Delayed frost days.
+
+    Number of days where daily minimum temperatures are below a threshold temperature
+    for April to June on Nouthern Hemisphere and for October to December on Southern Hemisphere.
```
coxipi, line 2657:
```suggestion
    for April to June on Northern Hemisphere and for October to December on Southern Hemisphere.
```

Zeitsperre:
Small reminder: When defining the docstring for an indice/indicator, `optional` means that the call signature explicitly accepts `None`.

Zeitsperre (approved):

Zeitsperre (approved):

Zeitsperre (approved):

Zeitsperre (approved):

Zeitsperre (approved):

Zeitsperre (approved):

coxipi:
Just to understand: If we ignore the North/South distinction, this is like computing `frost_days` with an indexer, correct? So I should  understand "delayed" as synonym of "late" here, correct?  
With the North/South addition, this is more complex because the indexer would need to be space-dependent?

aulemahal:
I share the questioning of @coxipi.

"delayed" made me confused, wouldn't something like "late" work better ?

And indeed, this is almost only an "indexed" `frost_days`. Without the hemisphere thing, this could be only an indicator (wrapping `frost_days` with an added indexer), or just be documented.

If you pass data with two hemispheres, you'll still sum over YS periods. Does that make sense ? In the south hemisphere, you'll have the number of frost days at the end of the summer summed with those as the beginning of the next one. The hemisphere-dependent code of xclim has mostly been concerned with numerical factors varying across latitudes, I'm not convinced this mechanic is useful enough to be explicitly implemented like this. Especially that this indicator is not the only one where it could be applied.

Zeitsperre (approved):

Zeitsperre:
@ludwiglierhammer 

Please excuse all the force pushes on this PR; I was trying to fix a permissions issue for the labeler when running against forks of the repository. I'll be continuing this work on my own fork to not pollute the history here.

ludwiglierhammer:
@aulemahal, @coxipi 

yes, `delayed_frost_days` is like `frost_days` with an indexer. 
I am fine with renaming `delayed` to `late`, no problem.

ludwiglierhammer:
I agree that the time resampling with two hemispheres is problematic. I think we need two different time resampling, one for the Northern and one for the Southern one:
```
freq_nh = "YS"
freq_sh = "AS-JUL"
tasmin_nh = tasmin.where(lat >= 0, drop=True)
tasmin_sh = tasmin.where(lat < 0, drop=True)
late_frost_days_nh = frost_days(tasmin_nh, thresh=thresh, freq=freq_nh)
late_frost_days_sh = frost_days(tasmin_sh, thresh=thresh, freq=freq_sh)
return xarray.merge([late_frost_days_nh, late_frost_days_sh]).tasmin
```
But with this approach we get another "problem" with two different time axes, e.g.:

```
array([cftime.DatetimeProlepticGregorian(1849, 7, 1, 0, 0, 0, 0, has_year_zero=True),
       cftime.DatetimeProlepticGregorian(1850, 1, 1, 0, 0, 0, 0, has_year_zero=True),
```
In generell, how should we handle the time axis if we have two different time resamplings? If the time axis is like above we have pretty much alternating NaN patterns for both the Northern and the Southern hemisphere. If we "combine" the two axes to something like `cftime.DatetimeProlepticGregorian(1849, 10, 1, 0, 0, 0, 0, has_year_zero=True)` this will yield to much confusion, I think.

Another aspect is that 6(!) parameters (`start_date_nh`, `start_date_sh`, `end_date_nh`, `end_date_sh`, `freq_nh` and `freq_sh`) handling one "problem" are not really user friendly and will yield to more confusion than clarification, I think.

ludwiglierhammer:
Hi @huard 

I totally agree with your suggestion. Is there anything I can do to help you in #1391?

For this PR, I simplified `late_frost_days` by deleting the `lat`, `start_date_sh` and `end_date_sh` and setting the default values for the Northern Hemisphere. I added a warning in the docstings similar to other indices that aslo depend on the hemisphere. Maybe we could merge this new indicator and close this PR afterwards.

xclim/indices/_threshold.py
```diff
@@ -3357,3 +3360,45 @@ def wet_spell_max_length(
...
+        The end date to consider.
+    freq : str
+        Resampling frequency.
+
+    Returns
+    -------
+    xarray.DataArray, [time]
+        Number of late frost days per period.
+    """
+    dates = [f"{date.dt.month:02d}-{date.dt.day:02d}" for date in tasmin.time]
```
huard, line 3400: This can be simplified with `tasmin.time.dt.strftime("%m-%d")`

huard:
Hi @ludwiglierhammer 

If you want to write documentation about this, it would be great!

I'm hesitant to add a new indice function with three lines of code. This is what we did at the beginning of xclim, but now we're trying to write "generic" indices, and use them in indicators that can be more specific. What I'd like is to keep your late_frost_days indicator, but manage to write it with a one-liner using existing functions. 

@coxipi suggested adding an indexer argument to the `frost_days` indice, would that work for you?

ludwiglierhammer:
Thanks @huard for your work.

I like those "generic" indices a lot. This makes the code much clearer, cleaner and flexible I think.
Where can or should I write more documentation about this?

huard:
Great !

I think a good place would be at the end of docs/notebooks/example.ipynb, but let's do this in another branch so we can merge this one.

Zeitsperre (approved):
LGTM!

coxipi:
Still need to remove some things (many "late_XYZ"), I will do this

Zeitsperre:
Fantastic work once again @ludwiglierhammer, thank you! And thanks, @huard, for the careful concern for the indexing operations!

ludwiglierhammer:
As always, I enjoyed working with you.

PR #1367, opened 2023-05-04
Support use of calendar version xclim-testdata tags

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Checks for outdated xclim now use the date of last modification to determine installation date of xclim to generate a CalVer string to compare against xclim-testdata

No.

### Other information:
See: https://github.com/Ouranosinc/xclim-testdata/pull/24

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: 2023.2.6
+  XCLIM_TESTDATA_BRANCH: 2023.5.4
 
 jobs:
   black:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: v0.41.0
+  XCLIM_TESTDATA_BRANCH: 2023.2.6
 
 jobs:
   black:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: 2023.5.4
+  XCLIM_TESTDATA_BRANCH: v2023.5.4
 
 jobs:
   black:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,9 @@ from __future__ import annotations
 import os
 import re
 import shutil
+import time
 import warnings
+from datetime import datetime as dt
 from functools import partial
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


Zeitsperre:
This PR is ready but demands the removal of an existing tag in xclim-testdata (`v0.41.0`). This will momentarily break tests for all existing pull requests, so changes should be performed rapidly.

tests/conftest.py
```diff
@@ -24,21 +26,31 @@
...
 
 if re.match(r"^v\d+\.\d+\.\d+", helpers.TESTDATA_BRANCH):
-    if parse_version(helpers.TESTDATA_BRANCH) > parse_version(__xclim_version__):
+    # Find the date of last modification of xclim source files to generate a calendar version
+    install_date = dt.strptime(
+        time.ctime(os.path.getmtime(working_set.by_key["xclim"].location)),
+        "%a %b %d %H:%M:%S %Y",
+    )
+    install_calendar_version = (
+        f"{install_date.year}.{install_date.month}.{install_date.day}"
```
tlvu, line 46: this regex will work for format `2023.05.12-r2` for example?

.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: v0.41.0
+  XCLIM_TESTDATA_BRANCH: 2023.5.4
```
tlvu: Maybe rename `XCLIM_TESTDATA_BRANCH` to `XCLIM_TESTDATA_COMMIT_IDENTIFIER` so you can use branch, or tag, or commit?

Anyways, don't change if risky.  I think we understand what it can do.

tlvu:
> demands the removal of an existing tag in xclim-testdata (`v0.41.0`).

Why?  Just keep the old tags there and add the new tag so you are both backward and forward compatible.  In fact you have to keep all the old tags in xclim-data so old xclim code that still relies on old tag format works.

tests/conftest.py
Zeitsperre: `pkg_resources.parse_version` can handle modifiers in the version string.

For this particular line, we're creating a CalVer string from the date of last modification of xclim, therefore there won't be any modifiers.

.github/workflows/main.yml
Zeitsperre: I'd rather settle on `BRANCH` as the use of this environment variable is to allow developers to test their branches using testing data in a different branch. This is currently documented as well.

tlvu: good for me.

Zeitsperre:
> > demands the removal of an existing tag in xclim-testdata (`v0.41.0`).
> 
> Why? Just keep the old tags there and add the new tag so you are both backward and forward compatible. In fact you have to keep all the old tags in xclim-data so old xclim code that still relies on old tag format works.

That was my mistake; My code to determine the last commit placed `v0.41.0` at the bottom, as the previous CalVer tags were sorted alphanumerically and did not have `v-`. This is now fixed.

I've created a new tag for `v0.41.0` that follows CalVer, so the `v0.41.0` is now redundant. I think it's safe (with the code changes here, preferable, even) to remove it, but not until we clear the existing PR queue.

tlvu (approved):

PR #1370, opened 2023-05-08
Use tasmin in all cold spell indices

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

``xclim.indices.cold_spell_frequency`` and ``xclim.indices.cold_spell_days`` now use `tasmin` instead of `tas`

Not really, `tas` can still be used as an input. But this might change the input from `tas` to `tasmin` in automated workflows, yielding different results for indices.


xclim/data/fr.json
```diff
@@ -449,13 +449,13 @@
     "abstract": "Nombre de jours de vagues de froid. Une vague de froid se produit lorsque la température minimale quotidienne est sous un centile donné pendant au moins {window} jours consécutifs."
   },
   "COLD_SPELL_FREQUENCY": {
-    "long_name": "Nombre total de séries d'au moins {window} jours consécutifs où la température quotidienne moyenne est sous {thresh}.",
+    "long_name": "Nombre total de séries d'au moins {window} jours consécutifs où la température minimale quotidienne moyenne est sous {thresh}.",
```
Zeitsperre, line 452: "température minimale quotidienne moyenne" seems confusing.

xclim/data/fr.json
```diff
@@ -449,13 +449,13 @@
...
     "description": "Nombre {freq:m} de vagues de froid. Une vague de froid se produit lorsque la température minimale quotidienne est sous {thresh} durant au moins {window} jours consécutifs.",
     "title": "Nombre de vagues de froid",
     "abstract": "Nombre de vagues de froid. Une vague de froid se produit lorsque la température minimale quotidienne est sous un seuil durant au moins un certain nombre de jours consécutifs."
   },
   "COLD_SPELL_DAYS": {
-    "long_name": "Nombre total de jours constituant des événements d'au moins {window} jours consécutifs où la température quotidienne moyenne est sous {thresh}",
+    "long_name": "Nombre total de jours constituant des événements d'au moins {window} jours consécutifs où la température minimale quotidienne moyenne est sous {thresh}",
```
Zeitsperre, line 458: Same as above.

coxipi:
`cold_spell_duration_index` is defined here  and uses `tasmin`. For `cold_spell_days` and `cold_spell_frequency`, I don't know a similar reference we use, I think we should simply use the same variable. @aulemahal , thoughts?

coxipi:
> This looks good. I'm not sure if we want to allow an intermediary version that accepts `tas` or `tasmin` (where using kwarg with `tas` emits a warning). It would not be absolutely needed, but could be a nice-to-have for users.
> 
> I leave the decision of whether we implement this to you.

I thought we had concluded to just leave this as-is (and if there is ever a reason we absolutely need `tasmin`, invent another index? )

Allowing both `tas` and `tasmin`,  I would still need to default to `tas` to avoid breaking changes? The advantage is that `tasmin` would now be allowed in a generalized workflow?

Issue #1135, opened 2022-07-21
Accelerate xclim import


aulemahal:
### Generic Issue

* xclim version: all
* Python version: any
* Operating System: any

### Description
Importing `xclim.sdba` is currently quite slow. I think this is due to the numba utilities being pre-compiled because we give explicit input types. 

### What I Did
In `perfimport.py`:
```python3
import xclim as xc
from xclim import sdba


if __name__ == '__main__':
    print('yeah')
```
and then:
```bash
python -X importtime perfimport.py 2> import.log
tuna import.log
```
Using [tuna](https://github.com/nschloe/tuna).

### What I Received
![image](https://user-images.githubusercontent.com/20629530/180317242-dc37cef6-51a2-4f89-a62f-71c141fa1e8b.png)

I believe the other long import,  "xclim/indices/_conversion.py", may be due to the UTCI function, which also uses numba.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Zeitsperre:
Do you mean to say that this can be solved by relaxing static typing in SDBA? Is this a "feature" of numba?

aulemahal:
Yep, this is what I read from the doc! https://numba.readthedocs.io/en/stable/user/jit.html#eager-compilation

IIRC, my goal with the static typing was avoid dtype promotions (float32 to 64). With this "eager" compilation, the tests would fail if the functions received mixed inputs, which I would then try to explicitly avoid. It may be thus be unnecessary to have them "in production". Also, a dtype promotion is not a tragic bug...

PR #1378, opened 2023-05-26
Accelerate import - prevent numba warnings

aulemahal:
### Pull Request Checklist:
    - This PR fixes an issue raised verbally by `xscen` developers.

PR #1135 removed all signatures where possible, so that `@jit`-decorated function would be compiled at run time (you know "just in time"...). However, `@guvectorize` -decorated functions _must_ have signatures and thus must be compiled at compile time, which means upon import.

This PR adds `cache=True` to these function, so numba compiles them once and than saves the compilation to a `__pycache__` folder if possible. The first import will be slow as before, but subsequent imports should be much faster.
I'm sure there are edgecases where the user of xclim doesn't have write access to the source folder, which might mean this caching will be unavailable. But for most users, this should be a nice performance boost.

On my machine I went from 13s to 2.5s.

This PR also forces `nopython=True` to these `@guvectorize` functions. Numba 0.59 will change the default value of this parameter and it was raising warning complaining about our lack of precision.

It shouldn't. But I do not know all the ramifications of caching these functions...

### Other information:
@RondeauG, fyi.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Announcements
 Bug fixes
 ^^^^^^^^^
 * Fixed a bug in ``xclim.core.calendar.time_bnds`` when using ``DataArrayResample`` objects, caused by an upstream change in xarray 2023.5.0. (:issue:`1368`, :pull:`1377`).
+* Accelerated import of xclim by caching the compilation of `guvectorize` functions.
 
 Internal changes
 ^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Announcements
 Bug fixes
 ^^^^^^^^^
 * Fixed a bug in ``xclim.core.calendar.time_bnds`` when using ``DataArrayResample`` objects, caused by an upstream change in xarray 2023.5.0. (:issue:`1368`, :pull:`1377`).
-* Accelerated import of xclim by caching the compilation of `guvectorize` functions.
+* Accelerated import of xclim by caching the compilation of `guvectorize` functions. (:pull:`1378`).
 
 Internal changes
 ^^^^^^^^^^^^^^^^
--- a/xclim/indices/fire/_cffwis.py
+++ b/xclim/indices/fire/_cffwis.py
@@ -136,7 +136,7 @@ from typing import Sequence
 
 import numpy as np
 import xarray as xr
-from numba import jit, vectorize
+from numba import njit, vectorize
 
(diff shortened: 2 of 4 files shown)


Zeitsperre:
> On my machine I went from 13s to 2.5s.

Wow!

Zeitsperre (approved):
LGTM

aulemahal:
In details.

# Previous behaviour
image
Notice the 3 main culprits : "sdba.nbutils" (2 guvectorize funcs), "indices.fire._ffdi" (2 guvectorize funcs) and "core.units".

# After this change, first import
image
Same as above.

# After this change, second import
image
Only `core.units` left as significant. Around 50% of what is left comes from our dependencies. I guess 50% of import time is still large...

Zeitsperre:
Would it be possible to deal with some of the numba-related warnings being raised right now? For instance:
```
$ xclim show_version_info

INSTALLED VERSIONS
------------------
/home/runner/work/xclim/xclim/xclim/indices/fire/_cffwis.py:207: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
python: 3.8.16
  def _day_length(lat: int | float, mth: int):  # pragma: no cover
boltons: installed
/home/runner/work/xclim/xclim/xclim/indices/fire/_cffwis.py:227: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
  def _day_length_factor(lat: float, mth: int):  # pragma: no cover
bottleneck: 1.3.7
cf_xarray: 0.8.1
cftime: 1.6.2
click: 8.1.3
clisops: None
dask: 2023.5.0
flox: None
jsonpickle: 3.0.1
lmoments3: 1.0.5
numba: 0.57.0
pandas: 1.5.3
pint: 0.21.1
scipy: 1.8.1
sklearn: 1.2.2
statsmodels: 0.14.0
xarray: 2023.1.0
xclim: 0.43.2-beta
Anaconda-based environment: no

aulemahal:
Of course! I forgot to look into this module (cffwis) as it had no pre-compiled numba funcs.

Issue #1268, opened 2023-01-12
Drop Python3.8


Zeitsperre:
Python3.8 has entered the security fixes stage and is gradually losing support from major scientific libraries: https://peps.python.org/pep-0569/

Dropping this version will open us up to some new features that were introduced in Puthon3.9 (https://peps.python.org/pep-0596/#features-for-3-9), notably `__future__.annotations` (https://peps.python.org/pep-0585/)

Zeitsperre:
Looks like xarray just dropped support for Python3.8 in main dev branch! https://github.com/pydata/xarray/pull/7461

huard:
Security fixes are going to be "provided irregularly on an “as-needed” basis until October 2024." Unless supporting 3.8 proves problematic, I would defer dropping support to as late as reasonable. Updating environments is not pleasant when you have a working setup.

Zeitsperre:
Sphinx just dropped Python3.8 support: https://www.sphinx-doc.org/en/master/changes.html#release-7-2-0-released-aug-17-2023

PR #1381, opened 2023-05-26
Small clean up of dependencies and environment

### LINKED ISSUE #1376: xESMF as indirect dependency
### Generic Issue

* xclim version: after 0.40
* Python version: all
* Operating System: Windows

### Description
Creating the xclim environment installs clisops, clisops depends on `xESMF`. Which means, that env can't be created on Windows.  However, this dependency is very thin, a mixture between legacy and documentation.

I suggest we remove the remaining explicit installation of clisops in `environment.yml`.

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1376 

* Removes `clisops` from the `environment.yml`, which removes the indirect dev dependency to `xESMF`. 
* While missing around I realized we make an explicit import of `filelock`, without listing it in the env (it's a dep of `tox`).
* `statsmodels` was in the "dev" deps section, but is a hard dep.
* On my machine at least, the pin of `pylint` was blocking the creation of an env with python 3.11. I removed it and it seems to work. I modified the rc file slightly to suppress one warning in the output. And, comble de l'ironie, running `pylint` in my older env (python 3.10, pylint 2.14) failed... So I leave the study of why it was pinned and if it is safe to unpin it to @Zeitsperre .

No

### Other information:
This might be a step towards #1268 !

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,7 @@ Internal changes
 * In order to ensure documentation can be rebuilt at a later time, errors raised by `sphinx` linkcheck are now set to be ignored when building the documentation. (:pull:`1375`).
 * With the publication of `xclim`, the code repository now offers a `CITATION.cff` configuration for users to properly cite the software (APA formatted and raw BibTeX) for academic purposes. (:issue:`95`, :pull:`250`).
 * Logging messages emitted when redefining units via `pint` (caused by `logging` interactions with dependencies) have been silenced. (:issue:`1373`, :pull:`1384`).
+* Fixed some annotations and dev dependencies issues to allow the development of xclim inside a python 3.11 environment. (:issue:`1376`, :pull:`1381`).
 
 v0.43.0 (2023-05-09)
 --------------------
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
     - conda-forge
     - defaults
 dependencies:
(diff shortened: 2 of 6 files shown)


Zeitsperre:
I just landed on this post thinking, "will this work, maybe?" https://github.com/tox-dev/sphinx-autodoc-typehints/issues/186#issuecomment-1450856669

It was me. I wrote the post. ...

Zeitsperre:
Theory: My workaround to deal with the fact that we can't supply `Parameter` to a static type definition (by just making it a string) is ignored under Python3.11.

There's a possibility that this could be fixed with the latest sphinx/sphinx-autodoc-typehints, but we can't install it due to limitations with sphinx_rtd_theme (it's effectively unmaintained). Changing to a different theme (like Furo, used in Miranda) involves some changes to our sphinx configuration, but maybe it would be worthwhile investigating soon? This merits its own issue, tbh.

In any case, problems resolved. It had nothing to do with the pylint pin.

Zeitsperre (approved):

Zeitsperre:
Update: The warnings appear to be due to an `__annotations__` quirk: https://docs.python.org/3/howto/annotations.html#annotations-quirks

Zeitsperre:
All this needs is an entry under `CHANGES.rst` and we can merge!

PR #1383, opened 2023-05-30
Faster escore

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* correlation functions in `_escore` now have an additional loop that basically replaces the numpy operation: `v1[:] - v2[:]`. This makes the function run much faster (~ 30 times faster)

No

### Other information:

There are some instances where nopython has a harder time for some numpy functions: https://stackoverflow.com/questions/42846313/numba-nopython-mode-slower-than-object-mode, this could be related.

### DIFF
--- a/xclim/sdba/nbutils.py
+++ b/xclim/sdba/nbutils.py
@@ -5,7 +5,7 @@ Numba-accelerated Utilities
 from __future__ import annotations
 
 import numpy as np
-from numba import boolean, float32, float64, guvectorize, njit
+from numba import boolean, float32, float64, guvectorize, jit, njit
 from xarray import DataArray
 from xarray.core import utils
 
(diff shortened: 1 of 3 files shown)


xclim/sdba/nbutils.py
```diff
@@ -107,6 +107,7 @@ def remove_NaNs(x):  # noqa
     return x[:, ~remove]
 
 
+# This is now unused
 @njit(fastmath=True)
 def _euclidean_norm(v):
     """Compute the euclidean norm of vector v."""
```
coxipi, line 113: I think this is not used elsewhere, not sure if we should keep it in case

aulemahal (approved):
The performance boost is quite counter-intuitive, but it works!
Well done!

Issue #1290, opened 2023-02-02
Add Hardiness Zones and generic functions for calculating climatologies


Zeitsperre:
Offshoot issue from #355, related to #1273 

The Hardiness Zones indicator (https://en.wikipedia.org/wiki/Hardiness_zone) would be a nice addition to the set of agriclimatic indicators on offer here. The metric helps determine climate-constrained agricultural suitability using a combination of climate factors. USDA is the most common/simple one (https://www.jstor.org/stable/23298457), but there are other versions, including a Canada-specific version developed by Natural Resources Canada (https://doi.org/10.4141/cjps67-044).

The intriguing part of this indicator is that it requires calculations/approaches that could also be applied for calculating SPI/SPEI and other indicators that are currently implemented in https://github.com/monocongo/climate_indices. It might be worthwhile to work on this indicator in part to achieve the objectives of #1273 (or vice versa).

aulemahal:
I had the following idea: we could think of "hardiness zones" as a climate _classification_, instead of an indicator.

Thus, it could be implemented under a new submodule of "climate classifier", which I would define as:
- Inputs are indicators, maybe even statistics of indicators (ex: 30-yr rolling mean)
- The `time` dimension is not special, classification does not perform temporal aggregation.
- Output is _discrete_
    - We could maybe have a special way of outputting correct [CF attributes for flags](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#flags).

In addition to `hardiness_zones`, we could have:
- `köppen` (yes with the umlaut)
- any others found here : https://en.wikipedia.org/wiki/Climate_classification

Zeitsperre:
Moving this from milestone. Hardiness indicators are ready, but an agreed-upon approach for calculating climatologies requires a bit more effort.

PR #1396, opened 2023-06-16
Hardiness zones

Zeitsperre:
### Pull Request Checklist:
    - This PR addresses part of #1290 

* Adds the Hardiness Zones index with options for both Australian and USDA classifications
* Adds separate indicators for each classification scheme.

No.


tests/test_indices.py
```diff
@@ -3371,6 +3371,30 @@ def test_dryness_index(self, atmosds):
...
+        (-47, "AU", np.NaN),
+    ],
+)
+def test_hardiness_zones(tasmin_series, tmin, meth, zone):
+    tasmin = tasmin_series(
+        np.zeros(10957) + K2C + 20, start="1997-01-01"
+    )  # 30 years at 20°C
+    tasmin = xr.where(
+        tasmin.time.dt.dayofyear == 1, tmin + K2C, tasmin, keep_attrs=True
+    )
```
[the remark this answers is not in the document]
Zeitsperre: I landed on:
```python
    tasmin = xr.where(
        tasmin.time.dt.dayofyear != 1, tasmin, tmin + K2C, keep_attrs=True
    )
```
But I like yours better.

xclim/indices/_agro.py
```diff
@@ -1418,3 +1419,75 @@ def effective_growing_degree_days(
...
+        Values are given at the end of the averaging window.
+
+    References
+    ----------
+    :cite:cts:`usda_2012`
+    :cite:cts:`dawson_plant_1991`
+    """
+    # Construct climatology
+    if parse_offset(freq) != (1, "A", True, "JAN"):
+        raise ValueError(f"Freq not allowed: {freq}. Must be `YS` or `AS-JAN`")
```
huard, line 1459: My understanding is that these checks are done at the indicator level using the `allowed_period` attribute.

xclim/data/fr.json
```diff
@@ -553,6 +553,18 @@
     "comment": "Formule originale prise de Jackson, D. I., & Cherry, N. J. (1988)",
     "abstract": "Indice climatique basé sur la température moyenne du mois le plus chaud et un coefficient basé sur la latitude pour tenir compte de la durée plus longue du jour favorisant les conditions de croissance. Développé spécifiquement pour la viticulture. Température moyenne du mois le plus chaud multipliée par la différence de coéffient de facteur latitude moins la latitude."
   },
+  "AUSTRALIAN_HARDINESS_ZONES": {
+    "long_name": "Échelle des températures minimales annuelles moyennes pendant une période donnée (Australian National Botanic Gardens)",
+    "description": "Indice climatique basé sur une moyenne mobile de la température annuelle minimale de {window} années. Développé spécifiquement pour aider à déterminer l'adéquation des plantes aux régions géographiques. Le système de classification des jardins botaniques nationaux australiens divise les catégories en zones de 5 degrés Celsius, allant de moins de -15 degrés Celsius à plus de 20 degrés Celsius.",
```
huard, line 558:
```suggestion
    "description": "Indice climatique basé sur une moyenne mobile de {window} années du minimum annuel de la température minimale. Développé spécifiquement pour aider à déterminer l'adéquation des plantes aux régions géographiques. Le système de classification des jardins botaniques nationaux australiens divise les catégories en zones de 5 degrés Celsius, allant de moins de -15 degrés Celsius à plus de 20 degrés Celsius.",
```

xclim/indices/_agro.py
```diff
@@ -1418,3 +1419,75 @@ def effective_growing_degree_days(
...
+
+    tnmin = tasmin.resample(time=freq).min().rolling(time=window).mean()
+
+    if method.lower() == "usda":
+        tnmin = convert_units_to(tnmin, "degF")
+        bins = np.append(
+            np.insert(np.arange(-65, 66, 5).astype("float"), 0, -np.inf), np.inf
+        )
```
huard, line 1467: I suspect that for very large datasets, it would be faster to convert the units of the bins to the tasmin units instead.

tests/test_indices.py
```diff
@@ -3371,6 +3371,28 @@ def test_dryness_index(self, atmosds):
...
+    "tmin,meth,zone",
+    [
+        (-6, "usda", 9.0),
+        (19, "usda", 13.5),
+        (-47, "usda", 1.5),
+        (-6, "anbg", 2),
+        (19, "anbg", 7),
+        (-47, "anbg", np.NaN),
+    ],
+)
```
huard, line 3384: Indicate in comment how the expected values are determined.

docs/references.bib
```diff
@@ -1922,3 +1922,28 @@ @article {frei_snowfall_2018
...
+
+@misc{usda_2012,
+	title = {{USDA} {Plant} {Hardiness} {Zone} {Map}},
+	url = {https://planthardiness.ars.usda.gov/},
+	abstract = {The 2012 USDA Plant Hardiness Zone Map allows gardeners and growers to determine which plants are most likely to thrive at a location. The map is based on the average annual minimum winter temperature, divided into 10-degree F zones.},
+	language = {en-ca},
+	urldate = {2023-06-16},
+	author = {{USDA Agricultural Research Service}},
+	year = {2012},
+	file = {Snapshot:/home/tjs/snap/zotero-snap/common/Zotero/storage/XT7HZ9YG/planthardiness.ars.usda.gov.html:text/html},
```
huard, line 1948: Remove

xclim/indicators/atmos/_temperature.py
```diff
@@ -1258,3 +1260,45 @@ def cfcheck(self, tas, snd=None):
...
     parameters={"lat": {"kind": InputKind.VARIABLE}, "lat_factor": 60},
 )
+
+
+australian_hardiness_zones = Temp(
+    title="Australian hardiness zones",
+    identifier="australian_hardiness_zones",
+    units="",
+    long_name="Australian National Botanic Gardens scale of average annual minimal temperatures over given time period",
+    description="A climate indice based on a rolling average of minimal annual temperature of {window} years. "
```
huard, line 1270:
```suggestion
    description="A climate indice based on a {window} years rolling average of the annual minimum temperature . "
```

huard (changes requested):

xclim/indicators/atmos/_temperature.py
huard: To validate to make sure the formulation is understood.

xclim/data/fr.json
huard: À valider

xclim/indices/_agro.py
Zeitsperre: Yes. I believe the issue is that if `freq` is not also present in the indice, we can't add `allowed_freq` to the wrapped indicator. This is also relevant for the indice as well (hardiness at the monthly timestep makes no sense).

I copied this from elsewhere, but something I'd like to confirm is if this prevents us from using the different annual start dates (`AS-JUL`). This should be allowed.

docs/references.bib
huard: I mean the `file` entry.

xclim/indices/_agro.py
huard: Indicator only checks if the base (here "A") is a match. 

See `ResamplingIndicator._preprocess_and_checks`

coxipi: `xc.core.calendar.parse_offset("AS-JUL")` yields `(1, 'A', True, 'JUL')`, so you will indeed raise a ValueError and "AS-JUL" won't be permitted as things are. If you just want yearly frequency, maybe you could use:
```python
_, base, _, _ = parse_offset(freq)
if base != "A":
    ...
```

xclim/indices/_agro.py
coxipi: That's a good point. We should probably be careful with the size of bins. To conserve the size of bins: 
```
        b0, b1 = "-65 degF", "65 degF"
        bins = np.linspace(convert_units_to(b0,tmin), convert_units_to(b1,tmin), int(1+(2*65)/5))
        bins = np.append(np.insert(bins, 0, -np.inf), np.inf)
```
and similarly for the degC part

coxipi: A more similar implementation with more call to `convert_units_to`  could be: 
```
bins = np.arange(-65, 66, 5)
bins = np.array([convert_units_to(str(bi) + " degF", tmin) for bi in bins])
bins = np.append(np.insert(bins, 0, -np.inf), np.inf)
```

xclim/indices/_agro.py
```diff
@@ -1418,3 +1419,75 @@ def effective_growing_degree_days(
...
+        bins = np.append(
+            np.insert(np.linspace(b0, b1, int(1 + (20 + 15) / 5)), 0, -np.inf), np.inf
+        )
+
+        def _zones(arr, bs):
+            c = np.digitize(arr, bs).astype(float)
+            c[c == 0] = np.nan
+            c[c == bs.size] = np.nan
+            return c
+
```
coxipi, line 1484: Taken USDA example as reference:
* Zone 1a is used for example in Alaska
* I don't see that Zone12a to 13b are used at all. 

For zones colder than 1a := 1rst zone, could we attribute the value 0? 
For zones hotter than 13b:= 26th zone, could we assign the value 27?

With window=30, we probably won't reach such extremes, but window=1 is a possible input, and we have seen temperature extremes (below the -60degF) that would be NaN in this case. Is there a reason it's better to put NaNs?

coxipi:
Currently,  70degF for USDA et  20 degC for anbg are not included in last zone, because bins are closed on left and open on right. Should I hack things so that the very last bin is also closed on the right or leave things are they are?

coxipi:
Also, angb zones were noted with half-integers in the previous code, but looking at https://www.cpbr.gov.au/gardens/research/hort.research/zones.html, I see 7 zones of 5 degC, noted 1,2,...,7. Not sure if I missed something?

Zeitsperre:
> Currently, 70degF for USDA et 20 degC for anbg are not included in last zone, because bins are closed on left and open on right. Should I hack things so that the very last bin is also closed on the right or leave things are they are?

If we want to follow the definition closely, it would be good to close them.

> Also, "angb" zones were noted with half-integers in the previous code, but looking at https://www.cpbr.gov.au/gardens/research/hort.research/zones.html, I see 7 zones of 5 degC, noted 1,2,...,7. Not sure if I missed something?

No, there were just a lot of problems in my code. Sorry!

xclim/data/fr.json
Zeitsperre: https://planthardiness.ars.usda.gov/: "The map is based on the average annual minimum winter temperature, divided into 10-degree F zones and further divided into 5-degree F half-zones"

Zeitsperre: From the website:

> "The statistic used by the USDA is the average annual minimum temperature. This causes immediate confusion. In Australia we use this term to mean the average minimum temperature over the whole year, whereas the USDA means the average, over ten or more years, of the very lowest temperature observed for each year for each meteorological station. I prefer to call the USDA term the average annual lowest temperature."

I suppose this suggests that the averaging window should be `1` for anbg`?

xclim/indicators/atmos/_temperature.py
Zeitsperre: See note about the averaging window for method `anbg`.

xclim/data/fr.json
Zeitsperre: Perhaps we could address this in the wrapped indicator with:
```python
    parameters={
        "method": "anbg"
        "window": {"default": 1}
    }
```

xclim/data/fr.json
```diff
@@ -553,6 +553,18 @@
...
+    "title": "Zones de rusticité australiennes",
+    "abstract": "Indice climatique basé sur une moyenne mobile de la température annuelle minimale d'un nombre donné d'années. Développé spécifiquement pour aider à déterminer l'adéquation des plantes aux régions géographiques. Le système de classification des jardins botaniques nationaux australiens divise les catégories en zones de 5 degrés Celsius, allant de moins de -15 degrés Celsius à plus de 20 degrés Celsius."
+  },
+  "USDA_HARDINESS_ZONES": {
+    "long_name": "Échelle de l'USDA des températures minimales moyennes pendant une période donnée",
```
aulemahal, line 563:
```suggestion
    "long_name": "Zones de rusticité",
```
Les long_name gagnent à rester court.

xclim/indices/_agro.py
```diff
@@ -1418,3 +1419,52 @@ def effective_growing_degree_days(
...
+    References
+    ----------
+    :cite:cts:`usda_2012`
+    :cite:cts:`dawson_plant_1991`
+    """
+    tnmin = tasmin.resample(time=freq).min().rolling(time=window).mean()
+
+    if method.lower() == "usda":
+        zones = get_zones(tnmin, "-60 degF", "70 degF", "5 degF")
```
aulemahal, line 1460:
```suggestion
        zones = 0.5 + get_zones(tnmin, "-60 degF", "70 degF", "5 degF") / 2
```
Non ?
That's what the docstring says and what feels more natural to me.

aulemahal (commented):
Haha, this is what we could call a complex simplification.

xclim/indices/_agro.py
Zeitsperre: This is covered in the wrapped indicator. We need to trust that users won't compute something that makes no sense with the index.

xclim/indices/_agro.py
Zeitsperre: I think we opted not to do this, since we add a coordinate that explains the zones in the returned DataArray.

xclim/indices/generic.py
```diff
@@ -925,17 +925,32 @@ def get_zones(
...
-    zone_next = f"{str2pint(zone_min).magnitude + str2pint(step_size).magnitude}  {pint2cfunits(str2pint(zone_min))}"
+    zone1 = zone_1 or zone_min
+
+    mn_p, mx_p, step_p, z1_p = (
+        str2pint(v) for v in [zone_min, zone_max, step_size, zone1]
+    )
+    units = [pint2cfunits(p) for p in [mn_p, mx_p, step_p, z1_p]]
+    if len(set(units)) != 1:
+        msg = f"Got different units among `zone_min` ({units[0]}), `zone_max` ({units[1]}), `step_size` ({units[2]})"
+        warnings.warn(msg + (f", `zone_1 ({units[3]})`." if zone_1 else "."))
```
Zeitsperre, line 942: Great!

aulemahal: Why a warning ? Isn't this fully supported by line 94 below ?

Zeitsperre: I asked Éric to add this. While it's fully supported (and who knows, maybe applicable to some studies?), I thought it would be good to have a warning for users to be aware that they're mixing units in their thresholds.

xclim/data/fr.json
coxipi: Therefore, we would be allowing a bit more general behaviour with the indice, since `window` could be changed even though we choose "usda" or "anbg". So, we allow small variations of the definition of the indices only though the use of the index. I think it could make sense indeed to allow more general behaviour with the index name `hardiness-zones`, which is less specific than the indicators : `usda-hardiness-zones` and so on. 

However, that is a weird control on the behaviour though, because we can only change the window, and moreover this is allowed to be changed only when a specific method is chosen (which in principle, already has a defined window). Maybe we should put the zone definitions (`zone_min`, `zone_max`, `step_size`) in the index signature, and allow to either call with "method==usda" or specify (`zone_min`, `zone_max`, `step_size`, `window`)? Another idea is to simply remove `method` from the index and control this exclusively through the indicator (by choosing appropriate  (`zone_min`, `zone_max`, `step_size`, `window`) ) , although I'm not sure this is similar to other comparable indicators in xclim, I'd have to compare with other cases

xclim/indices/generic.py
coxipi: Now that you ask, I'm also on the fence regarding the need for a warning. There are some instances where the same kind of behaviour is allowed in xclim. In this specific case, I do agree though, having different units for such closely related variables would make you wonder if the user has made an error of some sort. Perhaps it's not the kind of warning we want to see in xclim though.

As a comparison, `heat_wave_XYZ` has two thresholds (`thresh_tasmax` & `thresh_tasmin`). Although giving two thresh in "degC" and "degF" would be weird, it does not raise any warning.

aulemahal: I think a user error would look like a misspelled unit (genre :  `deg C`), but that would raise an error anyway. Passing different units seems to me much more likely to be deliberate.

xclim/data/fr.json
aulemahal: If I may, I suspect you are going too deep for this. Not that the generalization will never be useful, but... We only have two "zone indices" examples, this is not much to plan a generalization. You might be able to think out a plan and get something that looks flexible, but what if the next similar index does not fit ? I would suggest sticking to something simple and a bit specific, waiting for new similar indices to come along before generalizing.

xclim/indices/generic.py
Zeitsperre: Fair enough. We don't do that elsewhere, really. I think Pascal is right that we need to trust that the user knows what they're doing.

Sorry Éric, thanks again for adding it, but we should probably roll that back.

coxipi:
In `get_zones`,   `zone_min`, `zone_max` and `step_size` could have conflicting values if (zone_max - zone_min)/step_size is not a integer. Should I raise an error, or give a warning and extend zone_max?

xclim/data/fr.json
Zeitsperre: I think the complication here is that hardiness zones can become very specialized. That said, the binning logic proposed here is useful for all of them.

One would think that the Canadian Hardiness Zones would the next logical step, but it appears to be quite complex ([Wikipedia](https://en.wikipedia.org/wiki/Hardiness_zone#Canadian_hardiness_zones); [source article](https://cdnsciencepub.com/doi/10.4141/cjps67-064))

The Wikipedia page for hardiness zones also makes mention of the American Horticultural Society's [Heat Zones](https://ahsgardening.org/about-us/news-press/cool_timeline/heat-zone-map-developed/). This metric uses the average number of days exceeding a given threshold, much less complicated, but unfortunately, it seems this was never published academically, so I'm not sure how much effort should be spent including it.

In both cases, the zoning logic is something that could be very useful. In fact, many of the agricultural indices (Winkler and Huglin indices, Cool Night index, Dryness Index, others) could benefit from the averaging and zoning logic. I'm tempted to build on top of them with this new generic index, honestly. 

I think we should start somewhere, though. As far as I'm concerned, this implementation works as expected and is not overly-complicated. The indicator is clear enough for general users as well. Other thoughts?

xclim/indices/generic.py
```diff
@@ -945,7 +945,7 @@ def get_zones(
         Size of zones
     zone_1 : Quantified | None
         Input value whose zone should be labelled as "zone 1". It should be in the range [`zone_min`, `zone_max` [.
-        By default this is set to `None`, with "zone 1" starting `zone_min`
+        By default this is set to `None`, with "zone 1" aligned with `zone_min`
     bins : Quantified | None
```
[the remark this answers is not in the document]
coxipi: Isn't the same?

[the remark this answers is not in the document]
coxipi: It should be the same as a threshold in other indices e.g. "70 degC" and `Quantified` seems to be used in this context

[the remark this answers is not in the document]
Zeitsperre: Oh geez, maybe. I keep getting confused by that variable type. Hold on.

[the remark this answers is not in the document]
Zeitsperre: Oh sorry! I'm speaking specifically to `bins`!

I can see that `Quantified` is `xr.DataArray | str | Quantity`, but if we want to provide the option for users setting their own `bins`, this would be provided as `list[str | Quantitiy]`, no?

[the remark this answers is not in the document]
aulemahal: I suggest typing as `xr.DataArray` because it has to be a sequence. `Quantified` is for values that have units and can be of any dimensionality (scalar, vector, matrix, tensor, etc). Usually, when its not a scalar, the dimensions should not have any effect on the computation (like it should not be `time` in a normal indicator). 

But here, accepting `list[Quantified]` kinda means accepting a list of strings with different units. Sticking to an optional DataArray clears the possible confusion.

Also, "Optional[DataArray]" is a xclim-known type, which is good if it ever appears in an indicator's signature.

xclim/data/fr.json
aulemahal: I would say yes. May be it could be `Zones de rusticité (USDA)` and `Zones de rusticité (ANBG)` in that case ?

xclim/indices/generic.py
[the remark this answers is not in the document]
coxipi: I was the one confused after all!

What I had in mind for bins is really xr.DataArray | list : 
1. xr.DataArray([-60,-55, ... , 70], attrs={"units":"degF"})
2. [-60,-55, ... , 70].  units are the same as `da`

The code could easily accomodate  list[Quantity] e.g. ["5 degC", "10 degC", "10 degC", "20 degC"], but maybe we don't want that as you said. We do accept `zone_min` and `zone_max` as strings with different units though. Should we restrict this?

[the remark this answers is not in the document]
Zeitsperre: I think we're using `Quantity` and `Quantified` interchangeably here. I think having a list of bins with different units should be acceptable (or, at the very least, supported).

My understanding here is that we can potentially accept the following:
1. `xr.DataArray`
2. `list[str]` -> ["5 degC", "10 degC", "30 degF", "20 K"]
3. `list[pint.Quantity]` -> `[5 * pint.ureg.kelvin, 5 * pint.ureg.celsius]`

I hate the 3rd option, but my understanding is that it is _technically_ supported by `convert_units_to()`. Is this correct?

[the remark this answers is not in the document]
coxipi: I think keeping `zone_min`, `zone_max` makes sense, invoke them as we do with thresholds, allow different units. 

For bins, xr.DataArray makes more sense. List would be a nice shorthand, but not essential. list[Quantity] seems like a generalization of `zone_min` , `zone_max`, `zone_step` to [`zone_a`, `zone_b`, ..., `zone_z`], but at this point, I'm not sure if it could benefit anybody. I think most people would use: `xr.DataArray([value_a,value_b, ...,], attrs={"units":units})`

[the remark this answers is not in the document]
Zeitsperre: I agree. I think it's probably enough to support `zone_min` and `zone_max` with different units, but we should ask that `bins` consistently use the same unit. It's very difficult to imagine cases where a user would want something more than that.

We can let them open up an issue if ever that happens, which will likely be "never" haha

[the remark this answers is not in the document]
coxipi: Ok, let me check what is done elsewhere in the code. I just have the feeling this would be much more flexible than what is otherwise allowed elsewhere, but maybe it's not comparable

xclim/data/fr.json
Zeitsperre: > From the website:
> 
> > "The statistic used by the USDA is the average annual minimum temperature. This causes immediate confusion. In Australia we use this term to mean the average minimum temperature over the whole year, whereas the USDA means the average, over ten or more years, of the very lowest temperature observed for each year for each meteorological station. I prefer to call the USDA term the average annual lowest temperature."
> 
> I suppose this suggests that the averaging window should be `1` for anbg`?

Further down:

> "I have used the same statistic for Australia, but rather than use US zones directly I have modified the limits for each zone."

@huard 

Despite the confusing way this was written, I believe that the ANBG method is meant to be the exact same as that of USDA (mean of annual tn_min over years), averaging window included. I'll fix this up in the code and resolve this!

huard: I was mostly worried about the way the algorithm was translated into words (not the other way around), and made the change suggestion only for one both assuming they would apply to both.

xclim/indices/generic.py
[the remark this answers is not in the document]
coxipi: Ok, implemented `bins : xr.DataArray | list[Quantity] | None`. 

I agree now that a simple list without units was not a good idea. 

As for `xr.DataArray` vs. ` xr.DataArray | list[Quantity]`, I think there is merit to both. We can build indicators with `Optional[xr.DataArray]` and use bins in this way later. Once it's converted, I sort the "list[Quantity]", assuming a user could input a list of bins not necessarily ordered. I can revert changes if we don't agree on this.

xclim/indices/generic.py
```diff
@@ -897,3 +898,103 @@ def first_day_threshold_reached(
...
+    ----------
+    da : xarray.DataArray
+        Input data
+    zone_min : Quantity | None
+        Left boundary of the first zone
+    zone_max : Quantity | None
+        Right boundary of the last zone
+    zone_step: Quantity | None
+        Size of zones
+    zone_1 : Quantity | None
```
huard, line 946: The zone_1 thing is not super clear to me.

huard (approved):

xclim/indices/_agro.py
```diff
@@ -1418,3 +1419,53 @@ def effective_growing_degree_days(
...
+
+    elif method.lower() == "anbg":
+        zone_min, zone_max, zone_step = "-15 degC", "20 degC", "5 degC"
+
+    else:
+        raise NotImplementedError(
+            f"Method must be one of `usda` or `anbg`. Got {method}."
+        )
+
+    tn_min = tasmin.resample(time=freq).min().rolling(time=window).mean()
```
tlogan2000, line 1467: Do we not want to take min only in the time dime `min(dim='time')`?

Zeitsperre: We do.

xclim/indices/generic.py
coxipi: Take for example USDA method
in this case, `bins = [-60, -55, -50, ... 70]`

`[-60 degF, -55 degF[` is the first bin. By default, `np.digitize(-58, bins) ==1`
`[-55 degF, -50 degF[` is the second bin. `np.digitize(-52, bins) ==2`

This was an option to let choose which bin is labelled as "1". I think there were some dicussions allowing lower temperature at some point. This would let for instance extend the USDA zones to lower temperatures while keeping the original ordering. One would set the first bin to [-60 degF, -55 degF[. A lower bin [-65, -60] would then have index "0". THe lowet one "-1", etc.

This could be taken care of at higher level if it's better.

xclim/indices/_agro.py
tlogan2000: In the end this is exactly `xc.indices.tn_min(time=freq).rolling(time=window).mean()`

xclim/indices/generic.py
huard: What throws me off here is that zone_1 is a Quantity. What you're describing feels more like an integer offset.

xclim/indices/_agro.py
Zeitsperre: Essentially, haha

tlogan2000: I think xarray's behavior might have changed... `.min()` seems to only average across the resampled dim as wanted.  I thought you were going to end up averaging over all dims ... been a while since I coded an xclim index I guees

xclim/indices/generic.py
coxipi: Yes, that's what I had in mind by higher-level treatment, just take zones -> zones + zone_offset. 

Implementing this in the function would just be `return zones + zone_offset` instead of `return zones`. Is  it still worth keeping it?

huard: Since this logic is not user facing, I'm not sure the added complexity is worth it. 
You could keep the function private if we're not sure how exactly how it should be used and what signature it should have.

PR #1406, opened 2023-06-28
Convert doys

### LINKED ISSUE #1283: `convert_calendar` could convert day-of-year data
### Addressing a Problem?

Some indicators return `day-of-year` data. The values are thus constrained by the calendar of the input. A `360_day` timeseries will have a maximal value of 360. When mixing data from different calendar, like when doing ensemble statistics, this introduces a bias in the statistics. For example, if all members agree that this event happens on the last day of the year, some will have 360, some will have 365 and some 366. The mean will be lower than 365, and the standard deviation will be non-zero, even though every model agrees. 

### Potential Solution

`convert_calendar` could detect `day-of-year` data through the `is_dayofyear` attribute that xclim adds on such cases.

The basic conversion could be : `doy_out = out.time.dt.days_in_year * doy_in / doy_in.time.dt.day_in_year`.

### Additional context

Caveats:
1) This introduces floating point day of year, which might not be fully expected in some cases.
2) For 360_day calendar, this is equivalent to the `align_on='year'` timestamp conversion method. What about the `align_on='date'` ? That would need a mapping of doy values, not so complex, just a bit heavier. And should the user be able to decide on which method to use separately on the time coordinate and on the data?
3) `is_dayofyear` is xclim-specific. There is no CF standards about this, AFAIK. This would not be portable to `xarray`'s `convert_calendar`.

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1283 

* New `convert_doy` function to convert the calendar reference of doy data. Two options :
    - "year" : doy is seen as the fraction of the year. I.e. : 360 in `360_day` converts to 365 in `noleap`.
    - "date" : doy is seen as a date (MM-DD). 360 in `360_day` converts to 364 in `noleap` (December 30th).
* Convert calendar can automatically apply `convert_doy` if its new argument `doy=True` (or `doy='date'`).
* `convert_doy` will NOT convert the time coordinate.

No.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+v0.45.0 (unreleased)
+--------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* New function ``xclim.core.calendar.convert_doy`` to transform day-of-year data between calendars. Also accessible from ``convert_calendar`` with ``doy=True``. (:issue:`1283`, :pull:`1406`).
+
 v0.44.0 (2023-06-23)
 --------------------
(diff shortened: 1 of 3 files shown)


Zeitsperre (approved):
LGTM. Well done!

PR #1407, opened 2023-06-29
Adapt freq in training

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Frequency adaptation should sometimes be performed in the same step as training so that adapted grouped block are preserved for the training step. This PR allows this by invoking frequency adaptation directly in training functions.
* nbstripout in pre-commit now removes the metadata.kernelspec completely, which are annoying differences that come up when editing notebooks with different software (I think). This might be a problem if we want to mix different cell codes, e.g. python and R in a single notebook. If our use case one day evolves towards this, we can revisit this aspect.

Yes. The adaptation "in-training" is simply a new feature. But the adaptation "out-of-training" (or "modular" approach), akin to the old procedure, will also yield different results. This is because a multi-dimensional rank is now used in "_processing.py". 

### Other information:
I explain the motivation behind this PR (which is also explained below).

Using multi-dimensional rank changes the "out-of-training" adaptation (which already is "more" in lined with the expected scientific computation). But these results are "more" in-line with the scientific computation expected.  I say "more" in-line because if you really want the expected results, the adaptation should be done "in-training" so that each map block remains correctly adapted during the training. Otherwise, with the "out-of-training" adaptation we used to do: expand your data in blocks; frequency adapt them; de-expand to original data size; re-expand in blocks; train them. In the re-expansion, you could still have blocks that are not well adapted.


xclim/sdba/_processing.py
```diff
@@ -105,6 +105,94 @@ def _adapt_freq(
...
+
+
+def _adapt_freq_s(
+    ds: xr.Dataset,
+    *,
+    dim: Sequence[str],
+    thresh: float = 0,
+) -> xr.Dataset:
+    r"""
+    Adapt frequency of values under thresh of `sim`, in order to match ref.
```
coxipi, line 126: I defined `_adapt_freq_s` which is similar to `_adapt_freq.func`, with 2 changes:
* Rank can work on multiple dimensions (i.e. time & window)
* The `window` dimension is not removed from the output dimensions

Maybe this logic could be incorporated directly `_adapt_freq` with a boolean switch.

coxipi: I feel that reducing the window dimension in `_adapt_freq.func` is not what would be expected when `.func` is invoked? My `_adapt_freq_s` is what I expected `_adapt_freq.func` would do for me.

xclim/sdba/_processing.py
```diff
@@ -105,6 +105,94 @@ def _adapt_freq(
...
+                (rank < P0_ref) | (rank > P0_sim),  # Preserve current values
+                # Generate random numbers ~ U[T0, Pth]
+                (pth.broadcast_like(sim) - thresh)
+                * np.random.random_sample(size=sim.shape)
+                + thresh,
+            ),
+        )
+    return sim_ad
+
+
```
coxipi, line 195: No particular reason for not outputing `P0, pth`. This can be fixed once we decide the structure we want elsewhere

coxipi:
Just to share some motivations for this change: 

Adapt_freq modifies `hist` so that the dry-day frequency is equal to that of `ref` (`P0>0`), or if it's already smaller (`P0<=0`), then do nothing.
```
adapt_freq: ref, hist -> ref, hist'
```
Once this is done, we could re-apply the function and expect nothing to happen, since the frequency is already fixed.
```
adapt_freq: ref, hist' -> ref, hist'
```

This is not what happens when there is a window dimension:
```python
group = sdba.Grouper.from_kwargs(**{"group": "time.dayofyear", "window": 31})["group"]
adapt_freq = {"thresh":"1e-6 kg m-2"}
adapt_freq.setdefault("group", group)
hist, pth, dP01 = sdba.processing.adapt_freq(ref, hist, **adapt_freq)
hist, pth, dP02 = sdba.processing.adapt_freq(ref, hist, **adapt_freq)
```
image


This happens because we don't keep the `window` dimension explicitly. The comparison between `ref` and `hist` is done with expanded datasets `(window, time)`, but the correction is only done on the reduced dataset with `(time)`.   As a result, we can apply some corrections on a given doy, but on the second use of the function, as we re-expand with `(window, time)`, the can still be the same problems as before.

One possible fix is to simply do frequency adaptation with the window argument. This ensures that each doy is well adapted, so each block will be too. This result in potentially more modifications of the datasets on smaller samples. 

The other fix (proposed here) is to keep the `(window, time)` structure explicit, so that each block is adapted. As a result, frequency adaptation is less modular, since we want to call this within a map_blocks of the training functions. There is also perhaps a conceptual flaw: Given that adaptation can work differently on each block, it is not guaranteed that each block is coherent with each other. For example, with a `window=31`, the  doy=3, year=1990 `pr` value will be used in 31 blocks. The corrected value could differ from block to block. Granted, this is a mathematical trick, and the corrected value will be small values between 0 and the minimum non-zero value in `pr`. But I think this is still important to highlight.

xclim/sdba/_processing.py
aulemahal: Indeed, it would be cleaner if there was only one `_adapt_freq` func.

You could try removing the `sim = ds.sim.isel(window=(ds.sim.window.size - 1) // 2)` (L77-80) in the original function, as I commented, this was only a performance adjustment.

coxipi: Ok, but that still leaves the lines 
```
        dim = dim[0]

        # Get the percentile rank of each value in sim.
        rank = sim.rank(dim, pct=True)
```
`dim=dim[0]` seems motivated by the fact we preemptively select a window value. I will see what happens when using the multi-dimensional rank

xclim/sdba/_adjustment.py
```diff
@@ -9,21 +9,32 @@
...
+from ._processing import _adapt_freq
 from .base import Grouper, map_blocks, map_groups
 from .detrending import PolyDetrend
 from .processing import escore
 
 
+def _adapt_freq_hist(ds, adapt_freq_thresh):
+    with units.context(infer_context(ds.ref.attrs.get("standard_name"))):
+        thresh = convert_units_to(adapt_freq_thresh, ds.ref)
+    dim = ["time"] + ["window"] * ("window" in ds.hist.dims)
```
coxipi, line 26: Is this hard-coding fine? I'm assuming that hard-coding "window" at the very least is fine, because it was the case in the previous version of `_adapt_freq`. For "time", I could use `Grouper.DIM`

xclim/sdba/_processing.py
```diff
@@ -74,17 +75,19 @@ def _adapt_freq(
...
+            temp_dim = get_temp_dimname(sim.dims, "temp")
+            rank = (
+                sim.stack(**{temp_dim: dim})
+                .rank(temp_dim, pct=True)
+                .unstack(temp_dim)
+                .transpose(*sim.dims)
+                .drop_vars([dim_ for dim_ in dim if dim_ not in sim.coords])
+            )
+        else:
+            rank = sim.rank(dim[0], pct=True)
```
coxipi, line 90: Should I incorporate multi-dim ranking in the `sdba.rank` function?

I'm thinking : 
```python
def rank(...
dim : str | list[str]
)
...
dim = dim if isinstance(dim,list) else [dim]
# then use code above with general `pct` 
```

coxipi:
I should also note: Given the new multi-dimensional rank structure, even if we choose the same individual `window` value at the end, the processing on `sim_ad` is different because the ranks are different on this specific window value. Apparently, even without using the internal call to `adapt_freq` in the training function, the more modular call to `adapt_freq` now does a better job at creating a new `sim_ad` whose adaptation persists if we re-perform the `adapt_freq` routine. 

image

image

xclim/sdba/_adjustment.py
aulemahal: Translation of `Grouper.DIM` takes place in the decorator, so it won't be understood here, directly in the function. But hardcoding `time` is fine. (and window)

xclim/sdba/_processing.py
aulemahal: I think this is a good idea.

xclim/sdba/adjustment.py
```diff
@@ -352,6 +352,7 @@ def _train(
         nquantiles: int | np.ndarray = 20,
         kind: str = ADDITIVE,
         group: str | Grouper = "time",
+        adapt_freq_thresh: str | None = None,
```
aulemahal, line 355: I think this new arg merits documentation in the class doc string above.

aulemahal (approved):
Nice, thanks a lot!

And sorry for not reviewing earlier, it slipped through my radar and under the cracks.

PR #1424, opened 2023-07-24
Fix generic indicators registry entries and tests after sklearn update

### LINKED ISSUE #1423: Generic indicators not recognized as "core xclim indicator" in the registry.
### Setup Information

- Xclim version: master
- Python version: all
- Operating System: all

### Description

When creating a new indicator in a "virtual" module, and pointing the "base" indicator to one of the "xclim.generic.*" indicators, constructing the module fails with:

`ValueError: Requested base class stats is neither in the indicators registry nor in base classes registry.`

### Steps To Reproduce

The following yaml module specification builds "QMOY" from `xclim.generic.stats`.
```yml
realm: land
indicators:
  JQ1MAXHP:
    base: doy_qmax
    input:
      da: discharge
    parameters:
      indexer: {month: [12, 1, 2, 3, 4, 5]}
  QMOY:
    base: stats
    input:
      da: discharge
    parameters:
      op: mean
```
into  `test.yml`.

```
import xclim as xc
xc.build_indicator_module_from_yaml('test.yml')
```

### Additional context

This is because of https://github.com/Ouranosinc/xclim/blob/b96daa2a1d390782e71f07e930499a3b218d298b/xclim/core/indicator.py#L245, where "generic" is absent. Moreover, all "generic" indicators are declared by instantiation of base classes imported from `xclim.core.indicator`. Which means the module assigned by the interpreter will be `xclim.core.indicators`, which will fail the test linked above anyway.

The way we decide which indicator is "core" and which isn't should be changed. For reference, the only difference between both is that "core" indicator are referred to in the registry by their identifier, while the others have a module name prepended (ex: "cf.TMMEAN", "anuclim.P1_ANNMEANTEMP" vs "TG_MEAN"). The registry key is what must be written in the "base:" field of a virtual module. 

The short-term solution is to write "xclim.core.indicator.STATS" to use the generic "stats" indicator.

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1423 

* Adds shallow classes in `xclim/indicators/generic` so that the indicators declared there get a proper `__module__` attribute. 
* Adds `generic` to the list of "realms" that xclim considers "core".
* Add the "generic" indicators to the indicators doc page.
* Relax the friedman-rafsky analog test to avoid an unstable pixel.
* Check the sci-kit learn version in ensemble reduction tests to avoid a failing test with 1.3.0.

No.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ New features and enhancements
 Bug fixes
 ^^^^^^^^^
 * Fix `kldiv` docstring so the math formula renders to HTML. (:issue:`1408`, :pull:`1409`).
+* Fix the registry entries of "generic" indicators. (:issue:`1423`, :pull:`1424`).
 
 Internal changes
 ^^^^^^^^^^^^^^^^
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -81,7 +81,7 @@ def _indicator_table(module):
     return table
 
 
-modules = ("atmos", "land", "seaIce", "cf", "icclim", "anuclim")
+modules = ("atmos", "generic", "land", "seaIce", "cf", "icclim", "anuclim")
 indicators = {module: _indicator_table(module) for module in modules}
 
 # -- General configuration ---------------------------------------------
--- a/docs/indicators.rst
+++ b/docs/indicators.rst
@@ -67,6 +67,27 @@ seaIce: Sea ice
    {% endfor %}
    </dl>
 
+
+generic: Generic indicators
+===========================
+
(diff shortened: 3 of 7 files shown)


Zeitsperre (approved):
Nice

aulemahal:
Oh! haha thanks @Zeitsperre. This is not the reference I was hoping to point to, but it's good enough for the moment. Until I learn how to reference notebook sections in rst files ;).

Zeitsperre:
Woops!

PR #1443, opened 2023-07-27
Refactor indices with generic functions

### LINKED ISSUE #1386: Refactoring `maximum_consecutive_XYZ_days` indices
### Generic Issue

PR #1359 adds many new spell indices. Many indices like: `maximum_consecutive_dry_days` and other similar cases are just special cases of a spell index ([150eede](https://github.com/Ouranosinc/xclim/pull/1359/commits/150eede2c6b559056bd6c80db72b62e9ba99b320)). This might be something to consider as we try to refactor xclim (and perhaps have less indices/ more re-used assets).

#### List of indices 

- [x] `maximum_consecutive_{dry|frost|tx|wet}_days` could use: `{dry|cold|hot|wet}_spell_max_length`
- [x] `maximum_consecutive_frost_free_days`: There is no `frost_free_spell_max_length` but it could be added
- [x] `{tg|tn|tx}_{max|mean|min}` , `max_1day_precipitation_amount`, `{snw|snd}_max`, 

    ```python
    # these are all one-liner functions e.g. tx_min 
    return tasmax.resample(time=freq).min(dim="time").assign_attrs(units=tasmax.units)
    ```
    could use `generic.select_resample_op`. 

### Code of Conduct

coxipi:
### Pull Request Checklist:
    - This PR fixes #1386

* Refactoring + new indices (similar to already existing ones)

`integral` replaces `prod` and `delta_prod` options as `op` arguments in `to_agg_units` 

### Other information:
Few corrections in french translations. Also, `generic.select_resample_op` now keeps attrs in both use cases (`op` is a `str` or a function )


coxipi:
I left out the dry|wet cases because the general indices are actually a bit more complicated, it is not as straightforward to see the simple case from the generalization, I think it could be confusing.

coxipi:
A couple general remarks about what is not covered here, but related:
* Indices like `cold_spell_max_length` and similar allow the choice of any operator. So the condition given as input could be :
`tas < thresh`, or `tas <= thresh` or *even* `tas > thresh`, which is not what we should mean by a cold_spell. It's a bit weird. Also, it's essentially the same function as `hot_spell...` but with different variable. @aulemahal  Your recent PR should allow to write generic spell function, correct? Is this a direction we aim for?
* Right now, I think `dry_spell_max_length`  and `wet_spell_max_length` have the same behaviour, except one uses `>=`, the other `<`. In this case, they also have the same variable `pr`. 

Maybe those ideas could be incorporated in a second round of refactoring following @aulemahal 's PR.

CHANGES.rst
```diff
@@ -4,7 +4,11 @@ Changelog
 
 v0.45.0 (unreleased)
 --------------------
-Contributors to this version: David Huard (:user:`huard`), Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user: `juliettelavoie`), Gabriel Rondeau-Genesse (:user:`RondeauG`), Marco Braun (:user:`vindelico`).
+Contributors to this version: David Huard (:user:`huard`), Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user: `juliettelavoie`), Gabriel Rondeau-Genesse (:user:`RondeauG`), Marco Braun (:user:`vindelico`), Éric Dupuis (:user: `coxipi`).
```
Zeitsperre, line 7:
```suggestion
Contributors to this version: David Huard (:user:`huard`), Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user:`juliettelavoie`), Gabriel Rondeau-Genesse (:user:`RondeauG`), Marco Braun (:user:`vindelico`), Éric Dupuis (:user:`coxipi`).
```

CHANGES.rst
```diff
@@ -27,6 +31,9 @@ Internal changes
 * Increased the guess of number of quantiles needed in ExtremeValues. (:pull:`1413`).
 * Tolerance thresholds for error in ``test_processing::test_adapt_freq`` have been relaxed to allow for more variation in the results. (:issue:`1417`, :pull:`1418`).
 * Added 'streamflow' to the list of known variables (:pull:`1431`).
+* Refactoring (:pull:`1443`, :issue:`1386`) :
```
Zeitsperre, line 34:
```suggestion
* Refactoring of index backend calculations (:pull:`1443`, :issue:`1386`) :
```

xclim/indices/_hydrology.py
```diff
@@ -103,6 +104,27 @@ def rb_flashiness_index(q: xarray.DataArray, freq: str = "YS") -> xarray.DataArr
...
+def snd_max(snd: xarray.DataArray, freq: str = "AS-JUL") -> xarray.DataArray:
+    """Maximum snow depth.
+
+    The maximum daily snow depth.
+
+    Parameters
+    ----------
+    snw : xarray.DataArray
+        Snow depth (mass per area).
+    freq: str
```
Zeitsperre, line 117:
```suggestion
    freq : str
```

xclim/indicators/atmos/_temperature.py
```diff
@@ -932,6 +932,20 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
     parameters={"thresh": {"default": "0 degC"}},
 )
 
+frost_free_spell_max_length = Temp(
```
Zeitsperre, line 935: Needs to be added to `__all__`

Zeitsperre (approved):
Looks good!

aulemahal (commented):
Indeed, this is the aim of my PR : make the "generic" functions complete enough so that we can completely skip/minimize the "indices" part and directly construct indicators.

xclim/indices/generic.py
```diff
@@ -89,8 +89,8 @@ def select_resample_op(
     r = da.resample(time=freq)
     if isinstance(op, str):
         return getattr(r, op)(dim="time", keep_attrs=True)
-
-    return r.map(op)
+    with xr.set_options(keep_attrs=True):
+        return r.map(op)
```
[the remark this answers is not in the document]
aulemahal: Oh I see! Now that this function is used alone in some indices, I think it would make sense for it to manage the output units as well. There are many indices in this PR that could be removed if `select_resample_op` could manage units.

@Zeitsperre  As for preserving the other attributes, wouldn't this go against the usual behaviour of xarray ? It makes sense to me that indices drop the attributes, as this is what most xarray operations do. Users can easily change that with the decorator in their own code.

@coxipi Indeed, `doymax` is a good idea. It would set the xclim-specific `is_dayofyear=np.int32(1)`.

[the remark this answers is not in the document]
aulemahal: Thinking of it, I believe `select_resample_op` should explicitly _remove_ the `units` if "op" is unknown. Or `to_agg_units` should do that, but the goal would be to avoid as much as possible the inadvertent preservation of "units". An output without a units attr will make `declare_units` fail and hopefully, this would happen as early as possible, when developing the indicator.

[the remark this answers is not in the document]
Zeitsperre: @aulemahal Good point. I figured the use of `keep_attrs` above meant that we should be preserving attributes within these operations. Perhaps not neccessary.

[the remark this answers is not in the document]
coxipi: Maybe `doymax` is too specific for the current `to_agg_units`? I don't think it fits with the current paradigm:
```python 
    m, freq_u_raw = infer_sampling_units(orig[dim])
    out = out * m
```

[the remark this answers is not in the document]
aulemahal: Indeed it doesn't, my suggestion is to extend it so it does!

Genre:
```python
def to_agg_units(...):
    if op in ['doy', 'doymax', 'doymin']:
        # put doy stuff
    elif op in ['min', 'max', 'mean']:
        # put same units
    else:
        # to_agg_units as it exists currently
```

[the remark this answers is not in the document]
aulemahal: En fait, I could do this extension in my PR. This PR will work without it.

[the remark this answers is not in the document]
coxipi: I wanted to make sure that `select_resample_op` gives the correct units, so it can be used in indices, so I think I need this, no?

xclim/indices/generic.py
```diff
@@ -88,25 +88,26 @@ def select_resample_op(
...
+    return to_agg_units(out, da, op)
 
 
 def doymax(da: xr.DataArray) -> xr.DataArray:
     """Return the day of year of the maximum value."""
     i = da.argmax(dim="time")
     out = da.time.dt.dayofyear.isel(time=i, drop=True)
-    out.attrs.update(units="", is_dayofyear=np.int32(1), calendar=get_calendar(da))
-    return out
+    return to_agg_units(out, da, "doymax")
```
coxipi, line 103: using `to_agg_units` directly in `doymax`. Does that make things to obscure? I thought it might be better like this to avoid having the same info repeating in two places

xclim/core/units.py
```diff
@@ -581,19 +582,32 @@ def to_agg_units(
...
+    if op in ["min", "max", "mean"]:
+        out.attrs["units"] = orig.attrs["units"]
+
+    elif op in ["doymin", "doymax"]:
+        out.attrs.update(
+            units="", is_dayofyear=np.int32(1), calendar=get_calendar(orig)
+        )
+
+    elif op in ["count", "prod", "delta_prod"]:
+        m, freq_u_raw = infer_sampling_units(orig[dim])
```
coxipi, line 594: Was not sure what you meant by `doy`, but I implemented the rest of functions you suggested @aulemahal

xclim/indices/generic.py
aulemahal: Hum, looks kinda obscure to me as `doymax` is to be applied repeatedly on each resampling group. Seems strange to me that it would set the attributes itself ? (I see that this was the case before your change)

xclim/core/units.py
aulemahal: By "doy" I meant a generic "op" name for all operations that return a doy. Like season starts or ends. So that one day, those functions could make use of `to_agg_units`, and we would really centralized all unit-inferring to a single method.

xclim/indices/generic.py
coxipi: The units are always obvious following the doymax operation no? Is it a question of organizing philosophy, where units should be treated? Maybe I don't understand the subtlety if "applying this repeatedly on each resampling group"

coxipi:
Hmm,  maybe indeed I can see some problem in handling units with `select_resample_op`. For instance: 
```python
def daily_temperature_range(
    tasmin: xarray.DataArray,
    tasmax: xarray.DataArray,
    freq: str = "YS",
    op: str | Callable = "mean",
) -> xarray.DataArray:
    r"""
    ...
    """
    tasmax = convert_units_to(tasmax, tasmin)
    dtr = tasmax - tasmin
    out = select_resample_op(dtr, op=op, freq=freq)
    u = str2pint(tasmax.units)
    out.attrs["units"] = pint2cfunits(u - u)
    return out
```
if we wanted to drop lines setting units, we would need to change `op` to `op_delta`? But then, in `select_resample_op`, we would need to temporarily get rid of `_delta`:

```
    da = select_time(da, **indexer)
    r = da.resample(time=freq)
    if isinstance(op, str):
        out = getattr(r, op.replace("_delta",""))(dim="time", keep_attrs=True)
    else:
            out = r.map(op)
            op = op.__name__
    return to_agg_units(out, da, op)

File: xclim/indices/generic.py (hunk 6 above)
Comment (aulemahal) [reply xclim/indices/generic.py:103]: I meant by that that `doymax` is not a full operation, it's one that gets applied in a `resample` operation. But now that I think of it, there could be use cases where this is not true, and then we would absolutely want the units to be applied in `doymax`. I retract my comment, your implementation makes sense.

**Comment by aulemahal** [comment]
I'm not sure I fully understand your last comment @coxipi. But indeed, there's an issue for "degree days" where the op is "sum", but `to_agg_units` wants "delta_prod". It's not the only case, there's an ambiguity with "sum" : is it an integral or just a sum ?

Do you think the following idea could work : 
- Add "integral" to the valid "op" in `select_resample_op`.
- It gets translated to `sum` in the computation, but not in `to_agg_units`.
- When receiving "integral", `to_agg_units` checks the dimensions of the input units. If it is a temperature, the "delta_prod" method is taken, otherwise we choose "prod". 
Thus  "delta_prod" and "prod" are deprecated in favor of "integral" + automatic decision.

This automatic decision seems"safe" since an integral of non-delta temperatures doesn't mean anything anyway.

**Comment by coxipi** [comment]
> I'm not sure I fully understand your last comment @coxipi. But indeed, there's an issue for "degree days" where the op is "sum", but `to_agg_units` wants "delta_prod". It's not the only case, there's an ambiguity with "sum" : is it an integral or just a sum ?
> 
> Do you think the following idea could work :
> 
>     * Add "integral" to the valid "op" in `select_resample_op`.
> 
>     * It gets translated to `sum` in the computation, but not in `to_agg_units`.
> 
>     * When receiving "integral", `to_agg_units` checks the dimensions of the input units. If it is a temperature, the "delta_prod" method is taken, otherwise we choose "prod".
>       Thus  "delta_prod" and "prod" are deprecated in favor of "integral" + automatic decision.
> 
> 
> This automatic decision seems"safe" since an integral of non-delta temperatures doesn't mean anything anyway.

Never mind my comment, I was thinking about something else that has nothing to do with this, I don't think we need a `delta` distinction for these other operators. I'll give a read to your suggestion tomorrow.

**Comment by coxipi** [comment]
I understand what you're getting at, just not 100% on board with the nomenclature. Both are sums

```python
# sum (say tas is in degC)
cond = tas > 100  # Which days are boiling
Ndays = cond.sum("time")  # Number of boiling days

# integral 
delta = (tas - 16)
degdays = delta.clip(0).sum("time")
```
that could be expressed as integrals (could not get the "{ }" symbol due to GH below)

$$
\sum_{i} cond_{i} \to  \int_{t \in ( t | cond(t) = 1 ) } d t  = \int dt {} [cond(t)]
$$

For the other case, we have 

$$
\int dt [T(t) - 16 ^{\circ}C] cond_2(t)
$$

where  $cond_2(t) = (T(t) > 16 ^{\circ}C)$.

Maybe that would be an acceptable way to make a distinction, and "delta_prod" is not that crystal clear to start with. But the way I see it, both are sums, that could be expressed as integrals in the continuum limit, and what differs is the integrand, one is dimensionless, the other is dimensionful.

aulemahal:
I agree with the distinction your making, but I would have called the first equation a _count_ instead. A "sum" would be the summation of dimensionful quantity. I don't think we will ever have much of those in xclim, because it's too often ambiguous when dealing with timeseries. 

Thus again, my vision would be to have  `select_resample_op` receive "count" and translate it to a sum operation, and pass "count" to `to_agg_units`.

coxipi:
In fact, the first equation I showed is not correct, the left-hand side is dimensionless, while the right-hand side has dimension [time] (or at least, LHS & RHS are not equal). I can see the distinction you want to make between "count" and other quantities that we really think of as Riemann sums. 

So to summarize, at the time of the computation, "integral" and "count" would both be computed as sums. But in the first case, we really mean a Riemann sum. This means that we implicitly choose time units where the time step is 1, but this is not specified anywhere. `to_agg_units` completes this information when we specify the units, it can multiply the output accordingly if needed to have saner units.   This is why we need to input both the summed output, but also the original dataset, because as I just stated, the summed output doesn't have all the needed info on its units. 

I agree with this, I think it's a good way to do things.

coxipi:
In the example given in `to_agg_units`:

```python
    >>> degdays = (
    ...     (tas - 16).clip(0).sum("time")
    ... )  # Integral of  temperature above a threshold
    >>> degdays = to_agg_units(degdays, tas, op="delta_prod")
```

we have a function that does this: 

```python
def cumulative_difference(...):
        if op in ["<", "<=", "lt", "le"]:
        diff = (threshold - data).clip(0)
    elif op in [">", ">=", "gt", "ge"]:
        diff = (data - threshold).clip(0)
    else:
        raise NotImplementedError(f"Condition not supported: '{op}'.")

    if freq is not None:
        diff = diff.resample(time=freq).sum(dim="time")

    return to_agg_units(diff, data, op="delta_prod")
```

but we don't have a function wrapper like:
```python
def riemann_sum(da):
    return to_agg_units(da.sum(dim="time"), da, op="delta_prod")
```
correct?

aulemahal:
Correct. I'm not sure this would be of much use, though.

coxipi:
Also, this `cumulative_difference` example makes me realize that doing the `to_agg_units` is `select_resample_op` maybe debatable. The last `if` condition would have a redundant use of `to_agg_units` in this case:

```python 
    if freq is not None:
        diff = select_resample_op(diff, op="sum", freq=freq)

    return to_agg_units(diff, data, op="delta_prod")
```

unless we want to write this as: 

```python 
    if freq is not None:
        return select_resample_op(diff, op="sum", freq=freq)

    return to_agg_units(diff, data, op="delta_prod")
```

aulemahal:
The idea of having `to_agg_units` inside `select_resample_op`, is that the latter could then be used directly as the "compute" argument of an indicator, removing the need for a indice definition that would only serve the purpose of setting units.

I think we are currently implementing something alike a "change of paradigm" for xclim generic functions, feel free to break things here.

Zeitsperre:
@coxipi This discussion is well above my expertise, but I want to mention that the changes made in the public function `to_agg_units` should be identified in the PR description and in the CHANGES.rst as breaking changes!

xclim/core/units.py
```diff
@@ -582,15 +582,15 @@ def to_agg_units(
     >>> degdays.units
     'K d'
     """
-    if op in ["min", "max", "mean"]:
-        out.attrs["units"] = orig.attrs["units"]
+    if op in ["amin", "min", "amax", "max", "mean", "std"]:
+        out.attrs["units"] = orig.attrs["units"] if "units" in orig.attrs else ""
```
aulemahal, line 586:
```suggestion
        out.attrs["units"] = orig.attrs["units"]
```
This is not supposed to happen! Usually, xclim should have failed before, but we don't want to assume units if None are given (including "dimensionless").

aulemahal (approved):

xclim/core/units.py
```diff
@@ -566,12 +567,10 @@ def to_agg_units(
...
     ...     coords={"time": time},
-    ...     attrs={"units": "degC"},
     ... )
-    >>> degdays = (
-    ...     (tas - 16).clip(0).sum("time")
-    ... )  # Integral of  temperature above a threshold
-    >>> degdays = to_agg_units(degdays, tas, op="delta_prod")
+    >>> dt = (tas - 16).assign_attrs(units="delta_degC")
+    >>> degdays = dt.clip(0).sum("time")  # Integral of temperature above a threshold
+    >>> degdays = to_agg_units(degdays, dt, op="integral")
```
coxipi, line 573: @aulemahal This is a consequence of the reorganizing. We can't be "as" lazy as before and let "integral" know that we mean a temperature difference. We can still be kind of careless with the rest of operations, just have to be careful with the (T-T0) part.

coxipi: I'm thinking of a potential problem: 

Would this kind of computation sometimes happen?
```python
degdays = tas_degC.clip(tas_ref).sum("time") 
```

Essentially, instead of:
```
tas_degC - tas_ref > 0
```
you look at
```
tas_degC  > tas_ref
```

In this last case, it would be weird to define a temperature delta so that things work out.

Is it reasonable to expect users/developers should know: Ok, I want degdays, so this must be a temperature difference?

aulemahal: I don't think your example would happen. Summing relative units makes no sense to me!
Otherwise, it seems reasonable to me to expect developers to realize this.

PR #1446, opened 2023-07-31
Refactor base indicator classes

### LINKED ISSUE #1263: Frequency analysis misleadingly characterized as resampling indicator
https://github.com/Ouranosinc/xclim/blob/c8f45a74e9f20af00fcba9363d903a5dc987a9ae/xclim/indicators/land/_streamflow.py#L73
where `FA` is a subclass of `ResamplingIndicator`. However, the index reduces the  `time` axis completely, which is not what we would expect from a `ResamplingIndicator` that exposes a `freq` argument. 

Maybe we should have a new subclass for this type of indicator and rename the  `freq` argument to make it less misleading.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1263

Move around some of the base indicator classes code. Split the missing values handling from the resampling handling. This way, indicators like `return_level` can reduce the full time axis and still perform some missing values handling.

No, not yet.

### Other information:
@huard @RondeauG, if I am not mistaken, `return_level` and `fit` currently have missing values handling disabled because it was impossible with the previous classes to have both this and full reduction of "time". Now that it is possible, should we activate it ? 

Similarly, `stats` had the "Any" missing method forced. Is there a reason for that ? Here, I removed the argument, meaning it will use "from_context".


RondeauG:
> if I am not mistaken, `return_level` and `fit` currently have missing values handling disabled because it was impossible with the previous classes to have both this and full reduction of "time". Now that it is possible, should we activate it ?
> 
> Similarly, `stats` had the "Any" missing method forced. Is there a reason for that ? Here, I removed the argument, meaning it will use "from_context".

Absolutely ! For all 3, having full control on the `missing` argument would be a great quality-of-life change.

xclim/core/calendar.py
```diff
@@ -834,6 +835,58 @@ def construct_offset(mult: int, base: str, start_anchored: bool, anchor: str | N
     )
 
 
+def offset_is_divisor(freqA: str, freqB: str):
```
huard, line 838: I suggest using descriptive names for freqA and freqB. 
freqA -> divisor
freqB -> freq

Also, the function name could be `is_offset_divisor`, to emulate other `is_condition` functions found elsewhere.

xclim/core/indicator.py
```diff
@@ -1355,11 +1355,11 @@ def _show_deprecation_warning(self):
...
 
-class ResamplingIndicator(Indicator):
-    """Indicator that performs a resampling computation.
+class CheckMissingIndicator(Indicator):
+    """Class for indicators that completely reduce the time dimension, adding missing value checks.
 
-    Compared to the base Indicator, this adds the handling of missing data,
-    and the check of allowed periods.
+    A full reduction of the "time" dimension is expected by default: the missing step will fail if the output still has a time dimension.
+    To enable resampling, the :py:meth:`_get_missing_freq` method can be subclassed to return the resampling frequency. The method is always called with the indicator parameters.
```
huard, line 1362: Now that I've read the code I understand this comment, but it was cryptic on the first pass. 

It feels semantically weird that CheckMissingIndicator would be defined as an Indicator reducing the time dimension, but that ResamplingIndicator would then subclass it and revert that property. 

Should we bake missing values checks directly inside `Indicator` now that we're able to support them everywhere?

huard (changes requested):

huard (approved):

huard:
Issue https://github.com/CSHS-CWRA/RavenPy/issues/307 is probably due to the fact that this PR made `fit` a subclass of `ReducingIndicator`. It was previously a subclass of `Indicator`, which has no missing value logic in it. 

So the generic `fit` indicator essentially went from skipping missing value checks, to apply the missing algorithm from context (typically `any`). I think we should have highlighted this more clearly in the release notes.

PR #1492, opened 2023-10-05
Snow season length is 0 if there is 0 snow - fix solar_zenith_angle typo

### LINKED ISSUE #1491: Number of  days can be both nan and 0
### Generic Issue

* xclim version: 0.44.0
* Python version: 3.10
* Operating System: linux

### Description
We calculated [snow cover duration](https://xclim.readthedocs.io/en/stable/api.html#xclim.indicators.land.snd_season_length) for CORDEX data.

We noticed number of days can be both nan and 0 .
The calculation leads to nan if all values are equal 0. 
The calculation leads to 0 if at least one value is greater than 0 but all values are below threshold.
In both cases we expected a 0.
Is there a special purpose for this differentiation?

### Code of Conduct

### LINKED ISSUE #1493: "parallel" -> "parallelized"
### Generic Issue

* xclim version: 0.44.0
* Python version: 3.10
* Operating System: linux

### Description

We found a spelling mistake:
https://github.com/Ouranosinc/xclim/blob/master/xclim/indices/helpers.py#L299
"parallel" -> "parallelized"

here it is fine:
https://github.com/Ouranosinc/xclim/blob/master/xclim/indices/generic.py#L993

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1491 and fixes #1493

`snd_season_length` and `snw_season_length` got implemented with the same check as their `*_season_start` and `*_season_end` counter parts, with a mask where : `not at_least_n_valid(snw.where(snw > 0), n=1, freq=freq)`.

However, this means the output is NaN if all inputs are 0. This makes sense in the DOY case (can't a have a start/end date if there's no season), but it doesn't in the "length" case. Instead, I think (as does the issue raiser) an all-0 snow timeseries simply means a season length of 0. Thus, I removed the `.where(snw > 0)` part of the test for those two indicators.

EDIT: I also slipped in another fix for `xc.indices.helpers.cosine_of_solar_zenith_angle`. There was a typo in the `xr.apply_ufunc` call.

Yes because it will change the output of two indicators, but I think this is for the best. The previous behaviour was not documented as such in the doc, I believe the new one is actually more intuitive. No need for a warning, IMO.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 
 v0.46.0 (unreleased)
 --------------------
-Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), David Huard (:user:`huard`).
+Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), David Huard (:user:`huard`) and Pascal Bourgault (:user:`aulemahal`).
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -19,6 +19,7 @@ Bug fixes
 * Fixed an error in the `pytest` configuration that prevented copying of testing data to thread-safe caches of workers under certain conditions (this should always occur). (:pull:`1473`).
   * Coincidentally, this also fixes an error that caused `pytest` to error-out when invoked without an active internet connection. Running `pytest` without network access is now supported (requires cached testing data). (:issue:`1468`).
 * Calling a ``sdba.map_blocks``-wrapped function with data chunked along the reduced dimensions will raise an error. This forbids chunking the trained dataset along the distribution dimensions, for example. (:issue:`1481`, :pull:`1482`).
+* Indicators ``snd_season_length`` and ``snw_season_length`` will return 0 instead of NaN if all inputs have a (non-NaN) 0 snow depth (or water-equivalent thickness). (:pull:`1492`, :issue:`1491`)
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -19,6 +19,7 @@ Bug fixes
 * Fixed an error in the `pytest` configuration that prevented copying of testing data to thread-safe caches of workers under certain conditions (this should always occur). (:pull:`1473`).
   * Coincidentally, this also fixes an error that caused `pytest` to error-out when invoked without an active internet connection. Running `pytest` without network access is now supported (requires cached testing data). (:issue:`1468`).
 * Calling a ``sdba.map_blocks``-wrapped function with data chunked along the reduced dimensions will raise an error. This forbids chunking the trained dataset along the distribution dimensions, for example. (:issue:`1481`, :pull:`1482`).
+* Indicators ``snd_season_length`` and ``snw_season_length`` will return 0 instead of NaN if all inputs have a (non-NaN) 0 snow depth (or water-equivalent thickness). (:pull:`1492`, :issue:`1491`)
```
huard, line 22:
```suggestion
* Indicators ``snd_season_length`` and ``snw_season_length`` will return 0 instead of NaN if all inputs have a (non-NaN) zero snow depth (or water-equivalent thickness). (:pull:`1492`, :issue:`1491`)
```

huard (approved):

aulemahal:
@Zeitsperre do you know why the coveralls action is still hanging here ?

Zeitsperre:
Coveralls always expects a hook from GitHub with some stats to compare to. The problem is that because we aren't running all tests, the coverage will **always** be lower than expected, so the statistic is useless for developers until we build with the `approved` tag.

Maybe there's a way to send along a `skip` signal to coveralls when we're only building with Python3.9? I'll look into it.

aulemahal:
Hum... Ok, but then is that the reason I can't merge ?

Zeitsperre:
Oh, no, it isn't. For some reason, David's PR didn't trigger the `approved` label. Weird.

PR #1498, opened 2023-10-11
More generic `generic.stats` - Allow 3H inputs

aulemahal:
### Pull Request Checklist:
    - This PR fixes an issue raised in discussions with @vindelico 

* Relax indicator `generic.STATS` so it accepts non-daily inputs. I saw no particular reason for it to be limit to daily values, except from the metadata.
* Add variance to the supported `op` of `to_agg_units`.
* Add a `out_units` to `select_resample_op` so that one can use it with an `op` `to_agg_units` doesn't know, and explicitly pass the expected units.

Not a big one, but all YAML indicators previous constructed from `select_resample_op` should add `out_units: null` in their `parameters` section to remove the new argument from the signature. Altough its presence is not problematic in itself.

### DIFF
--- a/xclim/core/units.py
+++ b/xclim/core/units.py
@@ -520,7 +520,7 @@ def to_agg_units(
     orig : xr.DataArray
         The original array before the aggregation operation,
         used to infer the sampling units and get the variable units.
-    op : {'min', 'max', 'mean', 'std', 'doymin', 'doymax',  'count', 'integral'}
+    op : {'amin', 'min', 'amax', 'max', 'mean', 'std', 'var', 'doymin', 'doymax',  'count', 'integral', 'sum'}
         The type of aggregation operation performed. The special "delta_*" ops are used
         with temperature units needing conversion to their "delta" counterparts (e.g. degree days)
     dim : str
@@ -574,6 +574,9 @@ def to_agg_units(
     if op in ["amin", "min", "amax", "max", "mean", "std"]:
         out.attrs["units"] = orig.attrs["units"]
 
+    elif op in ["var"]:
+        out.attrs["units"] = pint2cfunits(str2pint(orig.units) ** 2)
+
     elif op in ["doymin", "doymax"]:
         out.attrs.update(
             units="", is_dayofyear=np.int32(1), calendar=get_calendar(orig)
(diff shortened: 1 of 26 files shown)


huard (approved):

huard:
Maybe add a quick test ?

aulemahal:
Woups good call.

Because the initial issue of Marco was with 3H input, I tested with that. Turns out `MissingBase` wasn't able to comprehend frequencies containing multiples. I fixed that too.

xclim/core/units.py
```diff
@@ -591,7 +594,7 @@ def to_agg_units(
             out.attrs["units"] = pint2cfunits(orig_u * freq_u)
     else:
         raise ValueError(
-            f"Aggregation op {op} not in [min, max, mean, std, doymin, doymax, count, integral]."
+            f"Unknown aggregation op {op}. Known ops are [min, max, mean, std, var, doymin, doymax, count, integral, sum]."
```
Zeitsperre, line 597: Missing `amax` and `amin`

xclim/indicators/generic/_stats.py
```diff
@@ -44,11 +44,11 @@ class GenericResampling(ResamplingIndicator):
 
 
 stats = GenericResampling(
-    title="Statistic of the daily values for a given period.",
+    title="Simple resampled statistic of the values.",
```
Zeitsperre, line 47: Does this need to be updated in the French translation?

xclim/core/units.py
aulemahal: @coxipi  what are "amin" and "amax" ?

xclim/indicators/generic/_stats.py
aulemahal: Indeed.

xclim/core/units.py
coxipi: https://numpy.org/doc/stable/reference/generated/numpy.amin.html
min/max of an array in numpy. 

When changing `to_agg_units` in https://github.com/Ouranosinc/xclim/pull/1443/files , I noticed that some functions use `amin/amax` instead of `min/max`, not including `amin/amax` caused some errors in the unit tests. The documentation in `to_agg_units` does not reflect this, only mentioning `min/max`, maybe  it would be better to also mention those functions. Would it be useful if I check which functions use `amin/amax`?

PR #1507, opened 2023-10-20
Fix data flags var name generation

aulemahal:
### Pull Request Checklist:
    - This PR fixes a issue raised in private communication by @RondeauG 

Changes how the data flags variable names are generated.

In the previous version, every kwargs was sent to `str2pint`. We were relying on a `try: except` to catch those kwargs that weren't quantities. This had the caveat of requiring an explicit list of possible errors. @RondeauG  had a case where `op='>='` was triggering a `ValueError`, which wasn't listed.

I changed this "implicit" parsing to an "explicit" one:
- Data flags declare the variable name as a templated string.
- Parameters are iterated through and handled according to their "InputKind"
    - "Quantified" inputs are handled as before, but all others are passed as-is.

This required:
- Changing the registering decorator to accept a templated string argument
- Correctly annotating the inputs, which lead to adding missing thresholds to the declared units

Yes, I made the arbitrary choice of changing how we stringify the minus sign:

```
Input thresh : -5.1 mm
Template : "values_greater_{thresh}"
Before : "values_greater__minus_5point1"
This PR: "values_greater_minus5point1"
```
However, I don't think this will affect many projects.

### Other information:
@RondeauG , could you try this new branch with your code ? (That's why I made you a reviewer)

### DIFF
--- a/xclim/core/dataflags.py
+++ b/xclim/core/dataflags.py
@@ -12,9 +12,9 @@ from inspect import signature
 from typing import Sequence
 
 import numpy as np
-import pint
 import xarray
 
+from ..indices.generic import binary_ops
 from ..indices.run_length import suspicious_run
 from .calendar import climatological_mean_doy, within_bnds_doy
 from .formatting import update_xclim_history
@@ -23,6 +23,7 @@ from .utils import (
     VARIABLES,
     InputKind,
     MissingVariableError,
+    Quantified,
     infer_kind_from_parameter,
     raise_warn_or_log,
 )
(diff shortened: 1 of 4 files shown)


xclim/core/dataflags.py
```diff
@@ -97,7 +108,7 @@ def _sanitize_attrs(da: xarray.DataArray) -> xarray.DataArray:
     return da
 
 
-@register_methods
+@register_methods()
```
Zeitsperre, line 111: Why are parentheses needed here?

Zeitsperre (approved):
LGTM. Nice to see existing functions being used to handle signatures.

xclim/core/dataflags.py
aulemahal: I added a decorator level to the function: `@register_methods()` actually returns the real decorator, using the default value (None, thus `func.__name__`) as variable name.

Zeitsperre: I don't get quite understand, but I'll give it a look on Monday with fresher eyes.

Issue #1255, opened 2022-12-09
Efficient Application of Bias Correction


kckornelsen:
* xclim version:
* Python version:
* Operating System:

### Description
I am trying to use xclim to bias correct precipitation from ECCC's GEPS product (ensemble NWP) and correcting it against CaPA data. The NWP model has 20 ensemble members and 16 lead times in the forecast. The data is applied to N different watersheds (each with a single timeseries and not gridded data). Through some research it was determined that each GEPS lead time has a slightly different bias. The EQM method was selected and parameters for EQM are pre-computed. The current code takes quite a bit of time to run and I would like to know if there are ways to speed it up. Particularly around the nested for loops. Is there a built in xarray function or xclim function?

### What I Did
Assume we start with a bit of house keeping, imports, identifying folders, etc.
```
def process_each_lead_time(leadtime_length,ini_time,geps_fcst,path_eqm_par_folder,path_output_geps,realizations,stations):
     
    # define the time slot for this leadtime 
    ltime_start = ini_time + np.timedelta64(leadtime_length*i_ltime, 'h')  + np.timedelta64(1,'h')
    ltime_end = ini_time + np.timedelta64(leadtime_length*(i_ltime+1), 'h')

    #select geps dataset within the time slot
    geps_fcst_ltime = geps_fcst.sel(time=slice(ltime_start, ltime_end)).copy(deep=True)

    # read eqm parameters for this lead time 
    eqm_pars = xr.open_dataset(os.path.join(path_eqm_par_folder,"Leadtime_"+str(i_ltime + 1) + "_EQM.nc")) 
    eqm_pars.close()

    # correct geps for each station and realization 
    for i_stations in range(0,len(stations)):
        i_station = stations[i_stations]
        geps_fcst_ltime_station = geps_fcst_ltime.sel(stations=i_station)
        
        stn = station_names[i_station]
        with open("../../Export/GEPS_FCST2/log.txt", "a") as f:
            print("INFO: processing station", i_station, "or", stn, ",at leadtime ",i_ltime, file=f)
            # print(i_station, file=f)
        for i_ens in range(0,len(realizations)):
            
            
            i_en = realizations[i_ens]
            
            # select parameter for this station and this realization 
            idx = i_stations*22 + i_ens
            EMQ_par = eqm_pars.sel(station_ens = idx)
            
            geps_ltime_uncorrect = geps_fcst_ltime_station.sel(realization = i_en)["PC_NWP"]

[...]

PR #1513, opened 2023-10-26
Speed up quantiles with sorting

SarahG-579462:
### Pull Request Checklist:
    - This PR will help #1255 

* `nbutils.quantile` has a speed-up of more than 2.5x by a combination of changes in `nbutils.quantile` and `nbutils._quantile`
* This does not cover `nbutils.vec_quantiles` (used for `adapt_freq`) but similar principles could be used
* It adds the possibility of using `fastnanquantile` module which is very fast

No

### Other information:
* The new low-level function to compute quantiles `nbutils._quantile`   is a 1d jitted version of `xclim.core.utils._nan_quantile` 
* Manual benchmarking can be performed in the notebook `benchmarks/sdba_quantile.ipynb`, attached to this PR.


xclim/sdba/nbutils.py
```diff
@@ -14,25 +14,14 @@
...
-
-
 @njit(
     [
         int64(float32[:]),
         int64(float64[:]),
     ],
     fastmath=False,
     nogil=True,
+    cache=False,
```
aulemahal, line 24: This was removed because it slows down the import of  `xclim` a lot (and by proxy, `xscen`). Many users will never use `sdba`, so it made sense to add the compilation burden only to those who did. And even then, waiting for the import seemed much more frustrating than waiting as few seconds more for your complex and heavy computation. See https://github.com/Ouranosinc/xclim/pull/1378.

SarahG-579462: An option would be to use [numba.cc](https://numba.readthedocs.io/en/stable/user/pycc.html) for ahead-of-time compilation during module install.

SarahG-579462: Oh I think I misunderstood, I thought you were talking about pre-compilation in general. Caching is not a problem, I just turned them off to make sure that my implementations for pre-compilation were working.

aulemahal: Ah! All good then, as long as we re-activate it before merging.

juliettelavoie:
For a 30 years period, over QC, the bias adjustment (first initial DQM + Npdf + restore) with these parameters:
```
n_iter: 30
nquantiles:50
group: time
n_escore: 1
```
took 8 hours for this branch and 8h40 for the master. (time for branch faster_mbcn is coming.)

Zeitsperre:
@SarahG-579462 

`checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/xclim/checkouts/1513/docs/notebooks/sdba-speed.ipynb: WARNING: document isn't included in any toctree`

Small FYI, you need to add the notebook to the `exclude_patterns` in `docs/conf.py` if you don't want it rendered.

xclim/sdba/nbutils.py
```diff
@@ -66,14 +151,13 @@ def quantile(da, q, dim):
...
-    if not hasattr(q, "dtype") or q.dtype != da.dtype:
-        qc = np.array(q, dtype=da.dtype)
-    else:
-        qc = q
-
+    # if not hasattr(q, "dtype") or q.dtype != da.dtype or q.shape[0] != da.shape[0]:
+    #    qc = np.array(q, dtype=da.dtype)
+    # else:
+    #    qc = q
+    qc = np.array(q, dtype=da.dtype)
```
coxipi, line 160: Revert or remove comments?

coxipi:
So I hijacked the PR a bit to test general benchmark tools. I needed to install ``pytest-benchmark``and ``pytest-benchmark[histogram]``.  The idea would be to have a benchmarking suite, similar to the structure of the testing suite. For a test similar to your current notebook, I run: 

```bash
pytest  -n0 --dist no benchmarks/sdba/test_nbutils.py
```
where `test_nbutils.py` is 

```python
from xclim.sdba import nbutils
import pytest 
import importlib
import os
import numpy as np

nq = 50
i = 0
@pytest.mark.parametrize(
    "use_nanquantile,size",
    [
    (True, 100),
    (True, 200),
    (True, 500),
    (True, 1000),
    (False, 100),
    (False, 200),
    (False, 500),
    (False, 1000),
    ],
)
def test_nanquantile_simple(benchmark,use_nanquantile, size):
    np.random.seed(0)
    arr = np.random.randn(size)
    os.environ["USE_NANQUANTILE"] = "True" if use_nanquantile else ""
    importlib.reload(nbutils)   
    def func():
        return nbutils._choosequantile(arr, np.linspace(0, 1, nq))
    # trigger jit compilation on first run
    global i
    if i==0: 
        func()
        i = i + 1
    benchmark(func)
```
(I should have named it `test__choosequantile` but anyways)

I can then have an histogram with the results:
image

It's a bit annoying that the order of parameters in `pytest.mark.parametrize` is not respected. It seems it's a known issue. Also, I don't know why there's always an outlier. It seems pretty plug and play. The visualization with the notebook is better, so we might want to keep a notebook for benchmarks too?

SarahG-579462:
> I did a test with EmpiricalQuantileMapping, 30 years, 3 locations, dayofyear-31. Sorting is still advantageous, but the effect is marginal. There is maybe an overhead which is just much greater than the actual computation of quantiles? `False/True` refers to `use_nanquantile`

This is really interesting, thanks. I totally support you taking over this PR, I don't really have time to work on it these days.

coxipi:
It seems that removing `stack` on time dimensions and use flatten/ravel instead saves about 25-30% when computing EmpiricalQuantileMapping!

coxipi:
See #1743 for additional tests. `_sortquantile` has performed very well, I think it's worth setting `allow_sortquantile=True` by default, which begs the question, do we need this switch now haha?

xclim/sdba/nbutils.py
coxipi: > Todo: turn on cache=True

seems to contradict the discussion above. We leave caching turned off?

coxipi:
Concerning `vecquantiles` we have the same problem that we had for the old implementation of `quantiles` : if `time` has object values, it is very slow. However, in other cases, it is surprisingly quite performant, which is a bit confusing since it seems to share a lot of structure with the old `quantiles` and this function was easier to optimize. I managed to obtain what seems a faster implementation by deactivating the njit:

```python
@njit
def jnq(arr,q):
    return np.nanquantile(arr, q)

# @njit
def _wrapper_vecquantiles1d(arr, rnk, ):
    out = np.empty((arr.shape[0]), dtype=arr.dtype)
    for index in range(out.shape[0]):
        if np.isnan(rnk[index]):
            out[index] = np.NaN
        else: 
            out[index] = np.nanquantile(arr[index], rnk[index])
    return out
# @njit
def _vecquantiles(arr, rnk, nreduce):#,  res):
        if nreduce == arr.ndim:
        #         # then rnk is a scalar
                return jnq(arr.flatten(), rnk)
        red_axis = np.arange(len(arr.shape) - nreduce, len(arr.shape))
        # return np.nanquantile(arr, rnk, red_axis)
        reduction_dim_size = np.prod([arr.shape[idx] for idx in red_axis])
        # kept dimensions
        keep_axis = np.arange(len(arr.shape) - nreduce)
        final_shape = [arr.shape[idx] for idx in keep_axis]
        # reshape as (keep_dims, red_dims), compute, reshape back
        arr = arr.reshape(-1, reduction_dim_size)
        rnk = rnk.reshape(-1)
        out = _wrapper_vecquantiles1d(arr,rnk) 
        return out.reshape(final_shape)

def vecquantiles(da, rnk, dim):
    """For when the quantile (rnk) is different for each point.

    da and rnk must share all dimensions but dim.

    Parameters
    ----------
    da : xarray.DataArray
        The data to compute the quantiles on.
    rnk : xarray.DataArray
        The quantiles to compute.
    dim : str or sequence of str
        The dimension along which to compute the quantiles.

    Returns
    -------
    xarray.DataArray
        The quantiles computed along the `dim` dimension.
    """
    dims = [dim] if isinstance(dim, str) else dim
    da = da.transpose(*rnk.dims, ...)
    return xr.apply_ufunc(
        _vecquantiles, 
        da, 
        rnk, 
        input_core_dims = [dims, []], 
        exclude_dims=set(dims), 
        dask="parallelized", 
        kwargs=dict(nreduce=len(dims)),
        output_dtypes = [da.dtype]
    )
``

**Comment by coxipi** [comment]
In the last implementation above, since I had no `njit` involved, I figured out it would be very similar to using `vectorize=True` in a `xr.apply_ufunc`

```python
def vecquantiles(da, rnk, dim):
    """For when the quantile (rnk) is different for each point.

    da and rnk must share all dimensions but dim.

    Parameters
    ----------
    da : xarray.DataArray
        The data to compute the quantiles on.
    rnk : xarray.DataArray
        The quantiles to compute.
    dim : str or sequence of str
        The dimension along which to compute the quantiles.

    Returns
    -------
    xarray.DataArray
        The quantiles computed along the `dim` dimension.
    """
    dims = [dim] if isinstance(dim, str) else dim
    da = da.transpose(*rnk.dims, ...)
    if USE_FASTNANQUANTILE: 
       func = fnq.nanquantile # this already treats q=NaN the way we want
    else: 
       func = lambda arr, q : np.NaN if np.isnan(q) else np.nanquantile(arr,q) 
    return xr.apply_ufunc(
        func, 
        da, 
        rnk, 
        input_core_dims = [dims, []], 
        output_core_dims = [[]], 
        dask="parallelized", 
        output_dtypes = [da.dtype], 
        vectorize=True
    )
    
```
the performance is a bit better than what was previously implemented in `sdba.nbutils.vecquantiles`. That's without `njit`. With a `njit`, the performance is actually worse.

SarahG-579462:
I took a look at `xclim.core.utils._nan_quantile`, which uses essentially the same algorithm as this PR, but I was getting much faster results. Turns out this was because this function does in-place computation, and so it much more memory-efficient. It also sorts the array in-place, so any subsequent sorts are very fast, which also affects _sortquantile.

I also tried numba-compiling this function. As a first pass, I was only able to make it work for 1d arrays, but the results still hold.

On a dataset of 11000 temporal points, here are my results, when including np.copy():

| Method | Time |
| ------------- | ------------- |
| sdba.nbutils._sortquantile(arr_m1,q) |877 µs ± 10.9 µs per loopc |
| xclim.core.utils._nan_quantile(np.copy(arr_m2),q)  | 858 µs ± 8.82 µs per loop  |
| sdba.nbutils.(np.copy(arr_m2),q)  | 901 µs ± 35.7 µs per loop  |

With the np.copy in the `timeit` setup, the numba'd _nan_quantile implementations end up completing in about 156 µs ± 1.41 µs per loop, while the non-numba version takes 221 µs ± 2.94 µs per loop.

SarahG-579462:
Ah, I was mistaken, aparently the setup code for timeit is only run once. Here's a more accurate version, with the setup code run every time:

Time (µs), no-copy, no-inplace, numba, sortquantile: avg 900.0 min 794.0 max 2746.7 std 525.23
Time (µs), copy, inplace, numba,  _nan_quantile: avg 973.7 min 810.4 max 9668.8 std 642.08
Time (µs), pre-copy, inplace, numba, _nan_quantile: avg 904.5 min 797.1 max 2006.6 std 529.12
Time (µs), once-copy, inplace, numba, _nan_quantile: avg 176.4 min 151.4 max 916.7 std 106.16
Time (µs), copy, inplace, abel, _nan_quantile: avg 948.2 min 801.8 max 3302.7 std 580.36
Time (µs), pre-copy, inplace, abel, _nan_quantile: avg 971.5 min 786.4 max 7285.8 std 619.52
Time (µs), once-copy, inplace, abel, _nan_quantile: avg 271.7 min 234.4 max 1473.6 std 165.02

coxipi:
@SarahG-579462 : As I understand, I should simply remove the %%timeit tests in the notebook as they don't make sense with Abel's function, correct?

Otherwise, we have simply used Abel's function as a backend in the end, and `fastnanquantile` is used if installed. PR is ready for review, I think it should be someone else than Sarah and me since we both worked on this.

xclim/sdba/nbutils.py
```diff
@@ -204,7 +217,17 @@ def _escore(tgt, sim, out):
     out[0] = w * (sXY + sXY - sXX - sYY) / 2
 
 
-@njit
+@njit(
+    #    [
```
bzah, line 221: Remove commented code ?
Also, I'm not well versed in numba but my understanding is that it mainly speeds up for loop over numpy arrays.
Does it accelerate anything else ? In particular, what is the purpose of numba for the functions of this module ?

xclim/sdba/nbutils.py
```diff
@@ -217,8 +240,18 @@ def _first_and_last_nonnull(arr):
     return out
 
 
-@njit
-def _extrapolate_on_quantiles(interp, oldx, oldg, oldy, newx, newg, method="constant"):
+@njit(
+    #    [
```
bzah, line 244: remove commented code ?

xclim/sdba/nbutils.py
```diff
@@ -239,7 +272,15 @@ def _extrapolate_on_quantiles(interp, oldx, oldg, oldy, newx, newg, method="cons
     return interp
 
 
-@njit
+@njit(
+    #    [
```
bzah, line 276: same, remove commented code ?

xclim/sdba/nbutils.py
```diff
@@ -62,77 +71,65 @@ def vecquantiles(
...
-    tem = utils.get_temp_dimname(da.dims, "temporal")
-    da = da.stack({tem: dims})
-
-    # So we cut in half the definitions to declare in numba
-    # We still use q as the coords, so it corresponds to what was done upstream
-    if not hasattr(q, "dtype") or q.dtype != da.dtype:
-        qc = np.array(q, dtype=da.dtype)
+def _quantile(arr, q, nreduce):
+    if arr.ndim == nreduce:
+        out = _nan_quantile(arr.flatten(), q)
```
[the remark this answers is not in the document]
coxipi: Thanks! I must admit I didn't put much thought in the method that is used. Since the function is hidden, I think it's reasonable to expect users to know what they're doing. But would it be better to set `alpha==beta==1/3`  as a default? Do I understand correctly that this specific method `alpha==beta==1/3` is better adapted for a general context where we have no a priori info on the distribution of data? Or are you saying: "if we don't much about the data, there is no reason to deviate from the numpy standard"?

xclim/sdba/nbutils.py
```diff
@@ -9,15 +9,24 @@
 
 import numpy as np
 from numba import boolean, float32, float64, guvectorize, njit
-from xarray import DataArray
+from xarray import DataArray, apply_ufunc
 from xarray.core import utils
 
+from xclim.core.utils import _nan_quantile
```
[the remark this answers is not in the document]
coxipi: on it, I will also change default to alpha==beta==1/3 (and revert back if you judge it's not appropriate)

xclim/sdba/nbutils.py
```diff
@@ -62,77 +71,65 @@ def vecquantiles(
...
+        final_shape = [arr.shape[idx] for idx in keep_axis] + [len(q)]
+        # reshape as (keep_dims, red_dims), compute, reshape back
+        arr = arr.reshape(-1, reduction_dim_size)
+        out = _wrapper_quantile1d(arr, q)
+        out = out.reshape(final_shape)
+    return out
+
 
+def quantile(da, q, dim):
+    """Compute the quantiles from a fixed list `q`."""
```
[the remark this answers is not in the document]
coxipi: It was removed by error,  I will recover the old docstring

xclim/sdba/nbutils.py
```diff
@@ -62,77 +71,65 @@ def vecquantiles(
...
-        out = np.empty((q.size,), dtype=arr.dtype)
-        out[:] = np.nanquantile(arr, q)
-    else:
-        out = np.empty((arr.shape[0], q.size), dtype=arr.dtype)
-        for index in range(out.shape[0]):
-            out[index] = np.nanquantile(arr[index], q)
+def _wrapper_quantile1d(arr, q):
+    out = np.empty((arr.shape[0], q.size), dtype=arr.dtype)
+    for index in range(out.shape[0]):
+        out[index] = _nan_quantile(arr[index], q)
```
coxipi, line 77: this is where numba has the best chance of helping, with the loop

xclim/sdba/nbutils.py
coxipi: Never mind, I think I simply removed the problematic headers in the concerned functions, the question remains on what to do with the rest.

xclim/sdba/nbutils.py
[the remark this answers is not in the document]
bzah: I agree. The current default are meant to reproduce numpy's default behavior (linear method) but I think it makes sense to use the best default as the actual default !

xclim/sdba/nbutils.py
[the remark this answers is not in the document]
bzah: > Do I understand correctly that this specific method alpha==beta==1/3 is better adapted for a general context where we have no a priori info on the distribution of data

Yes that's what I mean :)

xclim/sdba/nbutils.py
bzah: Ah yes indeed!

xclim/sdba/nbutils.py
SarahG-579462: Yes, I was exploring other compilation options for the jitting, particularly for testing, to avoid having to do a dry-run before every test, I believe.

It can be axed!

xclim/sdba/nbutils.py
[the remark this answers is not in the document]
SarahG-579462: Temperature data is often normally distributed, in which case 'normal unbiased' is likely better, with alpha = beta = 3/ 8.0, but otherwise I agree.

bzah (approved):
LGTM!

AUTHORS.rst
```diff
@@ -29,7 +29,7 @@ Contributors
 * David Caron `@davidcaron <https://github.com/davidcaron>`_
 * Carsten Ehbrecht <ehbrecht@dkrz.de> `@cehbrecht <https://github.com/cehbrecht>`_
 * Jeremy Fyke `@jeremyfyke <https://github.com/jeremyfyke>`_
-* Sarah Gammon <gammon.sarah@ouranos.ca> `@SarahG-579462 <https://github.com/SarahG-579462>`_
+* Sarah Gammon `@SarahG-579462 <https://github.com/SarahG-579462>`
```
coxipi, line 32:
```suggestion
* Sarah Gammon `@SarahG-579462 <https://github.com/SarahG-579462>`_
```

Zeitsperre (approved):
@SarahG-579462 or @coxipi Would it be possible for you to add fastnanquantile as a dependency that can be installed via `tox` (e.g. under deps: `fastnanquantile: fastnanquantile`) and add that to one of the PyPI/tox CI builds (in `main.yml`) so that we can test against it?

coxipi:
The present implementation is not right, the call to `nan_quantile` is not properly "jitterized". Among other things, some operations in nan_quantile use the argument `axis` in `nan_quantile` which is unsupported for many functions used with njit. The library `fastnanquantile` shows how those parts working on axis must be done outside the njit calls. Also, `np.asanyarray` doesn't seem to be supported. 

Since `sdba` and `xclim` are to be separated, I feel some redundancy would not be bad on this front, I was considering to simply revert some commits and continue to use `sortquantile`. Thoughts?

SarahG-579462:
> The present implementation is not right, the call to `nan_quantile` is not properly "jitterized". Among other things, some operations in nan_quantile use the argument `axis` in `nan_quantile` which is unsupported for many functions used with njit. The library `fastnanquantile` shows how those parts working on axis must be done outside the njit calls. Also, `np.asanyarray` doesn't seem to be supported. I'm not sure why this is triggered or not in certain contexts.
> 
> The function would either need to be properly jitterized, or keep it unjittered and study again the performance. Since `sdba` and `xclim` are to be separated, I feel some redundancy would not be bad on this front, I was considering to simply revert some commits and continue to use `sortquantile`. Thoughts?

During testing, I had made a version of Abel's `nan_quantile` that supports only 1 dimension, that is jitt'able. Would it be preferable to use this? Abel's unjitted version is really not much slower than the jitted version, because he's using quite pure numpy functions.

coxipi:
Should be good to go, I added a few tests

coxipi:
> Would it be possible for you to add fastnanquantile as a dependency that can be installed via tox (e.g. under deps: fastnanquantile: fastnanquantile) and add that to one of the PyPI/tox CI builds (in main.yml) so that we can test against it?

@Zeitsperre under `test-pypi`?

Zeitsperre (approved):

Zeitsperre:
@aulemahal I leave you to do the final review and merge. I won't pretend to be the expert here.

coxipi:
@SarahG-579462  can you inspect a last time? Should be good now

xclim/sdba/nbutils.py
```diff
@@ -141,7 +283,15 @@ def remove_NaNs(x):  # noqa
     return x[:, ~remove]
 
 
-@njit(fastmath=True)
+@njit(
+    [
+        float32(float32[:, :], float32[:, :]),
+        float64(float64[:, :], float64[:, :]),
+    ],
+    fastmath=True,
```
SarahG-579462, line 291: FYI, this is outside the scope of this PR, but there might be a bug here with fastmath=True, when there are NaNs in the series.

xclim/sdba/nbutils.py
```diff
@@ -158,7 +308,15 @@ def _correlation(X, Y):
     return d / (X.shape[1] * Y.shape[1])
 
 
-@njit(fastmath=True)
+@njit(
+    [
+        float32(float32[:, :]),
+        float64(float64[:, :]),
+    ],
+    fastmath=True,
```
SarahG-579462, line 316: Same here.

xclim/sdba/nbutils.py
SarahG-579462: https://llvm.org/docs/LangRef.html#fast-math-flags

fastmath=True implies nnan and ninf, both of which are not always true for our data. Is there a check in the function that calls this?

SarahG-579462 (commented):
Looks good to me! a note about fastmath below.

xclim/sdba/nbutils.py
coxipi: For those functions, there are no Nans involved, this is after calling `remove_NaNs` in `_escore` so all's good!

PR #1514, opened 2023-10-27
Refactor robustness methods

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add `ipcc-advanced` as a test to `ensembles.change_significance`. This corresponds to the simpler alternative of approach C as described in the Cross-Chapter Box 1 of the [IPCC Atlas (AR6, WG1)](https://www.cambridge.org/core/books/climate-change-2021-the-physical-science-basis/atlas/24E1C016DBBE4725BDFBC343695DE7DB).
* Add two examples in the "Ensembles" notebook to show how xclim can be used to get the IPCC-recommended hatching masks.
* Add a very simple `detrend` function.

No, but it might, see below.

### Other information:
As you can see in the examples, the IPCC decouples the "change significance" from the "sign agreement", while xclim does not. This make the second example more complex than expected.

I thought of :
- Changing the output of `change_significance` so that `pos_frac` is independent of change significance
- Adding a third output so we have: `significant_change_frac, significant_positive_change_frac, positive_change_frac`.
- Adding a keyword argument to switch between the two modes.
First two are breaking changes while the thirds feels overly heavy for something this simple...

### DIFF
--- a/docs/notebooks/ensembles.ipynb
+++ b/docs/notebooks/ensembles.ipynb
@@ -164,6 +164,7 @@
     "# Set display to HTML style (for fancy output)\n",
     "xr.set_options(display_style=\"html\", display_width=50)\n",
     "\n",
+    "import matplotlib as mpl\n",
     "import matplotlib.pyplot as plt\n",
     "\n",
     "%matplotlib inline\n",
@@ -282,9 +283,14 @@
    "source": [
     "### Change significance and model agreement\n",
     "\n",
-    "Aside from arithmetic reductions of the ensemble as shown above, xclim also implements helper functions characterizing the significance of simulated climatic change, and whether the different members agree or not over the sign of the change. Here we will use function [xc.ensembles.change_significance](../apidoc/xclim.ensembles.rstl#xclim.ensembles._robustness.change_significance) to generate masks replicating the ones in the [IPCC Atlas](https://interactive-atlas.ipcc.ch/), based on two model ensemble approaches outlined in Cross-Chapter Box 1 of the [IPCC Atlas chapter (AR6, WG1)](https://doi.org/10.1017/9781009157896.021).\n",
+    "When communicating climate change through plots of projected change, it is often useful to add information on the statistical significance of the values. A common way to represent this information without overloading the figures is through hatching patterns surimposed on the primary data. Two aspects are usually shown : \n",
     "\n",
-    "First we generate some fake annual mean temperature data. `ref` is the data on the reference period and `fut` is a future projection. In our example, there will be 5 different members in the ensemble."
+    "- change significance : whether most of the ensemble members project a climate change signal statistically significant in comparison to their internal variability.\n",
+    "- model agreement : whether the different ensemble members agree on the sign of the change.\n",
+    "\n",
+    "We can then divide the plotted points into categories each with its own hatching pattern, usually leaving the robust data (models agree and all show significant change) without hatching. \n",
+    "\n",
+    "Xclim provides some tools to help in generating these hatching masks. First is [xc.ensembles.robustness_fractions](../apidoc/xclim.ensembles.rst#xclim.ensembles._robustness.robustness_fractions) that can characterize the change significance and sign agreement accross ensemble members. To demonstrate its usage, we'll first generate some fake annual mean temperature data. Here, `ref` is the data on the reference period and `fut` is a future projection. There are be 5 different members in the ensemble. We tweaked the generation so that all models agree on significant change in the \"south\" while agreement and signifiance of change decreases as we go north and east."
    ]
   },
   {
@@ -342,9 +348,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#### Approach B: The simple method.\n",
+    "Change significance can be determined in a lot of different ways. Xclim provides some simple and some more complicated statistical test in `robustness_fractions`. In this example, we'll follow the suggestions found in the Cross-Chapter Box 1 of the [IPCC Atlas chapter (AR6, WG1)](https://doi.org/10.1017/9781009157896.021). Specifically, we are following Approach C, using the alternative for when pre-industrial control data is not available.\n",
     "\n",
-    "Zones where there is no agreement on the sign of the change are to be hatched. We call the function without any change significance test."
+    "We first compute the different fractions for each robustness aspect."
    ]
   },
   {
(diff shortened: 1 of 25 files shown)


docs/notebooks/ensembles.ipynb
```diff
@@ -275,6 +275,188 @@
...
+   "metadata": {},
+   "source": [
+    "#### Approach C: The advanced alternative.\n",
+    "\n",
+    "Here, we define three categories : robust change, conflicting change and no significant change or robust change. Change is said to be significant if the delta exceed an internal variability. The IPCC document suggests two different ways for computing this variability threshold:\n",
+    "\n",
+    "- $\\gamma = \\sqrt{2}\\cdot1.645\\cdot\\sigma_{20yr}$ where $\\sigma_{20yr}$ is the standard deviation of 20-year means, computed from non-overlapping periods in the pre-industrical control runs, after detrending with a quadratic fit.\n",
+    "- $\\gamma = \\sqrt{\\frac{2}{20}}\\cdot1.645\\cdot\\sigma_{1yr}$ where $\\sigma_{1yr}$ is the interanual standard deviation computed over the linearly detrended reference period.\n",
+    "\n",
+    "The first method involves data that is not always easily available. When such data is at hand, the xclim function could be used with a custom computation of the $\\gamma$ (`gamma`) threshold:\n",
```
huard, line 402:
```suggestion
    "The first method involves data from pre-industrial control simulations that are not always easily available. When such data is at hand, the xclim function could be used with a custom computation of the $\\gamma$ (`gamma`) threshold:\n",
```

huard: The second sentence is unclear. Does it mean the first approach is not implemented ?

docs/notebooks/ensembles.ipynb
```diff
@@ -275,6 +275,188 @@
...
+    "\n",
+    "```python\n",
+    "chg_frac, pos_frac = ensembles.change_significance(fut, ref, test='threshold', abs_thresh=gamma)\n",
+    "```\n",
+    "\n",
+    "The second method is more accessible and xclim implements it directly. However, `pos_frac` output of xclim is the fraction of the members showing _significant_ positive change, while the IPCC defines the agreement separately from the change significance. So we run the function twice."
```
huard, line 408: That's not super clear.

xclim/ensembles/_robustness.py
```diff
@@ -49,7 +50,7 @@ def change_significance(  # noqa: C901
         a distribution across the future period.
         `fut` and `ref` must be of the same type (Dataset or DataArray). If they are
         Dataset, they must have the same variables (name and coords).
-    test : {'ttest', 'welch-ttest', 'mannwhitney-utest', 'brownforsythe-test', 'threshold', None}
+    test : {'ttest', 'welch-ttest', 'mannwhitney-utest', 'brownforsythe-test', 'ipcc-advanced', 'threshold', None}
```
huard, line 53: I'd be more specific regarding naming, qqch comme `ipcc-ar6-c`.

xclim/ensembles/_robustness.py
```diff
@@ -323,7 +331,12 @@ def mwu_wrapper(f, r):  # This specific test can't manage an all-NaN slice
             changed = abs(delta / ref.mean("time")) > kwargs["rel_thresh"]
         else:
             raise ValueError("Invalid argument combination for test='threshold'.")
-
+    elif test == "ipcc-advanced":
+        # Detrend ref
+        refy = ref.resample(time="YS").mean()
+        ref_detrended = detrend(refy, dim="time", deg=1)
+        gamma = np.sqrt(2 / 20) * 1.645 * ref_detrended.std("time")
```
huard, line 338: In the NB, you say we can provide a custom gamma, but I don't see it here.

xclim/ensembles/_robustness.py
huard: Also, let's say we wanted to implement later the first gamma calculation with preindustrial data, would we have to break anything? Should we call it `ipcc-ar6-c-hist` and later implement `ipcc-ar6-c-pi` ? or something ?

xclim/ensembles/_robustness.py
```diff
@@ -111,13 +112,19 @@ def change_significance(  # noqa: C901
         Brown-Forsythe test assuming skewed, non-normal distributions. Same significance criterion as 'ttest'.
       'threshold' :
         Change is considered significative if the absolute delta exceeds a given threshold (absolute or relative).
+      'ipcc-advanced' :
+        An approximation of the "advanced approach" used in the IPCC Atlas and described in :cite:t:`ipccatlas_ar6wg1`.
+        Change is considered significant if the delta exceeds :math:`1.645\sqrt{\frac{2}{10}}\sigma_{ref'}$ where $\sigma_{ref'}` is
```
huard, line 117: Isn't it 2/20 ?

huard (commented):
- Looks good. 
- The nb text assumes the reader's familiarity with the topic and the IPCC approach. I suggest to take a more pedagogic approach. 
- I don't see a test ; )

docs/notebooks/ensembles.ipynb
aulemahal: The `gamma` computation is indeed not implemented for the case where pre-industrial control simulations are available. I'll rephrase.

xclim/ensembles/_robustness.py
aulemahal: Agreed to change `ipcc-advanced` to `ipcc-ar6-c`. 

Yes it would be breaking as this function would now need 3 inputs instead 2. : future, reference (for the deltas) and preindustrial (for `gamma`).

In fact, we could implement both as another function : `gamma(ref, input_type='pi'/'hist')` (or maybe a better name). The `ipcc-ar6-c` test in the main function would simply call `gamma(ref, 'hist')` while a user with PI data would call:
```python3
change_significance(fut, ref, test='threshold', abs_thresh=gamma(pi_data, 'pi'))
```
Does that make sense ?

xclim/ensembles/_robustness.py
aulemahal: See other comment, but indeed, I'll make the distinction clearer in the notebook.

docs/notebooks/ensembles.ipynb
aulemahal: What do you think of the apparent complexity of this ?  See main comment, but the double call comes from the our definition of `pos_frac` that is different from the IPCC's. This could be modified, but would be breaking.

huard: Here I'm not suggestion to change the code itself, just the explanations. In fact, You may not have to give so much information in the notebook. I suggest to focus on the method that's actually implemented (don't mention the first method), and explain how it differs from the IPCC's in a bit more sentences. 

The "So we run the function twice." is a bit cryptic.

xclim/ensembles/_robustness.py
huard: I _think_ having a third input (preindustrial) would be more user friendly. I'd implement the `gamma` function as a private function.

aulemahal:
@huard After our discussions and some thinking, I rewrote the function completely. 

I was not satisfied with the output of `change_significance` so I rewrote it as `robustness_fractions`. The former still works but it will warn the user that it is deprecated. Also added `robustness_categories` to help with making the actual map.

Finally, I simplified the notebook a lot. May be too much...

I'll add tests tomorrow.

aulemahal:
@RondeauG , would you have time to look at the new functions ? And may be @mccrayc, if you have time.

I'm simply looking for a review of the general idea and usage, as shown in the example : https://xclim--1514.org.readthedocs.build/en/1514/notebooks/ensembles.html#Change-significance-and-model-agreement

docs/notebooks/ensembles.ipynb
```diff
@@ -164,6 +164,7 @@
     "# Set display to HTML style (for fancy output)\n",
```
RondeauG, line 164: <p>Quelques détails de syntaxe ou typo:</p><ul><li><code>surimposed</code> --&gt; <code>superimposed</code>   </li><li>change significance : whether most of the ensemble members project <strong>a statistically significant climate change signal, </strong>in comparison to their internal variability.</li><li>There are<s> be</s> 5 different members in the ensemble</li></ul><p> </p><blockquote>(models agree and all show significant change)</blockquote><p>Je dirais plutôt <code>enough show a significant change</code>. L'IPCC utilise un seuil de X% (60?).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Ouranosinc/xclim/pull/1514/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

huard (approved):
Ça me semble beaucoup plus clair comme ça. Pas testé en vrai, juste lu en diagonale.

RondeauG:
This looks pretty good! `robustness_categories` will be very useful.

PR #1529, opened 2023-11-15
Add Lafferty and Sriver partition

### LINKED ISSUE #1497: Uncertainty partitioning method with support for multiple realizations and downscaling method
### Addressing a Problem?

Add uncertainty partitioning method that can support multiple realizations of the same model, and different downscaling method, which would let us analyse differences across ensemble products. 

### Potential Solution

https://www.nature.com/articles/s41612-023-00486-0

### Additional context

_No response_

### Contribution

### Code of Conduct

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #1497 

*  Add partition algo from Lafferty and Sriver (2023)

depends (see options below)

### Other information:
TODO list in this PR:

- [x] Add weights
- [x] add tests
- [x] Maybe add `num` to `hawkins_sutton` ?
- [x]  Clean up based on changes outside of this PR (eg. remove functions that are now elsewhere)

TODO list outside of this PR:

- [ ] Add a more general `graph_fraction_of_total_variance` to figanos (in progress by Juliette, https://github.com/Ouranosinc/figanos/pull/134)
- [ ] Add a function in xscen to prepare the data from a catalog. (https://github.com/Ouranosinc/xscen/pull/289)

In this function, I could rename the dimension to fit what the partition vocabulary (`model`, `scenario`, `downscaling`) OR we could change the partition vocab to the catalog/xscen vocab (`source`, `experiment`, `bias_adjust_project`). Option 2 is my personal preference, but that would be a breaking change for `hawkins_sutton`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: v2023.12.24
+  XCLIM_TESTDATA_BRANCH: v2023.12.14
 
 concurrency:
   # For a given workflow, if we push to the same branch, cancel all previous builds on that branch except on master.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,9 +40,9 @@ jobs:
         python-version:
           - "3.8"
     steps:
-      - uses: actions/checkout@v4.1.0
+      - uses: actions/checkout@v4.1.1
       - name: Set up Python${{ matrix.python-version }}
-        uses: actions/setup-python@v4.7.0
+        uses: actions/setup-python@v4.7.1
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install pylint and tox
@@ -64,9 +64,9 @@ jobs:
           - tox-env: "py39-coverage"
             python-version: "3.9"
     steps:
-      - uses: actions/checkout@v4.1.0
+      - uses: actions/checkout@v4.1.1
       - name: Set up Python${{ matrix.python-version }}
-        uses: actions/setup-python@v4.7.0
+        uses: actions/setup-python@v4.7.1
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install tox
@@ -108,14 +108,14 @@ jobs:
             python-version: "3.11"
             markers: -m 'not slow and not requires_internet'
     steps:
-      - uses: actions/checkout@v4.1.0
+      - uses: actions/checkout@v4.1.1
       - name: Install Eigen3
         if: contains(matrix.tox-env, 'sbck')
         run: |
           sudo apt-get update
           sudo apt-get install libeigen3-dev
       - name: Set up Python${{ matrix.python-version }}
-        uses: actions/setup-python@v4.7.0
+        uses: actions/setup-python@v4.7.1
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install tox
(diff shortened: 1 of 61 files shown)


juliettelavoie:
Right now, `da = variance / variance.sel(uncertainty="total") * 100` is inside the plotting function. But, this goes against the figanos philosphy "only plot, no calculation". I suggest adding a bool arg `fraction` which would replace the variance values in `uncertainty` by fractions or return a 3rd output with fractions.

Zeitsperre:
@juliettelavoie Be sure to follow the steps in the README for testing things on the PR before merging to master: https://github.com/Ouranosinc/xclim-testdata/blob/main/README.md

huard:
I'm feeling so-so regarding the fraction keyword. The issue I see is that there are a bunch of different figures typically created from those results, some using the fraction of variance, others using the variance itself. If the fraction calculation is embedded in the function, you need to run it twice.

My proposal would be to add a function, e.g. `to_fraction` that would do this calculation. 
The workflow would then be
```
uncertainty = lafferty_sriver(da)
fraction = to_fraction(uncertainty)
figanos.plot(fraction)
```

juliettelavoie:
@huard How about a `fraction` arg that creates a 3rd output instead of replacing the second?

huard:
Meh, I also find it annoying when the number of output changes based on the arguments you pass. It reminds me too much of matlab. ; )

huard:
Added a function for fractional uncertainty, but did not make any changes to the functions themselves. The test can reproduce the figure from the notebook. I think we're almost there.

tests/test_partitioning.py
```diff
@@ -67,3 +68,110 @@ def test_hawkins_sutton_synthetic(random):
...
+    time = xr.date_range("1970-01-01", periods=60, freq="Y")
+    da = xr.DataArray(
+        x, dims=("scenario", "model", "downscaling", "time"), coords={"time": time}
+    )
+    m, v = lafferty_sriver(da)
+    # Mean uncertainty over time
+    vm = v.mean(dim="time")
+
+    # Check that the mean relative to the baseline is zero
+    np.testing.assert_array_almost_equal(m.mean(dim="time"), 0, decimal=1)
```
juliettelavoie, line 100: I don't think this test makes sense in the context of lafferty and sriver because we are not looking at deltas like hawkins and sutton.

tests/test_partitioning.py
```diff
@@ -67,3 +68,110 @@ def test_hawkins_sutton_synthetic(random):
...
+
+    # Check that the scenario uncertainty is zero
+    np.testing.assert_array_almost_equal(vm.sel(uncertainty="scenario"), 0, decimal=1)
```
juliettelavoie, line 103: I don't understand why the time mean of scenario uncertainty is supposed to be 0 ?

juliettelavoie: just the saw the comment above saying that is is constant in time, but again lafferty doesn't use deltas. I don't understand why the test passed..

tests/test_partitioning.py
huard: You`re right, I copy-pasted the test a few weeks ago to exercise the function and didn't think it through.

tests/test_partitioning.py
huard: Same reason for the other point you mentioned, it's a dumb copy-paste.

huard: We need to come up with tests that actually make sense for the L-S method.

huard (approved):

juliettelavoie:
J'assume qu'il faut merge la PR dans xclim-testdata pour que le test passe ?

juliettelavoie:
> @juliettelavoie Be sure to follow the steps in the README for testing things on the PR before merging to master: https://github.com/Ouranosinc/xclim-testdata/blob/main/README.md

@Zeitsperre not sure what I am supposed to do in this PR  to make the test pass ?

.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: v2023.12.14
+  XCLIM_TESTDATA_BRANCH: v2023.12.24
```
Zeitsperre: ?

PR #1538, opened 2023-11-24
fix freq/offset/indexer attrs in SPI

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Fixes wrong attributes in SPI/SPEI functions, allow to save files in netcdf files.

No

### DIFF
--- a/xclim/indices/stats.py
+++ b/xclim/indices/stats.py
@@ -660,7 +660,7 @@ def standardized_index_fit_params(
     window: int,
     dist: str,
     method: str,
-    offset: Quantified = "",
+    offset: Quantified | None = None,
     **indexer,
 ) -> xr.DataArray:
     r"""Standardized Index fitting parameters.
@@ -712,7 +712,7 @@ def standardized_index_fit_params(
             f"The method `{method}` is not supported for distribution `{dist}`."
         )
 
-    if offset != "":
+    if offset is not None:
         with xr.set_options(keep_attrs=True):
             da = da + convert_units_to(offset, da, context="hydro")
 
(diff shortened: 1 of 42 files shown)


xclim/indices/stats.py
```diff
@@ -729,10 +729,12 @@ def standardized_index_fit_params(
         "scipy_dist": dist,
         "method": method,
         "group": group,
-        "time_indexer": indexer,
         "units": "",
         "offset": offset,
     }
+    if indexer != {}:
+        params.attrs["time_indexer"] = str(indexer)
+
```
coxipi, line 737: I think having info on the indexer used could be useful, unless it's already part of some automatically generated history info, but I don't think so.  

For instance, if `indexer == {"date_bounds":("01-01","07-01")}`, then this would be saved as:
`"{'date_bounds': ('01-01', '07-01')}"`
It's a bit clumsy but at least it keeps information. Feedback on this?

xclim/indices/stats.py
```diff
@@ -660,7 +660,7 @@ def standardized_index_fit_params(
     window: int,
     dist: str,
     method: str,
-    offset: Quantified | None = None,
+    offset: Quantified = "",
```
huard, line 663: Disclaimer: I don't really understand what the original issue is. 

In indices/_conversion.py, there are a lot of instances of a Quantified parameter set to None. I suspect it would be best to have consistency in how we represent default null Quantified values.  

Would doing something like
`offset = offset or ""`
let you keep None as the default but not mess with attribute records?

coxipi: Yes, that's a good way around, I'll adopt this instead, thanks

xclim/indices/stats.py
coxipi: Ok, in short to recover indexer, we need: 
```python
        if indexer == "":
            indexer = {}
        else:
            indexer = {indexer[0]:indexer[1:]}
```
Is this really the first time this kind of operation in xclim is needed?

xclim/indices/stats.py
coxipi: The original issue was what you describe, `None` and `{}` mess the attributes records, an array can'T be saved in netcdf. So I wanted to change the default values in the code (so that what is used in the code would be the same thing as in attrs), but as you suggest, simply changing was is stored in the attrs but keep using `None` in the rest of the code is also possible.

aulemahal (commented):
Looks good!

aulemahal (approved):
woups

xclim/indices/stats.py
huard: Do we need to do the same with other indices ?

huard (approved):

PR #1540, opened 2023-11-28
Temporary workaround for pandas and xarray frequency changes

### LINKED ISSUE #1534: Adopt new pandas frequency codes
### Addressing a Problem?

Pandas has made changes to the frequency string codes, which will begin in version 2.2.

The first step is the introduction of a new set of default codes:

- `A` -> `Y`
- `H` -> `h`
- `T` -> `min`
- `S` -> `s` (and so on for smaller periods)

And the addition of the "E" suffix to denote a frequency anchored at the end of the period (instead of it being the default when not mentioned) (applies to Q, and M).

All the previous codes are now deprecated, but will be valid _inputs_ for 2.2.

### Potential Solution

If we change all the hardcoded codes to the new ones, we will need to pin pandas, so a temporary dynamic solution might be necessary (as done in xarray, but more complex).

I suggest a `ensure_correct_freq()` function that is called everywhere a frequency code is sent to xarray/pandas. The function will translate using the old or the new version depending on the installed version of pandas.

### Additional context

The function could be removed when we pin pandas >= 2.2 and all codes written with the new syntax.

xscen could make use of that function also, if needed.

See xarray issue, PR and links to the Pandas issues : https://github.com/pydata/xarray/issues/8394

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1534

This PR became a bit more complicated when I realized that xarray's latest release implements the "new" standards, but pandas 2.2 has not yet been released. And some of those changes are not implemented at all yet in pandas[^1]. I.e. `pd.tseries.frequencies.to_offset(xr.infer_freq(da))` sometimes raises an error.

* New function  `fix_freq` : it takes in a frequency code and translates it to the appropriate syntax. There are, sadly, three cases:
    + xarray >= 2023.11.0, pandas >= 2.2 : The "new" syntax (see below). Some changes are not yet supported by pandas 2.1.
    + xarray >= 2023.11.0, but pandas < 2.2 : The "inter[mediate]" syntax. Changes the subdaily codes, but not the coarser ones.
    + xarray < 2023.11.0, pandas < 2.2 : The "old" syntax, the one you know.

By « syntax » I mean the _default_ codes. The "inter" version is supported by all previous xarray versions and pandas >= 1.0.

* All hard-coded frequencies have been adapted for the "inter" syntax. (in the code, tests and notebooks)

* `fix_freq` has been added everywhere conflict was possible
* Raised the minimum pin on pandas so we can at least use the "inter" codes and not have a fourth category. 

My goal here was to reduce the number of warnings for users. I wanted to avoid xclim's default to trigger user warnings or even errors. I don't think I achieved that goal, but I hope the transition will be smoother nonetheless.

For sure. I think a normal "beginner" usage of xclim would be unchanged, but I'm not so sure about things like `xscen` or custom indicator modules. But some of that is broken anyway with xarray 2023.11.0 and would further break with pandas 2.2. Hopefully, we are able to control the damage with this PR and makes minimum breaking changes here only.

### Other information:
The long term goal is to remove `fix_freq` when we pin pandas >= 2.2. This PR also introduces a few workarounds to remove at the moment, all denoted with "FIXME"  comments.

#### Major differences
| old    | inter  | new    |
|--------|--------|--------|
| AS-JAN | AS-JAN | YS-JAN[^2] |
| A-JAN  | A-JAN  | YE-JAN |
| Q-DEC  | Q-DEC  | QE-DEC |
| M      | M      | ME     |
| H      | h      | h      |
| T      | min    | min    |
| S      | s      | s      |

[^1]: and I was a reviewer on that PR in xarray. Woopsie!
[^2]: The "new" syntax is not yet supported by pandas (< 2.2), but xarray (>= 2023.11) will warn and return that in `xr.infer_freq`.


xclim/core/calendar.py
```diff
@@ -834,14 +914,99 @@ def construct_offset(mult: int, base: str, start_anchored: bool, anchor: str | N
...
+    "H": "h",
+    "T": "min",
+    "S": "s",
+    "L": "ms",
+    "U": "us",
+    "N": "ns",
+}
+
+
+# TODO: This can be removed when we pin pandas >= 2.2
```
Zeitsperre, line 948: Maybe we can integrate a small helper function here to check what the required pins are?

Something like:
```
from importlib.metadata import distribution

for req in distribution("xclim").requires:
    if req.split(";")[0].startswith("pandas>=2.2"):
        raise FixMeOrSomethingPleaseException()
```

Zeitsperre (approved):
Well played. What are the chances that we could push the freq fixing stuff upstream to `xarray`? This change must be affecting tons of users in addition to us.

aulemahal:
Closed as this seems to complicated. A drastic change in the pins seems cleaner.

PR #1558, opened 2023-12-11
Stack and unstack periods

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

New `stack_periods` and it's reverse `unstack_periods`.

The basic idea is to have `da.rolling(time=window).construct(dim, stride=stride)`, but with `window` and `stride` given as a multiple of a given `freq`, allowing the use of non-uniform temporal periods (like years or months). And to have a reverse method to go back to the linear timeseries.

We had `sdba.construct_moving_yearly_window`, this is the generalization of that logic for:
- other frequencies than 'YS'
- periods of different lengths (MS, QS)
- non-uniform calendars
Of course, the two latter points will generate an output where the "fake" time axis makes no sense at all, so this is to be use with precaution.

Yes.

`sdba.construct_moving_yearly_window` is transparently sent to `stack_periods` with a warning.

`sdba.unpack_moving_yearly_window` also warns and calls `unstack_periods` but the `append_ends` argument is not available on the latter. The behaviour is that same as `append_ends=True`.

### DIFF
--- a/docs/notebooks/sdba-advanced.ipynb
+++ b/docs/notebooks/sdba-advanced.ipynb
@@ -357,7 +357,13 @@
     "\n",
     "The \"unstack\" function does the opposite : it concatenates the windows together to recreate the original timeseries. It only works for the no-overlap case where `stride = window` and for the non-ambiguous one where `stride` divides `window` into an odd number (N) of parts. In that latter situation, the middle parts of each period are kept when reconstructing the timeseries, in addition to the first (last) parts of the first (last) period needed to get a full timeseries.\n",
     "\n",
-    "Here, as `ref` and `hist` cover 15 years, we will use a window of 15 on sim. With a stride of 5 years, this means the first window goes from 2000 to 2014 (inclusive). Then 2005-2019, 2010-2024 and 2015-2029. The last year will be dropped."
+    "Quantile Delta Mapping requires that the adjustment period should be of a length similar to the training one. As our `ref` and `hist` cover 15 years, we will transform `sim` by stacking windows of 15 years. With a stride of 5 years, this means the first window goes from 2000 to 2014 (inclusive). Then 2005-2019, 2010-2024 and 2015-2029. The last year will be dropped.\n",
+    "\n",
+    "<div class=\"alert alert-warning\">\n",
+    "\n",
+    "In the following example, `QDM` is configurated with `group=\"time.dayofyear\"` which will perform the adjustment for each day of year (doy) separately. When using `stack_periods` the extracted windows are all concatenated along the new `period` axis and they all share the same time coordinate. As such, for the doy information to make sense, we must use a calendar with uniform year lengths. Otherwise, the doys would shift one day at each leap year.\n",
+    "\n",
+    "</div>"
    ]
   },
   {
--- a/docs/notebooks/sdba-advanced.ipynb
+++ b/docs/notebooks/sdba-advanced.ipynb
@@ -346,21 +346,18 @@
     "\n",
     "Some Adjustment methods require that the adjusted data (`sim`) be of the same length (same number of points) than the training data (`ref` and `hist`). These requirements often ensure conservation of statistical properties and a better representation of the climate change signal over the long adjusted timeseries.\n",
     "\n",
-    "In opposition to a conventional \"rolling window\", here it is the _years_ that are the base units of the window, not the elements themselves. `xclim` implements `sdba.construct_moving_yearly_window` and `sdba.unpack_moving_yearly_window` to manipulate data in that goal. The \"construct\" function cuts the data in overlapping windows of a certain length (in years) and stacks them along a new `\"movingdim\"` dimension, alike to xarray's `da.rolling(time=win).construct('movingdim')`, but with yearly steps. The step between each window can also be controlled. This argument is an indicator of how many years overlap between each window. With a value of 1 (the default), a window will have `window - 1` years overlapping with the previous one. `step = window` will result in no overlap at all.\n",
+    "In opposition to a conventional \"rolling window\", here it is the _years_ that are the base units of the window, not the elements themselves. `xclim` implements `xc.core.calendar.stack_periods` and `xc.core.calendar.unstack_periods` to manipulate data in that goal. The \"stack\" function cuts the data in overlapping windows of a certain length and stacks them along a new `\"period\"` dimension, alike to xarray's `da.rolling(time=win).construct('period')`, but with yearly steps. The stride (or step) between each window can also be controlled. This argument is an indicator of how many years overlap between each window. With a value of 1, a window will have `window - 1` years overlapping with the previous one. The default (`None`) is to have `stride = window` will result in no overlap at all. The default units in which `window` and `stride` are given is a year (\"YS\"), but can be changed with argument `freq`.\n",
     "\n",
-    "By default, the result is chunked along this `'movingdim'` dimension. For this reason, the method is expected to be more computationally efficient (when using `dask`) than looping over the windows.\n",
+    "By chunking the result along this `'period'` dimension, it is expected to be more computationally efficient (when using `dask`) than looping over the windows with a for-loop (or a `GroupyBy`)\n",
     "\n",
     "Note that this results in two restrictions:\n",
     "\n",
     "1. The constructed array has the same \"time\" axis for all windows. This is a problem if the actual _year_ is of importance for the adjustment, but this is not the case for any of xclim's current adjustment methods.\n",
     "2. The input timeseries must be in a calendar with uniform year lengths. For daily data, this means only the \"360_day\", \"noleap\" and \"all_leap\" calendars are supported.\n",
     "\n",
-    "The \"unpack\" function does the opposite : it concatenates the windows together to recreate the original timeseries.\n",
-    "The time points that are not part of a window will not appear in the reconstructed timeseries.\n",
-    "If `append_ends` is True, the reconstructed timeseries will go from the first time point of the first window to the last time point of the last window. In the middle, the central `step` years are kept from each window.\n",
-    "If `append_ends` is False, only the central `step` years are kept from each window. Which means the final timeseries has  `(window - step) / 2` years missing on either side, with the extra year missing on the right in case of an odd `(window - step)`. We are missing data, but the contribution from each window is equal.\n",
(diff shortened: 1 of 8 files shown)


docs/notebooks/sdba-advanced.ipynb
```diff
@@ -346,21 +346,18 @@
     "\n",
```
coxipi, line 346: <p>Assuming that original full period is a total of 30 years, so:</p><p>[01-01-2000, 01-01-2030[</p><p>i.e. excluding 2030, including 2029.</p><p> </p><pre class="ql-syntax" spellcheck="false">Window=15, stride=5

[01-01-2000, 01-01-2015[
[01-01-2005, 01-01-2020[
[01-01-2010, 01-01-2025[
[01-01-2015, 01-01-2030[
</pre><p>So in this case I don't see how you must drop one year, am I missing something?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Ouranosinc/xclim/pull/1558/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

coxipi: Ok. Then I agree, having 31 years will be awkward with the stacking. 

It seems this was intentional, previously, there was a selection afterwards to restrict `sim` to a 30 years period. Anyways, I think reminding that the dataset here "covers 31 years" and, maybe say that a dataset with 30 years is more natural with those parameters (unchanged by stacking/unstacking)

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+        if frqM.onOffset(time):
+            return frq.onOffset(time)
+        return frq.onOffset(frqM.rollback(time))
+    # Pandas
+    time = pd.Timestamp(time)
+    frqM = pd.tseries.frequencies.to_offset("MS")
+    frq = pd.tseries.frequencies.to_offset(freq)
+    if frqM.is_on_offset(time):
+        return frq.is_on_offset(time)
+    return frq.is_on_offset(frqM.rollback(time))
```
coxipi, line 1608: Could we convert to a cftime and just recall the function? I tried and obtaining a cftime is not super obvious with the APIs?

coxipi: e.g. 
```suggestion
    # Pandas
    time = pd.Timestamp(time)
    time = cftime.datetime(*[getattr(time, meth) for meth in ["year", "month", "day", "hour", "minute", "second"]])
    _month_is_first_period_month(time, freq)
```

maybe it's more complicated or could lead to problem, feel free to ignore

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+        Output might be strange if this does not use an uniform calendar (noleap, 360_day, all_leap).
+    window : int
+        The length of the moving window as a multiple of ``freq``.
+    stride : int, optional
+        At which interval to take the windows, as a multiple of ``freq``.
+        For the operation to be reversible with :py:func:`unstack_periods`, it must divide `window` into an odd number of parts.
+        Default is `window` (no overlap between periods).
+    min_length : int, optional
+        Windows shorter than this are not included in the output.
+        Given as a multiple of ``freq``. Defaults is ``window`` (every window must be complete).
```
coxipi, line 1643:
```suggestion
        Given as a multiple of ``freq``. Default is ``window`` (every window must be complete).
```

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+            open_ended = win_slc.stop is None
+        if open_ended:
+            # Too short, we got to the end
+            break
+        if (
+            strd_slc.start == 0
+            and parse_offset(freq)[1] in "YAQ"
+            and min_length == window
+            and not _month_is_first_period_month(da.time[0].item(), freq)
+        ):
```
coxipi, line 1730: Just so I understand well. Can a good choice of `window, stride, min_length` avoid to trigger this condition? Or is this necessary because of the time unit implied with a freq e.g. `QS-DEC`. For instance, if I choose `freq="MS"`, with `window=stride=3`, I can reproduce `freq=QS-JAN` with `window=stride=1`? Having this if condition allows to cover certain cases that are not possible because of the inflexibility of `QS-JAN` vs. working with `MS` and bigger window/strides?

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+
+    try:
+        starts = da[dim]
+        window = starts.attrs["window"]
+        stride = starts.attrs["stride"]
+        freq = starts.attrs["freq"]
+        unequal_lengths = bool(starts.attrs["unequal_lengths"])
+    except (AttributeError, KeyError) as err:
+        raise ValueError(
+            f"`unstack_periods` can't find the window, stride and freq attributes on the {dim} coordiantes."
```
coxipi, line 1815:
```suggestion
            f"`unstack_periods` can't find the window, stride and freq attributes on the {dim} coordinates."
```

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+            )
+        return xr.concat(periods, "time")
+
+    # Difficult and ambiguous case
+    if (window / stride) % 2 != 1:
+        raise NotImplementedError(
+            "`unstack_periods` can't work with strides that do not divide the window into an odd number of parts."
+            f"Got {window} / {stride} which is not an odd integer."
+        )
+
```
coxipi, line 1866: This warning should be in the `stack_function` no?

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+    mult, *args = parse_offset(freq)
+    strd_frq = construct_offset(mult * stride, *args)
+
+    periods = []
+    for i, (start, length) in enumerate(zip(starts.values, lengths.values)):
+        real_time = _reconstruct_time(start)
+        slices = real_time.resample(time=strd_frq).groupers[0].group_indices
+        if i == 0:
+            slc = slice(slices[0].start, min(slices[mid].stop, length))
```
coxipi, line 1879: A bit confused here. There should be a special case like this for all stride windows where the full window is not included in the full period of interest. Hope this example shows well what I mean: 

window=5
stride=1
==> mid = 2


For i=2,3,4, using mid=2 correctly selects the middle of the window
But for i=1, the "o" should be in position "1", not mid=2 ? 
Agaub for i =0, mid=2 is not appropriate, the middle or the window (with NaNs on left) is 0.
  | i   | relative position of "o" among "x"'s | Not in time | 0   | 1   | 2   | 3   | 4   | 5   | 6   |
  | --- | -------------------------------------------- | ----------: | --- | --- | --- | --- | --- | --- | --- |
  | 4   | 2                                            |             |     |     | x   | x   | o   | x   | x   |
  | 3   | 2                                            |             |     | x   | x   | o   | x   | x   |     |
  | 2   | 2                                            |             | x   | x   | o   | x   | x   |     |     |
  | 1   | 1                           |           Nan | x   | o   | x   | x   |     |     |     |
  | 0   | 0                            |        Nan  Nan | o   | x   | x   |     |     |     |     |

coxipi: So, more briefly, could it be that we need the condition: `i<mid` ?

xclim/core/calendar.py
```diff
@@ -1588,3 +1589,302 @@ def get_doys(start, end, inclusive):
...
+        elif i == da.period.size - 1:
+            slc = slice(slices[mid].start, min(slices[Nwin - 1].stop or length, length))
```
coxipi, line 1881: probably the same remark applies here

xclim/core/calendar.py
aulemahal: This condition is needed because the anchor-based frequencies can start far in the past. `QS-JAN` is the same as `3MS` only if the data starts in january.  Otherwise, a "nMS" will always start on the first month of the data. 
The case I had in mind here was data starting in january and  `freq='QS-DEC'` or `'AS-JUL'`.

xclim/core/calendar.py
aulemahal: I wanted it to be possible to make irreversible stacking. 

We could add a warning.

xclim/core/calendar.py
aulemahal: Hmm, I think your example is not possible with the current function. The first window (`i = 0`) is always complete. Maybe you expected something looking `rolling(..., center=True)`, that's not what I implemented.

If we don't touch the `min_length` arg and only keep full windows, we get:

  | i   | extracted indexes |  0   | 1   | 2   | 3   | 4   | 5   | 6   |
  | --- | --------------------------------------------  | --- | --- | --- | --- | --- | --- | --- |
  | 2   | 2:                                            |     |     | x   | x   | o   | o   | o   |
  | 1   | 2                                            |     | x   | x   | o   | x   | x   |     |
  | 0   | :2                                            | o   | o   | o   | x   | x   |     |     |

If we allow `min_length=4`:

  | i   | extracted indexes |  0   | 1   | 2   | 3   | 4   | 5   | 6   |
  | --- | --------------------------------------------  | --- | --- | --- | --- | --- | --- | --- |
  | 3   | 2:                                            |     |     |      | x   | x   | o   | o   |
  | 2   | 2                                            |     |     | x   | x   | o   | x   | x   |
  | 1   | 2                                             |     | x   | x   | o   | x   | x   |     |
  | 0   | :2                                            | o   | o   | o   | x   | x   |     |     |

It was easier to code it like this than to imitate `center=True`, but indeed that might be confusing...

aulemahal: I think I'll add a table like this to the docstring!

xclim/core/calendar.py
coxipi: You can leave it as it is then, will try to think about `pd.Timestamp`

xclim/core/calendar.py
coxipi: Ah right. You have a test for this. Nevermind

xclim/core/calendar.py
coxipi: ah right, `rolling` by default does not center, oups.

coxipi (approved):
Nice job. I think this kind of logic could be useful to replace `map_blocks/map_groups` operations. Either to create arrays we directly work on, or to deduce time indices that can be used with the philosophy of groupies.

PR #1566, opened 2023-12-18
reorder with window

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* Allow reorder on group including a window

Previous behavior was to ignored the window (silently) and only do the reorder on the group (eg. dayofyear), now the window is considered.

### Other information:
This is for MBCn. In an email, Cannon said: "All parts of the process, including the reordering, are done on your subset, including the buffer around the day of interest. If you’re using 31 days x 30 years then you have 930 data points. Bias adjustment would be performed on these 930 cases. At the end, you’d reorder according to the QDM ranks and select the middle day. Repeat for the remaining days of the year."


xclim/sdba/_processing.py
```diff
@@ -158,17 +158,41 @@ def _reordering(ds, *, dim):
...
-            output_dtypes=[ds.sim.dtype],
+    def _reordering_2d(data, ordr):
+        data_r = data.ravel()
+        ordr_r = ordr.ravel()
+        reorder = np.sort(data_r)[np.argsort(np.argsort(ordr_r))]
+        return reorder.reshape(data.shape)[
+            :, int(data.shape[0] / 2)
+        ]  # pick the middle of the window
+
+    if "window" in ds.dims:
```
aulemahal, line 169:
```suggestion
    if "window" in dim:
```
I would decide according to the called list. `Grouper.apply` calls with `dims` as a list when `main_only` is false:
https://github.com/Ouranosinc/xclim/blob/e678308f2c5cde75a3c63cddcab7835dab1e422e/xclim/sdba/base.py#L389-L394

xclim/sdba/_processing.py
```diff
@@ -158,17 +158,41 @@ def _reordering(ds, *, dim):
...
+                input_core_dims=[["time", "window"], ["time", "window"]],
+                output_core_dims=[["time"]],
+                vectorize=True,
+                dask="parallelized",
+                output_dtypes=[ds.sim.dtype],
+            )
+            .rename("reordered")
+            .to_dataset()
+        )
+    else:
```
aulemahal, line 184: Missing a conditional where it fails if `len(dims) >  1` but the other one is not `window`.

xclim/sdba/_processing.py
```diff
@@ -158,17 +158,41 @@ def _reordering(ds, *, dim):
...
+        return (
+            xr.apply_ufunc(
+                _reordering_1d,
+                ds.sim,
+                ds.ref,
+                input_core_dims=[[dim], [dim]],
+                output_core_dims=[[dim]],
```
aulemahal, line 191:
```suggestion
                input_core_dims=[dim, dim],
                output_core_dims=[dim],
```

xclim/sdba/_processing.py
```diff
@@ -158,17 +158,47 @@ def _reordering(ds, *, dim):
...
-            output_core_dims=[[dim]],
-            vectorize=True,
-            dask="parallelized",
-            output_dtypes=[ds.sim.dtype],
+    def _reordering_2d(data, ordr):
+        data_r = data.ravel()
+        ordr_r = ordr.ravel()
+        reorder = np.sort(data_r)[np.argsort(np.argsort(ordr_r))]
+        return reorder.reshape(data.shape)[
+            :, int(data.shape[0] / 2)
```
coxipi, line 166: This should `data.shape[1]` no? `window` is the second dimension below. 

Also, I think that `map_blocks` does this automatically for you, depending on how you call it. I think `reordered=[Grouper.DIM]` would already do this, I'm not 100% sure. @aulemahal ?

juliettelavoie: oui oups en effet pour [1].

not sure how to do that ? it fails if I don't return something the size of time.

xclim/sdba/_processing.py
```diff
@@ -159,17 +159,47 @@ def _reordering(ds, *, dim):
...
-            output_dtypes=[ds.sim.dtype],
+    def _reordering_2d(data, ordr):
+        data_r = data.ravel()
+        ordr_r = ordr.ravel()
+        reorder = np.sort(data_r)[np.argsort(np.argsort(ordr_r))]
+        return reorder.reshape(data.shape)[
+            :, int(data.shape[1] / 2)
+        ]  # pick the middle of the window
+
+    if "window" in dim:
```
aulemahal, line 170:
```suggestion
    if {"window", "time"} == set(dim):
```
Since these are hardcoded in the `apply_ufunc` below.

aulemahal (approved):
This fell between the cracks, sorry. AFAIU, this looks good to go. Except for a nitpicky detail in the conditional.

xclim/sdba/_processing.py
aulemahal: Indeed, `map_groups` should slice to the middle of the window if the results still has a `window` dimension. That would require the `apply_ufunc` to declare `window` in its `output_core_dims`. However, as the result is the same, I think that slicing directly in the ufunc make more sense, and might yield better performances!

xclim/sdba/_processing.py
```diff
@@ -159,17 +159,47 @@ def _reordering(ds, *, dim):
...
+                dask="parallelized",
+                output_dtypes=[ds.sim.dtype],
+            )
+            .rename("reordered")
+            .to_dataset()
+        )
+    else:
+        raise ValueError(
+            f"Reordering can only be done along one dimension."
+            f" If there is more than one, one of the dimension should be `window`."
```
juliettelavoie, line 203:
```suggestion
            f" If there is more than one, they should be `window` and `time`."
```

PR #1580, opened 2024-01-09
Fast MBCn (a la groupies)

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

New `MBCn` TrainAdjust class. The train part finds adjustment factors for the npdf transform. The adjust part does the rest.

* A single numpy function to perform all rotations of the npdf_transform makes the process faster 
* Grouping is handled using the same logic as in numpy_groupies. I initially tried to stop using map_blocks by using what I call a the Big Dataset (BD) solution. It was a dataset that included the group windowed blocks. This was working well but sometimes caused  dask workers to die. Maybe a better chunking could have solved this problem. But instead of constructing a BD, we simply loop over blocks, and simply specify time indices in each block (à la groupies) in the original datasets. The resulting code is a bit more messy, but it seems to be working well performance-wise. 

The function also changes how windowed group blocks are handled throughout the computation. Now, a block is preserved its form from begin to start of the MBCn computation. 
* This is in contrast to the current way which was grouping and ungrouping block between each iteration of the NpdfTransform. 
* The standardization is performed on a block
* The univariate bias correction is maintainted as blocks, reordered, *then* the blocks are ungrouped
* In the sdba notebook, it was suggested that we should give the univariate bias corrected datasets in the npdf transform. But following (Cannon, 2018), we should input the raw datasets in the npdf transform. This change should not really matter that much, but still, to perform exactly the MBCn as presented by Cannon, this change is necessary.

All these changes will result in a different output for `window>1` and our implementation should now match that of Cannon.

No

### Other information

* It might be worthwhile to retest `map_blocks` to see if, with the rest of changes, it can offer a good performance. It would be cleaner code
* Using BD would also simplify many things, worth re-exploring if it can maintain the performance


xclim/sdba/adjustment.py
```diff
@@ -1113,79 +1157,120 @@ class NpdfTransform(Adjust):
...
-        rot_matrices = rot_matrices or rand_rot_matrix(
-            ref[pts_dim], num=n_iter, new_dim=rot_dim
-        ).rename(matrices="iterations")
-
-        # Call a map_blocks on the iterative function
-        # Sadly, this is a bit too complicated for map_blocks, we'll do it by hand.
-        escores_tmpl = xr.broadcast(
-            ref.isel({pts_dim: 0, "time": 0}),
-            hist.isel({pts_dim: 0, "time": 0}),
-        )[0].expand_dims(iterations=rot_matrices.iterations)
```
coxipi, line 1154: escores need to be reimplemented

.github/ISSUE_TEMPLATE/0001-GENERIC-ISSUE-TEMPLATE.yml
Zeitsperre, line 1: What happened here?

.github/DISCUSSION_TEMPLATE/questions.yml
```diff
@@ -6,7 +6,7 @@ body:
     attributes:
       value: |
         Thanks for opening this discussion!
-        Before you submit, please make sure you have read our [Code of Conduct](https://github.com/Ouranosinc/xclim/blob/master/CODE_OF_CONDUCT.md).
```
Zeitsperre: Good call, thanks!

xclim/sdba/utils.py
```diff
@@ -531,6 +563,13 @@ def rank(
     return rnk
 
 
+def _rank_np(arr, axis=None):
+    rnk = bn.nanrankdata(arr, axis=axis)
+    rnk = rnk / np.nanmax(rnk, axis=axis, keepdims=True)
+    mx, mn = 1, np.nanmin(rnk, axis=axis, keepdims=True)
+    return mx * (rnk - mn) / (mx - mn)
+
+
```
coxipi, line 572:
```suggestion
```

xclim/sdba/_adjustment.py
```diff
@@ -36,7 +38,9 @@ def _adapt_freq_hist(ds, adapt_freq_thresh):
     hist_q=[Grouper.PROP, "quantiles"],
     scaling=[Grouper.PROP],
 )
-def dqm_train(ds, *, dim, kind, quantiles, adapt_freq_thresh) -> xr.Dataset:
+def dqm_train(
+    ds, *, dim, kind, quantiles, adapt_freq_thresh, jitter_under_thresh_value
+) -> xr.Dataset:
```
coxipi, line 43: Maybe there's a better argument name. What seems more natural `jitter_under_thresh` is already used for the name of the function that does the jittering, so I thought it was best not to do this. 

Now I can't remember if this change was 100% necessary, but it was at least more practical to be able to include this option in the kwargs. I think that philosophically, 'adapt_freq' and 'jitter_under' are part of the 'training' and including 'adapt_freq' was necessary as we need to compare the zeroes of the grouped `ref, hist`. Again, I'm not sure this is 100% necessary for 'jitter_under_thresh', so I can revisit this aspect if preferable

xclim/sdba/_adjustment.py
```diff
@@ -161,8 +163,8 @@ def mbcn_train(
     Parameters
     ----------
     Dataset variables:
-        ref : training target
-        hist : training data
+        ref : training target, DataArray (stacked variables)
+        hist : training source, DataArray (stacked variables)
```
Zeitsperre, line 167: This should open with:
```
ds : xr.Dataset
    Dataset variables:
        ...

File: CHANGES.rst [hunk 6]
```diff
@@ -83,6 +88,7 @@ New features and enhancements
 * Added a new ``xclim.indices.generic.select_rolling_resample_op`` function to allow for computing rolling statistics. (:issue:`1480`, :pull:`1643`).
 * Add the possibility to use a group with a window in ``xc.sdba.processing.reordering``. (:pull:`1566`).
 
+
```
coxipi, line 91:
```suggestion

```

xclim/sdba/_adjustment.py
coxipi: ok, I will change all docstrings in _adjustment, that's not specific to my PR

coxipi: Hmm, I get "Unexpected indentation." for 
```python
    ds : xr.Dataset
        Dataset variables:
        ref : training target
        hist : training source
```
or
```python
    ds : xr.Dataset
        Dataset variables:
            ref : training target
            hist : training source
```
any other suggestion?

Zeitsperre: I guess we can't use ":" in the descriptions. What about:
```
ds : xr.Dataset
    Dataset variables must be :
    ref (training target)
    hist (training source)
```

coxipi: got it, I had to fight with my tabs

xclim/sdba/nbutils.py
```diff
@@ -61,6 +61,40 @@ def vecquantiles(
...
+    # compute the indices where each quantile should go:
+    # nb: nan goes to the end, so we need to subtract numnan to the size.
+    indices = q * (arr.size - 1 - numnan)
+    # compute the quantiles manually to avoid casting to float64:
+    # (alternative is to use np.interp(indices, np.arange(arr.size), sortarr))
+    frac = indices % 1
+    low = np.floor(indices).astype(np.int64)
+    high = np.ceil(indices).astype(np.int64)
+    return (1 - frac) * sortarr[low] + frac * sortarr[high]
+
```
coxipi, line 96: I just wanted something fast with my test with Juliette, don't know how we should proceed on this one

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1212,279 @@ def _adjust(
...
+        )
+        obj.set_dataset(out)
+        return obj
+
+    def adjust(
+        self,
+        sim: xr.Dataset,
+        *,
+        ref: xr.Dataset | None = None,
+        hist: xr.Dataset | None = None,
```
coxipi, line 1431: The structure here is not ideal: 

1. It needs `sim, ref,hist` but `ref,hist` were already needed in `train`. This is because I only do the npdf training in the training portion, so still need `ref,hist` after that.
2. Since it's still a BaseAdjust class, I think it expects only `sim` as an arg, but `ref,hist` are really not optional here, just a constraint of the class structrure. Perhaps I should define another class type.

xclim/sdba/adjustment.py
```diff
@@ -370,6 +374,7 @@ def _train(
         kind: str = ADDITIVE,
         group: str | Grouper = "time",
         adapt_freq_thresh: str | None = None,
+        jitter_under_thresh_value: str | None = None,
     ) -> tuple[xr.Dataset, dict[str, Any]]:
```
coxipi, line 378: not a fan of this keyword `jitter_under_thresh_value`, but there is already a function `jitter_under_thresh`, so not sure it's a good idea to use the same name for a string input

xclim/sdba/adjustment.py
```diff
@@ -370,6 +374,7 @@ def _train(
         kind: str = ADDITIVE,
         group: str | Grouper = "time",
         adapt_freq_thresh: str | None = None,
+        jitter_under_thresh_value: str | None = None,
```
coxipi, line 377: not a fan of this keyword jitter_under_thresh_value, but there is already a function jitter_under_thresh, so not sure it's a good idea to use the same name for a string input

coxipi:
I've cleaned the function a lot, it should be good for a read, hope this is digestible...

xclim/sdba/adjustment.py
```diff
@@ -318,6 +334,168 @@ def adjust(
...
+
+    @classmethod
+    def adjust(
+        cls,
+        ref: xr.Dataset,
+        hist: xr.Dataset,
+        sim: xr.Dataset | None = None,
+        **kwargs,
+    ) -> xr.Dataset:
+        r"""Return bias-adjusted data. Refer to the class documentation for the algorithm details.
```
coxipi, line 459: `sim` to be optional: I think it makes sense in a purely `adjust` Class to allow this and use directly `hist`  which is also in the arguments. And @LamAdr has the OTC function that performs an adjustment specifically on `hist`, not some other dataset (for another dataset, this is dOTC). 

In principle, I have the intution that OTC should in fact be generalized to accept a different `sim` than `hist`, say you train on a member-1, but you adjust member-2, in principles those two datasets have the same statistics, so it could be interesting. But it's not described in the literature I believe, it would required experiments.  For now, we can specifically ask for OTC that we use `sim == hist`

xclim/sdba/adjustment.py
```diff
@@ -121,13 +125,47 @@ def _check_inputs(cls, *inputs, group):
...
+                return inda
+
+            if target is None:
+                target = {
+                    v: inputs[0][dim].attrs["_units"][iv]
+                    for iv, v in enumerate(inputs[0][dim].values)
+                }
+            return (
+                _convert_units_to(inda, dim=dim, target=target) for inda in inputs
+            ), target
```
coxipi, line 157: I tried putting `_harmonize_units_multivariate` as another class function so that , but I was getting error about generators. This is because of the structure of `inputs`. I can re-try if we judge it would be better like this. 

Also, we now an output type which is not certain, target returned can be a str or dict[str]. I know it's not ideal, but it was very convenient to write like this.

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1245,253 @@ def _adjust(
...
+                    "Expected usage of the same group in adjust and training steps."
+                )
+            base_kws_vars[v].pop("group")
+
+            base_kws_vars[v].setdefault("nquantiles", self.ds.af_q.quantiles.values)
+            if np.isscalar(base_kws_vars[v]["nquantiles"]):
+                base_kws_vars[v]["nquantiles"] = equally_spaced_nodes(
+                    base_kws_vars[v]["nquantiles"]
+                )
+            # should maybe require that quantiles here match those from training?
```
coxipi, line 1471: `group` must be the same as in training, but otherwise different training/adjusting arguments can be given for the QDM and npdf.  To simplify things, I could require that we must use the same arguments. I would be a bit confusing because all `adj_kws`  would then be passed to the train function of `MBCn`

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1246,270 @@ def _adjust(
...
+                        )
+                    )
+                if "adapt_freq_thresh" in base_kws_vars[v]:
+                    base_kws_vars[v]["adapt_freq_thresh"] = str(
+                        convert_units_to(
+                            base_kws_vars[v]["adapt_freq_thresh"],
+                            self.train_units[v],
+                            context="hydro",
+                        )
+                    )
```
coxipi, line 1488: A consequence of using `multivar` is that the `thresh` are a bit more complicated to manage.
I could also unstack and restack variables inside the `mbcn_adjust`, I don't know which is preferable.

CHANGES.rst
```diff
@@ -19,8 +19,9 @@ Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Éric Dup
 
 New features and enhancements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-* New properties: Bivariate Spell Length (``xclim.sdba.properties.bivariate_spell_length``), Generalized Spell Lengths with an argument for `window`, and Specific Spell Lengths with `window` fixed to '1' (``xclim.sdba.properties.threshold_count``, ``xclim.sdba.properties.bivariate_threshold_count``). (:pull:`1758`).
-* New option `normalize` in ``sdba.measures.taylordiagram`` to obtain normalized Taylor Diagrams (divide standard deviations by standard deviation of the reference). (:pull:`1764`).
+* New properties: Bivariate Spell Length (``xclim.sdba.properties.bivariate_spell_length``), generalized spell lengths with an argument for `window`, and specific spell lengths with `window` fixed to 1  (``xclim.sdba.propertiies.threshold_count``, ``xclim.sdba.propertiies.bivariate_threshold_count``). (:pull:`1758`).
+* New option `normalize` in ``sdba.measures.taylordiagram`` to obtain normalized Taylor diagrams (divide standard deviations by standard deviation of the reference). (:pull:`1764`).
```
coxipi, line 23:
```suggestion
* New properties: Bivariate Spell Length (``xclim.sdba.properties.bivariate_spell_length``), generalized spell lengths with an argument for `window`, and specific spell lengths with `window` fixed to 1  (``xclim.sdba.propertiies.threshold_count``, ``xclim.sdba.propertiies.bivariate_threshold_count``). (:pull:`1758`).
* New option `normalize` in ``sdba.measures.taylordiagram`` to obtain normalized Taylor diagrams (divide standard deviations by standard deviation of the reference). (:pull:`1764`).
```

coxipi:
I have an implementation with `map_groups` in the branch `mbcn` which is much cleaner IMO. With 400 spatial points, 30 years data, a `group = sdba.Grouper("time.dayofyear", window=31)` was 'only' 2x slower with `map_groups`. Perhaps that `map_groups` could be modified to reproduce the speed of the gpies approach

tests/test_sdba/test_adjustment.py
```diff
@@ -554,6 +556,44 @@ def test_add_dims(self, use_dask, open_dataset):
...
+
+        MBCN = MBCn.train(
+            ref,
+            hist,
+            base_kws=dict(nquantiles=50, group=Grouper(group, window)),
+            adj_kws=dict(interp="linear"),
+        )
+        p = MBCN.adjust(sim=sim, ref=ref, hist=hist, period_dim=period_dim)
+        # 'does it run' test
+        p.load()
```
coxipi, line 594: I don't think there are simple tests that I could use assert. One thing I could do is fix the random seed, obtain some values, and hardcode them, so we will know in the future is some change modifies the way the function works...

juliettelavoie:
`train` creates a multiindex `win_dim`. You can't write it directly to file.

It has to  be encoded and decoded:

```

train ...
encoded = cfxr.encode_multi_index_as_compress(dtrain, "win_dim")
xs.save_to_zarr(encoded, ...)
```

```
dtrain= xr.open_zarr(...)
decoded = cfxr.decode_compress_to_multi_index(dtrain, "win_dim")
ADJ = sdba.adjustment.TrainAdjust.from_dataset(decoded)

...adjust

```

might be good to add that to the example. It might also be worth it to add the `stack_period` to the example.

coxipi:
Ah my bad,  I removed `.reset_index(dims_or_levels=[win_dim])`  so that incomplete or Gregorian years could also be used as input. Keeping this quantity is not crucial, it can be easily recomputed in adjust too, I will drop it. I forgot it was problematic with saving

xclim/sdba/adjustment.py
```diff
@@ -219,7 +258,6 @@ def adjust(self, sim: DataArray, *args, **kwargs):
             if "group" in self:
                 self._check_inputs(sim, *args, group=self.group)
 
-            sim = convert_units_to(sim, self.train_units)
```
coxipi, line 222: `_harmonize_units` above already converts units of `sim`

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        Dataset variables:
+            ref : training target  (stacked variables)
+            hist : training source (stacked variables)
+    rot_matrices : xr.DataArray
+        The rotation matrices as a 3D array ('iterations', <pts_dims[0]>, <pts_dims[1]>), with shape (n_iter, <N>, <N>).
+    pts_dims : str
```
aulemahal, line 224:
```suggestion
    pts_dims : sequence of str
```

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+
+    # npdf training core
+    af_q_l = []
+    escores_l = []
+
+    # loop over time blocks
+    for ib in range(gw_idxs[gr_dim].size):
+        # indices in a given time block
+        indices = gw_idxs[{gr_dim: ib}].astype(int).values
+        ind = indices[indices >= 0]
```
aulemahal, line 245: ¿ Por qué ?

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+    )
+    return out
+
+
+def _npdft_adjust(sim, af_q, rots, quantiles, method, extrap):
+    """Npdf transform adjusting.
+
+    Adjusting factors `af_q` obtained in the training step are applied on the simulated data `sim` at each iterated
+    rotation
+    """
```
aulemahal, line 290: Same comment about commenting a bit more on the inputs and numpiness.

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+        # loop over variables
+        for iv in range(sim.shape[0]):
+            af = u._interp_on_quantiles_1D_multi(
+                u._rank_bn(sim[iv], axis=-1),
+                quantiles,
+                af_q[ii, iv],
+                method=method,
+                extrap=extrap,
+            )
+            sim[iv] = u.apply_correction(sim[iv], af, "+")
```
aulemahal, line 309:
```suggestion
            sim[iv] = sim[iv] + af
```

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+    hist,
+    sim,
+    ds,
+    pts_dims,
+    interp,
+    extrapolation,
+    base,
+    base_kws_vars,
+    adj_kws,
+    period_dim,
```
aulemahal, line 330: Same comment about the annotations. The docstring also needs some numpydoc-ification I think. Section headers are missing.

xclim/sdba/adjustment.py
```diff
@@ -121,13 +125,48 @@ def _check_inputs(cls, *inputs, group):
...
+
+        dim_multivariate = None
+        for _dim, _crd in inputs[0].coords.items():
+            if _crd.attrs.get("is_variables"):
+                dim_multivariate = str(_dim)
+                break
+        if dim_multivariate is not None:
+            return _harmonize_units_multivariate(
+                *inputs, dim=dim_multivariate, target=target
+            )
```
aulemahal, line 168:
```suggestion
        for _dim, _crd in inputs[0].coords.items():
            if _crd.attrs.get("is_variables"):
                return _harmonize_units_multivariate(
                    *inputs, dim=str(_dim), target=target
                )
```
No ?

xclim/sdba/adjustment.py
```diff
@@ -121,13 +125,48 @@ def _check_inputs(cls, *inputs, group):
...
+                    inda[dim].attrs["_units"][iv] = target[v]
+                inda.attrs["units"], inda.attrs["standard_name"] = "", ""
+                return inda
+
+            if target is None:
+                target = {
+                    v: inputs[0][dim].attrs["_units"][iv]
+                    for iv, v in enumerate(inputs[0][dim].values)
+                }
+            return (
```
aulemahal, line 156: If it is not too messy, I would try to add a fast path if all inputs already have the same units. The univariate function relies on the fact that `convert_units_to` already does that, but here we are provoking operations even if the units are the same.

Also, does this work smoothly with dask ?

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1246,271 @@ def _adjust(
...
+        Small numbers result in less significant scores, but the execution time goes up quickly with large values.
+    n_iter : int
+        The number of iterations to perform. Defaults to 20.
+    pts_dim : str
+        The name of the "multivariate" dimension. Defaults to "multivar", which is the
+        normal case when using :py:func:`xclim.sdba.base.stack_variables`.
+    rot_matrices: xr.DataArray, optional
+        The rotation matrices as a 3D array ('iterations', <pts_dim>, <anything>), with shape (n_iter, <N>, <N>).
+        If left empty, random rotation matrices will be automatically generated.
+        The rotation matrices as a 3D array ('iterations', <pts_dims[0]>, <pts_dims[1]>), with shape (n_iter, <N>, <N>).
```
aulemahal, line 1284: Either the first or the third line feels duplicated.

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1246,271 @@ def _adjust(
...
+                        )
+                    )
+
+        adj_kws = adj_kws or {}
+        adj_kws.setdefault("interp", self.interp)
+        adj_kws.setdefault("extrapolation", self.extrapolation)
+
+        g_idxs, gw_idxs = grouped_time_indexes(ref.time, self.group)
+        self.ds["g_idxs"] = g_idxs
+        self.ds["gw_idxs"] = gw_idxs
```
aulemahal, line 1494: Why isn't this done in the `_train` ? The `_adjust` function should not modify the adjustment object's "state" I think.

aulemahal (approved):
First pass.

aulemahal:
Woups. I didn't mean to approve, only to comment.

xclim/sdba/adjustment.py
coxipi: it was until Juliette got a problem decoding `gw_idxs` because of its Multiindex nature...

xclim/sdba/_adjustment.py
```diff
@@ -43,6 +45,7 @@ def dqm_train(
     kind: str,
     quantiles: np.ndarray,
     adapt_freq_thresh: str | None = None,
+    jitter_under_thresh_value,
```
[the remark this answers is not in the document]
coxipi: Yep, I thought I corrected this earlier, that's weird

xclim/sdba/adjustment.py
aulemahal: Ah yes I see! Then I suggest adding these two to a copy of `self.ds` to avoid modifying the object.

xclim/sdba/adjustment.py
coxipi: yep, thanks

xclim/sdba/_adjustment.py
coxipi: ¿ que lo qué?

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
     return xr.Dataset(data_vars=dict(af=af, hist_q=hist_q))
 
 
+def _npdft_train(ref, hist, rots, quantiles, method, extrap, n_escore, standardize):
+    r"""Npdf transform to correct a source `hist` into target `ref`.
+
+    Perform a rotation, bias correct `hist` into `ref` with QuantileDeltaMapping, and rotate back.
+    Do this iteratively over all rotations `rots` and conserve adjustment factors `af_q` in each iteration.
+    """
+    if standardize:
```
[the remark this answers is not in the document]
coxipi: hmm we have a bunch of hidden function that are purely numpy, I thought it was sufficient. It's true that it looks a lot like other functions in `_adjustment.py` that are xarray.

xclim/sdba/_adjustment.py
coxipi: That's probably related to memory leaks. I guess it's not the most xarray-esque way of doing things, I have to improve this.

xclim/sdba/adjustment.py
coxipi: It's only attributing some dict values in the end, so there is not much gain with this fast pass, but hopefull the fast path is not too messy

xclim/sdba/_adjustment.py
aulemahal: I meant the `indices >= 0`, I don't think I understand.

xclim/sdba/_adjustment.py
```diff
@@ -126,32 +141,315 @@ def eqm_train(
...
+def mbcn_train(
+    ds,
+    rot_matrices,
+    pts_dims,
+    quantiles,
+    g_idxs,
+    gw_idxs,
+    interp,
+    extrapolation,
+    n_escore,
```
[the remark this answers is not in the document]
coxipi: typing, yes?

xclim/sdba/_adjustment.py
coxipi: it removes possible nans (which can come up with standard calendar because of leap years iirc, but I would not recommend using this with a non-regular calendar anyways)

aulemahal: So you are relying on that fact that NaN translates to a negative number with `astype` ? If so I suggest:
```python3
        indices = gw_idxs[{gr_dim: ib}].fillna(-1).astype(int).values
        ind = indices[indices >= 0]
```

docs/notebooks/sdba.ipynb
```diff
@@ -474,55 +474,16 @@
     "\n",
```
aulemahal, line 474: <p>This section's title should be changed as this is not "multi-step" anymore.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Ouranosinc/xclim/pull/1580/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

coxipi: <p>Ok, I thought that stacking variables could still count as an extra step, but granted</p><p> </p>

aulemahal (approved):
This is an actual approval.

Grosse et belle job @coxipi! Deux morceaux de robots : :palms_up_together:  :robot: :robot:.

xclim/sdba/_adjustment.py
```diff
@@ -212,7 +536,7 @@ def dqm_adjust(
         The kind of correction to compute. See :py:func:`xclim.sdba.utils.get_correction`.
     extrapolation : str
         The extrapolation method to use.
-    detrend : int | PolyDetrend
+    detrend : int or PolyDetrend
```
coxipi, line 539:
```suggestion
    detrend : int | PolyDetrend
```
unintended change

coxipi:
Nice! 

Merci pour la review qui a dû demander du jus aussi! J'essaie de faire plus de PRs moins grosses...

PR #1625, opened 2024-01-25
Option to output datasets

### LINKED ISSUE #1257: Return Datasets
### Generic Issue

It could be interesting to (optionally) return Datasets when computing indicators. This could be controlled by `set_options`. 

It would allow:
- Adding  `time_bounds`
- Having correct CF attributes : "Conventions" and "history" are supposed to be at the dataset-level.
- Preserve `grid_mapping` and other metadata/auxiliary vars when a `ds` is passed to the indicator.
- Multiple output indicators in a single dataset

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1257

* New option : `as_dataset`. When True, indicators return Datasets instead of DataArrays or tuples thereof.

No.

### DIFF
--- a/.github/workflows/label-on-approval.yml
+++ b/.github/workflows/label-on-approval.yml
@@ -74,7 +74,7 @@ jobs:
           (steps.fc.outputs.comment-id == '') &&
           (!contains(github.event.pull_request.labels.*.name, 'approved')) &&
           (github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name)
-        uses: peter-evans/create-or-update-comment@23ff15729ef2fc348714a3bb66d2f655ca9066f2 # v3.1.0
+        uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 # v4.0.0
         with:
           comment-id: ${{ steps.fc.outputs.comment-id }}
           issue-number: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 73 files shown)


Zeitsperre (commented):
Cool!

Is there a particular reason why we wouldn't want to offer this as a method/call signature of Indicator objects? e.g. `some_indicator(..., as_dataset=True)`?

aulemahal:
It's simply that I didn't want to add another argument to the indicators, as to now overcrowd the signature with parameters that do not impact the computation itself.

But, it could be done easily.

Zeitsperre:
That's a good point. The workflow example is clear enough, so I don't see any harm from keeping this as an option set via context.

The only other way I could see this being implemented (for conventional `xarray` users, anyway) would involve extending `xarray.DataArray.to_dataset()`, but that could be either very easy or an absolute nightmare. Is extending a `DataArray` class method an acceptable/desirable thing to do?

tlogan2000:
LGTM ... I think the context / set_options mode is a decent compromise

tlogan2000 (approved):

Zeitsperre:
New black came out. I'll see what can be done.

aulemahal:
Oh! I thought this was a python 3.8 issue...

Zeitsperre:
FYI: https://github.com/psf/black/releases/tag/24.1.0

Looks like a bunch of breaking changes. I think I'm going to pin `black` to the pre-commit latest version so we don't run into this again. Will happen in a new PR.

PR #1647, opened 2024-02-14
Better degree_days_exceedance_date and ensemble stats

### LINKED ISSUE #1459: `degree_days_exceedance_date` nans cause issue on climatedata.ca
### Generic Issue

If the input to `degree_days_exceedance_date` never reaches the `sum_thresh`, the output is `nan`.
This makes sense when a user is using this function on a single input.
However, when you do the calculation on different models and then calculate the ensemble stats, it can lead to issues.
If half the models never reach the threshold they will be ignored in the ensemble stats and the results will be biased too early.
For people using xclim directly, it is not so bad as they can just add a manual check (though it would be nice to have a warning), but it is worst when the function is used on the Analyze page of climatedata.ca.
People using the website only get the percentiles output and have no idea how many models were actually included in the ensemble. The intermediate data is invisible to them.

I'm not sure how to fix this. 
In my project, I manually set the nans to be at a specific day that make sense in my context. ( I am looking for date when ice roads season begins. I set the nans to the date of the end of the ice roads season, march 31st.)
Maybe this could be a new argument, `missing` or `never_reached` ?
Maybe just a warning that the threshold is not always reached is sufficient?
Maybe ensemble stats  can have a warning when there are nans ?
 
 Though, I don't really know how to transfer those warning to climatedata users.

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1459

Implements two solutions for #1459.

* `min_members` argument to ensemble stats functions. Simply mask elements when there are fewer valid (non-nan) members than this threshold. Similar to `min_periods` of xarray's `rolling`, but with a different default value.
* `never_reached` argument to `degree_days_exceedance_date`. A value (int, doy) to assign when the `sum_thresh` is not reached at the end of the period, to differentiate from missing values. 

No. Both default options preserve the previous behaviour (which means `min_members` doesn't have the same default as it's inspiration `min_periods`).

### Other information:
Both solutions are distinct. For `climatedata.ca`, I think only `min_members` would be useful ?

Another option for `never_reached` would be to assign the last doy of the period, would that be interesting ? @juliettelavoie 
To keep the annotation simple, we could do that if -1 is passed ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,8 @@ New features and enhancements
 * New ``xclim.core.calendar.stack_periods`` and ``unstack_periods`` for performing ``rolling(time=...).construct(..., stride=...)`` but with non-uniform temporal periods like years or months. They replace ``xclim.sdba.processing.construct_moving_yearly_window`` and ``unpack_moving_yearly_window`` which are deprecated and will be removed in a future release.
 * New ``as_dataset`` options for ``xclim.set_options``. When True, indicators will output Datasets instead of DataArrays. (:issue:`1257`, :pull:`1625`).
 * Added new option for UTCI calculation to cap low wind velocities to a minimum of 0.5 m/s following Bröde (2012) guidelines. (:issue:`1634`, :pull:`1635`).
+* Added option ``never_reached`` to ``degree_days_exceedance_date`` to assign a custom value when the sum threshold is never reached.
+* Added option ``min_members`` to ensemble statistics to mask elements when the number of valid members is under a threshold.
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,8 +22,8 @@ New features and enhancements
 * New ``xclim.core.calendar.stack_periods`` and ``unstack_periods`` for performing ``rolling(time=...).construct(..., stride=...)`` but with non-uniform temporal periods like years or months. They replace ``xclim.sdba.processing.construct_moving_yearly_window`` and ``unpack_moving_yearly_window`` which are deprecated and will be removed in a future release.
 * New ``as_dataset`` options for ``xclim.set_options``. When True, indicators will output Datasets instead of DataArrays. (:issue:`1257`, :pull:`1625`).
 * Added new option for UTCI calculation to cap low wind velocities to a minimum of 0.5 m/s following Bröde (2012) guidelines. (:issue:`1634`, :pull:`1635`).
-* Added option ``never_reached`` to ``degree_days_exceedance_date`` to assign a custom value when the sum threshold is never reached.
-* Added option ``min_members`` to ensemble statistics to mask elements when the number of valid members is under a threshold.
+* Added option ``never_reached`` to ``degree_days_exceedance_date`` to assign a custom value when the sum threshold is never reached. (:issue:`1459`, :pull:`1647`).
+* Added option ``min_members`` to ensemble statistics to mask elements when the number of valid members is under a threshold. (:issue:`1459`, :pull:`1647`).
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


Zeitsperre:
> Another option for `never_reached` would be to assign the last doy of the period, would that be interesting ? @juliettelavoie To keep the annotation simple, we could do that if -1 is passed ?

Would it be better to mark the value of `never_reached` as -1? I feel like the alternative would mean that users would need to mask different values depending on their inputs. Having both a NaN value for statistically untrustworthy and a negative value for `never reached` seems less confusing to me.

juliettelavoie:
I am not sure I understand what you mean by "last doy of the period" ? Is it the last day that it was reached on other years ? the day before `after_date` ?

aulemahal:
Oh I meant the last day of the period. Like with freq='YS', that would be 365 or 366 according to the calendar. So when the sum is never reached, the indicator would return the largest "doy" possible. I'm not sure if this makes sense at all though.

Zeitsperre:
As a GIS person, I'm not a fan of a having a dynamic value acting as a `NaN`. The fact that (if I understand correctly) depending on the frequency your `never_reached` value could be 360, 365, 366, 89, 90, 91, 92, 29, 30, 31... Feels nightmarish to me.

juliettelavoie:
I see. I guess it could make sense in some cases,  but may not make a lot a sense depending on the `after_date` or the context. My case study was freezing road with AS-JUL after oct 1st and most dates being in february/march, including june dates in an ensemble mean would not really make sense. 

Also, it is not that hard for the user to put "12-31" (in the case of YS). I guess the advantage of "last" is that is works even if "12-31" doesn't exist like for a 360_day calendar ?

juliettelavoie:
For context, the goal of having a "never_reached" value was not only to be able to filter it like a different NaN, but to include it in the ensemble mean in a way that made sense for the context (in climatedata.ca, only the ensemble mean is given). In my case study, where I was looking at the beginning of the ice road season, the `never_reached` was the date of the end of the ice roads season.

Zeitsperre:
> For context, the goal of having a "never_reached" value was not only to be able to filter it like a different NaN, but to include it in the ensemble mean in a way that made sense for the context (in climatedata.ca, only the ensemble mean is given). In my case study, where I was looking at the beginning of the ice road season, the `never_reached` was the date of the end of the ice roads season.

Ah, that helps clarifies things. It also leaves me feeling like this problem is even more complicated. What values could we accept in `never_reached` to cover our bases?
- `last` (end of period)
- `-1` (for easier filtering)
- `MM-DD` (for an arbitrary date; What happens if this date is beyond the date of a given period?)

If we offered these options, would we want to provide a default value? I feel like we wouldn't.

aulemahal:
ah! Well in that case, indeed making it possible to pass a "DayOfYearStr" makes sense. I'll abandon the idea of "last". 

So the options are now:

- None  (default) : assigns NaN (previous behaviour)
- int : Assigns the int.
- string "MM-DD" : finds the corresponding doy and assigns that.

(in reality, any non-none and non-string input is directly assigned, so floats are also possible)

aulemahal:
> I guess the advantage of "last" is that is works even if "12-31" doesn't exist like for a 360_day calendar ?

Yes, but also covers cases where `freq` is not annual.

Zeitsperre:
Side note: We're waiting on comments from the PCIC and Climatedata.ca collaborators. This PR should not be merged this before Friday (February 16th) to give them a chance to weigh in.

aulemahal:
We already return day-of-year as float because of that. We have the  `check_missing` mechanism that will potentially insert NaN afterwards, so we can't really promise an int dtype.

I put `int` as the annotation because it is a doy. But as always type checking is pretty loose.

cpomer10:
No additional comments from ECCC - thanks!

tlogan2000 (approved):

PR #1649, opened 2024-02-15
Sum is a sum

### LINKED ISSUE #1645: Which units should a "sum" aggregation return ?
### Generic Issue

Generic indices such as `select_resample_op` manage the units automatically using `to_agg_units`. This function decides how to transform the units according to the type of aggregation.

When it was first written, the only case for "sum" was degree-days, i.e. _integrating_ temperature over time. As such, the units are multiplied by the timestep. In most cases, this is a day and so it simply adds "d" to the existing units. Ex:
```
tas : daily temperature in K
out = to_agg_units(tas.sum(), tas, op='sum')
out.units == "K d"
```

However, an example where this is wrong has been raised in xhydro. https://xhydro.readthedocs.io/en/latest/notebooks/local_frequency_analysis.html#d)-Computing-volumes, see cell output 12.

The input data is the daily _volume_ of water flowing in a river. Thus, here we want the actual sum, not an integral.

Question:
Should we go back to considering `op='sum'` as an actual sum (and leave the units untouched) ? We already have `'integral'` as an available choice for `op`, it just isn't a common method name.

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1645

* `to_agg_units` will not change the units when the aggregation is `"sum"`.
* Fixed the units when the initial data has a temporal dimensionality (ex "m/s") and an `"integral"` is performed. The output units were previously not reduced, yielding things like "d m s-1". 

Yes, but `select_resample_op` is rarely used by the user directly. And we could consider that this is in fact a _fix_ from a previous ambiguous situation. No internal calls to `to_agg_units` were using `"sum"`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,7 +39,7 @@ Breaking changes
 * `black` formatting style has been updated to the 2024 stable conventions. `isort` has been added to the `dev` installation recipe. (:pull:`1626`).
 * The indice and indicator for ``winter_storm`` has been removed (deprecated since `xclim` v0.46.0 in favour of ``snd_storm_days``). (:pull:`1565`).
 * `xclim` has dropped support for `scipy` version below v1.9.0 and `numpy` versions below v1.20.0. (:pull:`1565`).
-* For generic function ``select_resample_op`` and ``core.units.to_agg_units``, operation "sum" will now return the same units as the input, and not implicitely translate to an "integral". (:issue:`1645`).
+* For generic function ``select_resample_op`` and ``core.units.to_agg_units``, operation "sum" will now return the same units as the input, and not implicitely translate to an "integral". (:issue:`1645`, :pull:`1649`).
 
 Bug fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,7 +39,7 @@ Breaking changes
 * `black` formatting style has been updated to the 2024 stable conventions. `isort` has been added to the `dev` installation recipe. (:pull:`1626`).
 * The indice and indicator for ``winter_storm`` has been removed (deprecated since `xclim` v0.46.0 in favour of ``snd_storm_days``). (:pull:`1565`).
 * `xclim` has dropped support for `scipy` version below v1.9.0 and `numpy` versions below v1.20.0. (:pull:`1565`).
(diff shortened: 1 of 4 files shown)


xclim/core/units.py
```diff
@@ -471,8 +471,8 @@ def to_agg_units(
         The original array before the aggregation operation,
         used to infer the sampling units and get the variable units.
     op : {'min', 'max', 'mean', 'std', 'var', 'doymin', 'doymax',  'count', 'integral', 'sum'}
-        The type of aggregation operation performed. The special "delta_*" ops are used
-        with temperature units needing conversion to their "delta" counterparts (e.g. degree days)
+        The type of aggregation operation performed. "integral" is mathematically equivalent to "sum",
+        but the units are multipled by the timestep of the data (requires an inferrable frequency).
```
Zeitsperre, line 475:
```suggestion
        but the units are multiplied by the timestep of the data (requires an inferrable frequency).
```

CHANGES.rst
```diff
@@ -39,6 +39,7 @@ Breaking changes
 * `black` formatting style has been updated to the 2024 stable conventions. `isort` has been added to the `dev` installation recipe. (:pull:`1626`).
 * The indice and indicator for ``winter_storm`` has been removed (deprecated since `xclim` v0.46.0 in favour of ``snd_storm_days``). (:pull:`1565`).
 * `xclim` has dropped support for `scipy` version below v1.9.0 and `numpy` versions below v1.20.0. (:pull:`1565`).
+* For generic function ``select_resample_op`` and ``core.units.to_agg_units``, operation "sum" will now return the same units as the input, and not implicitely translate to an "integral". (:issue:`1645`, :pull:`1649`).
```
Zeitsperre, line 42:
```suggestion
* For generic function ``select_resample_op`` and ``core.units.to_agg_units``, operation "sum" will now return the same units as the input, and not implicitly be translated to an "integral". (:issue:`1645`, :pull:`1649`).
```

aulemahal:
@Zeitsperre  All tests fixed. This is ready to go. It will solve the remaining bug in the rolling statistics PR.

Zeitsperre (approved):
Looks great. I'll be merging this in after #1650 !

PR #1653, opened 2024-02-19
Spei, proper use of `loc` parameter in scipy dists

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Spei used to rely on an offset to ensure that distribution of values are positive when using distributions such as `gamma, fisk`. Now, instead, we properly use the `loc` parameter in those distributions that generalize the distributions and allow negative values. For instance, the `gamma` disitribution is a 2-parameter function, but in ``scipy`` it's generalized to 3 parameters, with a `loc` parameter. This extends the support of `gamma` from $x \in [0,\infty]$ to $x \in [\text{loc}, \infty]$. 

The code was already compatible with this idea, just had to change ``_fit_start`` so it can compute estimates of the parameters in the cases with negatives values.

Yes, removing the offset from SPEI.

### DIFF
--- a/tests/test_indices.py
+++ b/tests/test_indices.py
@@ -695,6 +695,8 @@ class TestAgroclimaticIndices:
             tas = tasmax - 2.5
             tasmin = tasmax - 5
             wb = xci.water_budget(pr, None, tasmin, tasmax, tas)
+            # offset like in climate indices library
+            wb = wb + convert_units_to("1 mm/d", wb)
 
         params = xci.stats.standardized_index_fit_params(
             wb.sel(time=slice("1950", "1980")),
@@ -702,7 +704,6 @@ class TestAgroclimaticIndices:
             window=window,
             dist=dist,
             method=method,
-            offset="1 mm/d",
         )
         spei = xci.standardized_precipitation_evapotranspiration_index(
             wb.sel(time=slice("1998", "2000")), params=params
--- a/tests/test_indices.py
+++ b/tests/test_indices.py
@@ -696,7 +696,7 @@ class TestAgroclimaticIndices:
             tasmin = tasmax - 5
             wb = xci.water_budget(pr, None, tasmin, tasmax, tas)
             # offset like in climate indices library
-            wb = wb + convert_units_to("1 mm/d", wb)
+            wb = wb + convert_units_to("1 mm/d", wb, context="hydro")
 
         params = xci.stats.standardized_index_fit_params(
             wb.sel(time=slice("1950", "1980")),
--- a/xclim/indices/_agro.py
+++ b/xclim/indices/_agro.py
@@ -1,8 +1,6 @@
 # noqa: D100
 from __future__ import annotations
 
-import warnings
-
 import numpy as np
 import xarray
 
@@ -24,11 +22,7 @@ from xclim.indices._threshold import (
 )
 from xclim.indices.generic import aggregate_between_dates, get_zones
 from xclim.indices.helpers import _gather_lat, day_lengths
-from xclim.indices.stats import (
-    preprocess_standardized_index,
-    standardized_index,
-    standardized_index_fit_params,
-)
+from xclim.indices.stats import standardized_index
 
 # Frequencies : YS: year start, QS-DEC: seasons starting in december, MS: month start
 # See https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html
@@ -1094,12 +1088,10 @@ def rain_season(
 
 @declare_units(
     pr="[precipitation]",
-    pr_cal="[precipitation]",
     params="[]",
 )
 def standardized_precipitation_index(
     pr: xarray.DataArray,
-    pr_cal: Quantified | None = None,
     freq: str | None = "MS",
     window: int = 1,
     dist: str = "gamma",
@@ -1115,9 +1107,6 @@ def standardized_precipitation_index(
     ----------
     pr : xarray.DataArray
         Daily precipitation.
-    pr_cal : xarray.DataArray, optional
-        Daily precipitation used for calibration. Usually this is a temporal subset of `pr` over some reference period.
-        This option will be removed in xclim >=0.47.0. Two behaviour will be possible (see below)
     freq : str, optional
         Resampling frequency. A monthly or daily frequency is expected. Option `None` assumes that desired resampling
         has already been applied input dataset and will skip the resampling step.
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -44,6 +44,8 @@ Breaking changes
...
 * The indice and indicator for ``winter_storm`` has been removed (deprecated since `xclim` v0.46.0 in favour of ``snd_storm_days``). (:pull:`1565`).
 * `xclim` has dropped support for `scipy` version below v1.9.0 and `numpy` versions below v1.20.0. (:pull:`1565`).
+* For generic function ``select_resample_op`` and ``core.units.to_agg_units``, operation "sum" will now return the same units as the input, and not implicitly be translated to an "integral". (:issue:`1645`, :pull:`1649`).
+* `lmoments3` was removed as a dependency of `xclim` due to incompatible licensing (GPLv3 vs `xclim`'s Apache 2.0). Depending on the outcome of efforts to modify the licensing of `lmoments3`, this change may eventually be reverted. See `Ouranosinc/lmoments3#12 <https://github.com/Ouranosinc/lmoments3/issues/12>`_. See also the "frequency analysis" notebook for an example on how to continue using the probability weighted moments method for fitting distributions. (:issue:`1620`, :pull:`1644`).
 * Argument `offset` is removed  from ``xclim.indices.standardized_precipitation_evapotranspiration_index``. This will slightly change results when using SPEI with default values (:issue:`1477`  :pull:`1653`).
```
Zeitsperre, line 49: You're going to want to bring this entry up to v0.48.2!

coxipi: just pull origin master no?

coxipi:
I'm pretty much done here I think, but I would like to tests a bit more situations where we have negative values. I need to find realistic cases, not too extreme I suppose, although I don't see why the method would not work with very negative values. I have to see!

coxipi:
So I tried to obtain a proper estimation of `loc` parameter in the case of the gamma distribution, resorting to approximations as it's already the case when we have the two-parameter distribution. 

It becomes a mess. Even with a significantly worse approximation, the expression is quite messy. It can be done, but maybe there is a simple way I'm not seeing. 

`loc` is estimated in some of the distributions in `_fit_start` (e.g. `GEV`) so I will see if there's something obvious I'm not seeing. Otherwise, `_fit_start` will probably still work well to do what its name implies, start a fit, but otherwise should not be used as an approximate method (unless the intent IS to use a two parameter distribution).  So, either improve `_fit_start` or limit its application as a method for SPI/SPEI.

coxipi:
I will implement a very rudimentary estimation of the `loc` parameter using the quantiles:
Conisder an ordered time series  $X_1, X_2, \dots, X_n$. The rawest approximation would be: $\text{loc}^* = X_1$ which was the spirit of having an offset. By the definition of `loc`, we know we must have $\text{loc} \leq X_1$. A more refined approximation is: $\text{loc}^* = X_1 + (X_1-X_2)$, which is smaller. This might undershoot, i.e. $\text{loc}^* < \text{loc}^{\rm Truth}$. And so on.

If we want to use a specific `loc` parameter, this could be used through `fitkwargs`. For instance, in some cases, we might want to impose `loc==0`. In any cases, we should keep the zero inflated distributions for SPI/precips. I have a good idea of what would be good for many examples, I just need to figure out a proper way to organize things around these ideas.

coxipi:
I had to change some results in test_indices. The most significant changes are for the fisk distribution since I changed significantly the approximations. We now have a more rigorous implementation of the method of moments (I used <x> and <x^2>/<x>^2.

The choice of starting parameters can significantly affect even the full-fledged numerical approach (e.g. "ML" method)

coxipi:
I've been exploring the sensitivity to initial guesses in our procedure. 

Consider what is done in `test_indices.py` (`test_standardized_precipitation_index`, etc), computing a few values of SPI/SPEI. I show the example with the `fisk` distribution, but I also saw this phenomena for `gamma` dist

```method
freq = "D"
window = 1
dist = "fisk"
method = "ML"
ds = open_dataset("sdba/CanESM2_1950-2100.nc", cache=True, cache_dir=Path("/home/eridup1/tank/xclim_open_dataset/")).isel(location=1)
pr = ds.pr.sel(time=slice("1998", "2000"))
pr_cal = ds.pr.sel(time=slice("1950", "1980"))
params = xci.stats.standardized_index_fit_params(
    pr_cal, freq=freq, window=window, dist=dist, method=method, zero_inflated=True,
)
spi  = xci.standardized_precipitation_index(pr, params=params)
print(spi.isel(time=slice(-11, -1, 2)))
```

I used two ways to estimate `loc` in `_fit_start`, the `spi` results sometimes largely, sometimes barely
```python
# loc=0
>> [-0.12736419  1.30823431  0.67651811  0.10446318 -0.50246943]
# loc = (x1 * xn - x2**2) / (x1 + xn - 2 * x2)
# x1 is the smallest precip in the dataset, x2 the second smallest, xn the largest
>> [-0.15894928  1.30822495  0.44984607  0.14669885 -0.50273746]
```
It makes me think in some cases, the same minimum is found, in other cases no.

The difficulty of optimization for 3-parameter `gamma` distribution is discussed in this issue, confirms my suspicion.
https://github.com/scipy/scipy/issues/18775

CHANGES.rst
```diff
@@ -4,12 +4,20 @@ Changelog
...
-Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`).
+Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Éric Dupuis (:user:`coxipi`).
 
 Announcements
 ^^^^^^^^^^^^^
 * `xclim` has migrated its development branch name from `master` to `main`. (:issue:`1667`, :pull:`1669`).
 
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* Distribution with negative values are directly fitted without need for an offset for distributions such as `gamma, fisk` in ``xclim.indices.standardized_precipitation_evapotranspiration_index``. (:issue:`1477`  :pull:`1653`).
```
VascoSch92, line 15:
```suggestion
* Distributions with negative values are directly fitted without need for an offset for distributions such as `gamma, fisk` in ``xclim.indices.standardized_precipitation_evapotranspiration_index``. (:issue:`1477`  :pull:`1653`).
```

coxipi:
I did the computation for 20 values of loc0 between 0 and the guess I outlined above. There is some variability in the spi output, and that's true for any day of year:

### doy=361 (the 0.104 vs. 0.147 above):

image

### another random doy I took, 180

image

### The worst doy mentioned above, the one with 0.68 vs. 0.45

image


### BONUS: using `scipy.stats.fit` instead of `scipy.stats.fisk.fit`

All in all, it does not fare much better. Also, in this case, you have to input reasonable range of values to search within, so maybe this part could be optimized a bit more to avoid large span of values like this.

image


All in all, I think things are controlled well enough. As a last study, I will consider how the R package performs in these circumstances.

coxipi:
A bit of a tangent, but I was wondering how does this compare to the error in  a single fit. I generate 30 points in a gamma dist with R,  computed a fit + std errors, then computed a SPI for a hundred set of parameters within the error bars, this shows the inherent variability with this measure.

image

coxipi:
I've been battling with R trying to obtain a confirmation through another package that indeed, MLE is difficult for 3-parameters function. I've had good success working with 2-params distributions that come with `fitdistr` and `fitdistrplus`, but as I try to use 3-parameters in additional packages (e.g. `FAdist` or `qualityTools`), it becomes very difficult. I keep getting error message concerning `initial value in 'vmmin' is not finite`, which as I understand should be solved by giving different starting parameters. 

```R
library(FAdist)
library(fitdistrplus)
data <- rgamma3(30, shape = 0.5, scale = 1e-4, thres=0)  # Using Gamma distribution for sample data
params = fitdistr(data,dgamma3, start=list(shape=0.5,scale=1e-4, thres=0))
```

Maybe there is something simple I'm missing, I will simply wrap up the PR now

PR #1674, opened 2024-03-07
Fix map_blocks for DataArray with cftime and sdba_encode_cf

### LINKED ISSUE #1673: sdba map_blocks fails with non-dask inputs when sdba_encode_cf option is activated
### Generic Issue

* xclim version: 0.48.2
* Python version: all
* Operating System: all

### Description
See Ouranosinc/xscen#362

We cycle of the coords dict and modify it during the loop, which fails. 
https://github.com/Ouranosinc/xclim/blob/4546f28bddeac1d389de7a9a4d6820575aecd5c6/xclim/sdba/base.py#L676-L678

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1673 

### What kind of change does this PR introduced
**tl;dr  : me fix issue.**

I think this issue stems from a change in recent xarray, but the case was untested by xclim. The solution was to iterate over a list that is "disconnected" from the dictionary.

The issue occurs when a _DataArray_ that has a coordinate made of `cftime` objects is passed to a `map_blocks`-wrapped function and xclim's global option `sdba_encode_cf` is True.

The previous encoding code was iterating over the coordinates with `coords.items()`. When it would modify a the cftime coord, this would fail as you can't modify a dictionary while it is being iterated upon. 

Only happens if the option to encode is activated, only happens if there is a cftime coord and only happens on `DataArray` objects. I'm not 100% sure why, but I think `Dataset.coords.items` does not map directly to a dict.  We usually use `Dataset` objects in function wrapped by `map_blocks` (or `map_groups`). The culprit here was the polynomial's trend computation which uses a `DataArray`.

### Reason for the apparent dask-dependency
The interesting part is that this bug does not depend on the fact that the data is using dask or not.  But it led me to discover another "issue" or, rather, "edgecase" .

In `xscen.adjust` we wrap the xclim call with options:
```python3
        with xc.set_options(sdba_encode_cf=True, sdba_extra_output=False):
            out = ADJ.adjust(sim_sel, **xclim_adjust_args)
```
Here, for DQM, the `adjust` method is wrapped by `map_blocks`. When `sim_sel` uses dask, the actual adjustment is, of course, not executed here. Only when the computation is triggered will the adjustment function be run, and only then will it call the polynomial detrending. But at that moment, the `sdba_encode_cf` option is not set anymore! So the issue was not raised for the "dask" case.

When `sim_sel` was not using dask, all the code was executed under the option statement and thus the issue was raised. Which is a bit dumb because the encoding is a micro-optimization made for dask, it's no use otherwise.

No

### DIFF
--- a/xclim/sdba/detrending.py
+++ b/xclim/sdba/detrending.py
@@ -174,6 +174,7 @@ class PolyDetrend(BaseDetrend):
 @map_groups(trend=[Grouper.DIM])
 def _polydetrend_get_trend(da, *, dim, degree, preserve_mean, kind):
     """Polydetrend, atomic func on 1 group."""
+    print("polydetrend")
     if len(dim) > 1:
         da = da.mean(dim[1:])
     dim = dim[0]
--- a/xclim/sdba/detrending.py
+++ b/xclim/sdba/detrending.py
@@ -174,7 +174,6 @@ class PolyDetrend(BaseDetrend):
 @map_groups(trend=[Grouper.DIM])
 def _polydetrend_get_trend(da, *, dim, degree, preserve_mean, kind):
     """Polydetrend, atomic func on 1 group."""
-    print("polydetrend")
     if len(dim) > 1:
         da = da.mean(dim[1:])
     dim = dim[0]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,9 +10,14 @@ Announcements
 ^^^^^^^^^^^^^
 * `xclim` has migrated its development branch name from `master` to `main`. (:issue:`1667`, :pull:`1669`).
 
+Bug fixes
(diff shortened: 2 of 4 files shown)


xclim/sdba/detrending.py
```diff
@@ -174,6 +174,7 @@ def _get_trend(self, da):
 @map_groups(trend=[Grouper.DIM])
 def _polydetrend_get_trend(da, *, dim, degree, preserve_mean, kind):
     """Polydetrend, atomic func on 1 group."""
+    print("polydetrend")
```
huard, line 177:
```suggestion

```

huard (approved):
LGTM. Edgy edge case.

PR #1709, opened 2024-04-16
Reconfigure Builds and perform more doctests

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Splits the `notebooks` recipe in the `tox.ini` file
* Sets the GitHub Workflows to test the doctests across more builds
* Stages the `tox-gh` plugin for eventual integration

Yes, `tox-gh` is now a development dependency and the `"notebooks_doctests"` `tox` recipe has been replaced with `"notebooks"`, while `doctest` has been made into a modifier for typical `tox` runs to also run the documentation examples.

### Other information:

https://github.com/tox-dev/tox-gh

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,10 +10,6 @@ Announcements
 ^^^^^^^^^^^^^
 * `xclim` has migrated its development branch name from `master` to `main`. (:issue:`1667`, :pull:`1669`).
 
-Breaking changes
-^^^^^^^^^^^^^^^^
-* Added the `tox-gh` dependency to the development installation recipe. This will soon be required for running the `tox` test ensemble on GitHub Workflows. (:pull:`1709`).
-
 Bug fixes
 ^^^^^^^^^
 * Fixed an bug in sdba's ``map_groups`` that prevented passing DataArrays with cftime coordinates if the ``sdba_encode_cf`` option was True. (:issue:`1673`, :pull:`1674`).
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -10,6 +10,10 @@ Announcements
 ^^^^^^^^^^^^^
 * `xclim` has migrated its development branch name from `master` to `main`. (:issue:`1667`, :pull:`1669`).
 
+Breaking changes
+^^^^^^^^^^^^^^^^
+* Added the `tox-gh` dependency to the development installation recipe. This will soon be required for running the `tox` test ensemble on GitHub Workflows. (:pull:`1709`).
```
aulemahal, line 15: Honestly, I don't think this counts as a "breaking change".  More of an "internal change" no ?
I understand "breaking" as targeting a normal usage, not a development setup.

tox.ini
```diff
@@ -65,16 +73,15 @@ allowlist_externals =
...
 ;extras =
 
-[testenv:notebooks_doctests{-coverage,}]
-description = Run notebooks and doctests with pytest under {basepython}
+[testenv:notebooks{-prefetch,}]
+description = Run notebooks with pytest under {basepython}
 commands =
     pytest --no-cov --nbval --dist=loadscope --rootdir=tests/ --ignore=docs/notebooks/example.ipynb docs/notebooks
-    pytest --rootdir=tests/ --xdoctest xclim
+commands_post =
```
aulemahal, line 80: New empty directive ? Does this do something ?

aulemahal (approved):
Looks good!

CHANGES.rst
Zeitsperre: Good point. We need to have a defining document on what exactly is a "breaking change".

tox.ini
Zeitsperre: This is a feature in tox: Since this recipe is based on testenv, we can overwrite an existing directive to have nothing by simply creating an empty directive.

Why do that here? Because we can't run `pytest --nbval` with coverage (causes errors) so we don't want to bother with uploading coverage in case a user accidentally tries to. It already won't let them (`notebooks` doesn't accept the `coverage` modifier) but this is extra security.

PR #1713, opened 2024-04-17
Remove columns that are all NaNs in make_criteria

RondeauG:
### Pull Request Checklist:
    - This PR fixes #xyz

* Methods such as kmeans clustering will crash if there are NaNs in the data. This is something that should happen if only a subset of realizations are NaNs, but we should be safe to remove the columns that are NaN across all realizations.

* No

### DIFF
--- a/xclim/ensembles/_reduce.py
+++ b/xclim/ensembles/_reduce.py
@@ -57,6 +57,8 @@ def make_criteria(ds: xarray.Dataset | xarray.DataArray):
 
     `ds2` will have all variables with the same dimensions, so if the original dataset had variables with different
     dimensions, the added dimensions are filled with NaNs.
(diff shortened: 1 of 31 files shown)


xclim/ensembles/_reduce.py
```diff
@@ -113,6 +113,9 @@ def _make_crit(da):
     else:
         # Easy peasy, skip all the convoluted stuff
         crit = _make_crit(ds)
+
+    # drop criteria that are all NaN
+    crit = crit.dropna(dim="criteria", how="all")
```
aulemahal, line 118: Can you mention this change in the docstring ? For example, in the notes we show how to retrieve the original array, maybe you could add a mention that this will not be possible if there are all-nan criteria ?

aulemahal (approved):
Approved, with minor suggestion.

PR #1720, opened 2024-04-18
Proper use of `loc` parameter in `gamma`,`fisk` dists (SPI/SPEI)

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Spei used to rely on an offset to ensure that distribution of values are positive when using distributions such as `gamma, fisk`. Now, instead, we properly use the `loc` parameter in those distributions that generalize the distributions and allow negative values. For instance, the `gamma` disitribution is a 2-parameter function, but in ``scipy`` it's generalized to 3 parameters, with a `loc` parameter. This extends the support of `gamma` from $x \in [0,\infty]$ to $x \in [\text{loc}, \infty]$. 

The code was already compatible with this idea, just had to change ``_fit_start`` so it can compute estimates of the parameters in the cases with negatives values.

* Removing the offset from SPEI. Also, the `_fit_start` can now compute an estimate of `loc` which can affect what value we obtain when optimizing 3-parameters distributions. Even when the `loc` is fixed to 0 with `floc=0` in `fitkwargs`, the estimate of `fisk` was modified. The previous formulation was obtained by applying more or less the idea in the method of moments, but now I obtained an approximation that's more rigorous.

* The case of zero inflated distribution has been changed: Now, it is the calibration data that determines the probability of zero, as it should.

* Method `APP` can now only be used if `fitkwargs['floc']` is given as input by the user.


coxipi:
I started a new branch `spi_loc2` because there were many errors hard to track in `spi_loc`. Opening a new PR, this is the old one: #1653

xclim/indices/_agro.py
```diff
@@ -1094,16 +1090,15 @@ def _get_rain_season(pram):
...
     params="[]",
 )
 def standardized_precipitation_index(
     pr: xarray.DataArray,
-    pr_cal: Quantified | None = None,
     freq: str | None = "MS",
     window: int = 1,
     dist: str = "gamma",
     method: str = "APP",
+    fitkwargs: dict = {},
```
VascoSch92, line 1101: using an empty dictionary as a default is dangerous

https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

Zeitsperre: Good call! I tend to use a pattern that resembles the following:
```
my_func(arg: str, kwarg: dict | None = None):
    if kwarg3 is None:
        kwarg3 = dict{}
```

coxipi:
I was just trying to tie ends together, and was exploring the difference between fitting with zero inflated distributions vs. normal distributions. 

I'm considering the case where the dataset is inflated in some small value that I inject, but I don't do any special treatment for this value. I realize that things work well unless the inflated value is 0:
```python
pr0 = open_dataset("sdba/CanESM2_1950-2100.nc")
# Take the first 30 july 1rst 
pr0  = pr0.isel(location=1).isel(time=slice(179,365*30,365)).pr.values
pr0 = np.where(pr0>0, pr0, 1e-6)
min_prs = [-1e-10,-1e-20, 0, 1e-20, 1e-18, 1e-14, 1e-10]
print(["min_pr", "Pr(X>=min_pr)"])
for min_pr in min_prs:
    pr = pr0.copy()
    pr[0:11] = min_pr 
    a, loc,scale = sc.stats.gamma.fit(pr)
    print([min_pr, sc.stats.gamma.cdf(min_pr,a,loc=loc,scale=scale)])
```
yields:

```
['min_pr', 'Pr(X>=min_pr)']
[-1e-10, 0.07865019433968551]
[-1e-20, 0.07258213369500696]
[0, 5.552055328806407e-20]
[1e-20, 0.07787936999249587]
[1e-18, 0.07488806373878135]
[1e-14, 0.08371030972954428]
[1e-10, 0.08298754087251842]
```
and we can clearly see something is off with the 0-inflated case if we plot the pdfs:
image


Using a form of jitter_under_thresh, say that I inflate with precipitation values of 0, but then jitter between 10^(-21) and 10^(-20), I get something more in line with other results. 

If I inflate less values, 10 or less, it seems the problem goes away (I also added a small value in the cdf computation, but that doesn't explain the change in behaviour as far I could tell):
```python
import numpy as np
import scipy as sc
pr0 = open_dataset("sdba/CanESM2_1950-2100.nc")
# Take the first 30 july 1rst 
pr0  = pr0.isel(location=1).isel(time=slice(179,365*30,365)).pr.values
pr0 = np.where(pr0>0, pr0, 1e-6)
min_prs = [
    0,
    1e-20,
    1e-17,
    1e-16,
    ]
print(["min_pr", "Pr(X>=min_pr)"])
for min_pr in min_prs:
    pr = pr0.copy()
    pr[0:10] = min_pr
    a, loc,scale = sc.stats.gamma.fit(pr)
    print([min_pr, sc.stats.gamma.cdf(min_pr+1e-21,a,loc=loc,scale=scale)])
    x = np.linspace(0, 1e-15, 1000)
    plt.plot(x,sc.stats.gamma.pdf(x, a, loc=loc, scale=scale),label=min_pr)
    plt.yscale("log")
    plt.xlabel("x")
    plt.ylabel("pdf")
plt.legend(title="min_pr")
```

image

CHANGES.rst
```diff
@@ -4,12 +4,17 @@ Changelog
 
 v0.49.0 (unreleased)
 --------------------
-Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user:`juliettelavoie`), David Huard (:user:`huard`), Gabriel Rondeau-Genesse (:user:`RondeauG`).
+Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user:`juliettelavoie`), David Huard (:user:`huard`), Gabriel Rondeau-Genesse (:user:`RondeauG`),, Éric Dupuis (:user:`coxipi`).
```
coxipi, line 7:
```suggestion
Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user:`juliettelavoie`), David Huard (:user:`huard`), Gabriel Rondeau-Genesse (:user:`RondeauG`), Éric Dupuis (:user:`coxipi`).
```

coxipi:
Apparently the PR was never in draft mode? Just wanted to say it's ready for your eyes.

xclim/indices/_agro.py
```diff
@@ -1289,6 +1255,12 @@ def standardized_precipitation_evapotranspiration_index(
         `PWM` (probability weighted moments).
         The approximate method uses a deterministic function that doesn't involve any optimization. Available methods
         vary with the distribution: 'gamma':{'APP', 'ML', 'PWM'}, 'fisk':{'APP', 'ML'}
+    fitkwargs : dict, optional
+        Kwargs passed to ``xclim.indices.stats.fit`` used to impose values of certains parameters (`floc`, `fscale`).
+    offset : Quantified
+        For distributions bounded by zero (e.g. "gamma", "fisk"), the two-parameters distributions only accept positive
+        values. An offset can be added to make sure this is the case. This option will be removed in xclim >=0.49.0, ``xclim``
+        will rely on a proper use three-parameters distributions instead.
```
aulemahal, line 1263:
```suggestion
        will rely on proper use of three-parameters distributions instead.
```

xclim/indices/_agro.py
```diff
@@ -1289,6 +1255,12 @@ def standardized_precipitation_evapotranspiration_index(
         `PWM` (probability weighted moments).
         The approximate method uses a deterministic function that doesn't involve any optimization. Available methods
         vary with the distribution: 'gamma':{'APP', 'ML', 'PWM'}, 'fisk':{'APP', 'ML'}
+    fitkwargs : dict, optional
+        Kwargs passed to ``xclim.indices.stats.fit`` used to impose values of certains parameters (`floc`, `fscale`).
+    offset : Quantified
+        For distributions bounded by zero (e.g. "gamma", "fisk"), the two-parameters distributions only accept positive
+        values. An offset can be added to make sure this is the case. This option will be removed in xclim >=0.49.0, ``xclim``
```
aulemahal, line 1262:
```suggestion
        values. An offset can be added to make sure this is the case. This option will be removed in xclim >=0.50.0, ``xclim``
```

xclim/indices/stats.py
```diff
@@ -710,7 +739,28 @@ def standardized_index_fit_params(
...
+                    "Pass `floc=0` in `fitkwargs`."
+                )
+        else:
+            fitkwargs["floc"] = 0
+            warnings.warn(
+                "The APP method is only supported for two-parameter distributions with `gamma` or `fisk`."
+                "Location parameter `loc` will be set to 0. To avoid this warning, set `floc=0` manually in `fitkwargs`."
+            )
+    if offset:
+        warnings.warn("Inputting an offset will be deprecated in xclim>=0.49.0. ")
```
aulemahal, line 759:
```suggestion
        warnings.warn("Inputting an offset will be deprecated in xclim>=0.50.0. ")
```

aulemahal (commented):
This is not a thorough review, only superficial for now.

I'm not sure I understand the role of `zero_inflated`, I think it could be useful to explain when this option is needed in the notes of SPI ?

xclim/indices/stats.py
```diff
@@ -710,7 +739,35 @@ def standardized_index_fit_params(
...
+      \texttt{pdf}(X) = \pi  \texttt{ if }  X=0  \texttt{ else } (1-\pi) \texttt{pdf}_0(X)
     """
+    fitkwargs = fitkwargs or {}
+    if method == "APP":
+        if "floc" in fitkwargs.keys():
+            if fitkwargs["floc"] != 0:
+                raise ValueError(
+                    "The APP method is only supported for two-parameter distributions with `gamma` or `fisk`. `loc` parameter must be set 0."
+                    "Pass `floc=0` in `fitkwargs`."
+                )
```
aulemahal, line 758: I don't understand. Doesn't the implementation above actually  implement this ?

coxipi: Yes, we could in principle allow `APP` with `loc != 0`. This is just a concern about the quality of the fit: 
* This is taking into account how David was more or less on board with advertising `APP` method
* The way I implemented how to find the `loc` parameter is generic, independent of the distribution really, and by no means perfect. I tried finding a better formula with the method of moments and analytic approximations. It works very well with two parameters `shape, scale`, but not so much once the third parameter `loc` is added in. 

From what I have seen, I would not recommend using the method `APP` in this case

aulemahal: But then how do you use APP with SPEI in xclim >= 0.50, when `offset` is removed ?

aulemahal: As this the offset technique is what `climate_indices` does, I would let this option open, even if it's only through the `standardized_index_fit_params` method (but then, this would need to be documented!).

coxipi: Ok, I left the offset only in `standardized_index_fit_params`, and there are still some temporary checks in SPEI. 

What you had in mind for documentation? A notebook example to show how to reproduce `climate_indices`?

aulemahal: I'm not sure... I would say simply an example of how to passe an offset, I thought it would be through `floc` in `fitkwargs`.

coxipi: I'm looking into this. Perhaps allowing `APP` only when `floc` is given (`loc` is fixed, so effectively 2-parameter distribution). I'll see if using `floc` as a way to have the offset gives the expected results

coxipi: Seems to be working fine, I'll go with this

xclim/indices/_agro.py
```diff
@@ -1163,12 +1157,14 @@ def standardized_precipitation_index(
...
     -----
-    * The length `N` of the N-month SPI is determined by choosing the `window = N`.
+    * N-month SPI / N-day SPI is determined by choosing the `window = N` and the appropriate frequency `freq`.
     * Supported statistical distributions are: ["gamma", "fisk"], where "fisk" is scipy's implementation of
        a log-logistic distribution
     * If `params` is given as input, it overrides the `cal_start`, `cal_end`, `freq` and `window`, `dist` and `method` options.
+    * "APP" method only supports two-parameter distributions. Parameter `loc` needs to be fixed to use method `APP`.
     * The standardized index is bounded by ±8.21. 8.21 is the largest standardized index as constrained by the float64 precision in
       the inversion to the normal distribution.
+    * The results from `climate_indices` library can be reproduced with `method = "APP"` and `fitwkargs = {"floc": 0}`
```
coxipi, line 1167: @aulemahal perhaps a notebook in the future could be warranted, but I think this is still helpful in the meantime

xclim/indices/_agro.py
```diff
@@ -1326,44 +1296,63 @@ def standardized_precipitation_evapotranspiration_index(
...
-    --------
-    standardized_precipitation_index
-
     Notes
     -----
-    If results include NaNs, check that the `offset` parameter is larger than the minimum water budget values.
+    The results from `climate_indices` library can be reproduced with `method = "APP"` and `fitwkargs = {"floc": - offset}`
+    where the offset is a scalar that corresponds to "1 mm/d" in units of `wb`. This can be obtained with
+    `offset = xclim.core.units.convert_units_to("1 mm/d", wb, context="hydro")`
+
```
coxipi, line 1304: same here, info on how to obtain results from `climate_indices` library

xclim/indices/_agro.py
aulemahal: Indeed. A "xclim compared to other package" page in the doc could be useful.

coxipi: Good idea. This also could be a good place to discuss the wrapper on SBCK, eventually Ibicus if we implement some things, etc ... :upside_down_face:

tests/test_indices.py
```diff
@@ -702,10 +702,10 @@ def test_standardized_precipitation_evapotranspiration_index(
             tas = tasmax - 2.5
             tasmin = tasmax - 5
             wb = xci.water_budget(pr, None, tasmin, tasmax, tas)
-            wb = wb + convert_units_to("1 mm/d", wb, context="hydro")
-        fitkwargs = {}
         if method == "APP":
-            fitkwargs["floc"] = 0
+            # same offset as in climate indices
+            offset = convert_units_to("1 mm/d", wb, context="hydro")
```
aulemahal, line 707:
```suggestion
            offset = convert_units_to("1000 mm/month", wb, context="hydro")
```
Isn't this the hardcoded offset of `climate_indices` ?
https://github.com/monocongo/climate_indices/blob/d108eee982abae06a415e888319b8078af868558/src/climate_indices/indices.py#L280C62-L280C69

xclim/indices/stats.py
```diff
@@ -750,21 +750,15 @@ def standardized_index_fit_params(
...
+        if "floc" not in fitkwargs.keys():
+            raise ValueError(
+                "The APP method is only supported for two-parameter distributions with `gamma` or `fisk` with `loc` being fixed."
+                "Pass a value for `floc` in `fitkwargs`."
             )
-    if offset:
-        warnings.warn("Inputting an offset will be deprecated in xclim>=0.50.0. ")
     if offset is not None:
+        warnings.warn(
+            "Inputing an offset will be deprecated in xclim>=0.50.0. To achieve the same effect, use `fitkwargs['floc']` instead."
```
aulemahal, line 760:
```suggestion
            "Inputing an offset will be deprecated in xclim>=0.50.0. To achieve the same effect, pass `- offset` as `fitkwargs['floc']` instead."
```

aulemahal (approved):
I like this! thanks a lot !

tests/test_indices.py
coxipi: yes, climate_indices and xclim do match, but with the current datasets used in our test, we stil have negative values with the climate_indices offset. I wll wait for a next PR where we show many comparisons to many other libraries to sort this out

coxipi: but I agree, I didn't have the correct offset here

PR #1721, opened 2024-04-18
`mypy` compliance

### LINKED ISSUE #1719: Adopt `mypy` and stronger enforcement of `pylint`
### Addressing a Problem?

Why would we want to do this?
- Xclim is a scientific library that aims to be strongly typed and optimized; this library is made to do that specifically.
- Ensuring that variables are always typed and do not change their type during execution is a reliable way of optimizing our code base and evade undefined behaviours.
- Our code base is already close to compliant with full static typing and type checking that we would get from `mypy`.

### Potential Solution

We start using `mypy` in our linters and set some fairly high thresholds for `pylint`. We currently use `pylint` with some allowances (our code rating is hovering around 9.6/10, so we're nearly there).

`mypy` is a fairly involved check, so perhaps we would run the comprehensive checks from it as part of the `lint` job on GitHub, and run a more relaxed threshold for it as part of the `pre-commit` configurations (to catch obvious errors). `pylint` can be configured for stronger checks across both (runs very fast).

Although we've already ticked most of the boxes, this change might still need to occur progressively (as the documentation suggests). I can imagine this requiring effort from all maintainers to help determine what can easily be typed and what we can comfortably ignore.

### Additional context

I think we have been gradually making our way towards this for several years. The quality of our code has been improved thanks to the existing `pylint` checks (soft) as well as the battery of linters we use. The last steps needed to ensure `mypy`-compliant code will help address some finer issues that are often raised in other linters as well (`flake8`, `ruff`, `vulture`)

This goal is something that was also brought up by a few research developers I spoke with at PyCon Lithuania. This would be a medium-effort solution that would provide some security, code optimization, some potential speedups, and would benefit downstream projects hoping to adopt `mypy` as well for their code bases.

The comprehensive guide to `mypy`: https://dev.to/tusharsadhwani/the-comprehensive-guide-to-mypy-561m

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR addresses issues raised in #1719 

* Addresses several issues (around 250/1250 errors) associated with (re-)casting of variables and output typing.
* The `Indicator.__getitem__` method has been retired (to reduce instances of class methods returning `Any`)
* PEP8 rule N802 is now enforced (with `noqa` exceptions, mostly for functions based on `sfcWind` variable).
* The `wrapped_partial` function wrapper has been removed.

Yes. The Indicator `__getitem__` method has been removed. This has downstream impacts on `finch` and possibly other projects. Additionally, the obsolete `wrapped_partial` function (was previously used in Indicator construction) has also been removed.

### Other information:

@aulemahal

We currently use the `Quantified` type alias in almost all the threshold-based indicators to specify that this variable should conform to something resembling a string with units (`str | pint.Quantity`), but this alias also includes `xarray.DataArray`. Does this mean that we can pass (or would like to enable passing) `DataArrays` of grids denoting thresholds? If not, we may need to create a new, more constrained custom variable (`str2pint` often takes variables of this type but does not support `pint.Quantity` or `xarray.DataArray`). Solving this issue alone will address +250 errors.


xclim/__init__.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import pathlib
```
VascoSch92, line 5: `From pathlib import Path`

As you just use this class and nothing other in the module

Zeitsperre: Makes sense. I typically only ever import the class. Will modify, thanks!

Zeitsperre:
There seems to be a handful of trip-falls that we keep running into:
1. Many of the decorators that we employ do not constrain the types of variables passed in/out of them; This makes it so that regardless of the typing identified by a decorated function, the end results will always return `Any`. This can be fixed by using `-> TypeVar[...]` signatures.
2. Mentioned above, `Quantified` is used too often in places where it doesn't actually work (some functions demand `str`-like variables, but claim to support `Quantified`, which includes `DataArray`). This could become a breaking change but would address _many_ errors.

Changes as of this commit have us hovering at around 1050 errors total.

Zeitsperre:
Additional comment:

Due to the way that `numpy` is typed, any time that we pass `xarray.DataArray` to a `numpy.ufunc`-like operation (e.g. `np.logical_and(da1, da2)`) we need to specify the that the output is still a `xarray.DataArray` (expected output is an `ndarray` otherwise).

xclim/core/units.py
```diff
@@ -252,6 +252,7 @@ def str2pint(val: str) -> pint.Quantity:
         return units.Quantity(1, units2pint(val))
 
 
+# FIXME: The typing for this needs to be addressed. Maybe we need to use generics here to help specify what goes in/out?
```
aulemahal, line 255: Not sure I understand the comment. Doesn't `Quantified` cover exactly what this function accepts ? `DataArray`, `str` and `pint.Quantity`.

xclim/core/utils.py
```diff
@@ -71,8 +71,27 @@
...
+    def __init__(self, func: T) -> None:  # noqa: D102
+        self._func = func
+
+    def __call__(self, *args, **kwargs):  # noqa: D102
+        return self._func(*args, **kwargs)
+
+    def __getattr__(self, attr):  # noqa: D105
+        return getattr(self._func, attr)
+
+    _injected: dict[str, Any]
```
aulemahal, line 89: I know you told me about this, but I now understand less... What's the use of creating a full class that is never being used ? Isn't the output of `wrapped_partial` an actual callable and not a custom object ?

xclim/indices/_agro.py
```diff
@@ -132,7 +133,7 @@ def corn_heat_units(
         )
     ) / 2
 
-    chu.attrs["units"] = ""
+    chu = chu.assign_attrs(units="")
     return chu
```
aulemahal, line 137: Héhé this PR makes a clear point that shadowing variables isn't cool ;)

However, what's the issue with setting an attribute ?
```suggestion
    return chu.assign_attrs(units="")
```

Zeitsperre: The thing with returning an object that is being modified "in-transit" is that `mypy` can't effectively interpret its output type. I think it needs to be defined so that typing inference can happen, then it can be returned.

Since this is modifying `chu` here, maybe we could declare the type of that variable and see what happens? I'll test that out.

xclim/core/units.py
Zeitsperre: Good point. This comment should be placed elsewhere. `convert_units_to` is consistent/accurate.

xclim/core/utils.py
Zeitsperre: This entire wrapper code has been removed, as it has not been in use for several versions (replaced by `declare_units`)

xclim/indices/_multivariate.py
```diff
@@ -210,8 +210,10 @@ def cold_and_dry_days(
     thresh = resample_doy(pr_per, pr)
     pr25 = pr < thresh
 
-    cold_and_dry = np.logical_and(tg25, pr25).resample(time=freq).sum(dim="time")
-    return to_agg_units(cold_and_dry, tas, "count")
+    cold_and_dry = cast(xarray.DataArray, np.logical_and(tg25, pr25))
```
aulemahal, line 213: What's the difference between this and 
```suggestion
    cold_and_dry: xr.DataArray = np.logical_and(tg25, pr25)
```
?

xclim/indices/_multivariate.py
```diff
@@ -970,7 +978,7 @@ def precip_accumulation(
         pr = rain_approximation(pr, tas=tas, thresh=thresh, method="binary")
     elif phase == "solid":
         pr = snowfall_approximation(pr, tas=tas, thresh=thresh, method="binary")
-    pram = rate2amount(pr)
+    pram: xarray.DataArray = rate2amount(pr)
```
aulemahal, line 981: Isn't the output of  `rate2amount` already annotated as `DataArray` ?

xclim/indices/_multivariate.py
```diff
@@ -1804,7 +1812,8 @@ def winter_rain_ratio(
     """
     ratio = liquid_precip_ratio(pr, prsn, tas, freq=freq)
     winter = ratio.indexes["time"].month == 12
-    return ratio.sel(time=winter)
+    out: xarray.DataArray = ratio.sel(time=winter)
```
aulemahal, line 1815: Doesn't xarray already provide complete typing ? Isn't the output of `DataArray.sel` already annotated `DataArray` ?

xclim/indices/_threshold.py
```diff
@@ -292,8 +292,9 @@ def cold_spell_max_length(
         rl.longest_run,
         freq=freq,
     )
-    out = max_l.where(max_l >= window, 0)
-    return to_agg_units(out, tas, "count")
+    max_window = max_l.where(max_l >= window, 0)
+    out: xarray.DataArray = to_agg_units(max_window, tas, "count")
```
aulemahal, line 296: Same as for `rate2amount` : isn't `to_agg_units` already annotated correctly?

xclim/indices/_threshold.py
```diff
@@ -382,13 +383,16 @@ def snd_season_end(
...
         cond.resample(time=freq)
         .map(rl.season, window=window, dim="time", coord="dayofyear")
         .end
     )
-    out.attrs.update(units="", is_dayofyear=np.int32(1), calendar=get_calendar(snd))
-    return out.where(~valid)
+    resampled.attrs.update(
+        units="", is_dayofyear=np.int32(1), calendar=get_calendar(snd)
+    )
+    out: xarray.DataArray = resampled.where(~valid)
```
aulemahal, line 394: Same as above for `sel` : wouldn't xarray's typing make this clear ?

aulemahal (commented):
This looks good, but there are a lot of cases where it seems an annotation shouldn't be necessary ? I've commented on a few, but not all.

xclim/indices/_multivariate.py
Zeitsperre: I'm glad you asked ;)

The `variable: sometype =` approach is for specifying that the outputs from an assignment that would return `Any` is constrained to a specific `sometype` instead.

The `cast(sometype, ...)` approach is used to tell the type checker that you should treat the variable as `sometype` regardless of what the type hinting from the assignment would suggest. Casting here is needed because `numpy` does not understand that if passed `xarray` objects, it should output `xarray` objects (instead, it thinks they are `ndarray`).

xclim/indices/_multivariate.py
Zeitsperre: `rate2amount` is decorated with a `_register_conversion` and the typing for that hasn't been addressed yet, so whatever comes out of that function is seen as `Any`. I'll see if I can tackle that.

Zeitsperre: I can't seem to figure this one out. I'll make a note to return to this later on.

xclim/core/units.py
```diff
@@ -1016,10 +1028,17 @@ def check_units(val: str | xr.DataArray | None, dim: str | xr.DataArray | None)
...
+
     # In case val is a DataArray, we try to get a standard_name
-    context = infer_context(
-        standard_name=getattr(val, "standard_name", None), dimension=dim
-    )
+    if isinstance(val, xr.DataArray):
+        standard_name = val.attrs.get("standard_name", None)
+        context = infer_context(standard_name=standard_name, dimension=_dim)
+    else:
+        context = "infer"
```
aulemahal, line 1041:
```suggestion
        context = "none"
```

xclim/indices/_multivariate.py
```diff
@@ -696,7 +704,7 @@ def heat_wave_frequency(
         window=window,
         freq=freq,
     )
-    out.attrs["units"] = ""
+    out = out.assign_attrs(unit="")
```
aulemahal, line 707:
```suggestion
    out = out.assign_attrs(units="")
```

Zeitsperre: Ah, thanks!

xclim/indices/_threshold.py
Zeitsperre: This is true. Reverting.

xclim/indices/_threshold.py
Zeitsperre: This is an issue with the signature for `convert_units_to`; If we want to have consistency between what is passed and what is received, we need to be using Generics.

However, we can't use `Quantified` because the output is either a `DataArray` or a `float`. It's not possible for the checker to determine the combination of call signatures needed to get either of these, so the end result is that the received variable is potentially both types. Therefore, we need to constrain it to `DataArray` on reception.

Zeitsperre: The documentation on this is a bit technical. Maybe we can tackle this in another PR?

See: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

aulemahal: Honestly, I think it would be better to wait and fix these issues instead of introducing new things that we know will be removed. Or at least, this PR shouldn't add annotations that we know will be removed by the subsequent PR ?

Mypy tells you when types are not correctly used or are unparsable, but it does not tell you about unnecessary annotations, does it ?

Zeitsperre: It does not warn us about redundant annotations, no. Should I go forward with the removal of these specific changes, do you think?

xclim/indices/_multivariate.py
Zeitsperre: This is due to the decorator issue. FIXME's have been added.

Zeitsperre:
This is just missing the updates to the changelog about breaking changes, but I think we're better off now. Lots of small improvements.

xclim/indices/_conversion.py
```diff
@@ -1447,33 +1447,33 @@ def potential_evapotranspiration(
         pet = pet.clip(0)
 
     elif method in ["droogersallen02", "DA02"]:
-        tasmin = convert_units_to(tasmin, "degC")
-        tasmax = convert_units_to(tasmax, "degC")
-        pr = convert_units_to(pr, "mm/month", context="hydro")
+        _tasmin = convert_units_to(tasmin, "degC")
+        _tasmax = convert_units_to(tasmax, "degC")
+        _pr = convert_units_to(pr, "mm/month", context="hydro")
```
aulemahal, line 1452: Are those underscores a consequence of `convert_units_to` being hard to interpret by `mypy` ? If so, I would suggest not making the change here and rather wait for a typing fix in `convert_units_to`.

aulemahal (approved):
Thanks @Zeitsperre! This was a lot of work.

I have a last conservative comment, but this looks like a very good start for type-checked xclim.

xclim/indices/_conversion.py
Zeitsperre: Nope, this is needed because the type signature of all these variables are `Optional[xr.DataArray]`. Assigning them into new variables is needed to ensure that they don't retain the possibilty of being interpreted as `None`.

PR #1731, opened 2024-04-25
Use Furo documentation theme

### LINKED ISSUE #1693: Dark mode docs
### Addressing a Problem?

Could we have the docs for xclim be available in darkmode? with a toggle switch of course.

### Potential Solution

Change the readthedocs theme to something with a toggle for darkmode :) 

### Additional context

_No response_

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1693 

* Replaces the `sphinx-rtd-theme` for the `furo` documentation theme
* Adapts the `conf.py` to make use of specific variables for `furo`
* `pygments` themes and CSS overrides have been added to ensure adequate contrast of documentation elements in both "dark" and "light" modes
* Adds alternate logos for the "dark" theme
* Updates the docstrings for `xclim/data/__init__.py`, `xclim/core/locales.py`, `xclim/core/formatting.py`, and others
* Synchronizes tooling dependency versions (ruff, black, yamllint)
* `generate_indicator_docstring` has been adjusted to ensure that information blocks (e.g. `Parameters`, `Returns`) always have newlines between them and that call signature attributes are indented four (4) spaces instead of two (2)
* The `pygments`-affected code blocks in `extendxclim.ipynb` have been modified to produce nicer outputs

Yes. 
- The theme has been completely replaced with `furo`, which gives us a dark mode for free as well as better navigation scroll bars when viewing pages in landscape mode.
- Indicator docstrings have been modified to ensure that `Parameters` items conform strictly to `numpy`-docstring format (indentations of four spaces for attributes, changed from two spaces).

### Other information:

The load order of layouts appears to be slightly different, and I'm having trouble getting the `indicators.rst` search bar to load properly. Likely a very small change is needed.


Zeitsperre:
@aulemahal (and @SarahG-579462)

Tagging you here for the JavaScript magic. I think the `layout` files are handled differently or load later. It seems as though the header content that should be adding `minisearch.js` to the relevant pages isn't working as intended.

aulemahal:
@Zeitsperre Fixed the javascript and html injection! Furo simply uses a different base template name and jinja blocks.

Zeitsperre:
Dark mode is looking sharp! 
image

docs/_static/style.css
aulemahal, line 1: Indicator cards are not dynamic with the theme : 
image
 
The colours are set at the end of this file. I guess we should use css variables provided by `furo` ?

aulemahal: image

Same for table colours, set here too.

aulemahal (changes requested):
I like this! I think there's a little css fiddling needed still.

pyproject.toml
```diff
@@ -79,16 +79,17 @@ dev = [
   "pytest-cov",
   "pytest-socket",
   "pytest-xdist[psutil] >=3.2",
-  "ruff >=0.2.0",
+  "ruff >=0.3.0",
   "tokenize-rt",
-  "tox >=4.0",
+  "tox >=4.5",
```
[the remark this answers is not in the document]
SarahG-579462: Update all optional dependencies to match between pyproject and environment, so we don't have to deal with pip potentially rewriting conda-chosen installs

docs/_static/style.css
```diff
@@ -1,4 +1,4 @@
-@import url("theme.css");
+@import url("styles/furo.css");
```
SarahG-579462, line 1: this is not really usable, for jupyter outputs of xarray/pandas dataframes:
Screenshot_20240429_110047

(what information is supposed to be there:
Screenshot_20240429_110148
)

SarahG-579462: Note that this issue only applies when using the "auto" theme, not the "dark" or "light" theme. Probably a jupyter issue.

SarahG-579462: I suggest we just disallow the "auto" theme altogether, if possible. as noted elsewhere the icon is not intuitive.

SarahG-579462, line 1: the "copy output" button on jupyter cells is really unclear in darkmode:
Screenshot_20240429_110526

SarahG-579462, line 1: /notebooks/extendxclim.html
This image should be on a white background, or use grey text (there is black, invisible text here):

Screenshot_20240429_110720

Screenshot_20240429_110824

SarahG-579462, line 1: The yaml css's color for numbers is black. It should be something else:
Screenshot_20240429_110949

SarahG-579462, line 1: The css for indicators in darkmode is... shall I say... not great?
Screenshot_20240429_111219

SarahG-579462, line 1: slight theme regression, don't know if it's important: the "external link" css is gone
in the current docs:
Screenshot_20240429_112025
in furo:
Screenshot_20240429_112128

SarahG-579462 (changes requested):
Some issues with the css, and some suggestions.

docs/_static/style.css
Zeitsperre: The styling of external links was an option specifically for ReadTheDocs theme. Will see what I can do about this.

Zeitsperre: Will look into this.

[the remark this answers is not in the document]
Zeitsperre: This is a bit tricky. We have to set the theme statically, since this is hard-coded into the README. This logo is actually `xclim-logo-small.png`. I can make a dark version, though.

docs/conf.py
```diff
@@ -317,7 +337,7 @@ class XCStyle(AlphaStyle):
 # -- Options for LaTeX output ------------------------------------------
 
 latex_engine = "pdflatex"
-latex_logo = "logos/xclim-logo.png"
+latex_logo = "logos/xclim-logo-light.png"
```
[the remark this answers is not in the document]
Zeitsperre: Currently, there isn't a PDF version of the docs. This code block is redundant.

docs/_static/style.css
Zeitsperre: It appears that there's a discussion for this: https://github.com/pradyunsg/furo/discussions/775

Zeitsperre: This (I think) is set via the `pygments` library. I'll see what can be done.

Zeitsperre: We can modify the CSS using an override: https://sphinx-copybutton.readthedocs.io/en/latest/use.html#modify-the-copy-button-s-style-with-css.

Perhaps we can make this responsive in the conf.py? https://pradyunsg.me/furo/customisation/#light-css-variables-dark-css-variables

Zeitsperre: Alternatively, this button is responsive to mouseover to make it interactive/attention-grabbing: 
image

docs/_static/style.css
```diff
@@ -102,3 +102,64 @@ code > .indName {
 #incVirtModLbl {
     display: inline;
 }
+
+/* extend furo for inline ".only-dark" elements */
```
Zeitsperre, line 106: Are there any elements below here that we could move to the `conf.py`? Maybe the directives that style things handled by sphinx extensions/plugins?

SarahG-579462: everything here is standard css, and will work out of the box with any package. except the last bit, 
```
button.copybtn.copybtn svg {
  stroke: var(--color-content-foreground);
}
```
but it will be black if --color-content-foreground is not defined.

Zeitsperre: There's a `--color-foreground-primary` (and `secondary`) found in the `furo.css`. We can use that, I'm sure.

SarahG-579462: yes, but you're worried about maintenance issues. We can use the variables we self-define in conf.py, but we *do* need to use a var somewhere (unless we replace classes with js... which is just unnecessary here)

Zeitsperre: Okay, so long as we're building off of the `conf.py` configuration, I'm happy. Thanks!

docs/_static/style.css
Zeitsperre: For this, we can probably just choose a different dark theme. Examples: https://stylishthemes.github.io/Syntax-Themes/pygments/

Zeitsperre: I would propose "material" (since it's great for everything).

SarahG-579462:
I think that if we fix the jupyter notebook issue on `auto` (or remove auto), and fix the yaml issue for numbers, this is good to go.

docs/conf.py
```diff
@@ -283,31 +284,50 @@ class XCStyle(AlphaStyle):
...
-html_theme_options = {"logo_only": True, "style_external_links": True}
+html_title = "xclim Official Documentation"
+html_short_title = "xclim"
+
+html_theme = "furo"
+html_extra_path = ["variables.json"]
+
+# Theme options are theme-specific and customize the look and feel of a theme further.
+# For a list of options available for each theme, see the documentation.
+html_theme_options = {
```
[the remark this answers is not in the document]
Zeitsperre: It turns out that that's a bit complicated to perform. Having an "auto" option is integral to the functioning of the theme: https://github.com/pradyunsg/furo/issues/24.

docs/_static/style.css
Zeitsperre: @SarahG-579462 I've opted for `sas` and `lightbulb` as the `pygments` styles for code. Both of these themes are rated as being high contrast.

Zeitsperre: See: https://github.com/pradyunsg/furo/discussions/790

Zeitsperre: Existing issue specific to `xarray` objects: https://github.com/pydata/xarray/issues/8097#issuecomment-1688268711.

Zeitsperre: The theming for the lexer of the YAML block in the notebook simply needs to be adjusted:
```
    "highlighted_py = highlight(pydata, PythonLexer(), Terminal256Formatter(style=\"manni\"))\n",
    "highlighted_yaml = highlight(ymldata, YamlLexer(), Terminal256Formatter(style=\"manni\"))\n",
    "highlighted_json = highlight(jsondata, JsonLexer(), Terminal256Formatter(style=\"manni\"))"
```

SarahG-579462:
This latest commit fixes the issue with xarray not working in "auto" mode.

It is not the prettiest solution, since it requires redefining the xarray theme variables in the body tag. (in `_static/xarray.css`)

I chose to update the xarray css variables (e.g. `--xr-font-color0`) thru the externalized `--jp` variables that xarray seems to offer. I'm assuming these are jupyter variables. This fix should therefore be compatible with any jupyter output that uses variables set with a scope within `body`.

Other options would have been:
 - Copying the css that furo uses to set variables when swapping to light or dark mode, and adding a selector so that it updates in the root instead of the body.
 - Redefining the xarray colours in the body, instead of the offered `--jp` variables.
   - This might be a better solution, but it does not offer the nicety that is potentially fixing other jupyter notebook output issues.
 - wait for an upstream fix in either xarray or furo. 
   - The xarray upstream fix could be done in two ways:
     -  Add compatibility for the "auto" theme by using a similar css selector to furo, but targeting the :root pseudo-class instead of the body tag (using the [:has](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) pseudo-class)
     - Change the xarray default theming to be set in body instead of, or on top of, :root
   - the furo fix would be to:
     - Make the css variables apply to the :root pseudo-clas, instead of/on top of the body tag (once again with the :has pseudo-class.

SarahG-579462 (approved):
Approved! The logo looking unintuitive is very minor, and more a furo issue anyway.

docs/conf.py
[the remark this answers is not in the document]
SarahG-579462: I've raised an issue with furo about the unclarity of it. 

https://github.com/pradyunsg/furo/discussions/794

PR #1743, opened 2024-05-03
Use fastnanquantile

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

Use module `fastnanquantile` to compute quantiles in training functions of quantile mappings

No

### Other information:
Gotta go fast (see discussion for details, but we can expect a 4-5x speedup in the training step of quantile mappings)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+v0.50.0
+--------------------
+Contributors to this version: Éric Dupuis (:user:`coxipi`).
+
+New features and enhancements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* Great speed improvement in `sdba` quantile mappings training functions (e.g. ``sdba.DetrendedQuantileMapping.train``) due to using ``fastnanquantile`` as a backend.
+
+
 v0.49.0 (2024-05-02)
 --------------------
 Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Juliette Lavoie (:user:`juliettelavoie`), David Huard (:user:`huard`), Gabriel Rondeau-Genesse (:user:`RondeauG`), Javier Diez-Sierra (:user:`JavierDiezSierra`), Sarah Gammon (:user:`SarahG-579462`), Éric Dupuis (:user:`coxipi`).
--- a/environment.yml
+++ b/environment.yml
@@ -12,7 +12,6 @@ dependencies:
   - cftime >=1.4.1
   - Click >=8.1
   - dask >=2.6.0
-  - fastnanquantile
   - jsonpickle
   - numba
   - numpy >=1.20.0
@@ -81,5 +80,6 @@ dependencies:
   - yamllint
   - pip
   - pip:
+    - fastnanquantile
     - flake8-alphabetize
     - sphinxcontrib-svg2pdfconverter
--- a/environment.yml
+++ b/environment.yml
@@ -12,6 +12,7 @@ dependencies:
   - cftime >=1.4.1
   - Click >=8.1
   - dask >=2.6.0
+  - fastnanquantile
   - jsonpickle
   - numba
   - numpy >=1.20.0
--- a/xclim/sdba/_adjustment.py
+++ b/xclim/sdba/_adjustment.py
@@ -9,11 +9,11 @@ from __future__ import annotations
 
 import numpy as np
 import xarray as xr
+from fastnanquantile.xrcompat import xr_apply_nanquantile
 
 from xclim.core.units import convert_units_to, infer_context, units
 from xclim.indices.stats import _fitfunc_1d  # noqa
 
-from . import nbutils as nbu
 from . import utils as u
 from ._processing import _adapt_freq
 from .base import Grouper, map_blocks, map_groups
@@ -76,8 +76,12 @@ def dqm_train(
     refn = u.apply_correction(ds.ref, u.invert(ds.ref.mean(dim), kind), kind)
     histn = u.apply_correction(hist, u.invert(hist.mean(dim), kind), kind)
 
-    ref_q = nbu.quantile(refn, quantiles, dim)
-    hist_q = nbu.quantile(histn, quantiles, dim)
+    ref_q = xr_apply_nanquantile(refn, dim=dim, q=quantiles).rename(
+        {"quantile": "quantiles"}
+    )
+    hist_q = xr_apply_nanquantile(histn, dim=dim, q=quantiles).rename(
+        {"quantile": "quantiles"}
+    )
 
     af = u.get_correction(hist_q, ref_q, kind)
     mu_ref = ds.ref.mean(dim)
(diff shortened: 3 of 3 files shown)


coxipi:
## Small example of speed up

```python
import numpy as np
from fastnanquantile.xrcompat import xr_apply_nanquantile
from xclim.sdba import nbutils as nbu
ds = open_dataset("sdba/CanESM2_1950-2100.nc")
tx = ds.tasmax.sel(time=slice("1950", "1954"))
q = np.linspace(0, 1, 50)/100

# nbutils
%%timeit
nbu.quantile(tx, dim="time", q=q).compute()
>>> 10.1 ms ± 206 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)


# fastnanquantile
%%timeit
fnq.xrcompat.xr_apply_nanquantile(tx, dim="time", q=q).compute()
>>> 2.28 ms ± 26.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Zeitsperre (changes requested):
This is pretty neat (though I'm not sure why `fastnanquantile` hasn't been submitted upstream). Given that the library is in the early beta stage and doesn't have many maintainers (or a conda package), I'm a bit apprehensive about making this into a formal dependency. 

Would it be a reasonable ask to have this as an optional dependency and have our code base use it if it's available?

If you have the time to spare, you could also see if the creator would be open to placing his package on conda-forge? Could be as simple as opening an issue to ask.

coxipi:
`nbutils` could be improved a lot. We (@aulemahal, @SarahG-579462) were already doing good progress before I stumbled upon `fastnanquantile`. The bottleneck in `nbutils` can be reduced a lot by avoiding the temporal stacking and flatten the dimensions to be reduced before applying a quantile function. `fastnanquantile` also gave me ideas to further improve `nbutils`, this might be useful elsewhere.  By comparison, trying to use mix some parts of `nbutils` & `fastnanquantile`, it confirms my suspicions concerning the nature of the bottleneck. All in all, I think adding a `fastnanquantile` dependency is reasonable, but we can still achieve much better results if we instead decide to tweak `nbutils`.

coxipi:
```python
@njit
def _wrapper_sortquantile(arr,q): 
    out = np.empty((arr.shape[0], q.size), dtype=arr.dtype)
    for index in range(out.shape[0]):
        out[index] = _sortquantile(arr[index], q)
    return out

def _quantile2(arr, q, nreduce):
    if arr.ndim == nreduce :
        out = sdba.nbutils._sortquantile(arr.flatten(), q)
    else:
        axis = np.arange(len(arr.shape) - nreduce, len(arr.shape))
        drop_axis_idx = np.arange(len(arr.shape))[axis]
        final_shape = [arr.shape[idx] for idx in range(len(arr.shape)) if idx not in drop_axis_idx]
        final_shape = [len(q)] + final_shape
        reduction_dim_size = 1
        for axis_value in axis:
            reduction_dim_size *= arr.shape[axis_value]
        arr = arr.reshape(-1, reduction_dim_size)
        out = _wrapper_sortquantile(arr,q)
        out = out.reshape(final_shape)
    return out
def quantile2(da, q, dim):
    """Compute the quantiles from a fixed list `q`."""
    def _wrapper(npa, **kwargs):
        # move quantile axis to end. required for apply_ufunc
        return np.moveaxis(_quantile2(npa, **kwargs), 0, -1)
    qc = np.array(q, dtype=da.dtype)
    dims = [dim] if isinstance(dim, str) else dim
    kwargs = dict(nreduce = len(dims), q=qc)
    res = xr.apply_ufunc(
        _wrapper,
        da,
        input_core_dims=[dims],
        exclude_dims=set(dims),
        output_core_dims=[["quantiles"]],
        output_dtypes=[da.dtype],
        dask_gufunc_kwargs=dict(output_sizes={"quantiles": len(q)}),
        dask="parallelized",
        kwargs=kwargs,
    ).assign_coords(quantiles=q).assign_attrs(da.attrs)
    return res
 ```
 
 # nbutils-turbo (using lessons from fastnanquantile)
%%timeit
quantile2(tx, dim="time", q=q).compute()
>>> 2.47 ms ± 81.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

**Comment by aulemahal** [comment]
I agree with @Zeitsperre that having a single-maintainer, pip-only dependency is sketchy. As for why it isn't integrated upstream I think there's two reasons:
- can't go in `numpy` as it uses `jit` and numba.
- can't go in `numba` as it looks more complicated than a simple rewrite of numpy

It could optional, as for `SBCK`, for example. But if you achieve the same kind of performance with your own function, then that's good too.

**Comment by coxipi** [comment]
Replacing `stack` with `flatten` on temporal dims was already a very big improvement. My next idea was to try to avoid the spatial stacking of dimensions, I was thinking of investigating better use of `xr.apply_ufunc`, try to use `reshape` in numpy wisely etc. Those steps were already implemented well in `fastnanquantile` so it agrees with where I thought the next speed ups would be

**Comment by SarahG-579462** [comment]
If we can achieve the same without the dependency, I'm for either keeping it in-house, or we could ensure stability of the package by helping the maintainer with issues and adding it to conda-forge. Knowing that'd probably fall onto @Zeitsperre though, in-house is probably easier.

**Comment by coxipi** [comment]
I've adapted Sarah's PR with the functions above

```python
%%timeit
xr_apply_nanquantile(tx,dim ="time", q=q).compute()
>>> 2.27 ms ± 44 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
with xclim.set_options(allow_sortquantile=False):
    nbu.quantile(tx, q, dim="time").compute()
>>> 2.89 ms ± 44.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
with xclim.set_options(allow_sortquantile=True):
    nbu.quantile(tx, q, dim="time").compute()
>>> 2.52 ms ± 51.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Zeitsperre:
It doesn't seem like the project needs _too much_ help, honestly. If one among us would be willing to offer some aid to the dev and to get the package onto conda-forge, that would be enough for me.

Also, I can help with getting the package onto conda-forge, but I wouldn't want to commit to being the maintainer of it. This might be a great opportunity for @coxipi to learn conda-forge packaging ;)

coxipi:
I re-ran the same big data tests:

```
# nbutils (improved as above, but with np.nanquantile instead ) backend : 358 s
# fnq backend : 137 s
```
the ratio here is 2.6x instead of 1.4x for the small test above ... 

there is some important variability, but it seems `fnq` works better in this context. I'm re-running the tests. 

I will also give `da.quantile` a try, and might as well look at flox as suggested.

SarahG-579462:
Seems about right, my tests from way back when:
image

coxipi:
Yes, now it seems we can see your improvements with higher level functions, nice! The bottleneck I was mentioning (changing stack => flatten) doesn't seem so huge now if we compare the first three lines, I don't know why. Anways... For now I have modified the `quantile` function in the #1513 , and added `fastnanquantile` as an optional depedency.

Zeitsperre:
@coxipi 

For the dependency, I'd like you to add an installation recipe in tox for it if possible. What that looks like

- In the `tox.ini`, under `[testenv:deps]`: add an entry that says `fastnanquantile: fastnanquantile`
- In the `main.yml` GitHub Workflow, within the `[jobs:test-pypi:strategy:matrix:include]` entries: add the `fastnanquantile` command with a hyphen to one of the existing `tox-env` entries (maybe `py311-coverage-sbck`)

**Do not** add it to the `development-dependencies` of the `pyproject.toml`.

coxipi:
Closing for now, will open an issue.  `fastnanquantile` will be used optionally in #1513

PR #1764, opened 2024-05-29
Normalized taylordiagram

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add a `bool` option `normalize` and divide `ref_std, sim_std` by `ref_std` if True

No

### Other information:

I was planning a new plot option in `figanos`, but I was told we don't want to do any computation in `figanos`, it would be handy that `sdba` can manage this by itself.


xclim/sdba/measures.py
```diff
@@ -496,6 +500,13 @@ def _taylordiagram(
         }
     )
 
+    # Normalize the standard deviations byt the standard deviation of the reference.
+    if normalize:
+        out[{"taylor_param": [0, 1]}] = (
+            out[{"taylor_param": [0, 1]}] / out[{"taylor_param": 0}]
```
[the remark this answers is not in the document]
coxipi: If the field is homogeneous, yes.  I can't think of any case it would happen, and if so, doing a comparison with a Taylor diagram doesn't seem very interesting in this case? i.e. I would find this problematic even if `normalize=False` and we never do the division by 0.

coxipi:
```python
import figanos.matplotlib as fg
from xclim import sdba
import xclim
import numpy as np
ds = open_dataset("sdba/CanESM2_1950-2100.nc").sel(time=slice("1950", "2013"))
ds2 = open_dataset("sdba/ahccd_1950-2013.nc")
for v in ds.data_vars: 
    ds2[v] = xclim.core.units.convert_units_to(ds2[v], ds[v], context="hydro")
da = sdba.stack_variables(ds)
da2 = sdba.stack_variables(ds2)
out = sdba.measures.taylordiagram(ref=da, sim=da2, dim="time")
out[{"taylor_param":[0,1]}] = out[{"taylor_param":[0,1]}]/ out[{"taylor_param":0}]
# precip gives negatives correlations, just for plotting purposes
out[{"taylor_param":2}] = np.abs(out[{"taylor_param":2}])

# 1, old way of organizing points, still, using the dimensions in out
fg.taylordiagram(out)

# 2, new way of organizing points
fg.taylordiagram(out, colors_key="multivar", markers_key="location")


# 3, more like the old way, prepare a dictionary in advance
outs = out.stack(dimm=list(set(out.dims) - set(["taylor_param"])))
dd = {"-".join(v): outs.sel(dimm=v) for v in outs.dimm.values}
fg.taylordiagram(dd)

# 4, old way, with a dict, but use attrs to choose marker/color organization
outall = out.stack(dimm=list(set(out.dims) - set(["taylor_param"])))
dd = {"-".join(v): outs.sel(dimm=v) for v in outs.dimm.values}
for k in dd.keys():
    dd[k].attrs["attr_multivar"] = dd[k]["multivar"].values.item()
    dd[k].attrs["attr_location"] = dd[k]["location"].values.item()
fg.taylordiagram(dd, colors_key="attr_multivar", markers_key="attr_location")
```

coxipi:
With the old way, you are guaranteed to have unique points, it's not the case with the new way of organizing points ... this could be a feature, e.g. you would to scatter all your location points, here it seems a bit pointless, but for a more realistic dataset this might be interesting?

```python
fg.taylordiagram(out, colors_key="multivar") # there is still the location value that will not be discrimnated
```

image

xclim/sdba/measures.py
```diff
@@ -496,6 +500,17 @@ def _taylordiagram(
...
+    # Normalize the standard deviations byt the standard deviation of the reference.
+    if normalize:
+        if (out[{"taylor_param": 0}] == 0).any():
+            raise ValueError(
+                "`ref_std =0` (homogeneous field) obtained, normalization is not possible."
+            )
+        out[{"taylor_param": [0, 1]}] = (
+            out[{"taylor_param": [0, 1]}] / out[{"taylor_param": 0}]
+        )
+        out.attrs["normalized"] = True
```
coxipi, line 512:
```suggestion
        out.attrs["normalized"] = True
```
Assign this attribute if there is normalization, and don't add anything otherwise, is it reasonable?

xclim/sdba/measures.py
```diff
@@ -496,6 +500,17 @@ def _taylordiagram(
...
     )
 
+    # Normalize the standard deviations byt the standard deviation of the reference.
+    if normalize:
+        if (out[{"taylor_param": 0}] == 0).any():
+            raise ValueError(
+                "`ref_std =0` (homogeneous field) obtained, normalization is not possible."
+            )
+        out[{"taylor_param": [0, 1]}] = (
+            out[{"taylor_param": [0, 1]}] / out[{"taylor_param": 0}]
```
coxipi, line 510: Is throwing an error a good choice? 

If you send a homogenous field as a reference which will have `ref_std=0` , you should know you can't normalize this ... but technically, `ref_std=0` is computed inside the function, so you get a `ValueError` not from the input, something that's derived from the input within the function. I think it's all right, but I still want to point out this is a tricky ValueError that emerges from a computation inside the function

xclim/sdba/measures.py
Zeitsperre: I think this makes sense. This is solely for the user to consult manually, right?

xclim/sdba/measures.py
Zeitsperre: My stats-fu is not strong, so I can't effectively weigh in on this. @aulemahal?

Zeitsperre (approved):

xclim/sdba/measures.py
coxipi: Yes, it's not used elsewhere, thanks

Zeitsperre:
@coxipi Will merge this once the checks finish. All that's left is #1580 before the release!

PR #1778, opened 2024-06-13
Freezing rain events

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* ...

### DIFF
--- a/xclim/indices/_threshold.py
+++ b/xclim/indices/_threshold.py
@@ -37,7 +37,6 @@ from .generic import (
 # -------------------------------------------------- #
 
 __all__ = [
-    "black_ice_events",
     "calm_days",
     "cold_spell_days",
     "cold_spell_frequency",
@@ -54,6 +53,7 @@ __all__ = [
     "first_day_temperature_above",
     "first_day_temperature_below",
     "first_snowfall",
+    "freezing_rain_events",
     "frost_free_season_end",
     "frost_free_season_length",
     "frost_free_season_start",
@@ -105,91 +105,6 @@ __all__ = [
     "wetdays_prop",
     "windy_days",
 ]
-
-
-@declare_units(pr="[precipitation]", thresh="[precipitation]")
-def black_ice_events(
-    pr: xarray.DataArray,
-    thresh: Quantified = "1 kg m-2 d-1",
-    window_start: int = 3,
-    window_stop: int = 3,
-    freq: str | None = None,
-) -> xarray.Dataset:
-    r"""Black ice events.
-
-    Parameters
-    ----------
-    pr : xarray.DataArray
-        Black ice precipitation (`prfr`)
-    thresh : Quantified
-        Threshold that must be exceeded to be considered an event
-    window_start: int
-        Number of time steps above the threshold required to start an event
-    window_stop : int
-        Number of time steps below the threshold required to stop an event
-    freq : str
-        Resampling frequency.
-
-    Returns
-    -------
-    xarray.DataArray, [time]
-        Number of days with average near-surface wind speed below threshold.
-    """
-    freq = xarray.infer_freq(pr.time)
-    mag, units, _, _ = parse_offset(freq)
-    # condition to respect for `window_start` time steps to start a run
-    thresh = convert_units_to(thresh, pr)
-    da_start = pr >= thresh
-    da_stop = not da_start
-
-    # Get basic blocks to work with, our runs with holes and the lengths of those runs
-    # Series of ones indicating where we have continuous runs of black ice with pauses
-    # not exceeding `window_stop`
-    runs = rl.runs_with_holes(da_start, window_start, da_stop, window_stop)
-
-    # Compute the length of black ice events
-    # I think int16 is safe enough
-    ds = rl.rle(runs).to_dataset(name="run_lengths")
-    ds["run_lengths"] = ds.run_lengths.astype(np.int16)
-    ds.run_lengths.attrs["units"] = ""
-
-    # Time duration where the precipitation threshold is exceeded during an event
-    # (duration of complete run - duration of holes in the run )
-    ds["precipitation_duration"] = (
-        rl._cumsum_reset(
-            da_start.where(runs == 1), index="first", reset_on_zero=False
-        ).astype(np.int16)
-        * mag
-    )
-    ds["precipitation_duration"].attrs["units"] = units
-
-    # Cumulated precipitation in a given black ice event
-    pram = rate2amount(pr)
-    ds["cumulative_precipitation"] = rl._cumsum_reset(
-        pram.where(runs == 1), index="first", reset_on_zero=False
-    )
-    ds["cumulative_precipitation"].attrs["units"] = pram.units
-
-    # Reduce time dim to event dimension
-    mask = (ds.run_lengths > 0).any(dim=[d for d in ds.dims if d != "time"])
-    ds = ds.where(mask).dropna(dim="time").rename({"time": "event"})
-
-    # start time : with current implementation of time reduction above,
-    # this is not necessary, but it could be if we choose another way.
-    # ds["start"] = ds["time"].broadcast_like(ds)
-
-    # Other indices that could be completely done outside of the function, no input needed anymore
-    ds["number_of_events"] = (ds["run_lengths"] > 0).sum(dim="event").astype(np.int16)
-    ds.number_of_events.attrs["units"] = ""
-
-    ds["rate"] = ds["cumulative_precipitation"] / ds["precipitation_duration"]
-    units = (
-        f"{ds['cumulative_precipitation'].units}/{ds['precipitation_duration'].units}"
-    )
-    ds["rate"].attrs["units"] = ensure_cf_units(units)
-
-    ds.attrs["units"] = ""
-    return ds
 
 
 @declare_units(sfcWind="[speed]", thresh="[speed]")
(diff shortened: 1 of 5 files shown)


xclim/indices/_threshold.py
```diff
@@ -2654,6 +2656,123 @@ def wetdays_prop(
...
+        ds,
+        ds.run_lengths,
+        input_core_dims=[["time"], ["time"]],
+        output_core_dims=[["event"]],
+        kwargs=dict(max_event_number=max_event_number),
+        output_sizes={"event": max_event_number},
+        dask="parallelized",
+        vectorize=True,
+    ).assign_attrs(ds.attrs)
+
```
coxipi, line 2749: This change allows to filter the events, reorder this dimension. This requires xr.apply_ufunc and contiguous time dimension though. I don't think the previous implementation would have worked for without contiguous time anyways. 

However! The rest of the code uses calls to run_length and I believe this can be done with chunks over the time dimension. One other approach could to keep all time coords at first, and allow the use of yearly time chunks for instance, then save the output. Then, in a second pass, filter events using this apply_ufunc above.

xclim/indices/generic.py
```diff
@@ -1474,3 +1480,127 @@ def detrend(
...
+    data_is_rate : bool
+        True if the data is a rate that needs to be converted to an amount
+        when computing an accumulation.
+
+    Returns
+    -------
+    xr.Dataset
+    """
+    freq = xr.infer_freq(data.time)
+    mag, units, _, _ = parse_offset(freq)
```
coxipi, line 1517: `freq` is an input, but is simply redefined in the function?

tests/test_generic.py
```diff
@@ -768,3 +769,306 @@ def test_spell_length_statistics_multi(tasmin_series, tasmax_series):
...
+            0,
+            0,
+            0,
+            0,
+            0,
+            3,
+            2,
+            3,
+            2,
+        ]  # noqa
```
coxipi, line 1017: is it worth to make the notation less explicit but more compact in this case?
e.g.
```suggestion
        jan = [0]*3+[1,2, 3, 0, 3, 3,10] + [0]*6 + [2]*6 + [0]*5 + [3, 2, 3, 2]]  # noqa
```

aulemahal: dang the `#noqa` didn't work as expected... You idea works better!

aulemahal:
@coxipi  So I added the ability to find events for each resampling period.  I merged in `resample-map` thinking I could use that stuff here, but as the new functions are multivariate and output a Dataset, it doesn't work...

I edited to the comment with some more info.

tests/test_generic.py
coxipi: hmmm if you can make the #noqa work don't bother rewritting everything

xclim/indices/run_length.py
```diff
@@ -1683,3 +1690,144 @@ def suspicious_run(
...
+
+    Returns
+    -------
+    xr.Dataset, same shape as the data it has a new "event" dimension (and the time dimension is resample or removed, according to ``freq``).
+        event_length: The number of time steps in each event
+        event_effective_length: The number of time steps of even event where the start condition is true.
+        event_start: The datetime of the start of the run.
+        event_sum: The sum within each event, only considering the steps where start condition is true. Only present if ``data`` is given.
+    """
+    window_stop = window_stop or window
```
coxipi, line 1819: Do we want the default behaviour to reproduce a regular run length, or reproduce the concept of season? 

If the concept of season, then it's all right. If run_length, then we should take `window_stop=1`. I think both are fine, but we should add a note in the docstring to explain what the default behaviour reproduces

aulemahal: Ah! You're right I didn't think of that. I was thinking of run-like, so I'll change to 1. And edit the doc.

xclim/indices/generic.py
```diff
@@ -1478,3 +1478,70 @@ def detrend(
...
+    da_start = compare(data, op, thresh)
+    if thresh_stop is None and op_stop is None:
+        da_stop = ~da_start
+    else:
+        thresh_stop = convert_units_to(thresh_stop or thresh, data)
+        if op_stop is not None:
+            da_stop = compare(data, op_stop, thresh_stop)
+        else:
+            da_stop = ~compare(data, op, thresh_stop)
+
```
coxipi, line 1546: missing a "~" right? 
```suggestion
        if op_stop is not None:
            da_stop = ~compare(data, op_stop, thresh_stop)
        else:
            da_stop = ~compare(data, op, thresh_stop)

```

aulemahal: I was expecting `op_stop` to be the actual op here. It feels more intuitive ? 

So if you pass `op=">"`, you could pass `op_stop="<" `, for example.

coxipi: Ok, since `condition_stop` is not given as input here, I guess it makes sense to make it act on `data` rather than `~data`, but later on, `op_stop` on `~data`  so it will not be consistent

tests/test_generic.py
Zeitsperre:
```
some_really_long_line = [..., ..., ...]  # fmt: skip
```

aulemahal:
After discussion with Mrs Verglas herself, I decided that implementing a freezing rain indicator here  was not worth the hassle:
- No standard variable name for freezing rain
- Very specific and niche
- The best xclim implementation might not be the optimal performant one, considering that we don't need all outputs of the `find_events`
- The potential need to perserve the events would bypass an indicator implemented here anyway

But I did add an application for `runs_with_holes` by adding yet another parameter to the spell length statistics.

xclim/indices/generic.py
```diff
@@ -453,9 +457,17 @@ def spell_mask(
...
+        is_in_spell = (mask.rolling(time=window).sum() >= 1).shift(
+            time=-(window - 1), fill_value=False
+        )
         # Cut back to the original size
         is_in_spell = is_in_spell.isel(time=slice(0, data.time.size))
+
+    if min_gap > 1:
+        is_in_spell = rl.runs_with_holes(is_in_spell, 1, ~is_in_spell, min_gap).astype(
+            bool
+        )
```
coxipi, line 469: I think could be a if/else logic for `min_gap` which would be more efficient than this.
```suggestion
    if min_gap ==1 :
        # rest of the code
    else :
        mask = compare(data, op, thresh)
        is_in_spell = rl.runs_with_holes(mask, window, ~mask, min_gap).astype(
            bool
        )
```

This would be more efficient than first computing `is_in_spell`, just use directly `mask` instead. Now `min_gap` works on the condition instead of the spells, but I think it's equivalent?

aulemahal: I'm not sure it is equivalent, or maybe I'm not understanding the suggestion.

I think you are right for the window == 1 case, and the code already does what you suggest for that case. However when the spell has criteria on a moving-window statistic, we need to put True on all days within each spell before measuring the gaps, no ?

In the `win_reducer='mean'` , `window=3` case for example:
```
mask         = [F F F T F F F T F]
is_in_spell = [F F T T T F T T T]
```

coxipi: right I always forget about more complicated spells. I guess my point would be OK for the fast-track logic, but we can leave this as it is

xclim/indicators/atmos/_temperature.py
```diff
@@ -373,7 +375,8 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
     long_name="Total length of heat spells.",
     description="{freq} total length of heat spell events. "
     "A heat spell occurs when the {window}-day  averages of daily minimum and maximum temperatures "
-    "each exceed {thresh_tasmin} and {thresh_tasmax}.  All days of the {window}-day period are considered part of the spell.",
+    "each exceed {thresh_tasmin} and {thresh_tasmax}.  All days of the {window}-day period are considered part of the spell."
+    "Gaps of under {min_gap} days are allowed within a spell.",
```
coxipi, line 379: english? I'm not 100% but the formulation above seems weird
```suggestion
    "Gaps of less than {min_gap} days are allowed within a spell.",
```

tests/test_run_length.py
```diff
@@ -126,7 +126,7 @@ def test_rle(ufunc, use_dask, index):
 
 @pytest.mark.parametrize("use_dask", [True, False])
 @pytest.mark.parametrize("index", ["first", "last"])
-def test_extract_events_identity(use_dask, index):
+def test_runs_with_holes_identity(use_dask, index):
     # implement more tests, this is just to show that this reproduces the behaviour
     # of rle
```
coxipi, line 131:
```suggestion
    # this reproduces the behaviour or `rle`
```

coxipi:
@aulemahal  approved (can't approve this PR, I started it)

xclim/indicators/atmos/_temperature.py
aulemahal: @Zeitsperre what do you think ?

xclim/indicators/atmos/_temperature.py
```diff
@@ -307,7 +307,8 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
     long_name="Number of heat spells",
     description="{freq} number of heat spells events. A heat spell occurs when the {window}-day "
     "averages of daily minimum and maximum temperatures each exceed {thresh_tasmin} and {thresh_tasmax}. "
-    "All days of the {window}-day period are considered part of the spell.",
+    "All days of the {window}-day period are considered part of the spell. Gaps of under {min_gap} days are allowed "
```
Zeitsperre, line 310:
```suggestion
    "All days of the {window}-day period are considered part of the spell. Gaps of under {min_gap} day(s) are allowed "
```

xclim/indicators/atmos/_temperature.py
```diff
@@ -341,7 +342,8 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
     long_name="Longest heat spell",
     description="{freq} maximum length of heat spells. A heat spell occurs when the {window}-day "
     "averages of daily minimum and maximum temperatures each exceed {thresh_tasmin} and {thresh_tasmax}. "
-    "All days of the {window}-day period are considered part of the spell.",
+    "All days of the {window}-day period are considered part of the spell. Gaps of under {min_gap} days are allowed "
```
Zeitsperre, line 345:
```suggestion
    "All days of the {window}-day period are considered part of the spell. Gaps of under {min_gap} day(s) are allowed "
```

xclim/indicators/atmos/_temperature.py
Zeitsperre: My official Anglo opinion: `Gaps of {min_gap} day(s) or less are allowed within a spell.` or `Gaps of fewer than {min_gap} day(s) are allowed within a spell.`.

xclim/indices/run_length.py
```diff
@@ -1737,3 +1744,143 @@ def suspicious_run(
         keep_attrs=True,
         kwargs={"window": window, "op": op, "thresh": thresh},
     )
+
+
+def _find_events(da_start, da_stop, data, window_start, window_stop):
+    # Get basic blocks to work with, our runs with holes and the lengths of those runs
+    # Series of ones indicating where we have continuous runs of freezing rain with pauses
+    # not exceeding `window_stop`
```
Zeitsperre, line 1752: I would change this into a formal docstring, tbh.

xclim/indices/run_length.py
```diff
@@ -1737,3 +1744,143 @@ def suspicious_run(
...
+        The stopping boolean mask, true where the end condition of the event is fulfilled.
+        Defaults to the opposite of ``condition``.
+    window_stop : int
+        The number of consecutive True values in ``condition_stop`` for an event to end.
+        Defaults to 1.
+    data: DataArray, optional
+        The actual data. If present, its sum within each event is added to the output.
+    freq: str, optional
+        A frequency to divide the data into periods. If absent, the output has not time dimension.
+        If given, the events are searched within in each resample period independently.
```
Zeitsperre, line 1863:
```suggestion
    condition : DataArray of boolean values
        The boolean mask, true where the start condition of the event is fulfilled.
    window : int
        The number of consecutive True values for an event to start.
    condition_stop : DataArray of boolean values, optional
        The stopping boolean mask, true where the end condition of the event is fulfilled.
        Defaults to the opposite of ``condition``.
    window_stop : int
        The number of consecutive True values in ``condition_stop`` for an event to end.
        Defaults to 1.
    data : DataArray, optional
        The actual data. If present, its sum within each event is added to the output.
    freq : str, optional
        A frequency to divide the data into periods. If absent, the output has not time dimension.
        If given, the events are searched within in each resample period independently.
```

Zeitsperre (approved):
Some minor modifs, but looks good from my perspective

PR #1787, opened 2024-06-18
dOTC

LamAdr:
### Pull Request Checklist:
    - This PR fixes #xyz

* Introduce [dOTC](https://hess.copernicus.org/articles/23/773/2019/hess-23-773-2019.pdf) bias correction method into sdba.

No

### Other information:
Am I doing this right?


Zeitsperre:
> Am I doing this right?

You are!

Once you address the errors brought up by `pre-commit`, the tests should run. From what I can tell, there are a few programming errors and some style problems. To make things easier, try installing `pre-commit` beforehand to have the checks run when committing:
```shell
$ pip (or conda) install pre-commit
$ pre-commit install
```

LamAdr:
NB : I added "ot" to `codespell`'s ignore list

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1208,80 @@ def _adjust(
...
+                cov_factor=cov_factor,
+            ),
+            join="outer",
+            input_core_dims=[
+                ["time_cal", "multivar"],
+                ["time_cal", "multivar"],
+                ["time_tgt", "multivar"],
+            ],
+            output_core_dims=[["time_tgt", "multivar"]],
+            keep_attrs=True,
```
coxipi, line 1278: Which `attrs` are kept?  We should keep those of `sim`. Same remark with `hist` for OTC

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1208,80 @@ def _adjust(
...
+        cls,
+        ref,
+        hist,
+        bin_width=None,
+        bin_origin=None,
+        numItermax=100_000_000,
+    ):
+
+        out = xr.apply_ufunc(
+            otc_adjust,
```
coxipi, line 1225: minor nitpick, but I think I would have this apply_ufunc call in `_adjustment`, where the current `otc_adjust` is renamed as `_otc_adjust`

```python
# adjustment.py
# class OTC 
# _adjust
out = otc_adjust(ref, hist, kwargs = ....)

# _adjustment.py
# otc_adjust
out = xr.apply_ufunc(_otc_adjust, ref, hist, kwargs=..., input_core_dims=...)
```

Currently, `qdm_adjust, dqm_adjust`  etc in "_adjustment.py` take as input `xr.DataArray`, so I think it would be good to keep this structure. Eventually, when you add map_blocks, it will be easier for you if you follow more closely the current structure, you can just use other functions as template.

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,157 @@ def extremes_adjust(
...
+    """Optimal Transport Correction of the bias of X with respect to Y.
+
+    Parameters
+    ----------
+    Y : np.ndarray
+        Bias correction reference, target of optimal transport.
+    X : np.ndarray
+        Historical data to be corrected.
+    src : np.ndarray | None
+        If not None, X is still used to compute the optimal transport, but it is used to correct src.
```
coxipi, line 637: Am I understanding correctly: `src` is the same idea as having a `X1` in `dOTC`, but in this case, we consider `X` and `src` are in the same time period, so we don't apply special procedures as in dOTC?

coxipi:
Very nice! The code is very readable, well organized, good job! 

In one of my comment where I mention a possible function `_otc_adjust`, you can also see I mention `map_blocks`. Basically, your are currently doing the correction (for OTC): 
```
` tasmax_1950-1980, pr_1950-1980 -> [pr_1950-1980]', [pr_1950-1980]' ` 
```
where the full 30 years of climatology are used in training/adjusting. 

It is often useful to have different time grouping, e.g. put all January 1rst together and correct them apart, put all Jan. 2nd together, correct them separately, etc. 
```
` tasmax_01-01-19XY, pr_01-01-19XY -> [pr_01-01-19XY]', [pr_01-01-19XY]' ` 
` tasmax_02-01-19XY, pr_02-01-19XY -> [pr_02-01-19XY]', [pr_02-01-19XY]' ` 
```

If you look at `class EmpiricalQuantileMapping`, you will see that `eqm_train` can receive the argument "group". To reproduce the case you were doing, you set:
```
group = "time"
```
to have the second example I'm describing, you would set: 
```
group = "time.dayofyear"
```
this tells to the code, group data in days of year, i.e. Jan.1, Jan.2, Jan.3, etc.

xclim/sdba/_adjustment.py
LamAdr: This is used by dOTC at
```
    # Map hist to sim
    yX1 = otc_adjust(
        X1,
        X0,
        src=yX0,
        bin_width=bin_width,
        bin_origin=bin_origin,
        numItermax=numItermax,
    )
```
Here `yX0` is the transportation of `Y0` to `X0`, so `src` and `X0` are virtually the same. The only potential difference is the quantity of data. This comes from `SBCK`, but it might be possible to do `otc_adjust(X1, yX0)`.

coxipi: ah right, I missed that, makes sense, thanks.

xclim/sdba/utils.py
```diff
@@ -13,7 +13,9 @@
 import xarray as xr
 from boltons.funcutils import wraps
 from dask import array as dsk
+from ot import emd
```
coxipi, line 16: We should discuss at some point:
* if we want to include this method in SDBA by default, you should add `ot` as a requirement
* If we leave dOTC optional, you could take a look at SBCK's import and put conditional import for `ot`, and your functions/classes

xclim/sdba/adjustment.py
LamAdr: Thank you for taking a look @coxipi ! I believe this should be resolved now.

xclim/sdba/adjustment.py
LamAdr: I believe this is fixed as well.

xclim/sdba/adjustment.py
coxipi: Good! It makes think, there should really be information about both. Information about the original source (`hist, sim`  is the most crucial, but if we're talking about a scenario, we should also be informed about the target. This is not a point for your specific work here, just a thought I'm having, I will check how we handle this.

xclim/sdba/adjustment.py
```diff
@@ -1209,19 +1378,24 @@ def _adjust(
...
     def _adjust(
         cls,
-        ref,
-        hist,
-        bin_width=None,
-        bin_origin=None,
-        numItermax=100_000_000,
+        ref: xr.DataSet,
+        hist: xr.DataSet,
+        sim: None = None,  # TODO : ignored
```
LamAdr, line 1389: With the `MultivariateAdjust` class, I still need to list `sim` as input, otherwise `bin_width` gets set to `sim=hist`.
I'll see if I can think of something to make it look better.

coxipi: If I use this signature: 

```python
class OTC(MultivariateAdjust):
    """OTC"""

    @classmethod
    def _adjust(
        cls,
        ref: xr.DataSet,
        hist: xr.DataSet,
        sim: xr.DataSet | None = None,
        *,
        sim_is_hist: bool = True, 
        bin_width: list | None = None,
        bin_origin: list | None = None,
        numItermax: int = 100_000_000,
    ):
        if sim_is_hist == False:
            raise ValueError("OTC does not take a `sim` argument")
...
```

it seems to work. But I think we can simply use an attrs in `sim` instead, let me rewrite this

xclim/sdba/utils.py
```diff
@@ -922,3 +924,84 @@ def _skipna_correlation(data):
...
+    grid = (idx_grid + 1 / 2) * bin_width + bin_origin
+
+    return grid, mu, idx_bin
+
+
+def optimal_transport(gridX, gridY, muX, muY, numItermax):
+    """Computes the optimal transportation plan between X and Y."""
+    # Compute the distances from every X bin to every Y bin
+    C = distance.cdist(gridX, gridY)
+    C = C**2
```
coxipi, line 976:
```suggestion
    C = distance.cdist(gridX, gridY, 'sqeuclidean')
```
surely unimportant for overall performance, but why not

xclim/sdba/utils.py
```diff
@@ -922,3 +924,84 @@ def _skipna_correlation(data):
...
+
+    # Compute the optimal transportation plan
+    gamma = emd(muX, muY, C, numItermax=numItermax)
+    plan = (gamma.T / gamma.sum(axis=1)).T
+
+    return plan
+
+
+def eps_cholesky(M, nit=200):
+    """Cholesky decomposition."""
```
coxipi, line 986: Could you look at (Cannon, 2016), footnote on p.7048, there are some references for the smoothing of the Cholesky decomposition, it could be good to include those in the "References" section of the docstring.
Higham 1988; Knol and ten Berge 1989

coxipi: If it's a direct copy from SBCK giving some credit to them too could be good, I don't think there's a licensing problem it's GPLv3, @Zeitsperre any thoughts?

LamAdr: Absolutely. `u.bin_width_estimator` and `u.optimal_transport` are pretty much exact copies as well.

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,278 @@ def extremes_adjust(
...
+    Y : np.ndarray
+        Bias correction reference, target of optimal transport.
+    src : np.ndarray | None
+        If not None, X is still used to compute the optimal transport, but it is used to correct src.
+    bin_width : np.ndarray | None
+        Bin widths for all dimensions.
+    bin_origin : np.ndarray | None
+        Bin origins for all dimensions.
+    numItermax : int | None
+        Number of iterations of the earth mover distance algorithm.
```
coxipi, line 643:
```suggestion
        Maximum number of iterations used in the earth mover distance algorithm.
```

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,278 @@ def extremes_adjust(
...
+    return Z1
+
+
+@map_groups(scen=[Grouper.DIM])
+def dotc_adjust(
+    ds: xr.Dataset,
+    dim: str,
+    bin_width: np.ndarray | None = None,
+    bin_origin: np.ndarray | None = None,
+    numItermax: int | None = 100_000_000,
```
coxipi, line 834: We usually use snake_case convention
```suggestion
    num_iter_max: int | None = 100_000_000,
```
Also @aulemahal do we have other instances where we can give the power to users to control how long we should try to converge an algorithm? I don't think so.  I guess this could be compared to the number of rotations in MBCn, we could in principle set a target error value that we want to reach for convergence with a maximal number of rotations in MBCn. I think it's OK to give this control to users, thoughts?

xclim/sdba/adjustment.py
```diff
@@ -291,6 +295,10 @@ def adjust(
         xr.Dataset
             The bias-adjusted Dataset.
         """
+        if sim is None:
+            kwargs["sim_is_None"] = True
+            sim = hist.copy()
+
```
LamAdr, line 301: I used here the same trick used before in `MultivariateAdjust`. That's really not ideal but as the classes are defined right now I think this is our best choice.

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,289 @@ def extremes_adjust(
...
+        Adjusted data
+
+    References
+    ----------
+    :cite:cts:`sdba-robin_2021`
+    """
+    # Remove NaNs
+    Y0 = Y0[~np.isnan(Y0).any(axis=1), :]
+    X0 = X0[~np.isnan(X0).any(axis=1), :]
+    X1 = X1[~np.isnan(X1).any(axis=1), :]
```
SarahG-579462, line 792: Est-ce que c'est un problème que, après cette étape, Y0, X0, et X1 peuvent avoir des grandeurs différentes? ça l'air que tu les concatines après, donc probablement. Une façon de régler c'est de prendre l'intersection des indices non-NaN pour les données, mais ça diminuera la quantité de données. Quelquechose comme ça, peut-être?

```suggestion
    nonnan_ind = ~np.isnan(Y0).any(axis=1) &&  ~np.isnan(X0).any(axis=1) &  ~np.isnan(X1).any(axis=1)
    Y0 = Y0[nonnan_ind, :]
    X0 = X0[nonnan_ind, :]
    X1 = X1[nonnan_ind, :]
```

LamAdr: Je ne crois pas que ce soit un problème, les histogrammes sont standardisés de façon à sommer à 1. Je ne crois pas concaténer les données non plus, peux-tu spécifier où dans le code?

SarahG-579462: Ah, je voyais `[Y0, X0, X1]` plus bas, mais c'est pas une concatination, juste un array de np.array

xclim/sdba/adjustment.py
coxipi: Right, I think I have something a bit more elegant that will not require an extra input. This would mean changing the Adjust class so that `sim` is optional, and then:
```
# class Train
if sim is None:
   sim = hist.copy() 
   sim.attrs["_is_hist"] = True
...
# class DOTC
def _adjust(ref,hist, sim): 
   if "_is_hist" not in sim.attrs: 
       # throw an error
```

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,335 @@ def extremes_adjust(
...
+        spray_bins=False,
+    )
+
+    # Temporal evolution
+    motion = np.empty(yX0.shape)
+    for j in range(yX0.shape[1]):
+        if kind is not None and kind[j] == "*":
+            motion[:, j] = yX1[:, j] / yX0[:, j]
+        else:
+            motion[:, j] = yX1[:, j] - yX0[:, j]
```
coxipi, line 845: `input_core_dims` was `[dim_ref, "multivar" ]` previously, now it seems 'multivar' is the first coordinate. Is there a transpose somewhere or I'm not understanding something?

xclim/sdba/utils.py
```diff
@@ -922,3 +923,110 @@ def _skipna_correlation(data):
...
+    C = distance.cdist(gridX, gridY, "sqeuclidean")
+
+    # Compute the optimal transportation plan
+    gamma = emd(muX, muY, C, numItermax=numItermax)
+    plan = (gamma.T / gamma.sum(axis=1)).T
+
+    return plan
+
+
+def eps_cholesky(M, nit=200):
```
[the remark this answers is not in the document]
LamAdr: Ouain ça fait un peu gros. Avec nit=8 ça fait eps=2.56e-07, ce qui me semble petit par contre non?

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,323 @@ def extremes_adjust(
...
+        hist,
+        ref,
+        kwargs=dict(
+            bin_width=bin_width,
+            bin_origin=bin_origin,
+            num_iter_max=num_iter_max,
+            spray_bins=spray_bins,
+        ),
+        input_core_dims=[["dim_hist", "multivar"], ["dim_ref", "multivar"]],
+        output_core_dims=[["dim_hist", "multivar"]],
```
coxipi, line 745: following the change above, I think `multivar` should not be hardcode, 
`"multivar"` -> `dim_multivariate`

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,323 @@ def extremes_adjust(
...
+        out += np.random.uniform(low=-bin_width / 2, high=bin_width / 2, size=out.shape)
+
+    return out
+
+
+@map_groups(scen=[Grouper.DIM])
+def otc_adjust(
+    ds: xr.Dataset,
+    dim: list,
+    bin_width: list | None = None,
```
coxipi, line 698:
```suggestion
    dim: list,
    dim_multivariate: str,
    bin_width: list | None = None,
```

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,323 @@ def extremes_adjust(
...
+    -----
+    Dataset variables:
+      ref : training target
+      hist : training data
+      sim : simulated data
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        The dataset containing the data.
```
coxipi, line 882:
```suggestion
    Parameters
    ----------
    ds : xr.Dataset
        Dataset variables: 
            ref : training target
            hist : training data
            sim: data to adjust
```

xclim/sdba/utils.py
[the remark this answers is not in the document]
LamAdr: Deal. Merci!

Zeitsperre:
@LamAdr 

In order to pass the linting checks, we need to do two final things:

1. `POT` should be added somewhere to the `project.optional-dependencies`. I would suggest creating a new entry called `extras` and adding it there, then adding `xclim[extras]` to the `all` recipe.
2. We need to specify to `deptry` that `POT` provides the `ot` library. This can be done by adding the following to `tool.deptry.package_module_name_map`: `"POT" = "ot"`.

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,323 @@ def extremes_adjust(
...
+    Dataset variables:
+      ref : training target
+      hist : training data
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        The dataset containing the data.
+    dim : list
+        The dimensions that are not variables.
```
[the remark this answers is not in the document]
LamAdr: I went for `pts_dim` since it is already used by `NpdfTransform`. PrincipalComponents uses `crd_dim`.
Thanks!

Zeitsperre (commented):
I can't approve this (well beyond my expertise), but I'll set this to run the entire test suite. Well done.

LamAdr:
Thank you for your help Zeitsperre!

xclim/sdba/adjustment.py
```diff
@@ -1204,6 +1213,229 @@ def _adjust(
...
+        Will be estimated by default.
+    bin_origin : list | None
+        Bin origins for all dimensions.
+        Default is 0.
+    num_iter_max : int | None
+        Maximum number of iterations used in the earth mover distance algorithm.
+        Default is 100_000_000.
+    spray_bins : bool
+        If `False`, output points are located at the center of their bin.
+        If `True`, a random location is picked uniformly inside their bin. Default is `True`.
```
LamAdr, line 1238: @SarahG-579462 @coxipi 
Not sure about `spray_bins`. Contenders include `jitter_inside_bin` and `randomize_bin`.
Apart from that, I think it's ready.

SarahG-579462: I like jitter_inside_bins, personally

tests/test_sdba/test_adjustment.py
```diff
@@ -724,6 +726,226 @@ def test_real_data(self, open_dataset):
         new_scen.load()
 
 
+class TestOTC:
```
coxipi, line 729: Maybe add some `dask` testing (to chunk or not). See `TestQDM::test_mon_U` for example

xclim/sdba/_adjustment.py
```diff
@@ -615,3 +615,390 @@ def extremes_adjust(
...
+            if "units" not in ref.attrs["units"] or ref.attrs["units"] == "":
+                # Try to force units
+                units = str2pint(thresh).units
+                ref.attrs.update(units=str(units))
+            hist.loc[var] = _adapt_freq_hist(
+                xr.Dataset(
+                    {"ref": ref.sel({pts_dim: var}), "hist": hist.sel({pts_dim: var})}
+                ),
+                thresh,
+            )
```
coxipi, line 751: Forcing units could lead to errors (say you give `pr` in `mm/d` but the thresh is in `mm/s`). 

Adapted from my PR MBCn:

```python
# put this in adjustment.py
for var, thresh in adapt_freq_thresh.items():
    adapt_freq_thresh[var] = str(convert_units_to(thresh, self.train_units[var], context="hydro"))

# in _adjustment.py
for var, thresh in adapt_freq_thresh.items():
    hist.loc[var] = _adapt_freq_hist(
        xr.Dataset(
            {"ref": ref.sel({pts_dim: var}), "hist": hist.sel({pts_dim: var})}
        ),
        thresh,
    )
```
I can think I can push the branch MBCn today even if can get rid of some `tox` bugs that prevent me from merging, this will be necessary to have access to `train_units`

coxipi: The branch MBCn has been merged

LamAdr: Thanks! Works like a charm

xclim/sdba/adjustment.py
```diff
@@ -1267,6 +1276,243 @@ def _adjust(
...
+
+        scen = otc_adjust(
+            xr.Dataset({"ref": ref, "hist": hist}),
+            bin_width=bin_width,
+            bin_origin=bin_origin,
+            num_iter_max=num_iter_max,
+            jitter_inside_bins=jitter_inside_bins,
+            group=group,
+            pts_dim=pts_dim,
+        ).scen
```
coxipi, line 1378: `adapt_freq_thresh` missing

xclim/sdba/_adjustment.py
```diff
@@ -944,3 +944,378 @@ def extremes_adjust(
...
+
+    # Apply a variance dependent rescaling factor
+    if cov_factor == "cholesky":
+        fact0 = u.eps_cholesky(np.cov(Y0, rowvar=False))
+        fact1 = u.eps_cholesky(np.cov(X0, rowvar=False))
+        motion = motion @ fact0 @ np.linalg.inv(fact1)
+    elif cov_factor == "std":
+        fact0 = np.std(Y0, axis=0)
+        fact1 = np.std(X0, axis=0)
+        motion = motion @ np.diag(fact0 / fact1)
```
coxipi, line 1190: currently if `cov_factor` is not in {std, cholesky}, the algorithm silently applies no correction to the motion. 
1) If no correction to the motion is not intended, I think it should be an option (but it can just allow not be implemented since it's quite different from SBCK's implementation)
2) There should be an error if `cov_factor` is not in the proposed choices.
2)

LamAdr: Is that check satisfying?
```
if cov_factor not in ["identity", "std", "cholesky"]:
    raise ValueError(
        "`cov_factor` should be in ['identity', 'std', 'cholesky']."
    )
```

LamAdr: SBCK does have this option. Not sure of the use case, but since "cholesky" does not work with multiplicative kinds and "std" is univariate, maybe if something purely multivariate is needed?
Just thinking out loud, I'm open to remove the option if you think it would be useless.

LamAdr: Also maybe the default should be "cholesky"? SBCK defaults to "std".

coxipi: I don't have a strong opinion on it, we can keep it in as `None` since it's also in SBCK. I think your choice of `std` as default is fine. What I like about this is that you can set `kind={"pr":"*"}` without an error by defaulting with `std`.  Perhaps you can mention the option cholesky in the notebook?

xclim/sdba/_adjustment.py
```diff
@@ -944,3 +944,378 @@ def extremes_adjust(
...
+        if kind is not None and j in kind.keys() and kind[j] == "*":
+            motion[:, j] = yX1[:, j] / yX0[:, j]
+        else:
+            motion[:, j] = yX1[:, j] - yX0[:, j]
+
+    # Apply a variance dependent rescaling factor
+    if cov_factor == "cholesky":
+        fact0 = u.eps_cholesky(np.cov(Y0, rowvar=False))
+        fact1 = u.eps_cholesky(np.cov(X0, rowvar=False))
+        motion = motion @ fact0 @ np.linalg.inv(fact1)
```
coxipi, line 1186:
```suggestion
        motion = motion @ np.linalg.inv(fact1) @ fact0
```
no? matrices don't commute generally, and I believe we first need to remove the standard deviation, then add the one from the reference

Zeitsperre: Sidebar: What am I looking at here?? What does `@` do??

SarahG-579462: [matrix multiplication shorthand in python](https://peps.python.org/pep-0465/)

LamAdr: `@` is matrix multiplication. I thought it is simpler than SBCK's `apply_along_axis` but maybe not.
`fact0 @ np.linalg.inv(fact1) @ motion` is probably the best way to write it.

coxipi: yes that's another point, I find this tricky. `fact0, fact1` have shape `len(multivar), len(multivar)`, but they're not symmetric, is that right? so `fact0 @ np.linalg.inv(fact1) @ motion.T` and `fact0.T @ np.linalg.inv(fact1).T @ motion.T` will both work but only one is correct

xclim/sdba/adjustment.py
```diff
@@ -1267,6 +1276,247 @@ def _adjust(
...
+        - :math:`Y1_i` is the correction of the future simulated data mapped to :math:`i`.
+
+    Note that `POT <https://pythonot.github.io/>`__ must be installed to use this method.
+
+    This implementation is strongly inspired by :cite:t:`sdba-robin_2021`.
+
+    References
+    ----------
+    :cite:cts:`sdba-robin_2019,sdba-robin_2021`
+    """
```
coxipi, line 1461: A section "Notes" that describes the differences relative to SBCK would be good

docs/notebooks/sdba.ipynb
SarahG-579462, line 1: LGTM

xclim/sdba/_adjustment.py
```diff
@@ -944,3 +944,450 @@ def extremes_adjust(
...
+        Bin widths for specified dimensions.
+    bin_origin : dict | float | np.ndarray | None
+        Bin origins for specified dimensions.
+    num_iter_max : int | None
+        Maximum number of iterations used in the earth mover distance algorithm.
+    jitter_inside_bins : bool = True
+        If `False`, output points are located at the center of their bin.
+        If `True`, a random location is picked uniformly inside their bin. Default is `True`.
+    transform : str | None = "max_distance"
+        Per-variable transformation applied before the distances are calculated.
```
SarahG-579462, line 976: Peut-être ajouter des détails sur quels choix sont disponible? Je vois qu'ils sont dans les notes plus bas, mais juste une liste de chaque option serait bien pour les docs

xclim/sdba/adjustment.py
```diff
@@ -1267,6 +1276,334 @@ def _adjust(
...
+        .. math::
+
+            i_v' = \frac{i_v}{max\{i_v\}} \quad\quad\quad j_v' = \frac{j_v}{max\{j_v\}}
+
+    for variables :math:`v, w`. Default is `'max_distance'`.
+
+    Note that `POT <https://pythonot.github.io/>`__ must be installed to use this method.
+
+    This implementation is strongly inspired by :cite:t:`sdba-robin_2021`.
+    The differences from this implementation are :
```
SarahG-579462, line 1365: :+1:

SarahG-579462 (approved):
Looks good!

xclim/sdba/adjustment.py
Zeitsperre, line 1: Docstrings here also need to be updated.

xclim/sdba/_adjustment.py
```diff
@@ -944,3 +944,450 @@ def extremes_adjust(
...
+    X : np.ndarray
+        Historical data to be corrected.
+    Y : np.ndarray
+        Bias correction reference, target of optimal transport.
+    bin_width : dict | float | np.ndarray | None
+        Bin widths for specified dimensions.
+    bin_origin : dict | float | np.ndarray | None
+        Bin origins for specified dimensions.
+    num_iter_max : int | None
+        Maximum number of iterations used in the earth mover distance algorithm.
```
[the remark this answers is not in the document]
LamAdr: This is only for the docstring right? or in the actual signature too?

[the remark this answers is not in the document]
Zeitsperre: Just for the doctsrings, yes. There's more information in the spec document: https://numpydoc.readthedocs.io/en/latest/format.html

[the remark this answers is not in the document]
LamAdr: I think I got them all

LamAdr:
@Zeitsperre 
I guess the macos fails because SBCK isn't installed.
Do you understand why `Python3.9 (ubuntu-latest, conda)` fails?

Zeitsperre:
@LamAdr No worries, there's some behaviour change in the latest version of `dask` and it's impacting both conda and macOS builds right now. The fix is already in #1870 and will be merged soon.

Zeitsperre:
> @Zeitsperre I guess the macos fails because SBCK isn't installed.

Indeed. You need to add another `pytest.importorskip` on the affected tests. SBCK is only installed in two builds on CI.

Zeitsperre:
@LamAdr Is this pull request ready? If so, it'll be the last one before we tag a new release (ideally, tomorrow).

Let me know if you want more time to do any finishing touches!

LamAdr:
@Zeitsperre Yep, ready.
Thanks again

PR #1796, opened 2024-06-25
Growing Season operators

### LINKED ISSUE #1794: Ability to choose operator for Indicators growing_season_start/end
### Addressing a Problem?

It would be nice to be able to choose which operator to use for the indicators` growing_season_start/growing_season_end`. This is possible for the `growing_season_length` indicator. However, I find it a bit confusing if the default operator ">=" corresponds to the definition under **Notes** with `Tmean > 5 degC and Tmean < 5 degC` or if the default behaviour is ` Tmean >= 5 degC and Tmean <= 5 degC`.

### Potential Solution

_No response_

### Additional context

_No response_

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1794

* Adds the `op` keyword to `growing_season_{start | end}` indices and indicators, allowing for customizable threshold operators using `indices.generic.compare()`.
* Added the `op` operator to the indicator docstrings in English and French.
* Clarifies a docstring typo in `xclim.indices.growing_season_length`

Not really. The call signature has changed, but default behaviour is unchanged.


tests/test_indicators.py
```diff
@@ -488,6 +488,9 @@ def test_all_parameters_understood(official_indicators):
     if problems - {
         ("COOL_NIGHT_INDEX", "lat"),
         ("DRYNESS_INDEX", "lat"),
+        # TODO: How should we handle the case of Literal[str]?
+        ("GROWING_SEASON_END", "op"),
+        ("GROWING_SEASON_START", "op"),
```
aulemahal, line 493: I like the idea of `Literal`. 

I think we could simply map it to xclim's STRING ? Unless there's a simple way to detect if the allowed values' type, my guess is that we don't really need Literals for anything else than strings. So here : https://github.com/Ouranosinc/xclim/blob/98c251cf59f5e9c0e1573593342445e08ea9c957/xclim/core/utils.py#L659-L660

And if we do that, maybe we could apply the same for all `op` and `method`  arguments throughout the library ? And maybe use the annotation to parse the `choices` property of `xclim.core.indicator.Parameter` ?

aulemahal (approved):
Nice thanks!

My comment can be addressed in a subsequent PR.

PR #1804, opened 2024-06-26
Fix for Fahrenheit deltas, degree-days and others.

### LINKED ISSUE #1789: Degree day calculation unit handling errors
### Setup Information

- Xclim version: 0.47.0
- Python version:3.11
- Operating System:limux

### Description

Calculation of growing degree days using different units does not yield the same results : 

I believe that in `cumulative_difference` https://github.com/Ouranosinc/xclim/blob/a1a1a25550218330ea36ed428c058275c11579e2/xclim/indices/generic.py#L862  we should maybe convert the data to the threshold units instead of the opposite
doing this `threshold = convert_units_to(threshold, data)`  means that output units should be set to `'F days'` not '`K days`' that I am getting in my output

### Steps To Reproduce

```
import xarray as xr



from clisops.core import subset
import xclim

ds = xr.tutorial.load_dataset("air_temperature", chunks={})
ds = ds.rename({'air':'tas'}).resample(time='D').mean(dim='time')
ds['tas_F'] = xclim.units.convert_units_to(ds.tas, "degF")
# Single point for the example
ds_pt = ds.sel(lon=-73, lat=44, method='nearest')

out1 = xclim.atmos.growing_degree_days(tas=ds_pt.tas, thresh="5 degC", freq="MS")

out2 = xclim.atmos.growing_degree_days(tas=ds_pt.tas_F, thresh="5 degC", freq="MS")

out3 = xclim.atmos.growing_degree_days(tas=ds_pt.tas_F, thresh="278.15 K", freq="MS")

# Plot figure & tadaa! Everything is the same
fig, ax = plt.subplots(figsize=(12, 5))
fig = out1.plot(label="K and degC", linestyle="-", ax=ax)
fig = out2.plot(label="degF and degC", marker="s", markersize=10, linestyle="none", ax=ax)
fig = out3.plot(label="degF and K", marker="o", linestyle="none", ax=ax)
fig = ax.legend()
ax.set_title("Different input units");

```
image

### Additional context

_No response_

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1789

* New function `ensure_absolute_temperature` that converts temperature units to their absolute counterpart _assuming_ they represented a delta. So °C becomes K and °F becomes °R. Anything else goes unmodified (including compound units like `°C m` (whatever that means)).
* Use this function in `to_agg_units` to ensure a correct output for `op` "std", "var" and "integral".

No

### Other information:
Pint has "delta_degC" and "delta_degF" for temperature differences, but CF does not. So this new function helps in the specific case where we know the data is delta-like, which happens for the given `op`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bo
 
 Bug fixes
 ^^^^^^^^^
-* Units of degree-days computations with Fahrenheit input fixed to yield "°R d". Added a new ``core.units.ensure_absolute_temperature``  method to convert from delta to absolute temperatures (:issue:`1789`, :pull:`1804`).
+* Units of degree-days computations with Fahrenheit input fixed to yield "°R d". Added a new ``xclim.core.units.ensure_absolute_temperature`` method to convert from delta to absolute temperatures. (:issue:`1789`, :pull:`1804`).
 
 Internal changes
 ^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,11 @@ Changelog
 
 v0.51.0 (unreleased)
 --------------------
-Contributors to this version: Trevor James Smith (:user:`Zeitsperre`).
+Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Units of degree-days computations with Fahrenheit input fixed to yield "°R d". Added a new ``core.units.ensure_absolute_temperature``  method to convert from delta to absolute temperatures (:issue:`1789`, :pull:`1804`).
 
 Internal changes
 ^^^^^^^^^^^^^^^^
--- a/xclim/core/units.py
+++ b/xclim/core/units.py
@@ -463,6 +463,25 @@ def infer_sampling_units(
     return out
 
 
+DELTA_ABSOLUTE_TEMP = {
+    units.delta_degC: units.kelvin,
+    units.delta_degF: units.rankine,
+}
+
+
+def ensure_absolute_temperature(units: str):
+    """Convert temperature units to their absolute counterpart, assuming they represented a difference (delta).
+
+    Celsius becomes Kelvin, Fahrenheit becomes Rankine. Does nothing for other units.
+    """
+    a = str2pint(units)
+    # ensure a delta pint unit
+    a = a - 0 * a
+    if a.units in DELTA_ABSOLUTE_TEMP:
+        return pint2cfunits(DELTA_ABSOLUTE_TEMP[a.units])
+    return units
+
+
 def to_agg_units(
     out: xr.DataArray, orig: xr.DataArray, op: str, dim: str = "time"
 ) -> xr.DataArray:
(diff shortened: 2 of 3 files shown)


xclim/core/units.py
```diff
@@ -463,6 +463,25 @@ def infer_sampling_units(
...
+
+
+def ensure_absolute_temperature(units: str):
+    """Convert temperature units to their absolute counterpart, assuming they represented a difference (delta).
+
+    Celsius becomes Kelvin, Fahrenheit becomes Rankine. Does nothing for other units.
+    """
+    a = str2pint(units)
+    # ensure a delta pint unit
+    a = a - 0 * a
```
VascoSch92, line 479: I'm sorry, but this is no always `a` ?

aulemahal: Yes and no. The magnitude is always exactly the same, but the units can be different. 

Pint handles non-multiplicative (or _non-absolute_, or _offset_) units differently. A temperature substraction cannot return a temperature, instead is returns a "delta temperature" in pint. Ex: `1 degree_Celsius - 0 degree_Celsius = 1 delta_degree_celsius`.  This PR takes the concept one step further by converting the "delta temperature" in its absolute equivalent, since CF does not recognize delta temperatures.

See https://pint.readthedocs.io/en/stable/user/nonmult.html

VascoSch92: ah ok I see... thanks for the explanation :-)

CHANGES.rst
```diff
@@ -4,7 +4,11 @@ Changelog
 
 v0.51.0 (unreleased)
 --------------------
-Contributors to this version: Trevor James Smith (:user:`Zeitsperre`).
+Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Units of degree-days computations with Fahrenheit input fixed to yield "°R d". Added a new ``core.units.ensure_absolute_temperature``  method to convert from delta to absolute temperatures (:issue:`1789`, :pull:`1804`).
```
Zeitsperre, line 11:
```suggestion
* Units of degree-days computations with Fahrenheit input fixed to yield "°R d". Added a new ``xclim.core.units.ensure_absolute_temperature`` method to convert from delta to absolute temperatures. (:issue:`1789`, :pull:`1804`).
```

Zeitsperre (approved):
Elegant. LGTM.

tlogan2000 (approved):

xclim/core/units.py
```diff
@@ -518,19 +537,24 @@ def to_agg_units(
...
     >>> degdays = convert_units_to(degdays, "K days")
     >>> degdays.units
     'K d'
     """
-    if op in ["amin", "min", "amax", "max", "mean", "std", "sum"]:
+    if op in ["amin", "min", "amax", "max", "mean", "sum"]:
         out.attrs["units"] = orig.attrs["units"]
 
+    elif op in ["std"]:
+        out.attrs["units"] = ensure_absolute_temperature(orig.attrs["units"])
```
bzah, line 552: According to a recent change in CF (CF 1.11) you should also add an attribute `units_metadata="temperature: difference";` if you know that you are witting a difference of temperatures
See: 
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#use-of-units-metadata-ex

Zeitsperre: @aulemahal Do you want to address that here, or in another PR?

aulemahal: I think this goes in another PR.  OR rather it would be an alternative to this PR. Here we avoid the case `units_metadata="tempature: difference"` by converting to the absolute unit.

Zeitsperre: @bzah Would you mind opening a more elaborate issue for this particular feature?

bzah: Oups, I missed the replies, sorry!

@aulemahal Aren't you loosing information when converting from delta_degC to K ? 
Just looking at the units K, you can't tell if the data is on scale or a difference.

I think it would be complementary to this PR to add the `units_metadata="temperature: difference";` attribute.
Because then, you have both the workaround you wanted for pint delta units that are not understood by CF and also you have standardized metadata for people how would use the product itself.

aulemahal: I guess you're right. From my reading, I thought the `units_metadata` attribute only applied to offset units like degC, not to Kelvins. 
In most of the cases, though, I feel a human would infer the information from context : from the long name and other metadata fields. My fix was for the machines mainly.

PR #1806, opened 2024-06-26
Add Deptry, split `dev` into `dev` and `docs` recipes, address bandit security issues 

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds the `deptry` package to the linters for performing dependency analyses (unused or missing)
* Pinned a few linting dependencies
* Splits the `dev` dependencies into `dev` and `docs` (both can be installed with `all`)
* Enables `flake8-bandit` like checks in `ruff`, fixes a few security-related checks
* `xclim.testing.utils` now have more secure URL auditing checks

Yes, a new dependency has been added (`deptry`).

Also, `xclim` no longer bundles the documentation dependencies with the `dev` recipe. Splitting this reduces the number of dependencies needed to run the tests and also helps us know which dependencies are needed and where (determined with the aid of `deptry`).

### Other information:

https://deptry.com/
https://docs.astral.sh/ruff/rules/#flake8-bandit-s


xclim/core/units.py
```diff
@@ -349,10 +349,8 @@ def convert_units_to(  # noqa: C901
                         # The new cf standard name is inserted by the converter
                         try:
                             source = _CONVERSIONS[(convname, direction)](source)
-                        except Exception:
-                            # FIXME: This is a broad exception. Bad practice.
-                            # Failing automatic conversion
-                            # It will anyway fail further down with a correct error message.
+                        except KeyError:  # noqa: S110
+                            # There is no conversion available for this standard name
```
aulemahal, line 353: The previous problematic `except Exception` also caught any exception that were raised from the call of the conversion function. Values of `_CONVERSIONS` are functions like `rate2amount` and family.

If the conversion failed, then `source_unit` stays the original units and at line 367 (365 in this PR), the conversion fails with an error message given by pint. That was the objective of having a broad `except` : let pint print the error.

I agree that this is weird and not explicit at all. The solution might be to avoid the KeyError:
```python3
if (convname, direction) in _CONVERSIONS:
    try:
        source = _CONVERSIONS[(convname, direction)](source)
    except Exception as err:
        raise ValueError("some intelligent message") from err
    else:
        source_unit = units2pint(source)
```

aulemahal (changes requested):
The meta changes look good, as far as I understand, thanks!

However, I think the change in `convert_units_to` is to strict. It would yield incomprehensible error messages to the user.

xclim/core/units.py
```diff
@@ -349,9 +349,10 @@ def convert_units_to(  # noqa: C901
                         # The new cf standard name is inserted by the converter
                         try:
                             source = _CONVERSIONS[(convname, direction)](source)
-                        except KeyError:  # noqa: S110
-                            # There is no conversion available for this standard name
-                            pass
+                        except Exception as err:
+                            raise ValueError(
+                                f"No conversion available for this standard name: {standard_name}"
```
aulemahal, line 354:
```suggestion
                                f"There is a dimensionality incompatibility between the source and the target and no CF-based conversions have been found for this standard name: {standard_name}"
```

aulemahal (approved):
A suggestion to make the error message even more explicit, but otherwise, this is good to go!

PR #1431, opened 2023-07-26
Add streamflow as a known variable

RondeauG:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
    - No.
- [x] Tests for the changes have been added (for bug fixes / features)
  - [x] (If applicable) Documentation has been added / updated (for bug fixes / features)
- [x] CHANGES.rst has been updated (with summary of main changes)
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added

### What kind of change does this PR introduce?

* Adds 'streamflow' to the list of variables known to `xclim`.

### Does this PR introduce a breaking change?

No.

### Other information:

`discharge` is already there, but no standard name exists for that variable and `streamflow` is an alternative that is likely to become the standard for https://github.com/hydrologie (xdatasets & xhydro in particular). Rather than replacing `discharge` (which is not explicitely used anywhere to my knowledge, except than to declare units), we might as well support both.

PR #1832, opened 2024-07-08
Create _streamflow_flow_indices.py

faimahsho:
streamflow_flow_indice is a submodule in the xclim indices module to calculate streamflow signatures, representing individual watershed characteristics of large river/lake basins.

### Pull Request Checklist:
    - This PR fixes #xyz

* ...This PR introduces three hydrological signatures in xclim indices module. The functions included are to calculate the high and low flow variables of a data series of watersheds and generate a characteristic value of the catchment from the calculated hydrological signatures.

This can develop unique watershed attributes.


huard (commented):
If the flow_index is not in the literature, I suggest to remove it from this PR and keep it in your project scripts. 

Once this PR is merged in, I think it would be interesting to look at additional indices that requires other variables like precipitation, as I think it would provide better physical understanding of processes.

The next step will be to write a test for each index to make sure it behaves as expected. You'll see plenty of examples in the test directory.

.DS_Store
Zeitsperre, line 1: Remove this file and add `.DS_Store` to the `.gitignore`. I'm surprised it wasn't already there!

xclim/indices/_streamflow_flow_indices.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import annotations
+
+
+from xclim.core.units import declare_units
+
+
+declare_units(q="[discharge]")
```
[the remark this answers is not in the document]
Zeitsperre: Also, this is a decorator, so it should be placed directly above the `flow_index`.

xclim/indices/_streamflow_flow_indices.py
```diff
@@ -0,0 +1,126 @@
...
+    Parameters
+    ----------
+    q : xarray.DataArray
+        Daily streamflow data.
+    p is the percentile for calculating the flow index, specified as a float between 0 and 1, default is 0.95.
+
+    Returns
+    -------
+    xarray.DataArray
+        out = Normalized Qp, which is the p th percentile of daily streamflow normalized by the median flow.
```
Zeitsperre, line 27: The `out =` isn't needed since it's the only thing returned from the function.

xclim/indices/_streamflow_flow_indices.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@declare_units(q="[discharge]")
+def flow_index(q: xr.DataArray, p: float = 0.95) -> xr.DataArray:
+    """
+    Calculate the Qp  (pth percentile of daily streamflow) normalized by the mean flow.
+
+    Reference:
+    1. Addor, Nans & Nearing, Grey & Prieto, Cristina & Newman, A. & Le Vine, Nataliya & Clark, Martyn. (2018). A Ranking of Hydrological Signatures Based on Their Predictability in Space. Water Resources Research. 10.1029/2018WR022606.
+    2. Clausen, B., & Biggs, B. J. F. (2000). Flow variables for ecological studies in temperate streams: Groupings based on covariance. Journal of Hydrology, 237(3–4), 184–197. https://doi.org/10.1016/S0022-1694(00)00306-1
```
Zeitsperre, line 13: For references, we try to make use of `sphinxcontrib-bibtext`. This involves adding BibTeX entries to https://github.com/Ouranosinc/xclim/blob/main/docs/references.bib and citing them using the `cite:cts:`authornameyear` operator.

docs/references.bib
```diff
@@ -2152,3 +2152,35 @@ @article{droogers2002
...
+@article{article,
+    author = {Olden, Julian and Poff, N.},
+    year = {2003},
+    month = {03},
+    pages = {101 - 121},
+    title = {Redundancy and the Choice of Hydrologic Indices for Characterizing Stream Flow Regimes},
+    volume = {19},
+    journal = {River Research and Applications},
+    doi = {10.1002/rra.700}
+}
```
Zeitsperre, line 2186: The article codes (`article`) need to be unique entries here, e.g. `addor2018`, `clausen2000`, `olden2003`.

xclim/indices/_hydrology.py
```diff
@@ -2,11 +2,12 @@
 from __future__ import annotations
 
 import numpy as np
-import xarray
+import xarray as xr
```
Zeitsperre, line 5: This is fine. Don't forget to find and replace all instances of `xarray`

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,135 @@ def melt_and_precip_max(
...
+E           Arrays are not almost equal to 7 decimals
+E            ACTUAL: <xarray.DataArray 'q' ()> Size: 8B
+E           array(5.)
+E           Coordinates:...
+E            DESIRED: 0.08202247191011236
+
+/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/contextlib.py:81: AssertionError
+
+Process finished with exit code 1
+ calculating the flow index, between 0 and 1. Default of 0.95 is for high flows.
```
Zeitsperre: ?

docs/references.bib
```diff
@@ -2152,3 +2152,35 @@ @article{droogers2002
 	url = {https://www.scopus.com/inward/record.uri?eid=2-s2.0-0036464359&doi=10.1023%2fA%3a1015508322413&partnerID=40&md5=7322aaa4c6874878f5b1dab3c73c1718},
 	type = {Article}
 }
+
+@article{addor2018,
```
huard, line 2156: Good ! Now please use cite commands in the docstrings to refer to those. You'll see an example of how it's done in indices/_synoptic.py.

For example,
```
:cite:p:`addor2018`
```
to add a reference in parenthesis in the text (use `:cite:t:` for textual citations without parentheses), and 
```
:cite:cts:`addor2018`
```
 to add the citation in the References section.

huard (approved):
Looks good. Please review the changes I just pushed, and add tests for the "indicators" I created. 

Note that I added an indexer argument, which lets you pick a season or specific months. Let me know if you believe this is useful or not. 

I think the next step would be to have Louise review the finalized PR.

xclim/data/fr.json
```diff
@@ -1367,5 +1367,23 @@
...
+  },
+  "FLOW_INDEX": {
+    "long_name": "Indice de débit",
+    "description": "Le {p}e percentile du débit normalisé par le débit médian ({indexer}).",
+    "title": "Indice de débit",
+    "abstract": "Indice de débit, calculé comme le rapport entre un percentile donné du débit et la médiane."
+  },
+  "HIGH_FLOW_FREQUENCY": {
+    "long_name": "Fréquence des débits élevés",
+    "description": "Nombre de jours où le débit est {thresholf_factor} fois supérieur à la médiane.",
```
aulemahal, line 1379:
```suggestion
    "description": "Nombre de jours où le débit est {threshold_factor} fois supérieur à la médiane.",
```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,120 @@ def melt_and_precip_max(
...
+    qp = qt.quantile(p, dim="time")
+    q_median = qt.median(dim="time")
+    out = qp / q_median
+    out.attrs["units"] = "1"
+    return out
+
+
+@declare_units(q="[discharge]")
+def high_flow_frequency(
+    q: xr.DataArray, threshold_factor: int = 9, freq: str = "A-SEP", **indexer
```
aulemahal, line 326:
```suggestion
    q: xr.DataArray, threshold_factor: int = 9, freq: str = "YS-SEP", **indexer
```
Was the choice of a frequency anchored on the end date on purpose ? I think we never use that in xclim, I would suggest something anchored on the start date instead.

Also, "A" is deprecated in favor of "YE". "A-SEP" would be "YE-SEP".

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,120 @@ def melt_and_precip_max(
...
+        threshold = threshold_factor * median_flow
+
+    sel = select_time(q, **indexer)
+    out = threshold_count(sel, ">", threshold, freq=freq)
+    return to_agg_units(out, q, "count")
+
+
+@declare_units(q="[discharge]")
+def low_flow_frequency(
+    q: xr.DataArray, threshold_factor: float = 0.2, freq: str = "A-SEP", **indexer
```
aulemahal, line 368:
```suggestion
    q: xr.DataArray, threshold_factor: float = 0.2, freq: str = "YS-SEP", **indexer
```
Same comment as above.

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,120 @@ def melt_and_precip_max(
...
+        Calculated mean of low flow days per water year.
+
+    References
+    ----------
+    Olden, J. D., & Poff, N. L. (2003). Redundancy and the choice of hydrologic indices for characterizing streamflow regimes. River Research and
+    Applications, 19(2), 101–121. https://doi.org/10.1002/rra.700
+    """
+    mean_flow = q.mean(dim="time")
+    with xr.set_options(keep_attrs=True):
+        threshold = threshold_factor * mean_flow
```
aulemahal, line 399:
```suggestion
    threshold = threshold_factor * mean_flow
```
Why did you need `keep_attrs=True` here ? Usually, indices drop attributes. Indicators preserve them, but explicitly in the `Indicator` class.

xclim/indices/_streamflow_flow_indices.py
aulemahal, line 1: Is this file a duplicate?

aulemahal (changes requested):

xclim/indices/_hydrology.py
aulemahal: This issue is why the linting test is failing.

xclim/indices/_streamflow_flow_indices.py
huard: Yes, it should be removed.

xclim/indices/_hydrology.py
aulemahal: Just to be sure, "A-SEP" and "YS-SEP" are not equivalent. "A-SEP" goes from october to september while "YS-SEP" goes from september to august. My comment was about using "YS", feel free to chose the anchoring month as you feel is most appropriate for the indicator.

Also, the doc string still contains references to "A-SEP".

faimahsho: Got it. idea was to pick water year as default. So, I will keep it as A-Sep

aulemahal: Why not YS-OCT ?

faimahsho: yea, same concept. So I will change it to YS-OCT and update the docstring

.DS_Store
Zeitsperre: This file still needs to be removed, then I would add this file to the bottom of the `gitignore`:
```
# Apple
.DS_Store

```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,106 @@ def melt_and_precip_max(
...
+    out = threshold_count(sel, ">", threshold, freq=freq)
+    return to_agg_units(out, q, "count")
+
+
+@declare_units(q="[discharge]")
+def low_flow_frequency(
+    q: xr.DataArray, threshold_factor: float = 0.2, freq: str = "YS-OCT", **indexer
+) -> xr.DataArray:
+    """
+    Calculate the number of days in a given period with flows lower than a specified threshold. By default, the period is the water year starting on 1st October and ending on 30th September, as commonly defined in North America.
```
Zeitsperre, line 361:
```suggestion
    Calculate the number of days in a given period with flows lower than a specified threshold.
    
    By default, the period is the water year starting on 1st October and ending on 30th September, as commonly defined in North America.
```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,106 @@ def melt_and_precip_max(
...
+        Calculated mean of low flow days per water year.
+
+    References
+    ----------
+    :cite:cts:`Olden2003`
+    """
+    mean_flow = q.mean(dim="time")
+    threshold = threshold_factor * mean_flow
+    sel = select_time(q, **indexer)
+    out = threshold_count(q, "<", threshold, freq=freq)
```
Zeitsperre, line 387: I imagine you want this?
```suggestion
    out = threshold_count(sel, "<", threshold, freq=freq)
```

faimahsho: > I imagine you want this?

yes

xclim/indices/_hydrology.py
```diff
@@ -23,7 +27,7 @@
 
 
 @declare_units(q="[discharge]")
-def base_flow_index(q: xarray.DataArray, freq: str = "YS") -> xarray.DataArray:
+def base_flow_index(q: xr.DataArray, freq: str = "YS") -> xr.DataArray:
```
coxipi, line 30: What I will bring up was already there before the PR, but it would be good to address this while we're at it.

1. `discharge` is both a variable type defined in `units.py` (like "precipitation" & "temperature"), but also a variable name in `variables.yml` (like `pr` & `tas`). I think it should be one or the other
2. functions in `hydrology` should use something like `discharge` or `streamflow` instead of `q`, i.e. we should use a variable name defined in `variables.yml`
3. Speaking of which, `discharge` and `streamflow` in `variables.yml` seem to define the same quantity? Maybe we can remove `discharge` from `variables.yml` and keep it only as a variable type, then use `streamflow` as a variable name?

aulemahal: I found this : https://github.com/Ouranosinc/xclim/pull/1431
So it seems the issue is that there is no convention on the name. `q` is often used, but I agree that its ambiguous.

I'm not sure (1) is a showstopper. But as  `discharge` is not actually used in any indicator, we could indeed remove it from the `variables.yml`. 

@RondeauG how breaking would it be for downstream (xhydro) if we changed all the `q` inputs to  `streamflow` ? Does that make sense to you ?

RondeauG: >     3. Speaking of which, `discharge` and `streamflow` in `variables.yml` seem to define the same quantity? Maybe we can remove `discharge` from `variables.yml` and keep it only as a variable type, then use `streamflow` as a variable name?

This was because `discharge` was the legacy name used in `xclim`, but partners in `xhydro` much preferred `streamflow` (as `discharge` had a strong connotation with dams and regulated flow). 

However, I agree that `streamflow` is kinda long. `Q` is the standard in hydrology for streamflow in `m3 s-1`, and `q` for `m3 s-1 km-2`, but if it's too confusing with quantiles, we could find something else.

As to how breaking this would be for `xhydro`, this is completely fine for now. None of these indicators have been hard-coded anywhere in the package. The only explicit calls to `xclim` indicators are for  `stats` and the frequency analysis functions, both of which use `da`.

coxipi: If we sort out the situation with the variables used in the streamflow indices, then we can also address this comment here and simplify the `Streamflow` class

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,115 @@ def melt_and_precip_max(
...
+    Parameters
+    ----------
+    q : xr.DataArray
+        Daily streamflow data.
+    threshold_factor : int
+        Factor by which the median flow is multiplied to set the high flow threshold, default is 9.
+    freq : str, optional
+        Resampling frequency, default is 'YS-OCT' for water year starting in October and ending in September.
+    indexer
+        Indexing parameters to perform a temporal subset of the data.
```
aulemahal, line 342:
```suggestion
        Indexing parameters to perform a temporal subset of the data. The median flow is computed over the whole period, but the number of days only on the subset.
```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,115 @@ def melt_and_precip_max(
...
+        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
+
+    Returns
+    -------
+    xr.DataArray
+        Calculated mean of high flow days per water year
```
aulemahal, line 348:
```suggestion
        Number of high flow days
```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,115 @@ def melt_and_precip_max(
...
+    Parameters
+    ----------
+    q : xr.DataArray
+        Daily streamflow data.
+    threshold_factor : float
+        Factor by which the mean flow is multiplied to set the low flow threshold, default is 0.2.
+    freq : str, optional
+        Resampling frequency, default is 'YS-OCT' for water year starting in October and ending in September.
+    indexer
+        Indexing parameters to perform a temporal subset of the data.
```
aulemahal, line 381:
```suggestion
        Indexing parameters to perform a temporal subset of the data. The mean flow is computed over the whole period, but the number of days only on the subset.
```

xclim/indices/_hydrology.py
```diff
@@ -279,3 +283,115 @@ def melt_and_precip_max(
...
+        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
+
+    Returns
+    -------
+    xr.DataArray
+        Calculated mean of low flow days per water year.
```
aulemahal, line 387:
```suggestion
       Number of low flow days.
```

aulemahal (approved):
This looks good to me.

Here are a few "incoherences" I see, but that may well be expected.


- The high flow is computed with reference fo the _median_ while the low flow is done with reference to the _mean_. Normal ?
- The high and low flows have their reference computed over the whole period (year) and then the number of days is counted over the temporal subset. The flox index computes both the percentile and the median only over the subset.

xclim/indices/_hydrology.py
huard: The median vs mean part is ok, that's how it's done in the reference paper. 

For the subset, that's my mistake. I thought it would make more sense to compute the median and mean with respect to the entire year, but I think it would clash with how indexing is done elsewhere.

xclim/indices/_hydrology.py
```diff
@@ -345,15 +345,15 @@ def high_flow_frequency(
...
 
     References
     ----------
     :cite:cts:`addor2018,Clausen2000`
     """
-    median_flow = q.median(dim="time")
-    threshold = threshold_factor * median_flow
     sel = select_time(q, **indexer)
+    median_flow = sel.median(dim="time")
+    threshold = threshold_factor * median_flow
```
aulemahal, line 356: In that case, you could also remove the `indexer` here and simply create the indicator by inheriting from `ResamplingIndexingIndicator`.

But the result would be the same at the indicator level, so your choice.

huard: I removed indexing alltogether. I'm not sure anymore it even makes sense in those cases.

PR #1838, opened 2024-07-11
Fix wet spells - generic spell statistics

### LINKED ISSUE #1834: Wet spell max length indicator broken
### Setup Information

- Xclim version: 0.50.0
- Python version: 3.11
- Operating System: Linux

### Description

Computing the wet spell max length spell turned out to be incorrect.

### Steps To Reproduce

```
import numpy as np
import xarray as xr
import xclim as xc

a = np.zeros(366)
a[10:13] = 3
a[40:45] = 3
pr = xr.DataArray(a, dims=('time',), coords={'time': xr.date_range('2000-01-01', periods=366, freq='D')}, attrs={'units': 'mm/d'}

xc.atmos.wet_spell_max_length(pr, window=1, op='min')
```
Returns : 321.  The answer should be 5.

### Additional context

The issue is here https://github.com/Ouranosinc/xclim/blob/994a5a17e04d68d44317047f08cd90b98157ec5c/xclim/indices/_threshold.py#L3520.

It looks like the code was directly copied from `dry_spell_max_length` with all comparison operators inverted. But on that line, the comparison has to do with the number of days, so it should be ">=" instead. The current function completely inverts the run mask, so we get the longest run that is _not_ a wet spell.

I'd note that a fast path for the `window=1` case might be of interest here. Also, the documentation was not completely adapted when it was copied from `dry_spell` to `wet_spell`. For example, "max" as an "op" here makes no sense. It should be "min".

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1834

* Implements a generic spell length statistics function
* Use that function in dry and wet spells, which fixes wet spells
* That required me to implement the `reducer='count'` case in 1d `rle_statistics`.

Yes, results for all wet spells indicators will change. But they were not correct before.

Results for  `dry_spell_frequency` are also different, were also not correct previsouly.


xclim/indices/_threshold.py
```diff
@@ -3423,15 +3427,26 @@ def wet_spell_total_length(
         Accumulated precipitation value over which a period is considered dry.
     window : int
         Number of days when the maximum or accumulated precipitation is over threshold.
-    op : {"max", "sum"}
+    op : {"min", "max", "sum"}
```
coxipi, line 3430: I guess it should either allow "min" or "max" but not both, no?

xclim/indices/generic.py
```diff
@@ -352,6 +352,115 @@ def get_daily_events(
...
+        reducer=spell_reducer,
+        # The code above already ensured only spell of the minimum length are selected
+        window=1,
+        freq=freq,
+    )
+
+    if spell_reducer == "count":
+        return out.assign_attrs(units="1")
+    # All other cases are statistics of the number of timesteps
+    return to_agg_units(out, data, "count")
```
coxipi, line 461: should this be `"count"` -> `spell_reducer` ?

coxipi, line 461: should this be `"count"` -> `spell_reducer` ?

xclim/indices/generic.py
```diff
@@ -352,6 +352,115 @@ def get_daily_events(
...
+        resample_before_rl,
+        rl.rle_statistics,
+        reducer=spell_reducer,
+        # The code above already ensured only spell of the minimum length are selected
+        window=1,
+        freq=freq,
+    )
+
+    if spell_reducer == "count":
+        return out.assign_attrs(units="1")
```
coxipi, line 459: Is  "1" the new way of passing "" (dimensionless)?

xclim/indices/generic.py
```diff
@@ -352,6 +352,115 @@ def get_daily_events(
...
+        is_in_spell = compare(data, op, thresh)
+    else:
+        data_pad = data.pad(time=(0, window))
+        # The spell-wise value to test
+        # For example "win_reducer='sum'", we want the sum over the minimum spell length (window) to be above the thresh
+        spell_value = getattr(data_pad.rolling(time=window), win_reducer)()
+        # True at the end of a spell respecting the condition
+        mask = compare(spell_value, op, thresh)
+        # True for all days part of a spell that respected the condition (shift because of the two rollings)
+        is_in_spell = (mask.rolling(time=window).sum() >= 1).shift(time=-(window - 1))
```
coxipi, line 442: It should never be >1 no? There is at least a spacing of `window+1` between two end spells?

xclim/indices/generic.py
```diff
@@ -352,6 +352,115 @@ def get_daily_events(
...
+    thresh = convert_units_to(
+        threshold,
+        data,
+        context=infer_context(standard_name=data.attrs.get("standard_name")),
+    )
+
+    if window == 1:  # Fast path
+        is_in_spell = compare(data, op, thresh)
+    else:
+        data_pad = data.pad(time=(0, window))
```
coxipi, line 435: I guess all this rolling is only necessary for `window_reducer == ="sum"`, yes? If you want to include a faster algorithm for other cases, a modification of `rle` should do the trick: 

```python 
from xclim.indices.run_length import _cumsum_reset_on_zero
mask = compare(data, op,thresh)
# find sequences of consecutive respected constraints
cs_s = _cumsum_reset_on_zero(mask)
# end of these sequences
cs_s = cs_s.where(mask.shift({"time": -1}, fill_value=0) == 0)
# propagate these end of sequences
is_in_spell = (cs_s.where(cs_s>=window).where(mask>0, 0).bfill("time")>0)
```

coxipi:
Very nice to have this new generic function!

xclim/indices/_threshold.py
```diff
@@ -3346,10 +3347,11 @@ def wet_spell_frequency(
     freq: str = "YS",
     resample_before_rl: bool = True,
     op: str = "sum",
+    **indexer,
 ) -> xarray.DataArray:
     """Return the number of wet periods of n days and more.
 
-    Periods during which the accumulated or maximal daily precipitation amount on a window of n days is over threshold.
+    Periods during which the accumulated, minimal orr maximal daily precipitation amount on a window of n days is over threshold.
```
Zeitsperre, line 3354:
```suggestion
    Periods during which the accumulated, minimal, or maximal daily precipitation amount on a window of n days is over threshold.
```

xclim/indices/generic.py
```diff
@@ -352,6 +352,115 @@ def get_daily_events(
...
+        Statistic on the spell lengths.
+    freq: str
+        Resampling frequency.
+    resample_before_rl: bool
+        Determines if the resampling should take place before or after the run
+        length encoding (or a similar algorithm) is applied to runs.
+    indexer:
+        Indexing parameters to compute the indicator on a temporal subset of the data.
+        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
+        Indexing is done after finding the days part of a spell, but before taking the spell statistics.
```
Zeitsperre, line 395:
```suggestion
    r"""Statistics on spells lengths.

    A spell is when a statistic (`win_reducer`) over a minimum number (`window`) of consecutive timesteps respects a condition (`op` `thresh`).
    This returns a statistic over the spell's count or length.

    Parameters
    ----------
    data : xr.DataArray
        Input data.
    threshold : Quantified
        Threshold to test against.
    window : int
        Minimum length of a spell.
    win_reducer : {'min', 'max', 'sum', 'mean'}
        Reduction along the spell length to compute the spell value.
        Note that this does not matter when `window` is 1.
    op : {">", "gt", "<", "lt", ">=", "ge", "<=", "le", "==", "eq", "!=", "ne"}
        Logical operator. Ex: spell_value > thresh.
    spell_reducer : {'max', 'sum', 'count'}
        Statistic on the spell lengths.
    freq : str
        Resampling frequency.
    resample_before_rl : bool
        Determines if the resampling should take place before or after the run
        length encoding (or a similar algorithm) is applied to runs.
    \*\*indexer
        Indexing parameters to compute the indicator on a temporal subset of the data.
        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
        Indexing is done after finding the days part of a spell, but before taking the spell statistics.
```

xclim/indices/run_length.py
```diff
@@ -1129,7 +1129,7 @@ def statistics_run_1d(arr: Sequence[bool], reducer: str, window: int) -> int:
     ----------
     arr : Sequence[bool]
         Input array (bool)
-    reducer : {'mean', 'sum', 'min', 'max', 'std'}
+    reducer : {'mean', 'sum', 'min', 'max', 'std', 'count'}
```
Zeitsperre, line 1132:
```suggestion
    reducer : {"mean", "sum", "min", "max", "std", "count"}
```

CHANGELOG.rst
```diff
@@ -4,7 +4,11 @@ Changelog
 
 v0.52.0 (unreleased)
 --------------------
-Contributors to this version: David Huard (:user:`huard`).
+Contributors to this version: David Huard (:user:`huard`), Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Wet spell indicators fixed. New generic spell length statistic function ``xclim.indices.generic.spell_length_statistics`` used in wet and dry spells indicators. (:issue:`1834`, :pull:`1838`).
```
Zeitsperre, line 11:
```suggestion
* Addressed a bug found in ``wet_spell_*`` indicators that was contributing to erroneous results. A new generic spell length statistic function ``xclim.indices.generic.spell_length_statistics`` is now used in wet and dry spells indicators. (:issue:`1834`, :pull:`1838`).
```

Zeitsperre: Does this fix merit a Breaking Changes?

xclim/indices/_threshold.py
```diff
@@ -3153,6 +3154,7 @@ def dry_spell_frequency(
     freq: str = "YS",
     resample_before_rl: bool = True,
     op: str = "sum",
+    **indexer,
```
Zeitsperre, line 3157: An entry should be added to the docstring for `\*\*indexer`.

xclim/indices/_threshold.py
```diff
@@ -3346,10 +3347,11 @@ def wet_spell_frequency(
...
```
Zeitsperre, line 3350: ibid.

xclim/indices/_threshold.py
```diff
@@ -3423,15 +3427,26 @@ def wet_spell_total_length(
...
+        `min` means that all days within the minimum window must exceed the threshold.
+        `max` means that at least one day within the window must exceed the threshold.
+        `sum` means that the accumulated precipitation within the window must exceed the threshold.
+        In all cases, the whole window is marked a part of a wet spell.
     freq : str
         Resampling frequency.
+    resample_before_rl: bool
+        Determines if the resampling should take place before or after the run
+        length encoding (or a similar algorithm) is applied to runs.
     indexer
```
Zeitsperre, line 3441:
```suggestion
    \*\*indexer
```

xclim/indices/_threshold.py
```diff
@@ -3489,15 +3498,26 @@ def wet_spell_max_length(
...
```
Zeitsperre, line 3512:
```suggestion
    \*\*indexer
```

xclim/indices/_threshold.py
aulemahal: I mean, one could decide to define a "wet spell" with a max threshold, I'm not here to judge. I'll add the missing ones in the other functions.

xclim/indices/generic.py
aulemahal: No, because the `spell_reducer` is applied on the count of days. In all cases except `spell_reducer="count"`, we are returning a statistics on the number of days, so it is a "count" aggregation.

xclim/indices/generic.py
aulemahal: Added this as-is in the generic function. Did you see a way to generalize this futher then the two cases I saw ? See https://github.com/Ouranosinc/xclim/blob/8b75acd1cdeb2e5b989353825ae0e6ec2466b8e9/xclim/indices/generic.py#L434-L436

aulemahal: Even if it is weird, the "min" + "<" case is possible and I don't think it is feasible without both rolling. This fast path, if I understand it correctly, works when the condition on a the minimum window can only be true if it is true on each day of the window.

xclim/indices/generic.py
coxipi: Yes got it, I misread the comment above, a 'True' continuous spell can be composed of many subspells.

xclim/indices/generic.py
coxipi: What you say here is related to this comment, yes?:

> # For these two cases, a day can't be part of a spell if it doesn't respect the condition itself

or said otherwise, we must have "min" and ">"  / "max" and "<" correct? 

I agree that this fast-path is less flexible, but as you note it, it's more natural. I can see why the `op="sum"` could be interesting, but not the "min"&"<" frankly. 

Imagine you're trying to characterize wet spells, where the start condition is having 10 mm and say `window=3`.  

`pr = [0, 0, 20,20,20, 0,0] # in mm`

With the weird definition, the spell sequence is `[0,0,20,20,20]` because, why not haha?  You just end up adding `window-1` extra days no matter what, no?

aulemahal: In agree that the (min, <) and (max, >) are weird, but I just don't want to forbid a user from using it solely on this judgement.

aulemahal:
Sorry for the unrelated addition. Last commit fixes an issue I had in  `lazy_indexing` where the auxiliary coords were re-chunked when going through the `map_blocks` even though they were not modified. Further down the line, in the `resample(...).map()`, xarray would load them back for its check in the concatenation of the groups, which made the indicator not lazy.

xclim/indices/_threshold.py
```diff
@@ -3361,16 +3377,24 @@ def wet_spell_frequency(
...
-      length encoding (or a similar algorithm) is applied to runs.
-    op: {"sum","max"}
-      Operation to perform on the window.
-      Default is "sum", which checks that the sum of accumulated precipitation over the whole window is more than the
-      threshold.
-      "max" checks that the maximal daily precipitation amount within the window is more than the threshold.
-      This is the same as verifying that each individual day is above the threshold.
+        Determines if the resampling should take place before or after the run
+        length encoding (or a similar algorithm) is applied to runs.
+    op: {"sum","min", "max", "mean"}
```
Zeitsperre, line 3384:
```suggestion
    op : {"sum","min", "max", "mean"}
```

xclim/indices/_threshold.py
```diff
@@ -3230,11 +3238,18 @@ def dry_spell_total_length(
         Accumulated precipitation value under which a period is considered dry.
     window : int
         Number of days when the maximum or accumulated precipitation is under threshold.
-    op : {"max", "sum"}
-        Reduce operation.
+    op: {"sum", "max", "min", "mean"}
```
Zeitsperre, line 3241:
```suggestion
    op : {"sum", "max", "min", "mean"}
```

xclim/indices/_threshold.py
```diff
@@ -3410,27 +3432,37 @@ def wet_spell_total_length(
...
-    op : {"max", "sum"}
+    op : {"min", "sum", "max", "mean"}
         Reduce operation.
+        `min` means that all days within the minimum window must exceed the threshold.
+        `sum` means that the accumulated precipitation within the window must exceed the threshold.
+        In all cases, the whole window is marked a part of a wet spell.
     freq : str
         Resampling frequency.
-    indexer
+    resample_before_rl: bool
```
Zeitsperre, line 3455:
```suggestion
    resample_before_rl : bool
```

xclim/indices/generic.py
```diff
@@ -352,6 +352,129 @@ def get_daily_events(
...
+    ...     pram,
+    ...     thresh="20 mm",
+    ...     window=5,
+    ...     op=">=",
+    ...     win_reducer="sum",
+    ...     spell_reducer="max",
+    ...     freq="YS",
+    ... )
+
+    Here, a day is part of a spell if it is in any 5 day period where the total accumulated precipitation reaches or exceeds
```
Zeitsperre, line 423:
```suggestion
    Here, a day is part of a spell if it is in any five (5) day period where the total accumulated precipitation reaches or exceeds
```

xclim/indices/generic.py
```diff
@@ -352,6 +352,129 @@ def get_daily_events(
...
+    ...     tas,
+    ...     thresh="35 °C",
+    ...     window=7,
+    ...     op=">",
+    ...     win_reducer="min",
+    ...     spell_reducer="sum",
+    ...     freq="YS",
+    ... )
+
+    Here, a day is part of a spell if it is in any 7 day period where the minimum temperature is over 35°C. We then
```
Zeitsperre, line 409:
```suggestion
    Here, a day is part of a spell if it is in any seven (7) day period where the minimum temperature is over 35°C. We then
```

Zeitsperre (approved):
Nicely done!

PR #1845, opened 2024-07-18
Generic season index - no season is length 0

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Implements `generic.season`, a generic index function for computing the start/end/length of seasons. Based on `run_length.season` but with units management and the comparison.
* Divide `run_length.season` into `season_start`, `season_end`, plus the existing `season_length`. The idea of grouping code together was good, but it added a lot of unnecessary computations when only one facet is needed (i.e. most cases). I think I was able to divide it elegantly, using only one magic fast path.
* Rewrite `frost_free_season_start` and `frost_free_season_end`. Those were not respecting the "season" definition. Now they do.

`frost_free_season_start` and `frost_free_season_end` have changed. Their notion of "season" is not more strict. I think the difference not significative for most cases. No tests needed a change, but again our tests are very simple.

All season length indicators will now return 0 when no season is found (no start). Previously, when no start was found we returned 0 when an end was found and `nan` otherwise. I think the idea was to assume that neither end nor start meant the data was probably wrong. But that's the job of the missing checks, not the `run_length` helper.  Thus I applied the more logical rule : no season == 0 length.

I have reverted some changes of #1796, I'm sorry. I'm trying to generalize the indicators and it made more sense that all aspects of the same "growing season" were using the same exact arguments.

### Other information:
This allows indices to use run length function that return a DataArray after being called with a DataArray. This "type preservation" will be quite handy in my next PR, stay tuned!

Sadly, this branch is based off #1838 because I need both for PC.

### Ugh
The amount of copied docstring text and signatures is ridiculous. I really want to remove all those one-liner indices and only keep the indicators. But I fear this would be too breaking of a change as it would remove elements from the API.
Right now, we have inconsistent indicators with inconsistent documentation because everything was copy-pasted by hand with incremental changes that were not always ported back to the other indices.


CHANGELOG.rst
```diff
@@ -16,6 +16,11 @@ Breaking changes
 ^^^^^^^^^^^^^^^^
 * As updated in ``cf_xarray>=0.9.3``, dimensionless quantities now use the "1" units attribute as specified by the CF conventions, previously an empty string was returned. (:pull:`1814`).
 
+Breaking changes
+^^^^^^^^^^^^^^^^
+* Definition of the ``frost_free_season_start`` and ``frost_free_season_end`` have been slightly changed to be coherent with the ``frost_free_season_length`` and xclim's notion of `season` in general. Indicator and indices signature have changed. (:pull:`1845`).
```
Zeitsperre, line 21:
```suggestion
* The definition of the ``frost_free_season_start`` and ``frost_free_season_end`` have been slightly changed to be coherent with the ``frost_free_season_length`` and `xclim`'s notion of `season` in general. Indicator and indices signature have changed. (:pull:`1845`).
```

CHANGELOG.rst
```diff
@@ -26,6 +31,7 @@ Internal changes
 ^^^^^^^^^^^^^^^^
 * Changed the French translation of "wet days" from "jours mouillés" to "jours pluvieux". (:issue:`1825`, :pull:`1826`).
 * In order to adapt to changes in `pytest`, the doctest fixtures have been split from the main testing suite and doctests are now run using ``$ python -c 'from xclim.testing.utils import run_doctests; run_doctests()'``. (:pull:`1632`).
+* Added ``xclim.indices.generic.season`` to make season start, end and length indices. Added a `stat` argument to ``xclim.indices.run_length.season`` to avoid returning a dataset. (:pull:`1845`).
```
Zeitsperre, line 34: https://www.youtube.com/watch?v=P_i1xk07o4g

```suggestion
* Added ``xclim.indices.generic.season`` to make season start, end, and length indices. Added a ``stat`` argument to ``xclim.indices.run_length.season`` to avoid returning a dataset. (:pull:`1845`).
```

xclim/indices/_threshold.py
```diff
@@ -1050,21 +983,20 @@ def growing_season_end(
...
         Date of the year after which to look for the end of the season. Should have the format '%m-%d'.
     window : int
         Minimum number of days with temperature below threshold needed for evaluation.
     freq : str
         Resampling frequency.
-    op : {"<", "<=", "lt", "le"}
-        Comparison operation. Default: "<".
+    op : {">", ">=", "gt", "ge"}
+        Comparison operation. Default: ">". Note that this comparison is what defines the season.
+        The end of the season happens when the condition is NOT met for `window` consecutive days.
```
Zeitsperre, line 994: This makes much more sense. The fact that it was inverted for one release shouldn't be so bad. Behaviour is unchanged.

xclim/indices/generic.py
```diff
@@ -475,6 +477,83 @@ def spell_length_statistics(
...
+    window : int
+        Minimum number of days that the condition must be met / not met for the start / end of the season.
+    op : str
+        Comparison operation.
+    stat : {'start', 'end', 'length'}
+        Which season facet to return.
+    freq : str
+        Resampling frequency.
+    mid_date : DayOfYearStr, optional
+        An optional middle date to restrict the possible start and end of the season.
```
coxipi, line 512: here or below in a "Notes" section, it would be good to refer to `season_start/season_end`. The notion of `middle_date` is not super clear just by reading the `season` docstring

xclim/indices/_threshold.py
```diff
@@ -1003,42 +943,35 @@ def growing_season_start(
...
+        stat="start",
+    )
 
 
 @declare_units(tas="[temperature]", thresh="[temperature]")
 def growing_season_end(
     tas: xarray.DataArray,
     thresh: Quantified = "5.0 degC",
-    mid_date: DayOfYearStr = "07-01",
+    mid_date: DayOfYearStr | None = "07-01",
```
Zeitsperre, line 964: If this is an Optional field, it should be equal to `None` by default. Perhaps mention in the docstring that this can be set to `None`?

xclim/indices/_threshold.py
```diff
@@ -1078,37 +1010,35 @@ def growing_season_end(
...
-    return out
 
 
 @declare_units(tas="[temperature]", thresh="[temperature]")
 def growing_season_length(
     tas: xarray.DataArray,
     thresh: Quantified = "5.0 degC",
     window: int = 6,
-    mid_date: DayOfYearStr = "07-01",
+    mid_date: DayOfYearStr | None = "07-01",
```
Zeitsperre, line 1030: ibid.

xclim/indices/_threshold.py
```diff
@@ -978,21 +909,30 @@ def growing_degree_days(
 def growing_season_start(
     tas: xarray.DataArray,
     thresh: Quantified = "5.0 degC",
+    mid_date: DayOfYearStr | None = "07-01",
```
Zeitsperre, line 912: ibid.

xclim/indices/_threshold.py
```diff
@@ -1454,15 +1408,15 @@ def frost_free_spell_max_length(
 ) -> xarray.DataArray:
     r"""Longest cold spell.
```
Zeitsperre, line 1409:
```suggestion
    r"""Longest frost free spell.
```

xclim/indices/generic.py
```diff
@@ -475,6 +477,83 @@ def spell_length_statistics(
...
+    data: xarray.DataArray,
+    thresh: Quantified,
+    window: int,
+    op: str,
+    stat: str,
+    freq: str,
+    mid_date: DayOfYearStr | None = None,
+    constrain: Sequence[str] | None = None,
+) -> xarray.DataArray:
+    r"""Season
```
Zeitsperre, line 491:
```suggestion
    r"""Season.
```

xclim/indices/run_length.py
```diff
@@ -765,8 +867,9 @@ def season(
     """Calculate the bounds of a season along a dimension.
 
     A "season" is a run of True values that may include breaks under a given length (`window`).
-    The start is computed as the first run of `window` True values, then end as the first subsequent run
-    of  `window` False values. If a date is passed, it must be included in the season.
+    The start is computed as the first run of `window` True values, and the end as the first subsequent run
+    of  `window` False values. The end element is the first element after the season.
+    If a date is given, it must be included in the season, i.e. the start can't happen laterand the end can't happen earlier.
```
Zeitsperre, line 872:
```suggestion
    If a date is given, it must be included in the season, i.e. the start cannot occur later and the end cannot occur earlier.
```

xclim/indices/run_length.py
```diff
@@ -786,7 +889,7 @@ def season(
     Returns
     -------
     xr.Dataset
-        "dim" is reduced to "season_bnds" with 2 elements : season start and season end, both indices of da[dim].
+        A Dataset with three variables : start, end and length of the season.
```
Zeitsperre, line 892: Live by the Oxford comma, die by the Oxford comma.
```suggestion
        A Dataset with three variables : start, end, and length of the season.
```

xclim/indices/run_length.py
```diff
@@ -880,11 +967,9 @@ def season_length(
     date: DayOfYearStr | None = None,
     dim: str = "time",
 ) -> xr.DataArray:
-    """Return the length of the longest semi-consecutive run of True values (optionally including a given date).
+    """Length of a season.
 
-    A "season" is a run of True values that may include breaks under a given length (`window`).
-    The start is computed as the first run of `window` True values, then end as the first subsequent run
-    of  `window` False values. If a date is passed, it must be included in the season.
+    See :py:func:`season`.
```
Zeitsperre, line 972:
```suggestion
    Notes
    -----
    See :py:func:`season`.
```

xclim/indices/run_length.py
```diff
@@ -1053,6 +1133,54 @@ def last_run_before_date(
...
+    window : int
+        Minimum duration of consecutive run to accumulate values.
+    date : DayOfYearStr
+        The date before which to look for the run.
+    dim : str
+        Dimension along which to calculate consecutive run (default: 'time').
+    coord : Optional[Union[bool, str]]
+        If not False, the function returns values along `dim` instead of indexes.
+        If `dim` has a datetime dtype, `coord` can also be a str of the name of the
+        DateTimeAccessor object to use (ex: 'dayofyear').
```
Zeitsperre, line 1158:
```suggestion
        DateTimeAccessor object to use (e.g. 'dayofyear').
```

xclim/indices/run_length.py
```diff
@@ -1053,6 +1133,54 @@ def last_run_before_date(
...
+    ----------
+    da : xr.DataArray
+        Input N-dimensional DataArray (boolean).
+    window : int
+        Minimum duration of consecutive run to accumulate values.
+    date : DayOfYearStr
+        The date before which to look for the run.
+    dim : str
+        Dimension along which to calculate consecutive run (default: 'time').
+    coord : Optional[Union[bool, str]]
```
Zeitsperre, line 1155:
```suggestion
    coord : bool or str, optional
```

xclim/indices/run_length.py
```diff
@@ -755,6 +755,108 @@ def extract_events(
...
+    ----------
+    da : xr.DataArray
+        Input N-dimensional DataArray (boolean).
+    window : int
+        Minimum duration of consecutive values to start and end the season.
+    date : DayOfYearStr, optional
+        The date (in MM-DD format) that a run must include to be considered valid.
+    dim : str
+        Dimension along which to calculate consecutive run (default: 'time').
+    coord : Optional[str]
```
Zeitsperre, line 821:
```suggestion
    coord : str, optional
```

Zeitsperre (approved):
Looks good!

xclim/indices/generic.py
aulemahal: Done!
`mid_date` for all season functions.

Rewrote doc.

xclim/indices/_threshold.py
```diff
@@ -978,21 +909,30 @@ def growing_degree_days(
...
+    The default `freq` and `mid_date` parameters are valid for the northern hemisphere.
 
     Parameters
     ----------
     tas : xarray.DataArray
         Mean daily temperature.
     thresh : Quantified
         Threshold temperature on which to base evaluation.
+    mid_date : str, optional
+        Date of the year after before which the season must start. Should have the format '%m-%d'.
```
coxipi, line 935: > after before which

xclim/indices/_threshold.py
coxipi: I can't say for the type that would be better for this job, but I think that the logic is:

* we want the default behaviour that the season includes july 1rst
* but another possibility is to put no such constraint

So we don't want to use `None` as default? Perhaps calling it optional in this case is a bit weird, but there should be a way to allow this behaviour?

Zeitsperre: I suppose we can leave it as is; The warning about Northern Hemisphere locations is there. So long as the value `None` can be passed here, it's valid.

xclim/indices/run_length.py
```diff
@@ -786,7 +893,7 @@ def season(
     Returns
     -------
     xr.Dataset
-        "dim" is reduced to "season_bnds" with 2 elements : season start and season end, both indices of da[dim].
+        A Dataset with three variables : start, end and length of the season.
```
coxipi, line 896: I don't know why suggestion is not working.

You could follow the current convention in sdba

```python
    ds : xr.Dataset
        Dataset variables:
            start 
            end
            length
```

xclim/indices/_threshold.py
```diff
@@ -1280,143 +1213,164 @@ def frost_free_season_start(
...
 def frost_free_season_end(
     tasmin: xarray.DataArray,
     thresh: Quantified = "0.0 degC",
-    mid_date: DayOfYearStr = "07-01",
     window: int = 5,
+    mid_date: DayOfYearStr = "07-01",
+    op: str = ">=",
     freq: str = "YS",
 ) -> xarray.DataArray:
     r"""End of the frost free season.
```
coxipi, line 1262: I don't want to stir the pot more... but why not optional `mid_date` here too? is it different from `growing_degree...` ?

aulemahal: Indeed, I saw that too! It has been modified in the last commit.

xclim/indices/run_length.py
```diff
@@ -755,26 +755,133 @@ def extract_events(
...
+
+
 def season(
     da: xr.DataArray,
     window: int,
-    date: DayOfYearStr | None = None,
+    mid_date: DayOfYearStr | None = None,
     dim: str = "time",
     coord: str | bool | None = False,
 ) -> xr.Dataset:
```
coxipi, line 870: don't have much to say computation-wise, well done. I like how the fast path simplifies things in `season` (we don't mess with `no_end` now) well done

coxipi (approved):
Well done!

Zeitsperre (approved):
LGTM

PR #1848, opened 2024-07-19
Block-mapped resample with the help of flox

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Implements `resample_map`. This function is meant for all `da.resample(...).map(...)` calls. These, `flox` cannot improve automatically so we use some flox logic to help. The idea is to map the resample-map construct on each block in parallel. This is possible by first rechunking the array so that chunks boundary fit with resampling period boundaries (this is a flox function).

The main improvement should come from the fact that `map_blocks` hides much of the complexity to `dask`, so the resulting graph is much lighter.  I still have to better test the performance of this. My goal would be to have some short text in xclim's doc that highlights when the option is useful and when it is not. The option is activated through  `set_options`.

The current function works only when the input object is of the same type as the output one. So some functions couldn't be wrapped with this yet. The most important untouched code for the moment is the missing checks where I think this could help a lot.

It should not. This is completely optional.

### Other information:
In progress, I still need to prove the performance boost.

This depends on #1845 because I need all improvements for PC.


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=3.9,<4"]
+requires = ["flit_core >=3.8,<4"]
```
Zeitsperre, line 2: FYI, this should "technically" be okay for running in production, but this will likely cause some inconsistencies when building the package.

xclim/core/utils.py
```diff
@@ -760,3 +760,43 @@ def _chunk_like(*inputs: xr.DataArray | xr.Dataset, chunks: dict[str, int] | Non
...
+    Any coordinate that has a name different than its dimension(s) is flagged as auxiliary. All scalar coordinates are flagged as auxiliary.
+
+    Parameters
+    ----------
+    obj : DataArray or Dataset
+        Xarray object
+
+    Returns
+    -------
+    clean_obj :
```
Zeitsperre, line 780: Any?

xclim/indices/_threshold.py
```diff
@@ -30,6 +30,8 @@
     threshold_count,
 )
 
+from .helpers import resample_map
```
Zeitsperre, line 33: I think we should move towards absolute imports.

```suggestion
from xclim.indices.helpers import resample_map
```

xclim/indices/helpers.py
```diff
@@ -2,14 +2,15 @@
 Indices Helper Functions Submodule
 ==================================
 
-Functions that encapsulate some geophysical logic but could be shared by many indices.
+Functions that encapsulate logic but could be shared by many indices,
+but are not index-like themselves (these should go in the :py:mod:`xclim.indices.generic` module).
```
Zeitsperre, line 6:
```suggestion
Functions that encapsulate logic and can be shared by many indices,
but are not particularly index-like themselves (those should go in the :py:mod:`xclim.indices.generic` module).
```

xclim/indices/helpers.py
```diff
@@ -550,3 +552,82 @@ def _gather_lon(da: xr.DataArray) -> xr.DataArray:
...
+        Function to map on each resampled group.
+    map_blocks: bool or "from_context"
+        If True, the resample().map() call is wrapped inside a `map_blocks`.
+        If False, this does not do anything special.
+        If "from_context", xclim's "resample_map_blocks" option is used.
+        If the object is not using dask, this is set to False.
+    resample_kwargs: dict, optional
+        Other arguments to pass to `obj.resample()`.
+    map_kwargs: dict, optional
+        Arguments to pass to `map`.
```
Zeitsperre, line 587: I'm with the rest of the world on this that it's a bit silly, but the spaces around colons are important.

```suggestion
    obj : DataArray or Dataset
        The xarray object to resample.
    dim : str
        Dimension over which to resample.
    freq : str
        Resampling frequency along `dim`.
    func : callable
        Function to map on each resampled group.
    map_blocks : bool or "from_context"
        If True, the resample().map() call is wrapped inside a `map_blocks`.
        If False, this does not do anything special.
        If "from_context", xclim's "resample_map_blocks" option is used.
        If the object is not using dask, this is set to False.
    resample_kwargs : dict, optional
        Other arguments to pass to `obj.resample()`.
    map_kwargs : dict, optional
        Arguments to pass to `map`.
```
I'll be adding in a lint checks using `numpydoc-validate` in the future so it doesn't have to be me making the complaints, haha

xclim/indices/helpers.py
```diff
@@ -550,3 +552,82 @@ def _gather_lon(da: xr.DataArray) -> xr.DataArray:
...
+    if map_blocks == "from_context":
+        map_blocks = OPTIONS[MAP_BLOCKS]
+
+    if not uses_dask(obj) or not map_blocks:
+        return obj.resample({dim: freq}, **resample_kwargs).map(func, **map_kwargs)
+
+    try:
+        from flox.xarray import rechunk_for_blockwise
+    except ImportError as err:
+        raise ValueError(f"Using {MAP_BLOCKS}=True requires flox.") from err
```
Zeitsperre, line 604: Linters complain about f-strings in logging/exception calls.

```suggestion
        msg = f"Using {MAP_BLOCKS}=True requires flox."
        raise ValueError(msg) from err
```

tests/test_indices.py
```diff
@@ -29,6 +29,11 @@
 from xclim.core.options import set_options
 from xclim.core.units import convert_units_to, units
 
+try:
+    import flox
+except ImportError:
+    flox = None
```
Zeitsperre, line 35: Would it be better if we were adding `flox` to the `tox` configuration, then `parametrizing` the tests to use `flox` if it's there? I believe we have one build that tests the `extras` recipe already.

Zeitsperre (approved):
I'm far from an expert here, but the implementation looks good!

tests/test_indices.py
aulemahal: Yes, but my current "testing" only consist in hijacking an existing test and enabling the new option in a certain case. A `importorskip('flox')` would be much better, but I'll do that on a test specific to `resample_map`. To come in the next working days...

Zeitsperre: Okay, that sounds fine to me! Just keep it in mind.

pyproject.toml
Zeitsperre, line 1: I have an alternative suggestion here (and would save me from opening a PR):

1. Add `flox` (with a baseline pin version) to the `extras` recipe.
2. Add `flox` and `pot >=0.9.4` to the `environment.yml`.
3. Add `fastnanquantile >=0.0.2` to the `environment.yml` section for pip dependencies.
4. Remove `flox` from the DEP001 list.

aulemahal: I did the part about `flox`, should I still touch things related to pot and fastnanquantile?

xclim/core/utils.py
```diff
@@ -758,3 +758,44 @@ def _chunk_like(*inputs: xr.DataArray | xr.Dataset, chunks: dict[str, int] | Non
...
+    The auxiliary coordinates can be merged back with the dataset with
+    :py:meth:`xarray.Dataset.assign_coords` or :py:meth:`xarray.DataArray.assign_coords`.
+
+    >>> # xdoctest: +SKIP
+    >>> clean, aux = split_auxiliary_coordinates(ds)
+    >>> merged = clean.assign_coords(da.coords)
+    >>> merged.identical(ds)  # True
+    """
+    aux_crd_names = [
+        nm for nm, crd in obj.coords.items() if len(crd.dims) != 1 or crd.dims[0] != nm
```
coxipi, line 797: A "true" coordinate, i.e. it is a dimension of the array, always has the same name as the dimension and is 1-dimensional? So you verify the opposite of this condition?

xclim/core/options.py
```diff
@@ -71,6 +73,7 @@ def _valid_missing_options(mopts):
     SDBA_ENCODE_CF: lambda opt: isinstance(opt, bool),
     KEEP_ATTRS: _KEEP_ATTRS_OPTIONS.__contains__,
     AS_DATASET: lambda opt: isinstance(opt, bool),
+    MAP_BLOCKS: lambda opt: isinstance(opt, bool),
```
coxipi, line 76: is this an option because we're still not sure if the performance is good? Or there will always be situations where no wrapper is preferable?

xclim/indices/helpers.py
```diff
@@ -559,3 +561,83 @@ def _gather_lon(da: xr.DataArray) -> xr.DataArray:
...
+        return obj_chnk.resample({dm: frq}, **rs_kws).map(fun, **mp_kws)
+
+    # Template. We are hoping that this takes a negligeable time as it is never loaded.
+    template = obj_rechunked.resample(**{dim: freq}, **resample_kwargs).first()
+
+    # New chunks along time : infer the number of elements resulting from the resampling of each chunk
+    if isinstance(obj_rechunked, xr.Dataset):
+        chunksizes = obj_rechunked.chunks[dim]
+    else:
+        chunksizes = obj_rechunked.chunks[obj_rechunked.get_axis_num(dim)]
```
coxipi, line 633: I'm a bit confused by the comment. `obj_rechunked` has not been resampled...?

xclim/core/options.py
aulemahal: I think there would be situations where no wrapper is preferable, but it's mostly indeed because I'm not sure of the performance. In many cases, it should have a insignificant impact.

xclim/core/utils.py
aulemahal: Exact.

xclim/indices/helpers.py
aulemahal: The comment applies to the whole paragraph. It's a bit lengthy but here I take the flox-modified time chunksize and then I iterate over them. For each chunk, I look at the number of unique label in the resampling groups, which will correspond to the size of that chunk after the resample.

xclim/indices/helpers.py
```diff
@@ -562,6 +564,86 @@ def _gather_lon(da: xr.DataArray) -> xr.DataArray:
...
+    obj: xr.DataArray | xr.Dataset,
+    dim: str,
+    freq: str,
+    func: Callable,
+    map_blocks: bool | Literal["from_context"] = "from_context",
+    resample_kwargs: dict | None = None,
+    map_kwargs: dict | None = None,
+) -> xr.DataArray | xr.Dataset:
+    r"""
+    Wraps xarray's resample(...).map() with a :py:func:`xarray.map_blocks`, ensuring the chunking is appropriate using flox.
```
coxipi, line 577: missing first line?  and "Wraps" ->  "Wrap" on  first line

tests/test_atmos.py
```diff
@@ -642,11 +643,18 @@ def test_chill_units(atmosds):
...
     tasmax = atmosds.tasmax
     tasmin = atmosds.tasmin
     tas = make_hourly_temperature(tasmin, tasmax)
-    cp = atmos.chill_portions(tas, date_bounds=("09-01", "03-30"), freq="YS-JUL")
+    if use_dask:
+        tas = tas.chunk(time=tas.time.size // 2, location=1)
+
+    with set_options(resample_map_blocks=True):
+        cp = atmos.chill_portions(tas, date_bounds=("09-01", "03-30"), freq="YS-JUL")
+
```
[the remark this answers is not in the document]
aulemahal: For convenience. The option has no incidence on the computed values. Also, the option is actually ignored in the  `use_dask=False`, so this tests that shortcut while actually testing the difference between the block-mapped and the normal version.

PR #1850, opened 2024-07-22
Add additional checks for PWM fit existing values

### LINKED ISSUE #1849: indices.stats.fit fails for PWM with less than dist.numargs +2 values
### Setup Information

- Xclim version: 0.51.0
- Python version: 3.11.1
- Operating System: Ubuntu

### Description

Using the fit function with the `PWM` method and a lmoments3 dist fails if the dataarray contains slices with less than `dist.numargs + 2` non-null values (see below for an reproducible example). The simple solution would be to add another check in `_fitfunc_1d` when using the `PWM` method. I already implemented a monkeypatch for us so will provide a PR in a sec.

### Steps To Reproduce

```python
from xclim.indices.stats import _fitfunc_1d
import xarray as xr
import lmoments3.distr
import numpy as np

da = xr.DataArray([3,np.nan,1,np.nan], coords=[('x',[1,2,3,4])])
dist = lmoments3.distr.gum
_fitfunc_1d(da, dist=dist, method='PWM', nparams=3)
```

fails with 
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
python3.11/site-packages/xclim/indices/stats.py:51)     args, kwargs = _fit_start(x, dist.name, **fitkwargs)
python3.11/site-packages/lmoments3/distr.py:66) elif not lmom_ratios:

ValueError: At least 2 data points must be provided.
```

### Additional context

_No response_

### Contribution

### Code of Conduct

saschahofmann:
### Pull Request Checklist:
- This PR fixes #1849

### DIFF
--- a/xclim/indices/stats.py
+++ b/xclim/indices/stats.py
@@ -47,7 +47,7 @@ def _fitfunc_1d(arr, *, dist, nparams, method, **fitkwargs):
     elif method == "MM":
         params = dist.fit(x, method="mm", **fitkwargs)
     elif method == "PWM":
-        # In lmoments3 it will raise an error if only dist.numargs + 2 values are provided
+        # lmoments3 will raise an error if only dist.numargs + 2 values are provided
         if len(x) <= dist.numargs + 2:
             return np.asarray([np.nan] * nparams)
         params = list(dist.lmom_fit(x).values())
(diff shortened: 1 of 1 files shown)


xclim/indices/stats.py
```diff
@@ -47,6 +47,9 @@ def _fitfunc_1d(arr, *, dist, nparams, method, **fitkwargs):
     elif method == "MM":
         params = dist.fit(x, method="mm", **fitkwargs)
     elif method == "PWM":
+        # In lmoments3 it will raise an error if only dist.numargs + 2 values are provided
```
VascoSch92, line 50:
```suggestion
        # lmoments3 will raise an error if only dist.numargs + 2 values are provided
```

aulemahal (approved):
I thought about it and this is a good idea. Thanks!

PR #1864, opened 2024-08-02
cast high to float in jitter

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #xyz

* I am casting `high` to be a float, not an array. In the new version of dask, if `high` is an array, there are memory problems and we are unable to run jitter_under on the ESPO regions.
* I see that it was made explictly an array in #1721. @Zeitsperre do you remember why ?

no

### Other information:
This is only a problem when using dask.


aulemahal:
The docstring is a bit ambiguous as it says "a quantity with units". In the real xclim, this does mean that DataArrays are accepted. But we didn't use the  `Quantified` annotation here.

@RondeauG, to your knowledge, would there ever be a use for have the jitter limits change along a given dimension ? (Like a specific jitter for each latitudes).

My guess is no, so I suggest replacing the `np.array` cast above by `float`. And do this for all parameters. (lower + upper, + the other two functions).

PR #1881, opened 2024-08-13
Pylint fixes, remove deprecated functions

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adjust code base to address several `pylint`-related warnings.
* Removes several deprecated calendar functions.
* Pins the `vulture` dependency in `environment.yml`
* Employed casting to deal with a few `numpy` * `xarray` typing issues.
* Addressed a couple warnings related to `pint` usage.

Yes, several calendar-related functions that were deprecated and slated for removal in `xclim` v0.50 and v0.51 have been removed. These will need to be listed in the `CHANGELOG.rst`.

### Other Information:

I'm now a conda-forge maintainer for `vulture`, haha (https://github.com/conda-forge/vulture-feedstock)


xclim/indices/_agro.py
Zeitsperre, line 1: @coxipi Are we safe to remove the `offset` logic from `standardized_precipitation_evapotranspiration_index` here? Is it as simple as removing lines 1304 to 1325?

docs/conf.py
```diff
@@ -30,6 +30,7 @@
 xarray.CFTimeIndex.__module__ = "xarray"
 
 import xclim  # noqa
```
coxipi, line 32: now `xclim` is only used for :
```

# The short X.Y version.
version = xclim.__version__.split("-")[0]
# The full version, including alpha/beta/rc tags.
release = xclim.__version__
```
so perhaps we should also import `__version__` directly?

xclim/core/calendar.py
```diff
@@ -486,7 +378,7 @@ def percentile_doy(
         # Preserve chunk size
         time_chunks_count = len(arr.chunks[arr.get_axis_num("time")])
         doy_chunk_size = np.ceil(len(rrr.dayofyear) / (window * time_chunks_count))
-        rrr = rrr.chunk(dict(stack_dim=-1, dayofyear=doy_chunk_size))
+        rrr = rrr.chunk({"stack_dim": -1, "dayofyear": doy_chunk_size})
```
coxipi, line 382: using {} instead of dict: is that a personal stylistic choice, or it's best practices and some pre-commit check complains about it? Just so I know for the future

Zeitsperre: Best practices, apparently. Using `dict()` is good when you want to convert an object to a dict and relies on a C-lang internal, while using curly brackets seems to be slightly optimized for Python.

More context: https://stackoverflow.com/questions/664118/whats-the-difference-between-dict-and

docs/conf.py
Zeitsperre: We also access `xclim.indicators` on line 48. I think this is fine as is.

xclim/sdba/measures.py
```diff
@@ -144,9 +144,9 @@ def _postprocess(self, outs, das, params):
         """Squeeze `group` dim if needed."""
         outs = super()._postprocess(outs, das, params)
 
-        for i in range(len(outs)):
-            if "group" in outs[i].dims:
-                outs[i] = outs[i].squeeze("group", drop=True)
+        for out in outs:
+            if "group" in out.dims:
+                out = out.squeeze("group", drop=True)
```
coxipi, line 149: I don't think this is the same thing? 

```
outs = [1,2,3]
for out in outs: 
    out = out + 1 
print(outs)
>>> [1,2,3]
```

xclim/sdba/properties.py
```diff
@@ -91,15 +91,17 @@ def _postprocess(self, outs, das, params):
         """Squeeze `group` dim if needed."""
         outs = super()._postprocess(outs, das, params)
 
-        for i in range(len(outs)):
-            if "group" in outs[i].dims:
-                outs[i] = outs[i].squeeze("group", drop=True)
+        for ii, out in enumerate(outs):
+            if "group" in out.dims:
+                outs[ii] = out.squeeze("group", drop=True)
```
coxipi, line 96: here it's allright, compare to previous one. I might have missed other errors like the previous one, so maybe best to check again those modified loops

coxipi (changes requested):
looks good to me, but maybe double check the loops that you changed

xclim/indicators/land/_streamflow.py
```diff
@@ -16,10 +16,13 @@
 
 
 class Streamflow(ResamplingIndicator):
+    """Streamflow class."""
+
     context = "hydro"
     src_freq = "D"
     keywords = "streamflow hydrology"
 
+    # TODO: TJS: The signature of this method seems wrong. Should it be `def cfcheck(cls, q):` or something else? Is it a static method?
```
Zeitsperre, line 25: @aulemahal Any advice on this particular method ? Any way we can make this consistent ?

aulemahal: Indeed. all other implementations of `cfcheck` are normal methods. I would agree to do the same here.

However, I'll note that if we solve Eric's comment here https://github.com/Ouranosinc/xclim/pull/1832#discussion_r1725793960, then this function could be removed as the base  `cfcheck` will work.

xclim/core/units.py
```diff
@@ -57,7 +57,7 @@
 units = deepcopy(cf_xarray.units.units)
 # Changing the default string format for units/quantities. cf is implemented by cf-xarray
 # g is the most versatile float format.
-units.default_format = "gcf"
+units.formatter.default_format = "gcf"
```
aulemahal, line 60: This will require pinning pint (which will implicitly pin numpy > 2)!

If this is not ok to you, we could to a if-else with the version?

xclim/core/units.py
```diff
@@ -373,9 +370,7 @@ def convert_units_to(  # noqa: C901
     raise NotImplementedError(f"Source of type `{type(source)}` is not supported.")
 
 
-def cf_conversion(
-    standard_name: str, conversion: str, direction: Literal["to", "from"]
-) -> str | None:
+def cf_conversion(standard_name: str, conversion: str, direction: str) -> str | None:
```
aulemahal, line 373: I'm ok with this. But to understand, why was the stricter annotation removed ?

xclim/indicators/land/_streamflow.py
Zeitsperre: @coxipi Would it make more sense to address this in your PR then? I'll leave the `TODO` where it is.

xclim/core/utils.py
```diff
@@ -684,7 +683,7 @@ def adapt_clix_meta_yaml(  # noqa: C901
     freq_defs = {"annual": "YS", "seasonal": "QS-DEC", "monthly": "MS", "weekly": "W"}
 
     if isinstance(raw, os.PathLike):
-        with open(raw) as f:
+        with open(raw, encoding="utf-8") as f:
```
aulemahal, line 686: Is there a context where this isn't the default ?

Zeitsperre:
@aulemahal @coxipi This is good for final review!

xclim/indices/_conversion.py
```diff
@@ -773,11 +773,11 @@ def specific_humidity_from_dewpoint(
     ----------
     :cite:cts:`world_meteorological_organization_guide_2008`
     """
-    ε = 0.6219569  # weight of water vs dry air []
+    EPSILON = 0.6219569  # weight of water vs dry air []
```
aulemahal, line 776: :crying_cat_face:

xclim/sdba/utils.py
```diff
@@ -1025,7 +1026,13 @@ def optimal_transport(gridX, gridY, muX, muY, numItermax, transform):
     ----------
     :cite:cts:`sdba-robin_2021`
     """
-    from ot import emd
+    try:
+        from ot import emd  # pylint: disable=import-outside-toplevel
+    except ImportError as e:
+        raise ImportError(
+            "The optional dependency `POT` is required for optimal_transport. "
+            "You can install it with `pip install POT` or `pip install 'xclim[extras]'`."
```
aulemahal, line 1034:
```suggestion
            "You can install it with `pip install POT`, `conda install -c conda-forge pot` or `pip install 'xclim[extras]'`."
```

aulemahal (approved):

xclim/core/utils.py
Zeitsperre: Older Windows systems with unique locales can _potentially_ open files with different encodings or in str/bytes mode. This is a suggestion to make these reads bullet-proof.

xclim/indicators/land/_streamflow.py
coxipi: Not my PR, but I will add the relevant information in the PR so that the cfcheck function can be removed once this is fixed.

xclim/core/units.py
Zeitsperre: You know, I think that might've been a false-positive from `pylint`. The Literal method works fine in my IDE. I'll undo this change.

xclim/core/units.py
Zeitsperre: Sounds good! I'll adapt that.

xclim/indices/_conversion.py
Zeitsperre: The things we do for Windows users.

coxipi (approved):
approved

Zeitsperre:
The notebooks have been fixed in #1889. It likely makes more sense to merge that PR first before this one.

Zeitsperre:
@aulemahal 

I've made one or two small breaking changes:
- I've migrated `xclim.core` to use absolute imports instead of explicit relative imports (`from xclim.core import ...`)
- I've factored out a few small objects into hidden `_types.py` and `_exceptions.py` under `xclim.core`. The goal here being to reduce the number of time `xclim.core.utils` is imported throughout the module via its many members. Technically, this is breaking (and there are possibly other objects that might make sense being moved).

The rest of the changes are to clarify internal logic within functions via renaming. Thanks in advance for the re-review

aulemahal (approved):
Ah! Well done, thanks.

PR #1885, opened 2024-08-15
Heat spells

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* 3 new indicators `heat_spell_frequency`, `heat_spell_max_length` and `heat_spell_total_length`. The "heat_wave" indicators test that a bivariate condition on each day and then find consecutive runs of a minimum length. The new functions adhere to xclim's concept of "spell" : the condition is tested against a rolling N-day window and if it is fulfilled *all days* of the window are considered part of the spell.
* New generic function `spell_mask` that encapsulate the spell notion. It returns a boolean mask of the days part of spells. The code has been added two features:
    + Multivariate : input data can be a list (and so must be the thresholds). The way we check for the conditions across variables is decided by new argument `var_reducer` ('all' or 'any').
    + Weights : if the spell stat is a mean, weights can be given (used by the INSPQ, but not for the actual PC indicators).
* Indicator parameters can now be assigned new names. See examples in how the heat spell indicators are created. We could already do something similar with data variables. The internals of `Indicator` have changed accordingly (sorry).

No.

### Other information:
I kinda decided that "heat" meant "tasmax and tasmin".


xclim/indices/generic.py
```diff
@@ -354,6 +356,106 @@ def get_daily_events(
...
+        cs_s = cs_s.where(mask.shift({"time": -1}, fill_value=0) == 0)
+        # propagate these end of sequences
+        # the `.where(mask>0, 0)` acts a stopper
+        is_in_spell = cs_s.where(cs_s >= window).where(mask > 0, 0).bfill("time") > 0
+    else:
+        data_pad = data.pad(time=(0, window))
+        # The spell-wise value to test
+        # For example "window_reducer='sum'", we want the sum over the minimum spell length (window) to be above the thresh
+        if weights is not None:
+            spell_value = data_pad.rolling(time=window).construct("window").dot(weights)
```
coxipi, line 445: at the moment, `weights = [0.2, 0.2, 0.4]` and `weights = [2,2,4]`, even though they have the same ratios, won't give the same answers because we don't normalize weights. Is that intended? With `sum(weights) >1` you end up multiplying your input data by this factor. Do we want this kind of behaviour? The same spells can be achieved by diving the threshold by this same factor, I think users should play with the the threshold if they want to change the spells content, not with the input by playing with weights.  So I think normalizing the weights would be better. Thoughts?

coxipi: Also, just by curiosity, do we currently have indices that would use this mode (`mean` and non-trivial `weights`?)

aulemahal: Woups, indeed, we need to normalize the weights!

No we don't have official indicators using the weights, but this was a request for PC from the INSPQ. We decided not to go forward with the weights in PC anyway, but it was already coded.

xclim/indices/generic.py
```diff
@@ -424,41 +526,90 @@ def spell_length_statistics(
...
+@declare_relative_units(threshold1="<data1>", threshold2="<data2>")
+def bivariate_spell_length_statistics(
+    data1: xarray.DataArray,
+    threshold1: Quantified,
+    data2: xarray.DataArray,
+    threshold2: Quantified,
+    window: int,
+    win_reducer: str,
+    op: str,
+    spell_reducer: str,
```
coxipi, line 564: I think it's fine like this, maybe we don't want too much complexity, but let me point out a generalization that comes to mind: Having `op1` and `op2`.  For instance, `tasmax>thresh1`, `pr<thresh2`, i.e. `hot_and_dry_spells`.  See `_bivariate_spell_length_distribution` for instance.

coxipi (approved):
approved

tlogan2000 (approved):

PR #1889, opened 2024-08-20
Use pooch

Zeitsperre:
### Pull Request Checklist:
    - This PR relies on changes to be merged in https://github.com/Ouranosinc/xclim-testdata/pull/29

* Replaces the logic for file gathering and caching from the in-house developed version to instead use `pooch`.
  * In order to fetch testing data, one can now use the following:
  ```python
  from xclim.testing.helpers import nimbus

  n = nimbus()
  # from a fork of xclim-testdata: nimbus(repo="https://github.com/Me/My_Repo", branch="my_test_branch")
  file = n.fetch("some_folder/some_data.nc")
  ```
* Removes the remote GitHub calls for every file request (which was performed by `_get()`).
* Exports most of the file request and cache handling to `pooch`, while maintaining a relatively unchanged API for users.
* (To be confirmed) Speeds up the delivery of test data to tests by reducing the amount of redundant calls to fixtures and relying on a single pooch instance of pooch to prevent multiple setup stages.

Absolutely. `get_file` and `open_dataset` no longer fetch remote files from GitHub. Instead, a locally-stored `registry.txt` file contains all the checksums of all files needed to run the tests and returns the appropriate file from a locally-held cache. If the file checksum does not match the expected value, it will attempt to replace it from the remote storage.

Additionally, the `md5` files that accompanied all testing data files are now obsolete thanks to the use of the registry. The testing data is now versioned according to the `xclim-testdata` version/tag.

All the `prefetch` logic baked into the `pytest` calls has been removed, making the setup code much easier to follow. There is no longer a need to run `$ xclim prefetch_testing_data` unless users are running on Windows (for the very first run of `pytest` only).

There are now three environment variables to help developers:
- XCLIM_TESTDATA_BRANCH
    - Controls the branch name of `xclim-testdata`.
- XCLIM_TESTDATA_CACHE_DIR
    - Controls the local folder to be used when fetching the test data.
- XCLIM_TESTDATA_REPO_URL
    - Controls the repository URL for `xclim-testdata` (for forks) 

`platformdirs` is no longer a hard dependency. The default cache directory will only be determined if `pooch` is installed.

### Other information:

There is still a lot of potential here to tighten this up; I'd like to land on a design that is clean and easily portable to other projects.

What is unchanged is that `pytest` will still do the following on every run:
 1. Check that a locally stored copy of the test data exists in a platform-dependent default location and, if not found, will fetch a copy.
 2. Each worker of `pytest` creates its own copy of the test data, which is delivered by its own `pooch` instance, written to a threadsafe temporary directory
 3. The equivalent to the `get_file()` fixture is now `nimbus.fetch()`, providing the absolute paths to files, respective of platform and workers.

Many tests related to testing the file accessors have been removed (as these are now out of scope), but I'd like to test the implementation a bit before merging this (for both coverage and peace of mind).


.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ on:
       - submitted
 
 env:
-  XCLIM_TESTDATA_BRANCH: v2023.12.14
+  XCLIM_TESTDATA_BRANCH: cleanup
```
coxipi, line 22: what does this mean? XCLIM_TESTDATA_BRANCH will still exist no? The functions to interact with it have simply changed. Should we not keep some versioning like before?

tests/test_indices.py
```diff
@@ -2562,12 +2562,14 @@ def test_simple(self, open_dataset, ind, exp):
...
+    def test_indice_against_icclim(self, open_dataset):
         from xclim.indicators import icclim  # noqa
 
+        cmip3_tas = open_dataset("cmip3/tas.sresb1.giss_model_e_r.run1.atm.da.nc").tas
+
         with set_options(cf_compliance="log"):
-            ind = xci.tg_mean(cmip3_day_tas)
-            icclim = icclim.TG(cmip3_day_tas)
+            ind = xci.tg_mean(cmip3_tas)
+            icclim = icclim.TG(cmip3_tas)
```
coxipi, line 2572: I don't know what was `cmip3_tas` but I like this generalization of directly opening another folder on `xclim_testdata`

xclim/testing/conftest.py
```diff
@@ -11,12 +11,11 @@
...
 
 from xclim.testing import helpers
-from xclim.testing.utils import _default_cache_dir  # noqa
-from xclim.testing.utils import open_dataset as _open_dataset
+from xclim.testing.helpers import open_dataset as _open_dataset
 
 
 @pytest.fixture(autouse=True, scope="session")
-def threadsafe_data_dir(tmp_path_factory) -> Path:
+def threadsafe_data_dir(tmp_path_factory):
```
coxipi, line 18: why not specify the output type?

xclim/testing/helpers.py
```diff
@@ -10,13 +12,33 @@
...
+    SocketBlockedError = None
+
+try:
+    import pooch
+except ImportError:
+    warnings.warn(
+        "The `pooch` library is not installed. "
+        "The default cache directory for testing data will not be set."
+    )
+    pooch = None
```
coxipi, line 41: what happens then? not cache, we always download?

CONTRIBUTING.rst
```diff
@@ -269,21 +269,22 @@ Updating Testing Data
 
 If your code changes require changes to the testing data of `xclim` (i.e.: modifications to existing datasets or new datasets), these changes must be made via a Pull Request at the `xclim-testdata repository`_.
 
-`xclim` allows for developers to test specific branches/versions of `xclim-testdata` via the `XCLIM_TESTDATA_BRANCH` environment variable, either through export, e.g.::
+`xclim` allows for developers to test specific branches/versions or forks of the `xclim-testdata` repository via the `XCLIM_TESTDATA_BRANCH` and `XCLIM_TESTDATA_REPO` environment variables, respectively, either through export, e.g.::
```
coxipi, line 272: should we rename this repo at some point? if `xsdba` pulls from it too?

coxipi:
THe only change on user side is that `open_dataset` will automatically use caching if `pooch` is installed correct? 

I tried reviewing the code but there is a lot of things I don't really understand, my approval would not be much better than rubber stamp

xclim/testing/helpers.py
Zeitsperre: If pooch isn't providing a path, then a user can hardcode one and copy the data to a fixed location manually. If no data is found, then it (rightfully) will crash with an OSError.

CONTRIBUTING.rst
Zeitsperre: We could. My changes here make it much easier to do in the future (though I'm running into trouble with other projects due to hard coding of values; working through it).

I wouldn't oppose doing a huge cleanup of the testing data.

aulemahal (approved):
This looks good! Thanks.
A suggestion and yet again a nitpicky comment on the changelog ;).

docs/notebooks/analogs.ipynb
```diff
@@ -24,10 +24,10 @@
     "from __future__ import annotations\n",
     "\n",
     "import matplotlib.pyplot as plt\n",
+    "from xarray.coding.calendar_ops import convert_calendar\n",
     "\n",
     "from xclim import analog\n",
-    "from xclim.core.calendar import convert_calendar\n",
-    "from xclim.testing import open_dataset"
+    "from xclim.testing.helpers import open_dataset"
```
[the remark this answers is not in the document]
Zeitsperre: Good point. I'll see what can be done to make it simpler.

PR #1896, opened 2024-08-29
FIX: Standardize method applied on individual variables in OTC/dOTC

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* The mean and std in the 'standardize' normalization method of OTC/dOTC now is correctly applied on each variable individually
* While we're at it, I suggest changing the argument `transform` to `normalization` which I feel is more specific and informative. Also, in `optimal_transport`, we should follow the snake convention, `numIterMax` -> `num_iter_max` (I guess this was done to imitate the signature of `ot.emd`, but I think we need to stick to a standard in `xclim` functions)

* `transform` -> `normalization` in input arguments.


Zeitsperre (approved):
This seems like a fine change. The `transform` signature is very new, so no need for a deprecation warning message. Just need to update the CHANGELOG.

coxipi:
@SarahG-579462  Thoughts on these changes? Especially `transform` -> `normalization` which is a matter of taste.

coxipi:
It's a step in the right direction, but that's not even what I had in mind for the standardization. Right now, we just remove the mean/std of the cells, but those are simply evenly distributed. So, it doesn't say much about the distributions at hand, we don't have the information of the probability of each cell (or the count). Perhaps a proper standardization would not work well, but this remains to be seen

SarahG-579462 (approved):
This looks good. Did you want to include your work with the time step vs grid cell motion, or do you want that in a different PR?

coxipi:
Maybe the dOTC evolution through cells (more fidel to the paper) can wait another PR. But: 

```python
    def _standardize(grid, mu):
        mu0 = mu.reshape(len(mu), 1)
        mean = (mu0*grid).sum(axis=0)
        std =  np.sqrt((mu0*(grid**2-mean**2)).sum(axis=0))
        return (grid-mean)/std
   
    # this actually computes  `( X - mean(X) ) / std(X)`
    if normalization == "standardize":
        gridX = _standardize(gridX, muX)
        gridY = _standardize(gridY, muY)

    # this is the old way, it computes `(gridX - mean(gridX) ) / std(gridX)`
    elif normalization  == "standardize_cells":
        gridX = (gridX - gridX.mean(axis=0)) / gridX.std(axis=0)
        gridY = (gridY - gridY.mean(axis=0)) / gridY.std(axis=0)
        
```

The differences are not huge though:

image

coxipi:
I'll just merge this PR now, other changes can be in another PR.

PR #1920, opened 2024-09-13
Fix for xarray 2024.09.0

aulemahal:
Well, this could have been expected from my decision to use a private function that I myself removed... But I had forgotten that we were explicitly using it.

So this fixes xclim with xarray 2024.09.0 by having a try-except for the import of the decimal year stuff.

Also, removed the deprecated function.

As of xarray 2024.09, the "decimal year" can be calculated through `times.dt.decimal_year`, no need to use a private function.


Zeitsperre (approved):
Sounds good, but we need to make a new release, unfortunately.

aulemahal:
I'll let you decide how to manage the git stuff :sweat_smile:

Zeitsperre (changes requested):
~OK, we're going to need to rebase this PR onto the `v0.52.1` tag, if that's alright. If you want to do the following, this would be helpful:~
1. ~Start a new branch based on the `v0.52.1` tag (`v0.52.2`).~
2. ~`git cherry-pick` your commits to the new branch.~
3. ~Open a PR and do not merge main into it~

~I can take things from there.~

Scratch all of that. Working from here, see if you can get things working. There should be exactly two failing tests in all builds at the moment. I'll `mark.xfail` them here (they're removed in main) so that we can verify there aren't slower builds failing with unexpected failures.

aulemahal:
There's a bug with intersphinx for statsmodel, maybe just a retry could solve it ?

And the notebook CI doesn't seem aware that it should use the old version of xclim-testdata...

PR #1926, opened 2024-09-16
Heat wave magnitude indicator

### LINKED ISSUE #994: Heat Wave Magnitude Index
I would like to see the following indice considered as potential contributions to the xclim library:
The Heat Wave Magnitude Index is defined as the maximum magnitude of the heat waves in a year, where heat wave is the period ≥ 3 consecutive days with maximum temperature above the daily threshold for the reference period (https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2014JD022098/)

seblehner:
Closes #994 

### Pull Request Checklist:
    - This PR fixes #994 

* Adds a heat_wave_magnitude indicator based on a variable definition on heat waves (consecutive days of maximum temperature above a given threshold), which yields the highest magnitude in the form of accumulated maximum temperature differences above a threshold along a consecutive heat wave
* Adds a windowed_max_run_sum function to indices.run_length which yields the maximum accumulated values of runs per `freq` period
* Adds a rse (termed run sum encoding) function which sums positive values for runs

No

### TODO:
- [x] one of the test doesn't convert units properly
- [x] french translation (need help with that one)


seblehner:
Regarding the last test that fails, the problem arises because `to_agg_unit` converts the values from `Kelvin` to `deg C`, even though the input array already had `Kelvin` as unit. Maybe I am misunderstanding something.

The original data as per the following test is in `Kelvin`:
```python
def test_simple(self, tasmax_series):
        tx = np.zeros(366)
        tx[:5] = np.array([30, 30, 30, 30, 30])
        tx = tasmax_series(tx + K2C, start="1/1/2000")
        hwm = atmos.heat_wave_magnitude(tx, freq="YS")
        np.testing.assert_array_equal(hwm, [25])
```
Printing the threshold (default `deg C`) before and after the conversion:
```python
print(thresh)
thresh = convert_units_to(thresh, tasmax)
print(thresh)
```
yields first `25.0 degC` and second `298.15`. So the threshold values are correctly converted to `Kelvin`, which is correct since the input was also provided in `Kelvin`. So far so good.

Printing the output before the `heat_wave_magnitude` function returns, like so:
```python
    ...
    print(out)
    return to_agg_units(out, tasmax, op="max")
```
yields an `array` containing `25`, which is the expected and correct output.
However, when the results are returned, `to_agg_units` converts the result to `deg C`, subtracts 273.15 and hence provides the wrong output. What I don't understand is, why is the output converted, when the input was `Kelvin` and not `deg C`? Or is this expected behaviour that the output is converted to `deg C` in this case?

seblehner:
I changed the unit handling to 
```python
    out.attrs["units"] = tasmax.attrs["units"]
    return out
```
which seems more approriate compared to how other functions handle comparable output.
However, the output is still falsely converted. I suppose `declare_units` in that case modifies the output. What would the the prefered `xclim` way to handle the unit conversion properly in this case?

xclim/indicators/atmos/_temperature.py
```diff
@@ -215,6 +216,21 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
     compute=indices.heat_wave_frequency,
 )
 
+heat_wave_magnitude = Temp(
+    title="Heat wave magnitude",
+    identifier="heat_wave_magnitude",
+    units="degC",
```
aulemahal, line 222: You are requesting degC here for the indicator, would that be the source of your issue ?

xclim/indices/_threshold.py
```diff
@@ -1928,6 +1929,59 @@ def heat_wave_index(
...
+
+    out = rl.resample_and_rl(
+        over_values,
+        resample_before_rl,
+        rl.windowed_max_run_sum,
+        window=window,
+        freq=freq,
+    )
+    out.attrs["units"] = tasmax.attrs["units"]
+    return out
```
aulemahal, line 1982:
```suggestion
    return to_agg_units(out, tasmax, op="integral")
```
Shouldn't the output be "K d" ? A sum of temperatures doesn't mean much, but an integral does, no ? 
And xclim (in the last few versions) has made the decision to avoid using "degC day" as it is ambiguous.

xclim/indicators/atmos/_temperature.py
seblehner: That might be it indeed. What do you think should be the dedicated unit for it? "K d" like you suggested in the comment below (https://github.com/Ouranosinc/xclim/pull/1926#discussion_r1763248830)?

xclim/indices/_threshold.py
seblehner: Changing the units in https://github.com/Ouranosinc/xclim/pull/1926/files#diff-5d97458806cde0bbdc29e9a9b0e3209999b8f6d9b12b90ab9ac499a956d8ddd2R222 from `degC` to `K d` combined with this suggestion has indeed fixed the failing test. Thank you!

xclim/indicators/atmos/_temperature.py
seblehner: As mentioned in this comment https://github.com/Ouranosinc/xclim/pull/1926#discussion_r1763282932 this issue is resolved by changing the units to "K d".

xclim/indices/run_length.py
```diff
@@ -259,6 +259,50 @@ def get_rl_stat(d):
...
+        If 'first' (default), the run length is indexed with the first element in the run.
+        If 'last', with the last element in the run.
+
+    Returns
+    -------
+    xr.DataArray
+        Values are 0 where da is False (out of runs).
+    """
+    da = da.astype(float)
+
```
coxipi, line 285: I think we could simply remove this line and use this as `rle` since this function is more general. Then we only apply the `int` conversion if data type is a bool
```suggestion
    """
    if da.dtype == bool: 
       da = da.astype(int)
```

Does that sound reasonable @aulemahal ?

xclim/indices/run_length.py
```diff
@@ -1303,6 +1402,25 @@ def windowed_run_count_1d(arr: Sequence[bool], window: int) -> int:
...
+    window : int
+        Minimum duration of consecutive run to accumulate values.
+
+    Returns
+    -------
+    float
+        Accumulated values of positive values of a consecutive run at least `window` long.
+    """
+    # TODO: implement rse_1d function
+    raise NotImplementedError()
```
coxipi, line 1421: Do we still use these functions in general? I think it's worth keep the 1d-function in case if we have specific performance issues, this might be useful to have alternative implementations, but in general, we don't necessarily need to build all new functions with a 1d version?

xclim/indices/run_length.py
```diff
@@ -409,6 +453,61 @@ def windowed_run_count(
...
+        Total number of `True` values part of a consecutive runs of at least `window` long.
+    """
+    # TODO: implement ufunc function
+    # ufunc_1dim = use_ufunc(ufunc_1dim, da, dim=dim, index=index, freq=freq)
+
+    # if ufunc_1dim:
+    #     out = windowed_max_run_sum_ufunc(da, window, dim)
+
+    if window == 1 and freq is None:
+        out = da.max(dim=dim)
```
coxipi, line 498: Should this not be: 
```suggestion
    if window == 1 and freq is None:
        out = rse(da, dim=dim, index=index).max(dim=dim)
```
?

xclim/indices/_threshold.py
```diff
@@ -1928,6 +1929,58 @@ def heat_wave_index(
...
+    Returns
+    -------
+    DataArray, [time]
+        Heat wave magnitude.
+    """
+    thresh = convert_units_to(thresh, tasmax)
+    # condition where over threshold is fulfilled
+    cond = compare(tasmax, op, thresh, constrain=(">", ">="))
+    # difference for heatwave magnitude calculation
+    over_values = (tasmax - thresh).where(cond, 0)
```
aulemahal, line 1972: If you are putting 0 where tasmax does not respect the condition, then I'm not sure what use is the `op` argument ?

```suggestion
    over_values = (tasmax - thresh).clip(0)
```

xclim/indices/run_length.py
aulemahal: Yes that does absolutely make sense!

xclim/indices/run_length.py
aulemahal: Agree that this doesn't need a 1d function.

xclim/indices/_threshold.py
seblehner: Thanks for the suggestion, yes that's way more elegant.

xclim/indices/_threshold.py
```diff
@@ -1928,6 +1929,58 @@ def heat_wave_index(
     return to_agg_units(out, tasmax, "count")
 
 
+@declare_units(tasmax="[temperature]", thresh="[temperature]")
+def heat_wave_magnitude(
```
coxipi, line 1933: About the name. If I'm thinking about:
`heat_spell_max_length`
`heat_spell_total_length`

I think `heat_wave_magnitude` is more similar to `heat_spell_max_length`. Could there be a use for an indicator that computes the cumulative heat across the heat wave during a year, similar to what `heat_spell_total_length`?  If so, should we adopt a notation like:
`heat_wave_magnitude -> heat_wave_max_magnitude`
and potentially a new indicator in the future:
`heat_wave_total_magntitude` ?

coxipi: In fact, what I just described would compute the cumulative heat, and not the exceedance over a threshold.  Inspired by `degree_days_exceedance_date`, we could have:

`heat_wave_magnitude -> heat_wave_max_exceedance`

Does this sound clear?

xclim/indices/run_length.py
```diff
@@ -409,6 +453,61 @@ def windowed_run_count(
...
+
+    else:
+        d_rse = rse(da, dim=dim, index=index)
+        d_rle = rle(xr.where(da > 0, 1, 0), dim=dim, index=index)
```
coxipi, line 502: Not sure if it improves performance much but why not
```suggestion
        d_rle = rle(da>0, dim=dim, index=index)
```

tests/test_temperature.py
```diff
@@ -889,6 +889,32 @@ def test_nan_presence(self, tasmax_series):
...
+        np.testing.assert_array_equal(hwm, [25])
+
+    def test_nan_presence(self, tasmax_series):
+        tx = np.zeros(366)
+        tx[:5] = np.array([30, 30, 30, 30, 30])
+        tx[-1] = np.nan
+        tx = tasmax_series(tx + K2C, start="1/1/2000")
+
+        hwm = atmos.heat_wave_magnitude(tx, freq="YS")
+        np.testing.assert_array_equal(hwm, [np.nan])
```
coxipi, line 915: `heat_wave_max_length` would give a length of 5, not `np.nan`, I think it should also give a non-null result (25 in this case). However, I'm confused, I'm not sure what part of your code would produce `np.nan` instead of 25, do you know?

xclim/indices/run_length.py
seblehner: Ok, in that case should I simply remove the corresponding parts again?

tests/test_temperature.py
seblehner: I am not sure to be honest. I took this part from `TestHeatWaveIndex` and have assumed it follows some similar boilerplate checks that make the function return `np.nan`.

The following is the part that I then just adapted for magnitude.
```python

class TestHeatWaveIndex:

    ...

    def test_nan_presence(self, tasmax_series):
        tx = np.zeros(366)
        tx[:10] = np.array([29, 31, 31, 31, 29, 31, 31, 31, 31, 31])
        tx[-1] = np.nan
        tx = tasmax_series(tx + K2C, start="1/1/2000")

        hwi = atmos.heat_wave_index(tx, freq="YS")
        np.testing.assert_array_equal(hwi, [np.nan])
```

xclim/indices/_threshold.py
seblehner: Sounds clear, yes. I was mainly calling it magnitude because that's how it's described in the literature if I am not mistaken (e.g. https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2014JD022098). 

Distinguishing between and allowing for both, max and total, is a good point, since there are multiple different Heat Wave definitions (and perhaps Magnitude indices definitions) out there and I can imagine that the total quantity might be interesting to analyze as well.

xclim/indices/run_length.py
seblehner: Should I add these changes within this PR, or do you think it should be separate?

xclim/indices/run_length.py
aulemahal: Yes, I think removing the 1d stubs is ok.

xclim/indices/_threshold.py
coxipi: In the paper you shared, I don't see any threshold exceedance mentioned?  It seems that the definition that is often used   elsewhere is the Heat Wave Magntitude Index which is scaled between 0 and 1 (but still, it's tasmax and not tasmax -thresh that is transformed). This indicator seems to be the most standard.

Here : https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2020GL087097 the heatwave threshold exceedance is referred to as heat severity (there is also a spatial aspect to this)

Here: https://agupubs.onlinelibrary.wiley.com/doi/pdfdirect/10.1029/2022EA002567 the magnitude is defined in a similar way is the Tmax anomalies, but those anomalies are normalized before-hand. 

Similarly here: https://journals.ametsoc.org/view/journals/apme/56/9/jamc-d-16-0393.1.xml?tab_body=pdf, the magntitude refers to the "average departure from the heat wave day threshold (90th percentile as defined above) among grid points experiencing heat wave conditions."

All in all, it seems that "magnitude" is indeed often used to refer to the exceedance relative to a reference value. I think we can go with this since it seems standard. 

It doesn't seem to be the case that there is a well established definition for `heat_wave_magnitude`, but `heat_wave_magnitude_index` is more standard. I think it's fine that we use `heat_wave_max_magnitude`.

xclim/indices/run_length.py
coxipi: In this PR it's all right. You can just modify `rle` 
1. replace the int conversion with the condition above
2. `da == 1` -> `da>0`
and remove `rse`

tests/test_temperature.py
coxipi: Thanks, never mind. It must be the difference between the index, which is a simple function:
`xclim.indices.heat_wave_index`
and the class which performs a series of checks. 
`xclim.atmos.heat_wave_index`
I didn't realize it would not allow too many nans. 

Sorry about that!

xclim/indices/_threshold.py
coxipi: Just to say it clearly: 

I suggest changing the name of this current indicator to `heat_wave_max_magnitude`, to mimick `heat_wave_max_length`.  Magnitude is apparently often used as a synonym for "threshold exceedance" in the case of heat waves, so I think it's ok to keep this. @aulemahal @Zeitsperre if you have objections please let me know.

In the future,  `heat_wave_total_magnitude` could be considered in the future.

xclim/indicators/atmos/_temperature.py
```diff
@@ -215,6 +216,21 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
     compute=indices.heat_wave_frequency,
 )
 
+heat_wave_magnitude = Temp(
+    title="Heat wave magnitude",
+    identifier="heat_wave_magnitude",
+    units="K d",
+    long_name="Maximum difference between daily maximum temperature and {thresh} for days within a heat wave. "
+    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above "
```
coxipi, line 224:
```suggestion
    long_name="Maximum cumulative difference between daily maximum temperature and {thresh} for days within a heat wave. "
    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above "
```

xclim/indicators/atmos/_temperature.py
```diff
@@ -215,6 +216,21 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
+    "{thresh}.",
+    description="Magnitude of the most intensive heat wave per {freq} as accumulated deg C above a given threshold "
```
coxipi, line 226: `deg C` in description, but units above are specified as  `K d`.

seblehner: Fixed, thx.

coxipi: `deg C` in description, but units above are specified as  `K d`.  I suggest just not specify this, and add some details:  
```suggestion
    description="Magnitude of the most intensive heat wave per {freq}. The magnitude is the cumulative exceedance of  daily maximum temperature above {thresh}. A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above"
```

seblehner: Sorry my bad, I commited some changes in between and the suggestions you made are outdated now. I will wait and let you finish.

xclim/data/fr.json
```diff
@@ -377,6 +377,12 @@
     "title": "Indice de vague de chaleur",
     "abstract": "Nombre de jours de vagues de chaleur. Une vague de chaleur se produit lorsque la température maximale quotidienne excède un seuil donné durant un certain nombre de jours."
   },
+  "HEAT_WAVE_MAGNITUDE": {
+    "long_name": "TODO",
+    "description": "TODO",
+    "title": "TODO",
+    "abstract": "TODO"
+  },
```
coxipi, line 385: I never really know what information should go where ... 
```suggestion
  "HEAT_WAVE_MAGNITUDE": {
    "long_name": "Différence cumulative maximale entre la température maximale quotidienne et {thresh} "
    "pour les jours faisant partie d'une vague de chaleur. Une vague de chaleur est définie comme une série d'au moins "
    "{window} jours consécutifs où la température maximale quotidien est au-dessus de {thresh}",
    "description": "Cumul {freq:m} maximal de la différence entre la température et {thresh} pour les jours faisant partie "
    "d'une vague de chaleur. Une vague de chaleur est définie comme une série d'au moins {window} jours consécutifs où la "
    "température maximale quotidien est au-dessus de {thresh}",
    "title": "Magnitude de la vague de chaleur {freq:m} la plus intense. Une vague de chaleur est définie comme une série d'au moins "
    "{window} jours consécutifs où la température maximale quotidien est au-dessus de {thresh}.", 
    "abstract": "Magnitude de la vague de chaleur {freq:m} la plus intense. Une vague de chaleur est définie comme une série d'au moins "
    "{window} jours consécutifs où la température maximale quotidien est au-dessus de {thresh}.", 
  },
```

xclim/indicators/atmos/_temperature.py
coxipi: that's ok, I'll open a new comment

xclim/indicators/atmos/_temperature.py
```diff
@@ -215,6 +216,21 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
 
+heat_wave_magnitude = Temp(
+    title="Heat wave magnitude",
+    identifier="heat_wave_magnitude",
+    units="K d",
+    long_name="Maximum cumulative difference between daily maximum temperature and {thresh} for days within a heat wave. "
+    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above "
+    "{thresh}.",
+    description="Magnitude of the most intensive heat wave per {freq}. A heat wave occurs when daily maximum "
+    "temperatures exceed {thresh} for {window} consecutive days.",
```
coxipi, line 227: I was just suggesting a bit more details 
```suggestion
 description="Magnitude of the most intensive heat wave per {freq}. The magnitude is the cumulative exceedance of daily maximum temperature over {thresh}. A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above"
```

xclim/indices/_threshold.py
aulemahal: I just realized it,  but "heat_wave" is used elsewhere in xclim for the double-thresholded indicator (tasmax & tasmin) while the tasmax-only threshold is called "hot spell". I fear that the current name of the new indicator would bring confusion and I would suggest renaming to  `hot_spell_[...]`.

Sorry to come with this important change so late in the process!

As for the suffix, I agree with adding a "max" to leave the door open for a "total". 

"intensity" is used in 3 other indicators of xclim, all relating to the average value of precipitation.  I thus agree with using `magnitude` here, even more considering that it's already being used in the literature.

And @coxipi, I have yet to see a meaningful use of a sum of non-anomalies temperatures!

coxipi: It makes sense for `hot_spell`, it's unfortunate to have a name that's not closer to litterature but it's the price to pay to have coherent naming within xclim. 

> meaningful use of a sum of non-anomalies temperatures

That was more to reflect on the fact that in common language, "magntitude" makes me think of the temperature directly, not the anomalies. But I'm thinking more about "heat magnitude". Maybe it makes more sense like this, if there is no "heat wave", then the "heat wave magnitude" is zero.  

And yes, I guess summing temperatures in a given heatwave doesn't make much sense, the interesting information would be: what was the mean temperature / max temperature in a given heatwave. Anyways

xclim/indicators/atmos/_temperature.py
```diff
@@ -215,6 +216,20 @@ class TempWithIndexing(ResamplingIndicatorWithIndexing):
...
+heat_wave_magnitude = Temp(
+    title="Heat wave magnitude",
+    identifier="heat_wave_magnitude",
+    units="K d",
+    long_name="Maximum cumulative difference between daily maximum temperature and {thresh} for days within a heat wave. "
+    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above "
+    "{thresh}.",
+ description="Magnitude of the most intensive heat wave per {freq}. The magnitude is the cumulative exceedance of daily maximum temperature over {thresh}. A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above"
+    abstract="Magnitude of the most intensive heat wave per {freq}. A heat wave occurs when daily maximum "
+    "temperatures exceed given thresholds for a number of days.",
```
Zeitsperre, line 228:
```suggestion
    long_name="Maximum cumulative difference between daily maximum temperature and {thresh} for days within a heat wave. "
    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above {thresh}.",
 description="Magnitude of the most intensive heat wave per {freq}. The magnitude is the cumulative exceedance of daily maximum temperature over {thresh}. "
    "A heat wave is defined as a series of at least {window} consecutive days with daily maximum temperature above {thresh}",
    abstract="Magnitude of the most intensive heat wave per {freq}. A heat wave occurs when daily maximum "
    "temperatures exceed given thresholds for a number of days.",
```

coxipi:
@aulemahal I'm now wondering if having the float option in `rle` might cause problems with `rle_1d`, I will think about this

aulemahal:
I think we could simply explain the nuance in the docstring ? Like explain what happens when a bool / int / float is passed in the input of either functions. In most cases,  `rle` is not used by itself, but called from within one of the functions like `rle_statistics` or `windowed_max_run_sum`, and I would expect that function to choose the right function.

I can push such a commit.

coxipi (approved):
That looks good to me. Thanks for this nice work!

On a final note, do you have a specific paper in mind on which you base this indicator? I had a discussion on various papers I can't find it now. The indicator in the paper cited in the original issue is a bit different, it has a standardization aspect with a calibration period, so I would not cite this paper, or close the issue yet, the implementation is a bit different to what is done here.

seblehner:
> That looks good to me. Thanks for this nice work!

Thank you (and everyone else) also for helping! Happy to contribute 🙌 

> On a final note, do you have a specific paper in mind on which you base this indicator? I had a discussion on various papers I can't find it now. The indicator in the paper cited in the original issue is a bit different, it has a standardization aspect with a calibration period, so I would not cite this paper, or close the issue yet, the implementation is a bit different to what is done here.

A colleague cited this paper: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022EF002833 for the indicator. The paper is about a compound indicator for drought and heat, but the heatwave part follows the definition here (and is in turn referencing the Russo et al., 2014 paper, among others).

If I didn't miss anything, there are 2 points in particular about the Russo et al., 2014 definition:
- Reference period for threshold calculation: 1981--2010.
- Threshold calculated on a 31 day centered window.

The first point is already captured in the current implementation, because you can calculate the threshold for any reference period and supply the `xr.DataArray` as input for `thresh` (instead of specifying a static threshold like `25 degC`).

The second point however can not as easily be rebuilt I think. I don't know how much functionality under the `xclim` hood already exists for such operations and haven't looked into that yet to be honest. I think it boils down to something comparable to the following two indicators: 
- [`atmos.fraction_over_precip_doy_thresh`](https://xclim.readthedocs.io/en/stable/api_indicators.html#xclim.indicators.atmos.fraction_over_precip_doy_thresh)
- [`atmos.fraction_over_precip_thresh`](https://xclim.readthedocs.io/en/stable/api_indicators.html#xclim.indicators.atmos.fraction_over_precip_thresh)

Those two do in principle something similar, but the former calculates the threshold over `doy`, while the latter can have a static one. That would essentially be the same missing part (together with the added window around `doy`) to exactly capture the Russo et al., 2014 indicator.

coxipi:
Thanks! I just wanted to know which references should be cited, I've added both. I'll merge your PR as soon as the tests have re-passed

Zeitsperre:
@seblehner Thanks so much for your work on this! xclim v0.53 will be released with your changes either this or next week. Have a good day!

PR #1929, opened 2024-09-18
Update calendar.py

Balinus:
Seems like it should be worded as "biggest" (or a synonym) here and not "smallest"

### Pull Request Checklist:
Seems too straightforward as a change.

* Documentation

Nope


aulemahal (approved):
Indeed! Merci!

Zeitsperre (approved):

PR #1951, opened 2024-10-11
Using map groups in MBCn

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* MBCn internal functions now use map_groups instead of the groupies-like implementation

No

### Other information:

I need to do further tests, but so far it seems as fast, if not faster, than the old implementation. But I have to make tests with bigger datasets. 

I tried using the old way, only for training, for a dataset with resolution of 0.11deg over NAM. And the memory was busting, I think there was probably a memory leak in the old way, so that was my motivation to try again to use map_groups, rather than being focused solely on performance.


xclim/sdba/adjustment.py
```diff
@@ -1855,17 +1862,25 @@ def _adjust(
...
-            ref=ref,
-            hist=hist,
-            sim=sim,
-            ds=ds,
+        times = get_times(hist.time, group=group)
+        af_q[group.prop] = times.times.values
+        af_q = af_q.rename({group.prop: "time"})
+        af_q = af_q.broadcast_like(hist)
+        # a bit dirty, should avoid this situation before
+        af_q = af_q.chunk({"time": -1})
```
coxipi, line 1874: this is a hack: broadcast PROP dimension of af_q to the the full time dimension, so it can be used in `map_groups`. Perhaps there is already a built-in way to achieve this?

Issue #1892, opened 2024-08-26
Please add option for 'weekly' frequency in the SPI family of models


C15H24:
### Potential Solution

the function standardized_precipitation_index accepts freq = 'W' as signifying the nature of the groupings is weekly - and any downstream adjustments that would need to be made to validate the output

### Additional context

Many projects in government are beginning to operate and produce products on a weekly basis

### Contribution

- [ ] I would be willing/able to open a Pull Request to contribute this feature.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

coxipi:
Hi! We can add this indeed. Just a couple caveats:

1. It will only work with standard calendars, as `.resample(time="W")` and `.groupby("time.week")` is not yet compatible with cf-time (https://github.com/pydata/xarray/discussions/6375)
2. Since there is not an integer number of weeks in a year, using `.resample(time="W").mean()` the days included in week1 can  vary from year to year, e.g.
```
# days included in week 1 between 2000 and 2002
2000: [Jan 3, Jan 9]
2001: [Jan 1, Jan 7]
2002: [Dec31 2001, Jan 6]
...

```
sometimes Jan 1rst is not included, sometimes Week1 includes days from december, etc. I'm curious to know if this is what you envision when you consider weekly groupings.

coxipi:
@C15H24  We decided to support the weekly frequency after all, this will be in xclim's next release. Enjoy!

PR #1952, opened 2024-10-11
Support freq="W" in standardized indices

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add support for weekly standardized indices

No

### Other information:

I realize that xclim and climate_indices treat zero-inflated distributions differently. I will investigate which library does it correctly and include changes if needed

EDIT: I believe `xclim` has the correct implementation. The probability of zeroes should be determined in the fitting procedure, not by using the full dataset. The idea is when you compute a CDF, a zero value should be mapped to the `prob_of_zero` in the calibration period. That is the same logic as using the fitting params of the distribution in the calibration period to compute the CDF in the full dataset. 

I confirmed this is also how it's done in the case of the R package (SPEI), although I find some steps a bit weird, but I'm not too familiar with R.


CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ New features and enhancements
 * ``xclim.indices.run_length.runs_with_holes`` allows to input a condition that must be met for a run to start and a second condition that must be met for the run to stop. (:pull:`1778`).
 * New generic compute function ``xclim.indices.generic.thresholded_events`` that finds events based on a threshold condition and returns basic stats for each. See also ``xclim.indices.run_length.find_events``. (:pull:`1778`).
 * ``xclim.core.units.rate2amount`` and ``xclim.core.units.amount2rate`` can now also accept quantities (pint objects or strings), in which case the ``dim`` argument must be the ``time`` coordinate through which we can find the sampling rate. (:pull:`1778`).
+* `xclim.indices.stats.standardized_index` now supports a weekly resampling frequency. Only standard calendar is supported for this mode. (:issue:`1892`, :pull:`1952`)
```
Zeitsperre, line 31:
```suggestion
* ``xclim.indices.stats.standardized_index`` now supports a weekly resampling frequency. Only "standard" calendar is supported for this mode. (:issue:`1892`, :pull:`1952`)
```

coxipi:
I could not get "max-lik" to work with the R package, just got NaNs. But I hacked the xclim SPI to be able to use lmoments3 (it should be modified to be able to accept lmoments3) and the results are quite similar 
image

This is not the case as I add zeros, so the treatment must still be different

coxipi:
All in all, the weekly resampling seems to work well. On the question of the treatment of zeroes, climate-indices, R package, and xclim, all seem to work differently. I'm starting to think I have the right implementation, but if anyone wants to chip in, I would be glad if someone wants to give a fresh look on this

coxipi:
If the calendar is not appropriate, there will simply be problems with resampling or groupby ? Is that sufficient?

CHANGELOG.rst
```diff
@@ -29,6 +29,7 @@ New features and enhancements
 * ``xclim.indices.run_length.runs_with_holes`` allows to input a condition that must be met for a run to start and a second condition that must be met for the run to stop. (:pull:`1778`).
 * New generic compute function ``xclim.indices.generic.thresholded_events`` that finds events based on a threshold condition and returns basic stats for each. See also ``xclim.indices.run_length.find_events``. (:pull:`1778`).
 * ``xclim.core.units.rate2amount`` and ``xclim.core.units.amount2rate`` can now also accept quantities (pint objects or strings), in which case the ``dim`` argument must be the ``time`` coordinate through which we can find the sampling rate. (:pull:`1778`).
+* ``xclim.indices.stats.standardized_index`` now supports a weekly resampling frequency. Only "standard" calendar is supported for this mode. (:issue:`1892`, :pull:`1952`)
```
aulemahal, line 32:
```suggestion
* ``xclim.indices.stats.standardized_index`` now supports a weekly resampling frequency. Only "standard" calendars using numpy's `datetime64` dtype are supported for this mode. (:issue:`1892`, :pull:`1952`)
```

tests/test_indices.py
```diff
@@ -636,8 +668,11 @@ def test_standardized_precipitation_index(
             pytest.skip("Skipping SPI/ML/D on older numpy")
         ds = open_dataset("sdba/CanESM2_1950-2100.nc").isel(location=1)
         if freq == "D":
-            ds = ds.convert_calendar("366_day", missing=np.nan)
             # to compare with ``climate_indices``
+            ds = ds.convert_calendar("366_day", missing=np.nan)
+        elif freq == "W":
+            # only standard calendar supported with freq="W"
+            ds = ds.convert_calendar("standard", missing=np.nan, align_on="year")
```
aulemahal, line 675:
```suggestion
            ds = ds.convert_calendar("standard", missing=np.nan, align_on="year", use_cftime=False)
```

xclim/indices/_agro.py
```diff
@@ -1162,6 +1162,8 @@ def standardized_precipitation_index(
     * N-month SPI / N-day SPI is determined by choosing the `window = N` and the appropriate frequency `freq`.
     * Supported statistical distributions are: ["gamma", "fisk"], where "fisk" is scipy's implementation of
        a log-logistic distribution
+    * Supported frequencies are daily ("D"), weekly ("W"), and monthly ("MS").
+      Weekly frequency will only work if the input array has a "standard" calendar.
```
aulemahal, line 1166:
```suggestion
      Weekly frequency will only work if the input array has a "standard" (non-cftime) calendar.
```

xclim/indices/_agro.py
```diff
@@ -1298,7 +1298,7 @@ def standardized_precipitation_evapotranspiration_index(
     """
     fitkwargs = fitkwargs or {}
 
-    dist_methods = {"gamma": ["ML", "APP", "PWM"], "fisk": ["ML", "APP"]}
+    dist_methods = {"gamma": ["ML", "APP"], "fisk": ["ML", "APP"]}
```
aulemahal, line 1301: Why remove "PWM"?  I thought the conclusion was to change the accepted types of `dist` instead, to allow passing `lmoments3` classes ?

aulemahal (approved):
Ok, but I tought the solution for lmoments3 was the opposite.

xclim/indices/_agro.py
coxipi: Right now it's not working, so the smallest change possible was to forbid it for now. 

But anyways, the way I see it, the strings here will correspond to constrained and well tested flagships distributions. The PWM/gamma would use lmoments3 and fall in the category of uncontrolled distributions. Those changes would: 

* In SPI signature: `dist : str -> dist : str | scipy.stats.rv_continuous` 
* Only do checks if `dist` is a str:
```python
    if isinstance(dist,str):  # only check str, which are the flagship distributions/methods
        if dist in dist_methods.keys():
            if method not in dist_methods[dist]:
                raise NotImplementedError(
                    f"{method} method is not implemented for {dist} distribution"
                )
        else:
            raise NotImplementedError(f"{dist} distribution is not yet implemented.")
 ```

coxipi: So, in short, I was thinking I could add this in another PR, and keep this one focused and short and sweet. But I want to remove PWM right now, because it's not working.

aulemahal: This strategy is good with me!

coxipi:
I don't know why #1892 was not closed, I thought it was automatic

saschahofmann:
@coxipi I see that you explicitly removed PWM from the methods but in other functions an optional lmoments3 dependency can be used. Any reason why it shouldn't be the case here? (e.g. _fitfunc_1d)

coxipi:
I think it was just due to it being untested.  In #2045 I'm planning to remove any constraint on PWM. You can use any distribution you would like even with SPI (even if it's not the best choice for precipitations). My current way of thinking is that we should allow the use of PWM, but not give any warranty on its correctness since we don't support it officially

PR #1963, opened 2024-10-17
Fix subdaily rate2amount,  add hydro transformation from lwe thickness to amount

### LINKED ISSUE #1962: Can't `rate2amount` with smaller-than-daily frequencies
The line below has not been update since we switched to pandas' new frequency strings. In  v0.49.0, we removed many entries from the  `FREQ_UNITS` dict because the pandas strings are already understandable by `pint`. So at the minimum we should do `FREQ_UNITS.get(base, base)`. 

https://github.com/Ouranosinc/xclim/blob/2bb20af521b7c1c553972a69a07987df8631b0ce/xclim/core/units.py#L670

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1962

* The `_rate_to_amount_converter` relied on the `FREQ_UNITS`  directly, now it goes through `infer_sampling_units` instead. In exchange for slightly more code, this fixes the issue for sub-daily freqs and gives a better error for frequencies xclim can't understand.
* Add the "lwe thickness  -> amount" hydro transformation. The inverse was already there. This is to go from a thickness of liquid water (ex: mm of rain) to an amount (ex: kg m-2), by multiplying by the liquid water density (1000 kg m-3).

No.

### DIFF
--- a/xclim/core/units.py
+++ b/xclim/core/units.py
@@ -91,6 +91,11 @@ hydro.add_transformation(
     lambda ureg, x: x / (1000 * ureg.kg / ureg.m**3),
 )
 hydro.add_transformation(
+    "[length]",
+    "[mass] / [length]**2",
+    lambda ureg, x: x * (1000 * ureg.kg / ureg.m**3),
+)
+hydro.add_transformation(
     "[mass] / [length]**2 / [time]",
     "[length] / [time]",
     lambda ureg, x: x / (1000 * ureg.kg / ureg.m**3),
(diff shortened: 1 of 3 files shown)


xclim/core/units.py
```diff
@@ -461,12 +466,11 @@ def cf_conversion(
 FREQ_UNITS = {
     "D": "d",
     "W": "week",
-    "h": "h",
 }
 """
 Resampling frequency units for :py:func:`xclim.core.units.infer_sampling_units`.
 
-Mapping from offset base to CF-compliant unit. Only constant-length frequencies are included.
+Mapping from offset base to CF-compliant unit. Only constant-length frequencies that are not also pint units are included
```
Zeitsperre, line 473:
```suggestion
Mapping from offset base to CF-compliant unit. Only constant-length frequencies that are not also pint units are included.
```

Zeitsperre (approved):
Elegant.

PR #1971, opened 2024-10-22
Employ a `src`-based layout for code base

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adapts the code base to use a `src`-based layout (in lieu of a flat code base layout)
* Added a stricter `yamllint` configuration (no changes made to existing YAML files)
* Pins `pint` to >=0.24.4 in order to support Python3.13 (and evade failures from `flexparser>=0.4`).

Yes. The code base has been modified such that existing interactive (`--editable`) installations will need to be re-installed.

This change prevents developers from accidentally running tests or importing `xclim` from the root directory. That's essentially the major gain.

Otherwise, we can say that this brings `xclim` closer in line with the other `Ouranosinc/cookiecutter-pypackage` template-based projects.

### Other information:

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
https://www.pyopensci.org/python-package-guide/package-structure-code/python-package-structure.html#the-src-layout-and-testing


pyproject.toml
```diff
@@ -122,7 +123,7 @@ xclim = "xclim.cli:cli"
 "Source" = "https://github.com/Ouranosinc/xclim/"
 "Changelog" = "https://xclim.readthedocs.io/en/stable/history.html"
 "Issue tracker" = "https://github.com/Ouranosinc/xclim/issues"
-"About Ouranos" = "https://www.ouranos.ca/en/"
+"About Ouranos" = "https://ouranos.ca/en/"
```
coxipi, line 127: removing `www.` seems problematic in a web browser, is there a reason to do this?

coxipi (approved):
Looks good to me. Some lines add options (e.g. when `cli` is invoked, new options I don't know about), and I'm not too familiar with these commands

pyproject.toml
Zeitsperre: It used to be the case that `https://ouranos.ca` wouldn't resolve to our home page, but this was explicitly fixed last year. I find it more aesthetically pleasing to drop the `www.`, but I don't mind changing it back.

.yamllint.yaml
```diff
@@ -1,8 +1,39 @@
...
+    allow-non-breakable-words: true
+    allow-non-breakable-inline-mappings: true
     max: 120
     level: warning
+
+  new-lines:
+    type: unix
+
+  trailing-spaces: {}
+
```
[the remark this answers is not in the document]
Zeitsperre: These are rules around the spacing/placement of YAML fields. We already implicitly follow these rules, but this makes it abundantly clear that these are our conventions.

The most important ones here are:
- `allow-non-breakable-words`: Allow lines to surpass 120 characters if they can't be easily broken up (i.e. URLs).
- `new-lines.type`: When people develop on Windows, the encoding for new lines (`CRLF`) is different from Unix/Linux standards (`LF`). This can have impacts on formatting and sha256 sums, so we don't allow them.
- `forbid-duplicated-merge-keys`: All information pertaining to an entry should be found under one unique entry. Having multiple identical keys in a dictionary is apparently allowed in YAML representations. I really don't want that.
- `brackets.min|max-spaces-inside`: This is my personal preference, but I like seeing this: `[_'some-entry', 'another-entry'_]`, but this: `['some-entry', 'another-entry']` is also acceptable.
- The `indentation` rules specify that we don't care whether you use 2 or 3 or 4 spaces to denote entries under a key, so long as you use them consistently throughout the file, therefore the following would not be OK:
```YAML
entry1:
  - something1
  - something2
entry2:
    - otherthing1
    - otherthing2
```

Zeitsperre:
The changes to CLI were needed since the location of `CHANGELOG.rst` can change depending on how the library is installed. It won't necessarily be found two levels above anymore, so passing this file directly via the command line is now permitted.

pyproject.toml
coxipi: right now, I enter this domain in firefox or chromium, and the page never appears, but w/ `www.` it works no problem

Zeitsperre: Well then. Looks like the problem came back. Guess I'll resend the same email I sent last year. Will revert.

.yamllint.yaml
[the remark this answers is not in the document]
coxipi: Thanks for this explanation, interesting! Although, also, sorry you had to write all this, I just meant to say I'm not 100% familiar but it looks legit, so take my approval with a grain of salt on this particular file. It looks almost the same a xscen so I assume it will work fine.

aulemahal (approved):
Looks good!

PR #1973, opened 2024-10-24
Fix indicator outputting delta degC

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Before this fix: an indicator that outputs "degC" but as a "temperature: difference" would bug. The compute function would return a dataset with proper units and proper "units_metadata" attribute and then the indicator would try to convert the data to the expected units, but doing so it would only give the units to  `convert_units_to`. It would not give the `units_metadata` attribute.

thus, `convert_units_to` was taking the compute function output and getting "delta_degC", as expected, but then it was not able to convert to "degC", of course.

This change passes the whole attrs to  `convert_units_to` so it can understand that the expected units are a difference.

No.

We have no current indicator that output "degC" as "difference" (dtr outputs Kelvins), but the problem would emerge in custom indicators or with `sdba.measures.bias` where units are not explicitly set, so the "expected ones" are inferred from the compute function's output. Before this fix, the `units_metadata` attribute was dropped at that point, triggering the bug described above.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,11 +4,12 @@ Changelog
 
 v0.54 (unpublished)
 --------------------
-Contributors to this version: Éric (:user:`coxipi`).
+Contributors to this version: Éric (:user:`coxipi`), Pascal Bourgault (:user:`aulemahal`).
 
 Bug fixes
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


Zeitsperre (approved):
LGTM

PR #1988, opened 2024-11-08
Numpydoc Compliance

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Integrates the `numpydoc` library for docstring validation.
* Fixes the `codespell` configuration (previously was not parsing all relevant files).
* Drops Python3.9 coding conventions everywhere.
* Reformats existing docstrings using `\*` to use `*` instead.
    * On a related note, disabled `flake8`'s `RST210` and `RST213` checks that were flagging `*` symbols in docstrings as unterminated *emphasis* and **strong** tags. 
* Adds a step to the `Indictor.parse_indice` that sanitizes the `*` symbols from `Parameters` entries.
* Painstakingly modifies all docstrings to be closer to the `numpydoc` specification.

It should not. Call signatures and docstrings will be slightly changed.

### Other information:

https://numpydoc.readthedocs.io/en/latest/validation.html

Except for the custom fields that we use to populate our documentation, the docstrings of `xclim` are very consistent and virtually compliant with the standard now.

Both `xclim.sdba` and `xclim.core.missing` are not evaluated by `numpydoc` due to major refactoring efforts that are underway (https://github.com/Ouranosinc/xsdba) or needed (#2000).


xclim/core/indicator.py
```diff
@@ -219,7 +222,18 @@ def update(self, other: dict) -> None:
...
+
+        Parameters
+        ----------
+        other : dict
+            A dictionary of parameters.
+
+        Returns
+        -------
+        bool
+            Whether other can update a parameter dictionary.
```
coxipi, line 235:
```suggestion
        """Return whether `other` can update a parameter dictionary.

        Parameters
        ----------
        other : dict
            A dictionary of parameters.

        Returns
        -------
        bool
            Whether `other` can update a parameter dictionary.
```

xclim/core/indicator.py
```diff
@@ -1285,17 +1339,22 @@ def compute(*args, **kwds):
...
         Default cfchecks use the specifications in `xclim.core.utils.VARIABLES`,
         assuming the indicator's inputs are using the CMIP6/xclim variable names correctly.
         Variables absent from these default specs are silently ignored.
 
         When subclassing this method, use functions decorated using `xclim.core.options.cfcheck`.
+
+        Parameters
+        ----------
+        \*\*das : dict
+            A dictionary of DataArrays to check.
```
coxipi, line 1357: is this really a `dict`?  more like the arguments of a `dict`

xclim/core/missing.py
```diff
@@ -73,13 +86,53 @@ def __init__(self, da, freq, src_timestep, **indexer):
...
+        freq : str
+            Resampling frequency.
+        src_timestep : str
+            The expected input frequency. If not given, it will be inferred from the input array.
+        options : dict
+            Dictionary of options to pass to the instance.
+        indexer : dict
+            Time attribute and values over which to subset the array. For example, use season='DJF' to select winter
+            values, month=1 to select January, or month=[6,7,8] to select summer months.
+            If not indexer is given, all values are considered.
```
coxipi, line 112:
```suggestion
            If no indexer is given, all values are considered.
```
?

xclim/core/units.py
```diff
@@ -282,15 +294,16 @@ def pint_multiply(
     Returns
     -------
     xr.DataArray
+        The product DataArray.
     """
     a = 1 * units2pint(da)  # noqa
-    f = a * q.to_base_units()
+    _f = a * q.to_base_units()
```
coxipi, line 300: why change `f` but not `a` ?

xclim/core/units.py
```diff
@@ -896,11 +926,11 @@ def amount2rate(
 
     Parameters
     ----------
-    amount : xr.DataArray, pint.Quantity or string
+    amount : xr.DataArray or pint.Quantity or str
```
coxipi, line 929:
```suggestion
    amount : xr.DataArray | pint.Quantity | str
```

In other place in xclim we use the pipe, maybe we should uniformize this as well?

xclim/ensembles/_robustness.py
```diff
@@ -399,10 +413,10 @@ def robustness_coefficient(
...
     ----------
-    fut : Union[xr.DataArray, xr.Dataset]
-        Future ensemble values along 'realization' and 'time' (nr, nt). Can be a dataset,
-        in which case the coefficient is computed on each variable.
-    ref : Union[xr.DataArray, xr.Dataset]
+    fut : xr.DataArray or  xr.Dataset
+        Future ensemble values along 'realization' and 'time' (nr, nt).
+        Can be a dataset, in which case the coefficient is computed on each variable.
+    ref : xr.DataArray or xr.Dataset
         Reference period values along 'time' (nt). Same type as `fut`.
```
coxipi, line 420:
```suggestion
    fut : xr.DataArray |  xr.Dataset
        Future ensemble values along 'realization' and 'time' (nr, nt).
        Can be a dataset, in which case the coefficient is computed on each variable.
    ref : xr.DataArray | xr.Dataset
        Reference period values along 'time' (nt). Same type as `fut`.
```

xclim/indices/generic.py
```diff
@@ -1086,6 +1143,7 @@ def thresholded_statistics(
     Returns
     -------
     xr.DataArray
+        The DataArray for the given thresholded statistic.
```
coxipi, line 1146: A lot of time "The DataArray [...]" might just be replaced by "DataArray [...]", is there a constraint by numpydoc, or just a choice? If it's your stylistic preference that's all right.

xclim/indices/generic.py
```diff
@@ -1431,24 +1493,24 @@ def get_zones(
...
-        Zones to be used, either as a DataArray with appropriate units or a list of Quantity
-    exclude_boundary_zones : Bool
+        Input data.
+    zone_min : Quantity, optional
+        Left boundary of the first zone.
+    zone_max : Quantity, optional
+        Right boundary of the last zone.
+    zone_step : Quantity, optional
+        Size of zones.
+    bins : xr.DataArray or list of Quantity], optional
```
coxipi, line 1503: hum ... so leaving out pipe operators is intended?  or could we write: 

```suggestion
    bins : xr.DataArray | list[Quantity], optional
```

xclim/testing/utils.py
```diff
@@ -666,10 +677,6 @@ def populate_testing_data(
     local_cache : Path
         The path to the local cache. Defaults to the location set by the platformdirs library.
         The testing data will be downloaded to this local cache.
-
-    Returns
-    -------
-    None
```
coxipi, line 672: returns a Dataset ? no?

coxipi (approved):
Nice work! Following my comments you might want or not to change conventions, I can check again later

xclim/core/dataflags.py
```diff
@@ -215,11 +234,15 @@ def temperature_extremely_low(
     Parameters
     ----------
     da : xarray.DataArray
+        The DataArray of temperatures being examined.
```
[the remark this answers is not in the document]
coxipi: might keep "being examined". After seeing all the xclim code, I see that a short descriptive comment if often added, so that's all right with me. Stylistically, I think we currently try to avoid starting the description with "The" in xclim? The prototype, simplest example beign: 

```python
da : xr.DataArray
    Input dataset.
```
and the "The input dataset"

pyproject.toml
```diff
@@ -247,6 +248,35 @@ module = [
 ]
 ignore_missing_imports = true
 
+[tool.numpydoc_validation]
+checks = [
+  "all", # report on all checks, except the below
+  "ES01", # "No extended summary found"
+  "EX01", # "No examples section found"
+  "GL01", # "Docstring text (summary) should start in the line immediately after the opening quotes (not in the same line, or leaving a blank line in between)"
```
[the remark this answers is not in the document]
Zeitsperre: I suppose we could follow this. I forgot that I set this, tbh!

xclim/analog.py
```diff
@@ -435,17 +447,31 @@ def kolmogorov_smirnov(x: np.ndarray, y: np.ndarray) -> float:
     :cite:cts:`fasano_multidimensional_1987`
     """
 
-    def pivot(x, y):
-        nx, d = x.shape
-        ny, d = y.shape
+    def pivot(_x: np.ndarray, _y: np.ndarray) -> float:
+        """Pivot function to compute the KS statistic.
```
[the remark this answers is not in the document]
Zeitsperre: It's mostly to differentiate that the variables being used in this sub-function are not the same as those in the parent function. Also, this prevents situations where if someone were to remove the call signature of this function, it would become invalid (whereas previously, the function would have instead used the variables from the parent context).

xclim/core/calendar.py
```diff
@@ -677,18 +691,22 @@ def adjust_doy_calendar(
     max_target_doy = int(target.time.dt.dayofyear.max())
     min_target_doy = int(target.time.dt.dayofyear.min())
 
-    def has_same_calendar():
+    def has_same_calendar(_source, _target):  # numpydoc ignore=GL08
         # case of full year (doys between 1 and 360|365|366)
```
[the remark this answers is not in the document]
Zeitsperre: I agree. If we don't expect users to ever call these methods, they should be made private. I just didn't want to cause any breaking changes with this PR. 

The alternative is to figure out which methods should be explained and adequately document them.

xclim/core/datachecks.py
```diff
@@ -57,18 +59,23 @@ def check_freq(var: xr.DataArray, freq: str | Sequence[str], strict: bool = True
         )
 
 
-def check_daily(var: xr.DataArray):
+def check_daily(var: xr.DataArray) -> None:
     """Raise an error if not series has a frequency other that daily, or is not monotonically increasing.
```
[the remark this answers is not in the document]
Zeitsperre: Haha, yeah, what? Me neither.

xclim/core/dataflags.py
```diff
@@ -80,11 +80,21 @@ def __str__(self):
 ]
 
 
-def register_methods(variable_name=None):
-    """Register a data flag functioné.
+def register_methods(variable_name: str = None) -> Callable:
+    """Register a data flag as functional.
 
-    Argument can be the output variable name template. The template may use any of the stringable input arguments.
+    Argument can be the output variable name template. The template may use any of the string-like input arguments.
```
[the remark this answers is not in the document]
Zeitsperre: I think the complication here is that `xarray.DataArrays` don't return `str` when one is parsing variable names from them, but they instead give `Hashable` (which can be formatted to `str`).

In that sense, `stringable` is actually closer by some definitions, but it kind of also imply arguments that could be "strung together" (like objects with multiple strings?). I felt ad though "string-like" was a better option here.

As for "`str`-like" vs. "string-like", no strong opinions here.

xclim/core/dataflags.py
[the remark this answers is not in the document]
Zeitsperre: How do you feel about:
```python
da : xr.DataArray
    Input temperature DataArray.
```

xclim/core/dataflags.py
```diff
@@ -282,10 +309,12 @@ def negative_accumulation_values(
     Parameters
     ----------
     da : xarray.DataArray
+        The DataArray being examined.
```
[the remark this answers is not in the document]
Zeitsperre: Fair. My only issue is that this isn't a `Dataset` but a `DataArray`.

xclim/core/datachecks.py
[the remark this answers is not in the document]
aulemahal: oups, `not` is superfluous here.
```suggestion
    """Raise an error if series has a frequency other that daily, or is not monotonically increasing.
```

xclim/testing/utils.py
Zeitsperre: Nope, all this function does is fetches data and places it into the cache.

xclim/indices/generic.py
Zeitsperre: It's really minimal, but it's to prevent the styling linter from freaking out that two identical words are following each other ("DataArray DataArray").

Having rewritten many of these now, I find that having full sentence return value descriptions is nicer.

This is entirely a stylistic choice set by me. If you have feelings on this, I can revert.

xclim/ensembles/_robustness.py
Zeitsperre: The spec suggests using plaintext descriptions here: https://numpydoc.readthedocs.io/en/latest/format.html#parameters.

xclim/core/units.py
Zeitsperre: The spec suggests using plaintext descriptions here: https://numpydoc.readthedocs.io/en/latest/format.html#parameters.

xclim/core/units.py
Zeitsperre: Good point. Will look into that.

xclim/core/indicator.py
Zeitsperre: `das` is technically a dictionary because all arguments passed along are seen as members of a dictionary being populated in the call signature. We could be more specific about the types of the values accepted if we want (e.g. `dict of DataArray`).

The docstring needs _something_ and this is the closest generic type for the variable `das`.

xclim/core/indicator.py
```diff
@@ -506,12 +539,17 @@ def _parse_indice(compute, passed_parameters):  # noqa: F841
...
         docmeta = parse_doc(compute.__doc__)
         params_dict = docmeta.pop("parameters", {})  # override parent's parameters
 
         compute_sig = signature(compute)
+        # Remove the \\* symbols from the parameter names
+        _sanitized_params_dict = {}
+        for param in params_dict.keys():
+            _sanitized_params_dict[param.replace("\\*", "")] = params_dict[param]
+        params_dict = _sanitized_params_dict
+
```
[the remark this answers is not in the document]
Zeitsperre: There aren't too many projects that parse the docstrings of functions like we do. Maybe there are better ways, but this is a small but effective work-around. Once rendered, the docs like fine.

xclim/core/indicator.py
```diff
@@ -266,10 +292,17 @@ def __init__(self):
...
     @classmethod
-    def get_instance(cls):
+    def get_instance(cls) -> Any:  # numpydoc ignore=RT05
         """Return first found instance.
 
-        Raises `ValueError` if no instance exists.
+        Returns
+        -------
+        Any
+            FIXME: What is the return type?
```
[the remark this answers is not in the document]
Zeitsperre: @aulemahal Any ideas?

xclim/core/dataflags.py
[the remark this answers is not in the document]
coxipi: yes. my b

xclim/core/indicator.py
coxipi: From the document you linked:

> When documenting variable length positional, or keyword arguments, leave the leading star(s) in front of the name and do not specify a type

Does that apply to \*\* ?

xclim/core/units.py
coxipi: I just checked, this is not like the `_x` and `_y` earlier in a sub-function that you wanted a distinction from the parent function

xclim/core/dataflags.py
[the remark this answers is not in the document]
coxipi: You can keep the "The" and all, I don't have a strong opinion.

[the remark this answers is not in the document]
Zeitsperre: I'll think on whether there's something that fits better.

xclim/core/indicator.py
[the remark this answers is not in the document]
aulemahal: It returns an `Indicator` instance!
```suggestion
        Indicator
             First instance found of this class in the indicators registry.
```

xclim/core/indicator.py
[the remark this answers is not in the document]
aulemahal: Haha, parsing the docstring is indeed not quite orthodox, but it keeps the process of adding indicators « simple ».

xclim/core/indicator.py
aulemahal: Jumping in because I am confused too with the combination of both a r-string (`r""" .... """"`) _and_ the escaped stars `\*\*`. Shouldn't it be one of those, but no both ?

Zeitsperre: I think this may have come about originally from the Indicator docstring parsing logic. This has now been addressed. No more `\` needed when the docstring is raw.

xclim/core/units.py
Zeitsperre: The IDE was complaining because of this earlier in the file:
```python
with (files("xclim.data") / "variables.yml").open() as f:
    CF_CONVERSIONS = safe_load(f)["conversions"]
_CONVERSIONS = {}
```
Will rename this here and revert the changes in the function.

PR #1996, opened 2024-11-13
Add `q` (streamflow) to variables.yml

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add `q` as a proper variable in variables.yml

No

### Other information:

Discussing with @RondeauG , we were deliberating between `qstr`, `qstrm`, `strf`. I haven't seen any of these variables in many places in litterature, there doesn't seem to be a standard abbreviation apart from `Q` or `q` (flow or specific flow ), so I'm enclined to change `strf` if it doesn't feel right.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -22,11 +22,38 @@ jobs:
           egress-policy: block
           allowed-endpoints: >
             api.deps.dev:443
+            api.electricitymap.org:443
             api.github.com:443
+            api.green-coding.io:443
+            api.securityscorecards.dev:443
             github.com:443
+            ip-api.com:80
+            ipapi.co:443
+            proxy.golang.org:443
+            pypi.org:443
+            sum.golang.org:443
+
+      - name: Start Energy Measurement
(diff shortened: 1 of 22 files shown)


aulemahal:
But if `q` _is_ the most common standard, why would we not use that ?

coxipi:
> But if `q` _is_ the most common standard, why would we not use that ?

I just thought that the variable associated with the streamflow should have its standard naming declared with units in `variables.yml`. Then, if we define `q` as a streamflow in `variables.yml`, I'm not sure it's a good idea to use it as a variable for quantiles, no?

aulemahal:
Most references to `q` as "quantile" I see in xclim are in `sdba`, where I would assume context is enough to make the difference ? There are no indicators that refer to quantiles as `q`, only `xc.indices.stats.parametric_quantile` does it and I would argue again that context would be enough here.

I think my point is that such a breaking change doesn't seem justified if it is not based on convention.

coxipi:
Ok, I've just replaced the `streamflow` entry in variables.yml with `q`  then.

RondeauG (approved):
Sorry, I didn't see that you had requested a review from me!

Zeitsperre (approved):

PR #2019, opened 2024-12-11
Correctly interpolate seasons in Grouper

### LINKED ISSUE #2014: Linear interpolation for seasonal bias adjustment leads to non-smooth distribution
### Setup Information

- Xclim version: 0.52.2
- Python version: 3.11.6
- Operating System: Linux

### Description

If I change the grouper from the [first example of the docs](https://xclim.readthedocs.io/en/stable/notebooks/sdba.html) to use `time.season` instead of `time.month`. I get some drastic transitions when the seasons change:
image

I remember we did additions to enable the interpolation for these string columns of seasons but now I am not sure its working correctly. 

### Steps To Reproduce

Use the example from the tutorial or paste the following
```python
import nc_time_axis  # noqa
import numpy as np
import xarray as xr

%matplotlib inline
plt.style.use("seaborn-v0_8")
plt.rcParams["figure.figsize"] = (11, 5)

# Create toy data to explore bias adjustment, here fake temperature timeseries
t = xr.cftime_range("2000-01-01", "2030-12-31", freq="D", calendar="noleap")
ref = xr.DataArray(
    (
        -20 * np.cos(2 * np.pi * t.dayofyear / 365)
        + 2 * np.random.random_sample((t.size,))
        + 273.15
        + 0.1 * (t - t[0]).days / 365
    ),  # "warming" of 1K per decade,
    dims=("time",),
    coords={"time": t},
    attrs={"units": "K"},
)
sim = xr.DataArray(
    (
        -18 * np.cos(2 * np.pi * t.dayofyear / 365)
        + 2 * np.random.random_sample((t.size,))
        + 273.15
        + 0.11 * (t - t[0]).days / 365
    ),  # "warming" of 1.1K per decade
    dims=("time",),
    coords={"time": t},
    attrs={"units": "K"},
)

ref = ref.sel(time=slice(None, "2015-01-01"))
hist = sim.sel(time=slice(None, "2015-01-01"))
QM_mo = sdba.QuantileDeltaMapping.train(
    ref, hist, nquantiles=15, group="time.season", kind="+"
)
scen = QM_mo.adjust(sim, extrapolation="constant", interp="linear")

ref.groupby("time.dayofyear").mean().plot(label="Reference")
hist.groupby("time.dayofyear").mean().plot(label="Model - biased")
scen.sel(time=slice("2000", "2015")).groupby("time.dayofyear").mean().plot(
    label="Model - adjusted - 2000-15", linestyle="--"
)
scen.sel(time=slice("2015", "2030")).groupby("time.dayofyear").mean().plot(
    label="Model - adjusted - 2015-30", linestyle="--"
)
plt.legend()
```

### Additional context

_No response_

### Contribution

### Code of Conduct

saschahofmann:
### Pull Request Checklist:
    - This PR fixes only part of #2014 

This PR adds a line to correctly interpolate seasonal values. It also changes the test_timeseries function that now accepts a `calendar` argument instead of `cftime`. Not providing it or providing `None` is equivalent to `cftime=False` and `calendar='standard` to the previous `cftime=True`. This allows for testing different calendar implementations e.g. 360_day calendars


saschahofmann:
I just realised that the factor of 1/6 is assuming that all seasons have the same length which in gregorian calendars is not necessarily true but I am not sure it matters too much at least the function should be smooth.

saschahofmann:
Just to prove that this leads to a smooth result, same input as in the issue: 
image

saschahofmann:
Weirdly and contrary to what I showed yesterday, today I am still getting clear transitions as if there still wasn't any linear interpolation.

saschahofmann:
I reinstalled xclim but I am still getting very similar results to before the "fix". You have any advice on where else I could look?

Zeitsperre:
> I reinstalled xclim but I am still getting very similar results to before the "fix". You have any advice on where else I could look?

Could it be that you have obsolete `__pycache__` folders still among your cloned folders? @coxipi is looking into recreating your example based on your branch for validation, but if the tests are working as intended on CI, then it's likely a caching/installation issue.

coxipi:
I managed to install the environment

* I confirmed that the function has the appropriate modifications inside the notebook I'm using 

```python
import inspect
print(inspect.getsource(sdba.base.Grouper.get_index))
```

* I also find that the interpolation is wrong. 

I'll try to have look later. Maybe the `interp` boolean condition is not triggered properly?

saschahofmann:
I am pretty sure that the `get_index` function is updated in my notebook. Either I am wrong in expecting a smoother result (it seems to have changed slightly to what I got earlier) or there is something else going on. I will keep investigating

coxipi:
It's simply `interp` which can't be "nearest", otherwise no interpolation takes place ... coming up with an example, but just wanted to comment now so you take a pause if you're troubleshooting

coxipi:
There is clearly something wrong going on. Comparing 
hist - scen_month
scen_time - scen_month
scen_season - scen_month
scen_month - scen_month

scen_season is way off

image

saschahofmann:
@coxipi I think only mention this in the original issue: my analysis is done with `QuantileDeltaMapping` instead of `EmpiricalQuantileMapping`. Here the equivalent chart to yours for that: 
image
season still seem kinda weird

saschahofmann:
A similar trend becomes apparent when looking at the adjusted - historical (now for EmpiricalQuantileMapping) 
image


The first season still seems to track the monthly adjustment quite well, but the behaviour completely changes after the second adjustment.

coxipi:
Yes, I have seen simlilar things by playing with the choice of how `get_index`. I feel this should not be this sensitive. Let me try and get this back

saschahofmann:
Weirdly, the abrupt changes dont coincide with the seasons 
image

saschahofmann:
Maybe little surprising but the interpolated `af` also shows these weird patterns.
image

What I don't fully understand: I assume that the new quantile is in respect to the current season? Which in this case would mean that the during the transitions the values are extremely different ?

coxipi:
Ok, I'll try to interpolate season adjusting factors to months to compare

coxipi:
Working the season adjustment factors interpolated to the months, then use the interpolated adjustment factors as if I was doing a monthly adjustment

image

saschahofmann:
I guess that makes sense. If I am right above and the problem is that we use a single quantile determined for a season when interpolating the af than it'd make sense that when you interpolate to month the difference between the two is smoother.

saschahofmann:
My suggestion is for a single future value:
1. Create a timeseries (read season series) by grabbing the quantiles **respective** to each season. So for example: the 28th Feb in our example timeseries is around the 99th percentile for season 0 but in the 1st percentile for season 1.
2. For this series we interpolate to the season float index. E.g in the case of 29 Feb to 0.5

coxipi:
Same idea with dayofyear:

image

image


I have been very sloppy in implementing in this, but I just wanted to confirm we can have decent interpolation results from the season. So it's most likely and the way we do the interpolation that is not appropriate. Although what you discuss about quantiles vs. dayofyear is interesting, I think the problem is more basic

coxipi:
> What I don't fully understand: I assume that the new quantile is in respect to the current season? Which in this case would mean that the during the transitions the values are extremely different ?

I am not sure what you mean by the "new quantile is in respect to the current season"

saschahofmann:
Ok I think my comment is only true for QuantileDeltaMapping. I am not understanding why it happens for EmpireQuantileMapping.

In the former case, the first thing we do is compute the quantile for a value in respect to a season
```python
sim_q = group.apply(u.rank, ds.sim, main_only=True, pct=True)
```

I could imagine that when we do linear interpolation that doesn't make too much sense. Instead the quantile for a value should be computed for each season separately and then we should interpolate these values at the decimal season point.

saschahofmann:
I do agree that there is something in the interpolation. I am looking at the plot of the interpolated `af` and I am very confused. 
image
For two periods it seem to interpolate something correctly but the rest of the time it throws all sense overboard 😓 

You think its in `_interp_on_quantiles_2D`?

coxipi:
> I could imagine that when we do linear interpolation that doesn't make too much sense. Instead the quantile for a value should be computed for each season separately and then we should interpolate these values at the decimal season point.

I see your point.  

* Basic interpolation of EmpiricalQuantileMapping (EQM): Discrete data {sim_q, af_q} -> Interp to continuous function: af = f(sim).
* Basic interpolation of QuantileDeltaMapping: Discrete data {q, af_q} -> Interp to continuous function: af = f(rank).
* After we add a time grouping index in the mix, e.g. for EQM the discrete data is {season, sim_q,af_q}. The idea is to interpolate this to a bivariate function `af = f(dayofyear, sim)`. In this case, it makes sense to think of the sim values as being distributed along dayofyear, we have the orginal timeseries for these values.
* If we try this in QDM, the final function will look like `af = f(dayofyear, rank)`. But the notion of rank is dependent on the time grouping. Saying that we want a mix of the adjustment of season-1 and season-2 when we are the boundary of the two seasons is a bit weird, since the ranks were computed for season-1 or season-2, but not the period in between. 

To give a concrete example, consider tasmax on Feb.28. Its rank is computed in DJF. But if we included it in MAM, it would be a lower rank. We are saying that we have an interpolated continuous function that we want to apply on those ranks, but these values are still segmented, there are four groups of ranks in a year.

saschahofmann:
For the QDM case, I imagined it to be like this: for every sim value you get the `af` per season (or month) and then interpolate this "timeseries" to the time of the sim value. In your example of Feb 28, you would grab the `af` value for DJF at rank 0.9 and for MAM at rank 0.3 add the values and divide by 2 (since Feb 28 is at season 0.5). 


Understand I correctly, that you would do the interpolation for EQM by first creating a yearly distribution per dayofyear?

coxipi:
> Understand I correctly, that you would do the interpolation for EQM by first creating a yearly distribution per dayofyear?

Yes, that is what I was doing. There is a first interpolation, convert the series of af_q obtained seasonally to a group of af_q interpolated to each dayofyear. Then, simply proceed as if the adjustment had been on dayofyears. Not temporal interpolation left, only a second uni-dimensional interpolation on the quantiles. For QDM, this means that the ranks of sim would be obtained individually for each dayofyear.

I was only trying to see if there is an interpolation that can work reasonnably well in this context, I'm not sure if it's the correct way to go.  It's not a 2D interp, so it's quite different from the current implementation.

> For the QDM case, I imagined it to be like this: for every sim value you get the af per season (or month) and then interpolate this "timeseries" to the time of the sim value. In your example of Feb 28, you would grab the af value for DJF at rank 0.9 and for MAM at rank 0.3 add the values and divide by 2 (since Feb 28 is at season 0.5).

I was thinking something like this, but what if you're more in season-1, e.g.  Jan 31rst 20XX. Do you just add the specific value among the MAM group of values to compute "a rank of  Jan 31rst 20XX within MAM"? It might make more sense to always compute the rank for a window of ~90 days around the day in question. In this case, this is starting to ressemble an adjustment with "time.dayofyear" with a window of 90. The difference is simply that the training data, you obtain adjustment factors not for every day, but the four ranges of ~90 days (AKA seasons)

saschahofmann:
I was thinking that the timeseries you build for every value would be linearly interpolated like right now.

You build a timeseries with the respective quantiles for DJF at 0, MAM at 1, JJA at 2 and SON at 3 and then proceed to interpolate the value for the day of year ( I think Jan 31 is something like 0.173)

saschahofmann:
I guess that will probably lead to similar results? Only tricky thing are different calendar year lengths e.g. we work a lot with 360_day calendars

coxipi:
> What I still dont understand why we see a similar problem for the EmpiricalQuantileMapping? If I read the code correctly the simu values are directly interpolated to the historical values? not sure how the quantiles are considered in this?

They aren't. I think the interpolation makes more sense as-is in this context, I agree. Maybe the interpolation is not working as expected

coxipi:
## interpolating the results of training on dayofyear first, then perform a dayofyear adjustment (`adj-season-interp_on_doy`)
image
image

## interpolating the results of training on month first, then perform a month adjustment, with the 2d interpolation (`adj-season-month`)

image
image

coxipi:
Even with this alternative implementation, the results from "time" are closer to "month" than "season", and that seems wrong

coxipi:
I think the season adjustment just doesn't work, forget about interpolation?

saschahofmann:
Uff ok. I guess we would expect to be the same, wouldn't we?

saschahofmann:
af's are exactly the same so it's not the training:
image

saschahofmann:
I think it leads to the same problem: the interpolated af's are just wrong:
```python
ds = eqm_time.ds
group = "time"
sim = hist_spring
af = u.interp_on_quantiles(
    sim,
    ds.hist_q,
    ds.af,
    group=group,
)

ds = eqm_season.ds
group = "time.season"
sim = hist
af_season = u.interp_on_quantiles(
    sim,
    ds.hist_q,
    ds.af,
    method='nearest',
    group=group,
)
af_season = af_season.where(af_season.time.dt.month.isin([3,4,5]), drop=True)

af.groupby('time.dayofyear').mean().plot(label='interpolated af time')
af_season.groupby('time.dayofyear').mean().plot(linestyle='--', label='interploated af time.season')
plt.legend()
```
image

coxipi:
Great point. Just to summarize:

|af agree|scen disagree|
|-|-|
|image|image|


The problem exists even without interpolation though, with interp="nearest". That why I wanted to compare the "time" adjustment with only MAM in the timeseries vs. the "time.season" adjustment, WITHOUT  interpolation. I believe in this specific case, this should be equivalent

saschahofmann:
Sorry my bad, interpolation in this case means `nearest` interpolation for each time value, as opposed to the `af` value provided by the training (with dimensions quantile [and season])

coxipi:
In any case, the mean over the 15-year period should be closer to the target, time.season is way off (and in fact, I maintain, it should be equal to the time adjustment.)

image

> Sorry my bad, interpolation in this case means nearest interpolation for each time value, as opposed to the af value provided by the training (with dimensions quantile [and season])

Got it, sorry I read too fast. Good observation. It seems `_interp_on_quantiles_2D` is messing up `season` even without interpolation, I will have to look into this

saschahofmann:
I think I am getting closer. I tried to replicate whats happening inside `_interp_on_quantiles_2D`, for a single value on 2000-04-15. Above that value would be -1.23223145
```python
from scipy.interpolate import griddata
oldx = qdm_season.ds.hist_q.sel(season='MAM')
oldg = np.ones(20)
oldy = qdm_season.ds.af.sel(season='MAM')
value = hist.sel(time='2000-04-15')
newx = value
newg = u.map_season_to_int(value.time.dt.season)
griddata(
    (oldx.values, oldg),
    oldy.values,
    (newx.values, newg),
    method='nearest'
)
```
If I didn't make mistake that should mimick the function but it leads to 0.20145109.

saschahofmann:
Ok I just need to find the difference between this code and the one running `_interp_on_quantiles_2D`. Chart shows the af obtained from EQM with group='time' as above and the one manually computed.
```python
from scipy.interpolate import griddata

oldx = qdm_season.ds.hist_q.sel(season="MAM")
oldg = np.ones(20)
oldy = qdm_season.ds.af.sel(season="MAM")
value = hist_spring
newx = value
newg = u.map_season_to_int(value.time.dt.season)
afs = griddata((oldx.values, oldg), oldy.values, (newx.values, newg), method="nearest")

af.groupby("time.dayofyear").mean().plot(label="interpolated af time")
af_corrected = xr.DataArray(afs, coords=dict(time=hist_spring.time))
af_corrected.groupby('time.dayofyear').mean().plot( linestyle="--",
                                                   label='seasonal af manually computed with griddata')
plt.legend()
```

image

coxipi:
Great, I have reproduced your example

```python
def _interp_on_quantiles_2d(newx, newg, oldx, oldy, oldg, method, extrap):
    mask_new = np.isnan(newx) | np.isnan(newg)
    mask_old = np.isnan(oldy) | np.isnan(oldx) | np.isnan(oldg)
    out = np.full_like(newx, np.nan, dtype=f"float{oldy.dtype.itemsize * 8}")
    if np.all(mask_new) or np.all(mask_old):
        warn(
            "All-nan slice encountered in interp_on_quantiles",
            category=RuntimeWarning,
        )
        return out
    out[~mask_new] = griddata(
        (oldx[~mask_old], oldg[~mask_old]),
        oldy[~mask_old],
        (newx[~mask_new], newg[~mask_new]),
        method=method,
    )
    # if method == "nearest" or extrap != "nan":
    #     # 'nan' extrapolation implicit for cubic and linear interpolation.
    #     out = _extrapolate_on_quantiles(out, oldx, oldg, oldy, newx, newg, extrap)
    return out

def bad_af_season(afq, hist):
    oldg = np.ones(20)
    oldy = afq.sel(season="MAM")
    value = hist.where(hist.time.dt.month.isin([3,4,5]), drop=True)
    newx = value
    newg = u.map_season_to_int(value.time.dt.season)
    afs = _interp_on_quantiles_2d(newx,newg,oldx,oldy,oldg, "nearest", "nan")
    return xr.DataArray(afs, coords=dict(time=value.time))
    
af.groupby("time.dayofyear").mean().plot(label="interpolated af time")
good_af_season(qdm_season.ds.af,hist).groupby('time.dayofyear').mean().plot( linestyle="--",
                                                   label='seasonal af manually computed with griddata')
bad_af_season(qdm_season.ds.af,hist).groupby('time.dayofyear').mean().plot( linestyle="--",
                                                   label='seasonal af manually computed with xclim internals')
plt.legend()
```

And I get the same. I was having some problems with numba, I commented the "extrapolate", but we have problems when method!="nearest" too, so that should not be the problem. So now I suspect that it might be because of the arguments 
`newx, newg, oldx, oldy, oldg`, I've built them in your way, I have to check what is done in xclim

image

saschahofmann:
Yes I'm also still trying to figure out the difference but I guess I ran into the same problem as you with the extrapolation

coxipi:
> I am quite surprised that there are no more spikes in the nearest after all I would have thought that this would kinda be expected.

I see your point. Maybe the sufficiently high number of quantiles (20 or 50) and the fact that you average over 15 years is enough to make this smooth. If you look directly at the time series, the "nearest" should be less smooth?

Anyways, great work, thanks a lot!

coxipi:
> Ah of course I saw the OR but not !='nan' 🤦 . 

I made the same error above when I commented the extrapolate haha...

coxipi:
Hum, in the QDM case, the linear interpolation seems to have some issues?

saschahofmann:
Ok here the linearly interpolated `af`s for QDM: 
image

as comparison nearest:
image


I believe this is due to the problem I mentioned in December and better summarised by [you](https://github.com/Ouranosinc/xclim/pull/2019#issuecomment-2554134740)

saschahofmann:
Do you have any resources on someone else doing this? The originally paper Cannon et al. (2015 ) doesn't seem to look at monthly or seasonal adjustments/ the only thing I could find was them saying to use a sliding window:
> To correct biases in the seasonal cycle, the quantile mapping algorithms are applied to pooled daily data falling within sliding 3-month windows centered on the month of interest. For example, correction of data from December would include days from November to January, correction of data from January would include days from December to February, and so forth. Time-dependent means and empirical quantiles of projected data are calculated over 30-yr sliding windows centered on the year of interest

coxipi:
No, unfortunately, I'm searching right now. @aulemahal , Sascha fixed one problem, in the extrapolation, it was assumed that values of season/month would start at 0, but for season with periodic condition, it can go below zero, this needed a change. 

But now, we have the problem I describe here, e.g.:

> To give a concrete example, consider tasmax on Feb.28. Its rank is computed in DJF. But if we included it in MAM, it would be a lower rank. We are saying that we have an interpolated continuous function that we want to apply on those ranks, but these values are still segmented, there are four groups of ranks in a year.

Do you agree this is a problem? Do you know if people explored specifically the use of QDM with seasons in the litterature?

saschahofmann:
I was also thinking the other option for a fix for the extrapolation without needing to change the extrapolate function is changing the mapping of the seasons to start at 1 (so that cyclic_bounds would add 0 and 5). Not sure which of the two you prefer. I think the current fix might be more robust to future changes because it simply uses the coordinate values.

aulemahal:
> Do you know if people explored specifically the use of QDM with seasons in the litterature?

I don't think I have ever read such a paper (neither for QDM nor for any other QMs). Maybe @huard remembers if we had sources in mind when implementing it ?

My "fear" is that we implemented it because it was possible, because  `time.dt.season` existed in the code.

saschahofmann:
The same problem exist for months I think.

saschahofmann:
I am moving the discussion to #2048 and will leave the issue #2014 open for now. As discussed in the development meeting I will add warnings that link to that discussion. Where should I place these warnings?

I guess I could raise a `logging.warning` in `qdm_adjust', when grouping is not None and interp='linear'? I was thinking to also put one in the docs, but I am not sure what would be the right place to do that.

Once this is done, I suggest we merge this in since all changes still apply IMO.

src/xclim/sdba/_adjustment.py
```diff
@@ -587,6 +587,13 @@ def qdm_adjust(ds: xr.Dataset, *, group, interp, extrapolation, kind) -> xr.Data
             sim : Data to adjust.
     """
     sim_q = group.apply(u.rank, ds.sim, main_only=True, pct=True)
+    if group.prop and interp != "nearest":
+        import logging
+
+        logging.getLogger("xclim").warning(
+            f"Using a {interp} interrpolation with QuantileDeltaMapping might create sudden jumps between different"
+            " groups. See discussion https://github.com/Ouranosinc/xclim/discussions/2048 for more information "
+        )
```
Zeitsperre, line 596: Looks good!

I think the `warnings` library is a better fit here: 
```suggestion
        msg = (
            f"Using a {interp} interrpolation with QuantileDeltaMapping might create sudden jumps between different"
            " groups. See discussion https://github.com/Ouranosinc/xclim/discussions/2048 for more information."
        )
        warnings.warn(msg)
```

saschahofmann:
I discovered another issue when running QDM with a 360_day calendar. To reproduce you can just convert ref and hist with `.convert_calendar('360_day', align_on='year')`.

One of the issues is related to @aulemahal PR #2038, here I am using a different approach.
Could someone check changes in base.py get_coordinates and in utils.pyL478?

saschahofmann:
And another fix for using sdba.Scaling in utils.pyL222

src/xclim/sdba/base.py
```diff
@@ -196,8 +198,7 @@ def get_coordinate(self, ds: xr.Dataset | None = None) -> xr.DataArray:
             if ds is not None:
                 cal = get_calendar(ds, dim=self.dim)
                 mdoy = max(
-                    xr.coding.calendar_ops._days_in_year(yr, cal)
-                    for yr in np.unique(ds[self.dim].dt.year)
+                    xr.coding.calendar_ops._days_in_year(ds[self.dim].dt.year, cal)
```
aulemahal, line 201: Indeed, this fixes it, but it has the disadvantage that it continues to use a private xarray function. The solution in #2038 is independent of xarray, even if not as fancy. The main difference is in the case where the passed dataset only covers non-leap years (and has a standard calendar). Do you think that case is probable ?

saschahofmann:
Merged @aulemahal fix and this is now ready for merging

aulemahal (approved):
Looks good! Thanks!

I think we can avoid a breaking change in `test_timeseries`.

src/xclim/testing/helpers.py
```diff
@@ -208,16 +208,18 @@ def test_timeseries(
         The frequency of the time dimension. Default is daily/"D".
     as_dataset : bool
         Whether to return a Dataset or a DataArray. Default is False.
-    cftime : bool
-        Whether to use cftime or not. Default is False.
+    calendar : str or None
+        Whether to use a calendar. If a calendar is provided, cftime is used.
```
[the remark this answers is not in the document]
saschahofmann: Reverted the removal of the cftime argument and added a line to the changelog

src/xclim/sdba/utils.py
```diff
@@ -475,9 +477,10 @@ def interp_on_quantiles(
         return out
 
     if prop not in xq.dims:
-        xq = xq.expand_dims({prop: group.get_coordinate()})
+        prop_coords = group.get_coordinate(newx)
```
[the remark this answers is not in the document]
saschahofmann: Good point. I am not sure whether this can happen or not put I put it outside the if stagement

saschahofmann:
From my side this looks good not sure how to trigger the coverage pipeline?

Zeitsperre (approved):

saschahofmann:
@Zeitsperre I added these two lines to allow linear interpolation with `dayofyear` but I am wondering whether this combination makes sense? 
https://github.com/Ouranosinc/xclim/pull/2019/files#diff-3fdadf9776d54afe05de332ab8810fc3d2939d75bd98273f3661f05d71f296a4R314-R315

coxipi:
> @Zeitsperre I added these two lines to allow linear interpolation with `dayofyear` but I am wondering whether this combination makes sense? https://github.com/Ouranosinc/xclim/pull/2019/files#diff-3fdadf9776d54afe05de332ab8810fc3d2939d75bd98273f3661f05d71f296a4R314-R315
> 
> It was failing in a test

I don't think there would be a point in doing linear interpolation for dayofyear. We have adjusting factors for each dayofyear, so we don't have in-between values where we need to interpolate the training data. It was failing in a test before?

saschahofmann:
Yes. Its this test https://github.com/Ouranosinc/xclim/actions/runs/12985440090/job/36210393696 

I fixed it by allowing again linear interp of dayofyear. I think it was allowed previously but I slightly restructured the code so now by default it was disabled.

saschahofmann:
I reverted the change that allows linear interp with dayofyear grouping and instead changed the test to use nearest

saschahofmann:
@coxipi I think maybe reenabling linear interpolation from dayofyear for now is a better solution. Since otherwise it represents a breaking change, e.g. right now the tutorial does some similar (hence the breaking test in sdba-advanced.ipynb) where we do EQM with dayofyer with linear interp.

saschahofmann:
@aulemahal or @coxipi . Would love to get this over the line I think if you agree I add a DeprecationWarning and then ready to merge?

coxipi:
Hi @saschahofmann sorry for the late response, we were busy with a seminar last week. 

One thing I'm starting to realize is that the linear interpolation can act on two fronts, on the grouping dimension (e.g. season, month) but also on the quantiles. We really want to leave open the possibility of linear interpolation on quantiles I think. 

If you want to leave things as they were in the most economical way possible (if you need to deactivate a test, go for it), I really think this is beyond the scope of your PR anyways, you fixed many things indepently of this. I can take care of this in another PR. 

@aulemahal, I think that in `utils`, we don't use the 2D interp function for `group=="prop"`. We should not use it for `group=="dayofyear"` either, agreed? `interp` should still be possible to use with quantiles, it's just implicit that dayofyear grouping will also be 1-dimensional
```python
    if prop in ["group", "dayofyear"]:
        if prop =="group":
            if "group" in xq.dims:
                xq = xq.squeeze("group", drop=True)
            if "group" in yq.dims:
                yq = yq.squeeze("group", drop=True)
        out = xr.apply_ufunc(
            _interp_on_quantiles_1D,
            ...
```

saschahofmann:
I guess we can merge then 😬

coxipi (commented):
just commenting so I can commit my suggestion

coxipi:
Feel free to merge when it becomes possible again, we're just waiting on tests

saschahofmann:
I dont have write access so cant merge

coxipi:
> I dont have write access so cant merge

Ah, right.

Thanks again for your nice contribution!

sylvainmarchi:
Hello,
I’m having trouble understanding this thread and the one titled "Correctly interpolate seasons in Grouper #2019." Could someone clarify the role of the interp argument, particularly in the context of MBCn as shown below? Does it apply to the group dimension, the quantiles, or both?
Thank you,
Sylvain

group = sdba.Grouper("time.dayofyear", window=31)
```
ADJ = sdba.MBCn.train(
	ref,
	hist,
	base_kws={"nquantiles": 50, "group": group},
	adj_kws={"interp": "linear", "extrapolation": "constant"}, 
	n_iter=20,  # perform 20 iteration
	n_escore=1000,  # only send 1000 points to the escore metric
)
```

coxipi:
Hi Sylvain, 

In this case of `sdba.Grouper("time.dayofyear", window=31)`, a linear interpolation will only be performed on the quantiles. Consider the adjustment on dayofyear-1. I will label the subset with rolling windows as ref1w31, hist1w31. We want to adjust the simulation on dayofyear-1 with this, sim1. You find the values of 50 equally spaced quantiles:

{ref1w31_q01, ref1w31_q03, ... , ref1w31_q99}
{hist1w31_q01, hist1w31_q03, ... , hist1w31_q99}

Compute adjustment factors from those
{af1_q01, af1_q03, ..., af1_q99}

Then the linear interpolation gives you a continuous function of the adjustment factors: 
af1(q). You want to apply these adjustment factors on the ranks of sim1.

In the particular case of day of year grouping, no more interpolation is needed, we have adjustment factors af1 for sim1, af2 for sim2, af3 for sim3, etc.

When working with seasons, we have less adjustments factors, instead of af1,af2, ... af365, we only have afDJF, afMAM, afJJA, afSON. But we would still like to have distinct adjustment factors for each dayofyear, so apply different adjustment to sim1, sim2, ... sim365. To achieve this, we interpolate the results between seasons {afDJF, afMAM, afJJA, afSON}. So in this sense, there is also an interpolation on the grouping dimension, not only on quantiles. My description of what is done is not exact, because the two-fold interpolation on quantiles and season is done in one step, but I think it still schematizes well what is happening. 

Hope this clears this up!

Éric

sylvainmarchi:
Hello Éric,
Thank you for your very clear explanation.
Sylvain

src/xclim/sdba/base.py
```diff
@@ -293,21 +295,39 @@ def get_index(
...
+                    else 365 + (0 if calendar == "noleap" else ind.is_leap_year)
+                )
+                # This is assuming that seasons have the same length. The factor 1/6 comes from the fact that
+                # the first season is shifted by 1 month the but the middle of the season is shifted in the other direction
+                # by half a month so -(1/12-1/24)*4 = -1/6
+                i = ind.dayofyear / length_year * 4 - 1 / 6
+            else:
+                raise ValueError(
+                    f"Interpolation is not supported for {self.dim}.{self.prop}."
+                )
```
coxipi, line 317: darn, I thought we had removed this in the end, oupsy

PR #2040, opened 2025-01-10
Clean and export the units registry

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Remove support for non-CF aliases : "cms" (m3 s-1), "mmday" (mm d-1) and "pct" (%). These are not supported by `udunits2` and thus I don't see why we should support them ? @huard, @RondeauG what do you think ? I'm not strongly opinionated on this.

* Set the unit registry as pint's "application registry" to make it usable by non-xclim aware applications. To be obtained with `ureg = pint.get_application_registry()`.

* Units of quantities printed in indicator attributes are now formatted using the CF syntax (ex: "mm d-1" instead of "mm/d"). Not sure how I missed that earlier...

* Clean up in the code of the registry declaration and addition of comments to explain what is happening. 

* A small change in `dataflags` that was actually unneeded at the end, but it felt cleaner to me, so I kept it.

Some units are not supported anymore (`cms`, `pct` and `mmday`).
Some attributes will be different as the units are printed differently (for fractions mostly).

### Other information:
The registry export allows an application/module that is units-agnostic in itself to be used with xclim. For those applications / usages, the import order might be significant though : `xclim` has to be the last `set_application_registry` done.  Specifically in the climate case, as  `xclim.core.units` imports `cf_xarray.units` (which also sets the app registry), a subsequent `import cf_xarray.units` is not problematic as python will _not_ actually execute it. 

@coxipi  Incoming PR to `xsdba` to use exactly that.


tests/test_units.py
```diff
@@ -58,16 +57,13 @@ def test_dimensionality(self):
         with units.context("hydro"):
             fu = 1 * units.parse_units("kg / m**2 / s")
             tu = 1 * units.parse_units("mm / d")
-            fu.to("mmday")
-            tu.to("mmday")
+            fu.to("mm/day")
+            tu.to("mm/day")
```
Zeitsperre, line 61: I'm a bit confused here; Are we testing to see that transforming this twice doesn't raise an error?

src/xclim/core/indicator.py
```diff
@@ -1311,7 +1311,7 @@ def _format(
         # Add formatting {} around values to be able to replace them with _attrs_mapping using format.
         for k, v in args.items():
             if isinstance(v, units.Quantity):
-                mba[k] = f"{v:g~P}"
+                mba[k] = f"{v:gcf}"
```
Zeitsperre, line 1314: I'm sure this works, but I'm not clear what the change here is.

src/xclim/core/units.py
```diff
@@ -64,28 +63,21 @@
...
-    units.default_format = "gcf"
-# Switch this flag back to False. Not sure what that implies, but it breaks some tests.
-units.force_ndarray_like = False  # noqa: F841
-# Another alias not included by cf_xarray
-units.define("@alias percent = pct")
-
-# Default context.
-null = pint.Context("none")
-units.add_context(null)
+units.formatter.default_format = "gcf"
```
Zeitsperre, line 66: `xclim` still supports `numpy>=1.23`; Won't this cause issues?

Zeitsperre (approved):
LGTM, but some clarifications would be nice to have!

huard:
The non-standard units are stuff that we've seen in the wild. I suspect that removing them here will force us to make explicit unit changes in RavenPy and maybe xhydro. 

I'm not sure I see why we'd remove them if it doesn't cause us pain. I think our stance is "we'll make sure CF-compliant files work", not "non CF-compliant files will fail".

aulemahal:
I can put them back. 
But my long-term goal is to move units-handling elsewhere, most probably `cf-xarray`. And I doubt that this is a good place for those aliases (they already copied the registry from us and _did not_ copy those).

src/xclim/core/units.py
aulemahal: The numpy condition was a confusion from me! The issue was actually with pint, which we already pin to the correct version.

src/xclim/core/indicator.py
aulemahal: The units are formatted with the `cf` formatter (defined in `cf-xarray`), instead of the short (`~`) pretty (`P`) formatter.  They look a lot alike, the main difference being in how they express negative powers / fractions.

tests/test_units.py
aulemahal: I blindly modified the weird unit to the good one. The last line is useless, but the other tests the hydro context.

huard:
> I can put them back. But my long-term goal is to move units-handling elsewhere, most probably `cf-xarray`. And I doubt that this is a good place for those aliases (they already copied the registry from us and _did not_ copy those).

Ahh, I see now.  @Zeitsperre expect some breakage in notebooks and tests from ravenpy and pavics-sdi.

Zeitsperre:
I'm prepared for that if it means that things are improving overall. PAVICS is pins xclim, so no current images should be impacted.

tests/test_units.py
Zeitsperre: Remove the line and we're good?
```suggestion
            fu.to("mm/day")
```
Or should we add an assertion here for completion's sake?

RondeauG:
Sorry, I skipped this conversation.

> Remove support for non-CF aliases : "cms" (m3 s-1), "mmday" (mm d-1) and "pct" (%). These are not supported by udunits2 and thus I don't see why we should support them ? @huard, @RondeauG what do you think ? I'm not strongly opinionated on this.

These are not anywhere in xHydro, so personally, I don't mind removing support for them.

PR #2045, opened 2025-01-13
Change lmoments3 behaviour with fitkwargs

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add a warning if `fitkwargs` is given with `lmoments` method
* Except if we use the `gamma` distribution is used with `floc`: This is accepted and used to shift the distribution

No 

### Other information:

I realize that SPI/SPEI is still heavily constrained in what distribution can be used. For now, only gamma/fisk (log-logistic) are allowed. At some point we could revisit this. I think we reached a point where we could be more permissive, since we ironed out many of the potential problems in these functions. There is still potential that problems could arise if we allow more distributions, but we might have to accept that some problems might arise when working with these optimization procedures, and fix them on the fly (or not, if it's just a limitation of the method).


coxipi:
@aulemahal Sascha suggested an error instead of a warning, thoughts?

src/xclim/indices/stats.py
```diff
@@ -59,6 +59,21 @@ def _fitfunc_1d(arr, *, dist, nparams, method, **fitkwargs):
...
+            )
+        if "floc" in fitkwargs and type(dist).__name__ == "GammaGen":
+            # lmoments3 assumes `loc` is 0, so `x` may need to be shifted
+            # note that `floc` must already be in appropriate units for `x`
+            params = dist.lmom_fit(x - fitkwargs["floc"])
+            params["loc"] = fitkwargs["floc"]
+            params = list(params.values())
+        else:
+            params = list(dist.lmom_fit(x).values())
         params = list(dist.lmom_fit(x).values())
```
aulemahal, line 77:
```suggestion
```

aulemahal (approved):
I was thinking of a warning because I was thinking of a usage where one calls the function multiple time with different distributions but the same other arguments. I realize now that this does not make much sense, woups.

Thinking of it, an error might indeed be better. Sorry for the confusion!

src/xclim/indices/stats.py
coxipi: oupsie

coxipi:
Now, we can use PWM in SPI/SPEI functions directly.

I added tests: The results with lmoments3 (PWM) / gamma is close to what we get with ML/gamma or APP/gamma, so I'm satisfied, things work well.

tests/test_indices.py
```diff
@@ -672,6 +704,13 @@ def test_standardized_precipitation_index(
             and Version(__numpy_version__) < Version("2.0.0")
         ):
             pytest.skip("Skipping SPI/ML/D on older numpy")
+
+        # change `dist` to a lmoments3 object if needed
+        if method == "PWM":
+            lmom = pytest.importorskip("lmoments3.distr")
+            scipy2lmom = {"gamma": "gam"}
+            dist = getattr(lmom, scipy2lmom[dist])
```
aulemahal, line 712: @Zeitsperre are we sure that any `*-lmoments3` tox test suite is running here ? 

@coxipi Because I think the test should fail : `dist` is passed as a `lmoments3` object in the test but in the SPI and SPEI functions there are tests like `dist in dist_methods`, which would fail. I'm not sure how to do it, but the function/indicator must now be able to accept scipy/lmoments3 distribution objects if the `method=='PWM'` is to be enabled!

coxipi: Ah you're right! in the tests I have called the low-level function, so it's really `params` that counts in the SPI call. The default distribution that is sent in the warning message does not matter.

Zeitsperre: Confirming here that tests using `lmoments3` are being run.

coxipi:
`rv_continuous` now allowed generally, with the usual constraints remaining with a the use of a string. 

Should be good to go, let's see the tests

PR #2060, opened 2025-01-27
Enable select_resample_op for unitless data

saschahofmann:
### Pull Request Checklist:

* `to_agg_units` now returns the resampled data without units if the original data was unitless
I could imagine that maybe you don't ever want unitless data in which this can be closed but if not then this is an easy fix.

I don't think so.

### Reproducible example
Before this change this example was breaking with a `KeyError`
```python
from xclim.indices.generic import select_resample_op
import xarray as xr
import numpy as np

time = xr.cftime_range("2000-01-01", "2000-12-31", freq="D", calendar="noleap")
da = xr.DataArray(np.random.random_sample((time.size,)), coords={"time": time})

select_resample_op(da, op='max', freq='MS')
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,7 +25,8 @@ Internal changes
 
 Bug fixes
 ^^^^^^^^^
-* Fixed an issue in ``xclim.core.units.to_agg_units`` that broke the resampling when computing on unitless data
+* Fixed an issue in ``xclim.core.units.to_agg_units`` that broke the resampling when computing on unitless data (:pull:`2060`).
+
 v0.54.0 (2024-12-16)
 --------------------
 Contributors to this version: Trevor James Smith (:user:`Zeitsperre`), Pascal Bourgault (:user:`aulemahal`), Éric Dupuis (:user:`coxipi`), Sascha Hofmann (:user:`saschahofmann`).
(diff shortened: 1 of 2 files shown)


Zeitsperre (approved):
Thanks again @saschahofmann! I agree that we should handle this case rather than error here.

aulemahal (changes requested):
Hmm, not sure I agree here!

To me, `select_resample_op` and `to_agg_units` are part of the "unit-aware" section of xclim, passing unitless data is an error. "unitless" doesn't mean "dimensionless", it means "insufficient information". 

This condition ensures, `select_resample_op` can be used as a "generic compute function", directly  wrapped in an `Indicator`. 

What was your use case that required unitless-support ? Could `da.assign_attrs(units="1")` solve the issue ?

coxipi:
@aulemahal sorry I'm late to the dance. My reasoning was the opposite, I thought these functions outside of `indices` `agro`, etc. could be more general. But I was thinking more in terms in the various indices that we have that already put a constraint on variables having units, so it would be all right to call this more general function because the decorators put a constraint on the input. But, in fact, even though `select_resample_op` is often called by `indices` functions, it can also itself be considered at the same level as functions in `indices`, correct? In that case I agree. But the indicator class already does some sanity checks on the inputs, right? Perhaps the responsibility of filtering bad or good input should be at higher level?

aulemahal:
>  it can also itself be considered at the same level as functions in indices

This is my reasoning indeed! Even though it is used a lot within unit-aware functions, I'd prefer to keep it unit-aware itself, so simple indicators can be created from it.

If you all agree, I suggest we close this. Thanks @saschahofmann , sorry for thwart your plans!

PR #2068, opened 2025-02-07
No interpolation on dayofyear 

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* During the adjusting part of  quantile methods, a interpolation on the grouping dimension is done if the interpolation "interp" is chosen. This should not be the case when `group="time.dayofyear", so this option is now blocked

No? Even if interpolation was applied before, it should have done nothing, just a more complicated algorithm for nothing

### Other information:

See the discussion here https://github.com/Ouranosinc/xclim/pull/2019#issuecomment-2632429653


coxipi:
see the discussion here https://github.com/Ouranosinc/xclim/pull/2019#issuecomment-2632429653

aulemahal (approved):
That makes sense! Have you seen a change in performance ?

coxipi:
> That makes sense! Have you seen a change in performance ?

good question, let me see ... darn, wish I had made that benchmark suite...

coxipi:
Don't know how to transfer this to xsdba? I can transfer  an issue but not a PR? I will simply close it and open a new issue in xsdba. Maybe. It's not really a priority. It would be nice to avoid useless 2d interpolation when possible though, probably it might save some time on DQM computations. This would not affect MBCn as it is currently, and that's the big bottleneck computation-wise.

PR #2090, opened 2025-02-28
Add ECO-CI Project token

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds project tracking to the ECO-CI Action with project-level tracking
* Re-enables `step-security` connection audits for everything.

No.

### Other information:

Project and organization-level tracking is a "paid" service. We have access to a relatively liberal monthly allowance of minutes thanks to the generous offer of the creator of the project.

One thing worth looking at is reducing the number of full builds we do. The monthly minute usage of `xclim` alone is really high (60-70% of minutes used in Ouranosinc are here). We can reduce this and still maintain effective testing/coverage.

For the `step-security` action, there area handful of unnamed IP addresses listed. Will see if we can drop a few of these before merging to `main`.


Zeitsperre:
This PR is waiting on some clarifications from the ECO-CI developers concerning the ElectricityMaps integration.

.github/workflows/main.yml
```diff
@@ -33,13 +33,14 @@ permissions:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
coxipi, line 36: should the ubuntu field in `dependency-review.yml` also be replaced with this kind of logic?

coxipi: I mean, defining a matrix category? I guess if it's only the name of the OS it's not very useful

coxipi (approved):
looks good

coxipi (approved):
looks good

.github/workflows/main.yml
Zeitsperre: I think for something pretty OS-agnostic like linting/dependency analysis, we probably don't need to test using a matrix. I'll clear some of this up.

Zeitsperre: Ah, I know why I set this: The OS is supplied to multiple steps in the job (the runner itself and as metadata for ECO-CI). Better to have it set in one place.

PR #2114, opened 2025-03-20
Fix issue with xr.core.ops renamed in versions of xarray>=2025.3.0

jensdebruijn:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes [#2113 ](https://github.com/Ouranosinc/xclim/issues/2113)
- [ ] Tests for the changes have been added (for bug fixes / features)
  - [ ] (If applicable) Documentation has been added / updated (for bug fixes / features)
- [x] CHANGELOG.rst has been updated (with summary of main changes)
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added

### What kind of change does this PR introduce?

* This PR fixes an issue with an operation not being available in xarray versions later than 2025.3.0

### Does this PR introduce a breaking change?

No breaking changes

PR #2116, opened 2025-03-21
Temporarily revert star annotation in call signatures - remove all internal xarray imports

### LINKED ISSUE #2113: 'xarray.core' has no attribute 'ops' in xarray 2025.03.0
### Setup Information

- Xclim version: 0.55.1
- Python version: python 3.13.2
- Operating System: linux

### Description

When running `standardized_precipitation_evapotranspiration_index` xclim runs into the error

>   File ".../.venv/lib/python3.13/site-packages/xclim/indices/generic.py", line 282, in get_op
    return xr.core.ops.get_op(binary_op)  # noqa
           ^^^^^^^^^^^
AttributeError: module 'xarray.core' has no attribute 'ops'

The error does not occur with the previous version of xarray 2025.01.2

I believe `xr.computation.ops.get_op` should become `xr.core.ops.get_op`. My suggested fix would be:

```
    if xr.__version__ <= "2025.1.2":
        return xr.core.ops.get_op(binary_op)
    else:
        return xr.computation.ops.get_op(binary_op)  # noqa
```

### Steps To Reproduce

```
import xarray as xr
import xclim.indices as xci

wb = xr.DataArray(
    data=range(100),
    coords={"time": xr.cftime_range("2000-01-01", periods=100, freq="M")},
    dims=["time"],
    attrs={"units": "kg m-2 s-1"},
)

xci.standardized_precipitation_evapotranspiration_index(wb=wb)
```

### Additional context

_No response_

### Contribution

### Code of Conduct

### LINKED ISSUE #2120: 'xarray.core' has no attribute 'merge' when running with dask 2025.3.0
### Setup Information

- Xclim version: 0.56.x-dev
- Python version: All
- Operating System: Any

### Description

I feel like I'm getting deja vu.

It would appear that the latest `dask` isn't playing nicely with our current usages. I'm surprised this wasn't caught by upstream tests. 

### Steps To Reproduce

_No response_

### Additional context

https://github.com/Ouranosinc/xclim/actions/runs/14037153854/job/39297938203

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #2120
    - This PR fixes #2113 

* Reverts star-annotated call signatures (not supported for Python 3.10)
*`xclim` now uses directly `operator` instead of using `xarray`'s derived `get_op` function. A refactoring in `xarray` had changed the position of `get_op` which caused a bug.
    + All other uses of `xarray`'s internal API were also removed (:pull:`2116`).
* Updates the pre-commit hooks.

No.


aulemahal (approved):

Zeitsperre:
This PR needs the changes from #2114

aulemahal:
@Zeitsperre My last commit removed all xarray imports that were not from their "public" API, hoping that avoids all further problemes like we add with xarray 2025.3.

To do so, I had to copy a few lines from xarray for the `merge_attrs_drop_conflicts` and `get_temp_dimname` functions. @coxipi  you might want to the the same as I did for xsdba, even though this is not a bug right now, just futurproofing.

I also had to remove `DataArrayResample` and `DatasetResample` from the signature of `xc.core.calendar.time_bnds`. They are still mentionned in the docstring... Not the best, but I don't see how I can have them in the signature without importing them at runtime....

Zeitsperre:
> I also fixed an issue with setting options that are deep dictionaries. And fixed the issue with the missing methods. Flox is now enabled by default for resampling with `first` and `last`, but the way xarray calls flox is broken for datetime data in some cases. I simply used another resampling method as the line in `expected_count` only cares about the output coordinates. (This call is done because the alternative was to dig into private xarray objects)

Thanks for the fixes! Think you could mention this in the Changelog as a (potentially) breaking change?

aulemahal:
> as a (potentially) breaking change?

I don't think the user is impacted at all! This change does not change anything in the function's outputs.

I found more xarray internals and removed them. I had to reimplement a version of  `days_in_year` to do so, but the version I copied from xarray is pretty short.

The last xarray internal calls I see, but couldn't/wouldn't remove are:
- `xr.core.common._contains_cftime_datetimes` and `xr.conventions.encode_cf_variable` in `xclim.sdba.map_blocks` : I think the whole block of code can be removed, but I will do so in a PR to `xsdba` instead.
- `xr.coding.cftime_offsets.to_offset` in a few places : we need the returned `Offset` object, which itself isn't public API anyway. No way I am reimplementing this whole machinery in xclim.

Finally, I also had to remove the annotations from `xclim.sdba.Grouper.group`. I think returning a `GroupBy` object is ok, but it isn't itself public API, only its methods are... It could be moved around in the module without warning.

PR #2122, opened 2025-03-24
Fix 'degF d' conversion in growing degree days

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* ...

### DIFF
--- a/src/xclim/core/units.py
+++ b/src/xclim/core/units.py
@@ -665,15 +665,20 @@ def to_agg_units(out: xr.DataArray, orig: xr.DataArray, op: str, dim: str = "tim
 
     elif op in ["count", "integral"]:
         m, freq_u_raw = infer_sampling_units(orig[dim])
-        # TODO: Use delta here
         orig_u = units2pint(orig)
         freq_u = str2pint(freq_u_raw)
+
         with xr.set_options(keep_attrs=True):
             out = out * m
 
         if op == "count":
             out.attrs["units"] = freq_u_raw
+
         elif op == "integral":
+            if "[temperature]" in orig_u.dimensionality:
+                # ensure delta_temperature
+                orig_u = 1 * orig_u - 1 * orig_u
+
             if "[time]" in orig_u.dimensionality:
                 # We need to simplify units after multiplication
 
--- a/tests/test_temperature.py
+++ b/tests/test_temperature.py
@@ -592,6 +592,16 @@ class TestGrowingDegreeDays:
 
         assert np.isnan(gdd.values[0, -1, -1])
 
+    def test_conversion(self, open_dataset):
+        ds = open_dataset("ERA5/daily_surface_cancities_1990-1993.nc")
+
+        ds["tas_F"] = convert_units_to(ds.tas, "degF")
+        ds_pt = ds.isel(location=1)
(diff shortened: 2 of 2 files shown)


src/xclim/core/units.py
```diff
@@ -665,15 +665,20 @@ def to_agg_units(out: xr.DataArray, orig: xr.DataArray, op: str, dim: str = "tim
...
         with xr.set_options(keep_attrs=True):
             out = out * m
 
         if op == "count":
             out.attrs["units"] = freq_u_raw
+
         elif op == "integral":
+            if "[temperature]" in orig_u.dimensionality:
+                # ensure delta_temperature
+                orig_u = 1 * orig_u - 1 * orig_u
```
VascoSch92, line 680: But this is equal to zero or?

coxipi: We are merely interested in getting a "delta" unit for degF or degC (e.g. degC -> delta_degrees_celcius), this accomplishes this.

coxipi: For instance, before we had something like this:
```python
orig_u = xclim.core.units.units2pint("degF")
freq_u = xclim.core.units.str2pint("day")

print(orig_u)
>>> °F
print(freq_u*orig_u)
>>> 255.928 d K
print(xclim.core.units.pint2cfattrs(freq_u*orig_u))
>>> {'units': 'd K', 'units_metadata': 'temperature: unknown'}
```

but if we first do this trick, then
```python
print(orig_u)
>>> 0 delta_degF
print(freq_u*orig_u)
>>> 0 d delta_degF
print(xclim.core.units.pint2cfattrs(freq_u*orig_u))
>>> {'units': 'd degF', 'units_metadata': 'temperature: difference'}
```

coxipi: @aulemahal 

While writing the comment above, I realized that using a `units2pint` to define `freq_u` instead of `str2pint`, I get the correct result without the little `1*orig_u - 1*orig_u` trick

```python
# error
orig_u = xclim.core.units.units2pint("degF")
freq_u = xclim.core.units.str2pint("d")
print(orig_u)
print(freq_u*orig_u)
print(xclim.core.units.pint2cfattrs(freq_u*orig_u))

>>> °F
255.928 d K
{'units': 'd K', 'units_metadata': 'temperature: unknown'}
```

vs. 

```
# no error
orig_u = xclim.core.units.units2pint("degF")
freq_u = xclim.core.units.units2pint("d")
print(orig_u)
print(freq_u*orig_u)
print(xclim.core.units.pint2cfattrs(freq_u*orig_u))

>>>°F
d °F
{'units': 'd °F', 'units_metadata': 'temperature: unknown'}
```

perhaps this something in `units` was changed with this unforeseen consequence?

aulemahal: oh! That may be it. `str2pint` returns a Quantity, and not a unit. When two units are multiplied together, it seems that `pint` does not try to perform further conversions!

However, may be the trick is still better because it makes it clear this is a delta ? It protects us from further pint modifications ?

coxipi: Yes, I think this trick works well. It just makes me wonder if there could be problems related to this. The issue also comes up from a mix of our units tools, and performing a pint calculation `freq_u*orig_u`. I don't think we have to worry about problems like this elsewhere?

aulemahal (approved):

PR #2130, opened 2025-03-26
Datetime fixes

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Removes usage of `xr.cftime_range` which is deprecated. Most calls were in the tests.
* Avoids `strptime` without year information.
* Add `pd.DatetimeIndex` case to  `get_calendar` (I think this worked previously, but it wasn't tested).
*  `test_timeseries` defaults to `cftime=None` (usage of cftime is determined by the requested calendar)

No

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Bug fixes
     + All other uses of `xarray`'s internal API were also removed (:pull:`2116`).
 * Fixed an issue with star-annotated call signatures to maintain Python 3.10 compatibility. (:pull:`2116`).
 * Fixed `to_agg_units` that was converting units of temperature differences prematurely, without changing accordingly the values in the related DataArrays. (:issue:`2121`, :pull:`2122`).
-* ``get_calendar`` now supports pandas' ``DatetimeIndex``. xclim no longer uses ``xarray.cftime_range``, which has been deprecated. (:pull:`2130`).
+* ``get_calendar`` now supports pandas' ``DatetimeIndex``. `xclim` no longer uses ``xarray.cftime_range``, which has been deprecated. (:pull:`2130`).
 * Avoid unnecessary time resampling in `preprocess_standardized_index` when `freq` is not None but the same as the input data. (:issue:`2111`, :pull:`2112`)
 * Fixed an issue with ``fire_season`` that made it fail with datasets having non-uniform chunks. (:issue:`2129`, :pull:`2132`).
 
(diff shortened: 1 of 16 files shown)


coxipi (approved):
approved!

Zeitsperre:
@aulemahal If you want to just add an entry to the changelog, this'll be good for tomorrow morning. Thanks again!

CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Bug fixes
     + All other uses of `xarray`'s internal API were also removed (:pull:`2116`).
 * Fixed an issue with star-annotated call signatures to maintain Python 3.10 compatibility. (:pull:`2116`).
 * Fixed `to_agg_units` that was converting units of temperature differences prematurely, without changing accordingly the values in the related DataArrays. (:issue:`2121`, :pull:`2122`).
+* ``get_calendar`` now supports pandas' ``DatetimeIndex``. xclim no longer uses ``xarray.cftime_range``, which has been deprecated. (:pull:`2130`).
```
Zeitsperre, line 17:
```suggestion
* ``get_calendar`` now supports pandas' ``DatetimeIndex``. `xclim` no longer uses ``xarray.cftime_range``, which has been deprecated. (:pull:`2130`).
```

Do you know wince when it has been deprecated? Do we need a lower pin on `xarray` for this?

src/xclim/indices/run_length.py
```diff
@@ -1660,12 +1660,13 @@ def index_of_date(
     """
     if date is None:
         return np.array([default])
-    try:
+    if len(date.split("-")) == 2:
+        date = f"2000-{date}"
```
Zeitsperre, line 1664: "2000"? Is this solely to get a valid datetime string?

CHANGELOG.rst
aulemahal: We don't need a pin ! `date_range` has been there for some time, the solution should  be valid for all supported xarray.

The deprecation was announced in 2025.03. There are no "terminal" version indicated in the warning, nor in the PR.

src/xclim/indices/run_length.py
aulemahal: Yep. Now that I think of it, we should use a leap year so we can parse `02-29`...

```suggestion
        date = f"1840-{date}"
```
In honor of the Act of Union adopted by the British Parliament in July 1840.

Issue #2131, opened 2025-03-26
keep_longest_run is ressource-heavy


SarahG-579462:
### Generic Issue

* xclim version: 0.54.0
* Python version:
* Operating System:

### Description

xc.indices.run_length.keep_longest_run is slow and expensive in memory. With no resampling, for a 200 MB boolean dataset, it runs in 50 seconds and temporarily consumes 6 GB of memory.

### What I Did
```

>>> season.nbytes / 1e6 # MB
206.7168

>>> run_season_longest_native = xc.indices.run_length.keep_longest_run(season)
CPU times: user 27.9 s, sys: 24.7 s, total: 52.6 s
Wall time: 52.9 s

>>> %memit keep_longest_run(season)
peak memory: 14061.95 MiB, increment: 6505.66 MiB
```

<details>

<summary>
Memory profile information
</summary>

```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   757   7552.0 MiB   7552.0 MiB           1   def keep_longest_run(
   758                                             da: xr.DataArray, dim: str = "time", freq: str | None = None
   759                                         ) -> xr.DataArray:
   760                                             """
   761                                             Keep the longest run along a dimension.
   762                                         
   763                                             Parameters
   764                                             ----------
   765                                             da : xr.DataArray
   766                                                 Boolean array.
   767                                             dim : str
   768                                                 Dimension along which to check for the longest run.
   769                                             freq : str
   770                                                 Resampling frequency.
   771                                         
   772                                             Returns
   773                                             -------
   774                                             xr.DataArray, [bool]
   775                                                 Boolean array similar to da but with only one run, the (first) longest.
   776                                             """
   777                                             # Get run lengths
   778   9129.1 MiB   1577.1 MiB           1       rls = rle(da, dim)
   779                                         
   780   9129.1 MiB      0.0 MiB           2       def _get_out(_rls):  # numpydoc ignore=GL08
   781  10706.4 MiB   -197.1 MiB           2           _out = xr.where(
   782                                                     # Construct an integer array and find the max
   783   9326.4 MiB    197.3 MiB           1               _rls[dim].copy(data=np.arange(_rls[dim].size)) == _rls.argmax(dim),
   784  10903.5 MiB   1577.1 MiB           1               _rls + 1,  # Add one to the First longest run
   785  10903.5 MiB      0.0 MiB           1               _rls,
   786                                                 )
[...]

PR #2136, opened 2025-03-28
Faster rle

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Faster rle, especially when data is unchunked along "dim"

Not really, but output may have smaller int types than before when possible

### Other information:

This is joint work with @SarahG-579462, see her issue #2131


src/xclim/indices/run_length.py
```diff
@@ -200,26 +232,20 @@ def rle(
...
     if index == "first":
-        out = out[{dim: slice(None, None, -1)}]
-
+        sl, slc = slc, sl
+    mask_near_zero = (1 - da)[slc]
+    mask_near_zero[dim] = cs_s[sl][dim]
+    cs_s[sl] = cs_s[sl] * mask_near_zero
+    # 2) And only keep numbers part of a run. This also reinserts 0's
+    out = cs_s * da
+    # Result: 1) & 2) -> Only keep the largest number in any given run, the run length
```
coxipi, line 248: that's a bit less readable than before, but since `rle` is more of a low-level function used in many places, I think it's acceptable

src/xclim/indices/run_length.py
```diff
@@ -124,6 +124,40 @@ def resample_and_rl(
...
+            return dtype
+    return np.uint64
+
+
+# TODO: Compare this with _rle_1d, but I think this has the edge. A single operation
+# a*b + b effectively replaces a cumsum and diff operations
+def _cumsum_reset_np(arr, index):
+    # run the cumsum and prod backwards or forward
+    it = range(1, len(arr), 1) if index == "last" else range(len(arr) - 2, -1, -1)
+    for i in it:
```
VascoSch92, line 139: I think you can optimise that... You are using lumpy right?

```python
step_indices = np.array(it) 
prev_indices = step_indices - it.step

arr[step_indices] *= arr[prev_indices] + 1
```

coxipi:
Sorry, this is a lot bigger than it needed to be. I just had a couple of `rle` ideas that accumulated over time

src/xclim/indices/run_length.py
```diff
@@ -125,6 +131,40 @@ def resample_and_rl(
...
+def _smallest_uint(da, dim):
+    for dtype in [np.uint8, np.uint16, np.uint32, np.uint64]:
+        if np.iinfo(dtype).max < da[dim].size:
+            return dtype
+    return np.uint64
+
+
+# The `one` argument was weirdly necessary to avoid some type problems with jit
+@njit
+def _cumsum_reset_np(arr, index, one):
```
huard, line 143:
```suggestion
def _cumsum_reset_np(arr, index, one):
"""100110111 -> 100120123"""
```

huard (commented):
T'would be useful to add a simple example in the rle docstring as well.

coxipi (commented):
.

src/xclim/indices/run_length.py
```diff
@@ -465,11 +506,13 @@ def windowed_max_run_sum(
...
-        out = rle(da, dim=dim, index=index).max(dim=dim)
+        # using _cumsum_reset_xr instead of rle to be able to handle a float
+        out = _cumsum_reset_xr(da, dim=dim, index=index, reset_on_zero=True).max(dim=dim)
 
     else:
-        d_rse = rle(da, dim=dim, index=index)
-        d_rle = rle((da > 0).astype(bool), dim=dim, index=index)
+        # using _cumsum_reset_xr instead of rle to be able to handle a float
+        d_rse = _cumsum_reset_xr(da, dim=dim, index=index, reset_on_zero=True)
+        d_rle = rle(da > 0, dim=dim, index=index)
```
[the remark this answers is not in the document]
huard: I must say that I'm confused by this function. The docstring mentions "consecutive float values", but da is documented as a boolean. Then the window criteria is computed on da > 0, which only makes sense if da is a bool, but if it's bool, then why do we need to call both cumsum_reset_xr(da) and rle(da>0) ?

src/xclim/indices/run_length.py
```diff
@@ -465,11 +507,13 @@ def windowed_max_run_sum(
         Total number of `True` values part of a consecutive runs of at least `window` long.
     """
     if window == 1 and freq is None:
-        out = rle(da, dim=dim, index=index).max(dim=dim)
+        # using _cumsum_reset_xr instead of rle to be able to handle a float
+        out = _cumsum_reset_xr(da, dim=dim, index=index, reset_on_zero=True).max(dim=dim)
```
huard, line 511: Why not use the generic `_cumsum_reset` ?

src/xclim/indices/run_length.py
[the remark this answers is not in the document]
huard: The function will fail if you're interested in runs of negative values.

src/xclim/indices/run_length.py
```diff
@@ -125,6 +131,43 @@ def resample_and_rl(
...
+    cond = da == 0 if reset_on_zero else da.isnull()  # reset condition
+    cs2 = cs.where(cond)  # keep only numbers at positions of zeroes e.g. N11NN3NNN (default)
+    cs2[{dim: 0}] = 0  # put a zero in front e.g. 011NN3NNN
+    cs2 = cs2.ffill(dim=dim)  # e.g. 011113333
+    out = cs - cs2
+    if index == "first":
+        out = out[{dim: slice(None, None, -1)}]
+    return out
+
+
```
huard, line 170: Maybe add corner tests for those core functions? I understand they are implicitly tested by many tests, but I think it'd be good to add tests for weird inputs, ie 0-length arrays, arrays with only NaNs, infs, etc. Stuff that would be not be exercised by other tests and that could provide safeguards when we try to optimize further.

huard (approved):
This looks fine, but I think there's something wrong in `windowed_max_run_sum`. I'll let you decide if you want to open another issue for this or fix it here.

src/xclim/indices/run_length.py
coxipi: The generic `_cumsum_reset` may use the fast track (`_cumsum_reset_np`) depending on the chunking. The fast track explicitly assumes that we have 1's and 0', and we don't want tha. Here, we use `_cumsum_reset_xr`, which allows this kind of behaviour:
2 2 0 0 2 2 2 0 0  -> 2 4 0 0 2 4 6 0 0 

Perhaps having the _cumsum_reset function that should be used only with 1's and 0's is not so intuitive. Again, this is because the fast track exploits the idea of having only 1's and 0's.

coxipi:
> I must say that I'm confused by this function. The docstring mentions "consecutive float values", but da is documented as a boolean. Then the window criteria is computed on da > 0, which only makes sense if da is a bool, but if it's bool, then why do we need to call both cumsum_reset_xr(da) and rle(da>0) ?

> da is documented as a boolean
that's an error, I'll fix it

>  Then the window criteria is computed on da > 0, which only makes sense if da is a bool
Well. we transform this as a boolean. The idea is: 

1. Compute  threshold exceedances: 
 `da = (tasmax - thresh).clip(0,None)`
This is the input of windowed_max_run_sum.

2.Then, the idea is to find run_lengths where we have exceedances for at least `window` days (heat waves), that's why we compute 
```python
d_rle = rle(da > 0, dim=dim, index=index)
d_rle >= window
```

3. In these, we want the intensity of the heat wave, that's why we need to compute the cumulative exceedances with the actual values of exceedances
`d_rse = _cumsum_reset_xr(da, dim=dim, index=index, reset_on_zero=True)`

src/xclim/indices/run_length.py
coxipi: > ie 0-length arrays

I'm not sure why we would be interested in this? It's not a test I've ever seen in our testing suite at least.

> only NaNs

For rle, we expect boolean arrays. Even if you had nans in your data, typically da> thresh will get rid of nans and transform them to False. Currently, all-nan will just give all 0's, no runs. I can modify the return statement at the end of rle so that nans are reinserted at the end:
`return out.where(da)`

src/xclim/indices/run_length.py
```diff
@@ -152,20 +195,19 @@ def _cumsum_reset(
...
+        # only int can be used in this case
+        typ = _smallest_uint(da, dim)
+        out = xr.apply_ufunc(
+            _cumsum_reset_np,
+            da.astype(typ),
+            input_core_dims=[[dim]],
+            output_core_dims=[[dim]],
+            dask="parallelized",
+            kwargs={"index": index, "one": typ(1)},
+        ).astype(float)
```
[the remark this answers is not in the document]
coxipi: In the fast track, this is not supported. The fast track could always stay as integer. But I thought, since the slow-track (_cumsum_reset_xr) will yield float, I should cast the fast_track as float too for consistency of output

tests/test_run_length.py
```diff
@@ -80,6 +80,12 @@ def test_dask(self, use_dask):
         sus = rl.suspicious_run(da)
         assert sus.all()
 
+    def test_empty(self):
+        da = xr.DataArray(np.array([[1, 0], [0, 1]]), dims={"time": 2, "loc": 2})
+        da = da.isel(time=slice(None, 0))
+        rlength = rl.rle(da).size
+        assert da.size == rlength.size == 0
```
Zeitsperre, line 87:
```suggestion
        rlength = rl.rle(da)
        assert da.size == rlength.size == 0
```

coxipi: thanks

src/xclim/indices/run_length.py
huard: You could call the function `binary_cumsum_with_reset`.

src/xclim/indices/run_length.py
huard: I'll let you judge what kind of edge cases are useful and which are not, but I've seen xclim functions trip up on 0-length arrays resulting from all NaNs that were dropped (e.g. statistical fits), so I thought maybe this could happen here as well. 

Not sure I measure well the implications of this choice, so do what you think is best.

src/xclim/indices/run_length.py
```diff
@@ -152,20 +195,19 @@ def _cumsum_reset(
...
-
-    if index == "first":
-        out = out[{dim: slice(None, None, -1)}]
-
+    if not _is_chunked(da, dim) and reset_on_zero:
+        # only int can be used in this case
+        typ = _smallest_uint(da, dim)
+        out = xr.apply_ufunc(
+            _cumsum_reset_np,
+            da.astype(typ),
```
aulemahal, line 203: Ok le truc du `_smallest_uint` c'est bon! Malheureusement, bon n'est pas toujours élégant :sweat_smile: . Est-ce qu'on sait si y a un réel impact sur la performance ? Si non, j'aurais tendance a y aller avec un int normal ?

src/xclim/indices/run_length.py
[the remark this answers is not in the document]
aulemahal: I'm ok with the conversion, even though I personally don't see a need to be stiff on the return type of a private function. The conversion is almost sure to happen somewhere down the line.

PR #2139, opened 2025-03-31
Removed deprecated functions, deprecate `testing.open_dataset`

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Removes the deprecated `sfcwind_2_uas_vas` and `uas_vas_2_sfcwind` converter functions.
* Adds a deprecation notice to `xclim.testing.open_dataset` to be removed in a future version
* Changes the signature of `xclim.testing.helpers.generate_atmos` to only accept a `nimbus` object.
* Fixes a bug that was causing cache directories to store generated files in a directory that duplicated the version string (e.g. `XDG_CACHE_DIR/xclim-testdata/v2077.1.1/v2077.1.1`).

Yes. Functions that were previously deprecated have been removed. Suggestions have been made to use the `nimbus` class to manage fetching testing data.

When opening either an absolute path location on disk or an OPeNDAP link, developers are expected to use the typical `xarray.open_{mf}dataset()` function. Paths provided by `nimbus.fetch()` will always be absolute paths.

### Other information

This change moves most of the xclim-testdata fetching mechanism to rely on `pooch.Pooch` as that object is much better at handling caching and versioning.

It is very important to never use `nimbus.fetch()` on a `pathlib.Path` object. The `registry` of a `Pooch` object is synonymous with a dictionary, and some issues that needed to be dealt with stemmed from that misusage.

An example of what this does for Windows:
```python

# Actual file is "NRCANdaily/nrcan_canada_daily_tasmax_1990.nc" in registry

file = Path("NRCANdaily", "nrcan_canada_daily_tasmax_1990.nc")
xr.open_dataset(nimbus.fetch(file))

>>> ValueError: File 'NRCANdaily\nrcan_canada_daily_tasmax_1990.nc' is not in the registry.
```

### Tiniest caveat

In order to support `no external sockets` mode, I needed to slightly the internal `_fetch` method of `nimbus`:

```python
try:
    return _nimbus.fetch_diversion(*args, **kwargs)
except SocketBlockedError as err:
    raise FileNotFoundError(
        "File was not found in the testing data cache and remote socket connections are disabled. "
        "You may need to download the testing data using `xclim prefetch_testing_data`."
    ) from err
```

I might try proposing a "No external connections" mode to the `pooch` developers.


src/xclim/testing/utils.py
```diff
@@ -567,8 +568,8 @@ def _downloader(
     return _nimbus
 
 
-# FIXME: This function is soon to be deprecated.
 # idea copied from raven that it borrowed from xclim that borrowed it from xarray that was borrowed from Seaborn
+@deprecated(from_version="0.57.0", suggested="nimbus")
```
aulemahal, line 572: I understand the idea of switching to a more modern tool, but why the need to remove this function ?

Could we simply rewrite it so:
```python3
def open_dataset(name, nimbus_kwargs, **xr_kwargs):
    return xr.open_dataset(nimbus(**nimbus_kwargs).fetch(name), **xr_kwargs) 
```
?  (ok, maybe with a bit more doc and  annotations, but you see what I mean)

Yes the function looks simple, but having a very convenient function identical to `xr.open_dataset` seems a bonus to me! It makes doc tests and examples much more readable. And testing is also made simpler.

Zeitsperre: Yeah, alright. It's simple and elegant enough to remain useful. I can't fight you on this, haha.

I'll make the changes.

tests/test_sdba/test_detrending.py
```diff
@@ -32,7 +32,7 @@ def test_poly_detrend_and_from_ds(series, tmp_path, open_dataset):
     file = tmp_path / "test_polydetrend.nc"
     fx.ds.to_netcdf(file, engine="h5netcdf")
 
-    ds = open_dataset(file)
+    ds = xr.open_dataset(file)
```
coxipi, line 35: so the tmp_path function differently, they don't need nimbus?

coxipi: where exactly are they stored anyways?

coxipi (approved):
approved!

src/xclim/core/utils.py
```diff
@@ -865,10 +865,11 @@ def split_auxiliary_coordinates(
...
 
-    >>> # xdoctest: +SKIP
-    >>> clean, aux = split_auxiliary_coordinates(ds)
-    >>> merged = clean.assign_coords(da.coords)
-    >>> merged.identical(ds)  # True
+    .. code-block:: python
+
+        clean, aux = split_auxiliary_coordinates(ds)
+        merged = clean.assign_coords(da.coords)
+        merged.identical(ds)  # -> True
```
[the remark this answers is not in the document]
Zeitsperre: Yes, this is no longer a test. Perpetually skipped tests are useless IMO.

tests/test_sdba/test_detrending.py
Zeitsperre: Good question! It depends on how the tests are launched:
- If the tests are launched with `numprocesses` (no distributed testing), the files are found in `XDG_CACHE_HOME` (or equivalent for macOS/Windows).
- If the tests are launched with parallel execution, the first worker able to start downloading test data places the data into `XDG_CACHE_HOME` (if not already there) and afterwards, all workers copy this folder to a new temporary location. This prevents synonymous reads of shared files by multiple workers (which cause problems for `python-NetCDF4`).

Zeitsperre: Woops, hold on. In this case, yes, the `tmp_path` is worker-dependent. `nimbus` is used for resolving dataset locations that identify true testing data. Within a test, the temp filepath is a full location, self-contained to that test * worker.

src/xclim/testing/utils.py
Zeitsperre: For complete clarity, the new function will work as intended, so long as the string being passed to it is in the registry. Anything else will error out.

Zeitsperre: For complete clarity, the new function will work as intended, so long as the string being passed to it is in the registry. Anything else will error out.

Zeitsperre:
I've made quite a few changes, so I feel like a second review is warranted. Thanks again!

src/xclim/testing/helpers.py
```diff
@@ -114,7 +108,7 @@ def add_example_file_paths() -> dict[str, str | list[xr.DataArray]]:
 
     Returns
     -------
-    dict of str or dict of list of xr.DataArray
+    dict of str or list of xr.DataArray
         A dictionary of doctest-relevant datasets.
```
coxipi, line 112: this is not coherent with the return type just above?

Zeitsperre: Good catch! Why did I change thisÉ

aulemahal (approved):

PR #2144, opened 2025-04-08
Support DataTree

### LINKED ISSUE #2127: Datatree compatability
### Addressing a Problem?

As xarray has officially adopted the [DataTree](https://docs.xarray.dev/en/latest/generated/xarray.DataTree.html) structure it would be nice to have integration with xclim. 
In particular, it would be handy to be able to do something like:
```python
#With dt as an xarray datatree
dt_tnn = xclim.indicators.cf.tnn(dt, vars=["tas"], *arg, **kwargs)
```
Thus calculating the xclim indicator for all datasets in the datatree and resulting in a new datatree.

Similarly for unit conversion.

Is this something that xclim would be interested in developing? 
This will also require some upstream developments in xarray datatree.

### Potential Solution

My current solution which is using the old [archived datatree repo](https://xarray-datatree.readthedocs.io/en/latest/) is as follows. 

Note that the decorator @map_over_subtree did not make it to xarray... However, I found it very usefull.

```python
@map_over_subtree
def xclim_indicator(ds, indicator, vars, **kwargs):
    if isinstance(vars, str):
        return indicator(ds[vars], **kwargs).to_dataset()
    elif isinstance(vars, list): #Order is important!
        data_arrays = [ds[var] for var in vars]
        return indicator(*data_arrays, **kwargs).to_dataset()

@map_over_subtree
def convert_units_to(ds, var, target_unit, context="infer"):
    if var in ds:
        ds = ds.copy()
        ds[var] = xclim.units.convert_units_to(ds[var], target_unit, context=context)
    return ds

dt_tnn = xclim_indicator(dt, xclim.indicators.cf.tnn, vars="tas", freq="YS")
dt_txx = xclim_indicator(dt, xclim.indicators.cf.txx, vars="tasmax", freq="YS")

dt = convert_units_to(dt, var, unit)
```

### Additional context

_No response_

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2127

Support `xarray.Datatree` objects in the  `ds` argument of indicators. When receiving such an object, the indicator computation is mapped over the datasets of the tree.

Passing a `Datatree` implies setting the `as_dataset` option to True.

No, it shouldn't. `xr.Datatree` was added recently, and xclim will not fail if it absent. However, I can't promise it will work with the old standalone `xarray-datatree` module!

### Other information:
This is missing some tests, but most importantly I have a dilemma here : what do we do when a node doesn't have the appropriate variables ?

A Datatree can have nodes with no data, I think it is quite common, depending on your structure. Currently, the code will skip any node that has no "data vars", returning the node untouched. However, any other node that doesn't have the required variables will trigger a failure with a `MissingVariableError`. 

Alternatives would be to return an empty node or return the node untouched.

What does @kobebryant432 think of this question ?


kobebryant432:
Hey @aulemahal, 

Thanks for tackling this so fast! Really appreciated. 

- Skipping empty nodes seems logical, this also maintains the tree structure. 
- For nodes that don't have the required variables, I would prefer being explicit. Triggering a failure with a MissingVariableError will inform the user that there is missing data (which may be unexpected).

It may be nice to illustrate how one could than filter the DataTree to include only those leaves (nodes with datasets) that have the required dataset.

kobebryant432:
Something like this would do the job and allow applying the datatree where possible.

```python
var = "air"
DT.filter(lambda ds: var in ds.data_vars)
```

kobebryant432:
Note that currently, this PR does not address convert_units_to functionality.

kobebryant432:
Started an attempt in #2152

huard (approved):

PR #2152, opened 2025-04-18
Add convert_units_to wrappers for datasets and datatrees

### LINKED ISSUE #2127: Datatree compatability

kobebryant432:
### Pull Request Checklist:
    - This PR fixes #2127 

First attempt to extend add convert_units_to xarray Datatree's

No

### Other information:
Here again the question arises "What to do with datasets that do not have the specified variable?". Currently implemented to give an Error if one dataset in the Datatree does not have the specified variable as done with the indicators. 

Willing to write tests, and docs but waiting for some feedback first because I am not convinced of this implementation.


aulemahal:
What do you think of a polymorphism in the same vein as the one I suggested in #2144 ?

Ex:
```python3
from xclim.core.units import convert_units_to

# Variable, normal case
out = convert_units_to(da, 'degC')

# Dataset
out = convert_units_to(ds, {'tas': 'degC', 'pr': 'mm/d'})

# DataTree
out = dt.map_over_datasets(
        convert_units_to,
        {'tas': 'degC', 'pr': 'mm/d'},
)
```

I'm not a frequent user of datatrees, but it seems to me that the `dt.map_over_datasets` construct would be the normal thing to do in most cases ? I'm asking because I'm worried about adapting multiple functions with the same snippet (or duplicating functions like in this PR). I feel we can let the user do the job of wrapping, as they will have to do anyway with most functions ?

The `Indicator` case was a bit different as the  `ds` arg was not the first one, and thus the call necessitated a special path.

src/xclim/core/units.py
```diff
@@ -324,6 +331,81 @@ def str2pint(val: str) -> pint.Quantity:
...
 
+def convert_units_to_ds(
+    ds: xr.Dataset,
+    var: str,
+    target: Quantified | units.Unit | dict,
+    context: Literal["infer", "hydro", "none"] | None = None,
+) -> xr.Dataset:
+    """
+    Convert the units of a variable in a dataset to the target units.
+    This function will use :py:func:`xclim.core.units.convert_units_to` to perform the conversion.
```
Zeitsperre, line 342:
```suggestion

    This function will use :py:func:`xclim.core.units.convert_units_to` to perform the conversion.
```

aulemahal:
@kobebryant432 Thanks for your contribution. Unless you have issues with the version implemented in #2144, I'll close this.

PR #2164, opened 2025-05-06
Fix `create_ensemble` calendar argument docstring

### LINKED ISSUE #2108: create_ensemble return dataset with wrong calendar
### Generic Issue

* xclim version: 0.55.1

### Description
When using `ensemble.create_ensemble` with `noleap` and `propleptic_gregorian` calendar realizations, and computing `atmos.growing_degree_days`, the output has `nan` for leap year for `noleap` realization.

### What I Did

```
clim1 = xr.Dataset() # proleptic_gregorian
clim2 = xr.Dataset() # noleap
clim_data = ensembles.create_ensemble([clim1, clim2], realizations=['Model1', 'Model2'])

gdd = growing_degree_days(clim_data.tas, thresh='10 degC', freq='YS')
```

### What I Received

Image

###  Why ?
While the bug is easy to fix providing `calendar="noleap"` argument to `ensemble.create_ensemble`, the documentation says :
> calendar : str, optional
        The calendar of the time coordinate of the ensemble.
        By default, the smallest common calendar is chosen.
        For example, a mixed input of "noleap" and "360_day" will default to "noleap".
        'default' is the standard calendar using np.datetime64 objects (xarray's "standard" with `use_cftime=False`).

From my understanding, a dataset with a `noleap` calendar (smallest) should be returned in the case above.
The bug seems to come from the fact that `ensemble.create_ensemble` uses the default `calendar` argument of  `_ens_align_datasets`, which determines the calendar from `common_calendar` with `join="outer"` while it should uses `join="inner"`
https://github.com/Ouranosinc/xclim/blob/123ba600f53d87e098101063b81d0dfc256b17c3/src/xclim/ensembles/_base.py#L454

### Fixing ?
I think either the documentation should be edited or the code changed to return the smallest calendar. The later case makes more sense for me and I already have the code to fix the issue, let me know if you want me to do the PR.

### Code of Conduct

baptistehamon:
### Pull Request Checklist:
    - This PR fixes #2108

* This PR modifies the docstring of `create_ensemble` to avoid confusion about how the function aligns realizations with different calendars.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 
 v0.57.0 (unreleased)
 --------------------
-Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), Juliette Lavoie (:user:`juliettelavoie`), Pascal Bourgault (:user:`aulemahal`), Armin Hofmann (:user:`HofmannGeo`).
+Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), Juliette Lavoie (:user:`juliettelavoie`), Pascal Bourgault (:user:`aulemahal`), Armin Hofmann (:user:`HofmannGeo`), Baptiste Hamon (:user:`baptistehamon`).
 
 New indicators
 ^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


Zeitsperre (approved):
Thanks again!

PR #2168, opened 2025-05-09
Use Literals in operator `op` arguments

### LINKED ISSUE #1810: Use string `Literal` for str-based matching
I think we could simply map it to xclim's STRING ? Unless there's a simple way to detect if the allowed values' type, my guess is that we don't really need Literals for anything else than strings. So here : https://github.com/Ouranosinc/xclim/blob/98c251cf59f5e9c0e1573593342445e08ea9c957/xclim/core/utils.py#L659-L660

And if we do that, maybe we could apply the same for all `op` and `method`  arguments throughout the library ? And maybe use the annotation to parse the `choices` property of `xclim.core.indicator.Parameter` ?

_Originally posted by @aulemahal in https://github.com/Ouranosinc/xclim/pull/1796#discussion_r1655579649_

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1810 

* Changes the call signatures for `op` arguments to use the `Literal` type when providing a limited set of acceptable strings.
* Updates the `infer_kind_from_input_parameter` to accept a `Literal` sequence of string options.

No.

### Other information:

This change will make it so that coding behaviour in IDEs will show warnings if a string being supplied as an argument is not within the list of acceptable options. This is not just an aesthetic/behavioural change but can also help with `mypy` type checking.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,10 +41,10 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.11.8
     hooks:
-      - id: ruff
-        args: [ '--fix', '--show-fixes' ]
       - id: ruff-format
         exclude: '(src/xclim/indices/__init__.py|docs/installation.rst)'
+      - id: ruff
+        args: [ '--fix', '--show-fixes' ]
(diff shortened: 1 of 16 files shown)


aulemahal (approved):
Looks good!

PR #2178, opened 2025-05-20
Fix robustness fractions and categories for no data

### LINKED ISSUE #2167: Robustness categories broken if all realizations are invalid
### Setup Information

- Xclim version: latest
- Python version: 3.12
- Operating System: Linux

### Description

When all realizations are NaN (invalid) the output of `xclim.ensembles.robustness_fractions` is NaN for all fields, except `valid` which is 0. This seems valid to me. However, then we pass that to `robustness_categories`, it results in bogus data.

Then way the method is coded, everything must be integers, but NaN values will force xarray to use  a `float` dtype for the output. The result is that some numbers get translated to there binary equivalent. Like `2 => -9223372036854775709`.

### Steps To Reproduce

Too lazy to come up with a MWE.

### Additional context

In my specific case, where I am only using the `None` and `'ipcc-ar6-c'` method, this can be circumvented by passing `changed.fillna(99)`.

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2167

* Instead of `NaN`, fractions are returned as 0 when there are no valid members.
* The categories are masked if no valid members were found.

Yes as the previous NaN output might have been expected (at least a test expected it). But all the information used here was already there before.

### DIFF
--- a/src/xclim/ensembles/_robustness.py
+++ b/src/xclim/ensembles/_robustness.py
@@ -310,6 +310,7 @@ def robustness_fractions(  # noqa: C901
 def robustness_categories(
     changed_or_fractions: xr.Dataset | xr.DataArray,
     agree: xr.DataArray | None = None,
+    valid: xr.DataArray | None = None,
     *,
     categories: list[str] | None = None,
     ops: list[tuple[str, str]] | None = None,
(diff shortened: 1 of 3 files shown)


src/xclim/ensembles/_robustness.py
```diff
@@ -253,7 +254,7 @@ def robustness_fractions(  # noqa: C901
     out = xr.Dataset(
         {
             "changed": change_frac.assign_attrs(
-                description="Fraction of members showing significant change. " + test_str,
+                description="Fraction of valid members showing significant change. " + test_str,
```
Zeitsperre, line 257:
```suggestion
                description=f"Fraction of valid members showing significant change. {test_str}",
```

Zeitsperre (approved):

PR #2188, opened 2025-06-04
Allow pooch to be mocked

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Catches a really specific corner case that can affect downstream projects that rely on mocking of `pooch`.

No.

### Other information:

OK, this is wild. Let's say we want to install a very minimal amount of dependencies in order to reduce the memory load on ReadTheDocs; We can very safely mock most libraries and not need to worry unless one of the libraries expects to raise a specific error in the normal operation. 

If we mock `pooch` here, then the `pooch` call resolves but doesn't provide the expected data structure needed for `pathlib.Path` and instead of raising an Attribute error (`None` has no attributes), it raises instead a `TypeError` (the mocked `pooch.os_cache._type` is not a parsable string or Path).

Relevant Pull Request (`miranda`): https://github.com/Ouranosinc/miranda/pull/241

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,11 +4,15 @@ Changelog
 
 v0.58.0 (unreleased)
 --------------------
-Contributors to this version: Sebastian Lehner (:user:`seblehner`).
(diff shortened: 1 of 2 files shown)


aulemahal (approved):
Poor pooch, being mocked like that. :cry:

PR #2189, opened 2025-06-05
Season length from boundaries

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Generic function to compute season length from boundaries (start and end)

No


CHANGELOG.rst
```diff
@@ -4,11 +4,12 @@ Changelog
...
 --------------------
-Contributors to this version: Sebastian Lehner (:user:`seblehner`), Trevor James Smith (:user:`Zeitsperre`).
+Contributors to this version: Sebastian Lehner (:user:`seblehner`), Trevor James Smith (:user:`Zeitsperre`), Éric Dupuis (:user:`coxipi`).
 
-New indicators
-^^^^^^^^^^^^^^
+New indicators and features
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * New indicator ``xclim.atmos.antecedent_precipitation_index`` computes the `antecedent_precipitation_index` (weighted summation of daily precipitation amounts for a given window). (:issue:`2166`, :pull:`2184`).
+* New helper ``xclim.indices.generic.season_length_from_boundaries`` takes `season_start` and `season_end` as input and gives `season_length`. This is used when starts and ends are computed different resampling frequencies: `days_since` are used to compute temporal lengths in this case. (:pull:`2189`).
```
```suggestion
* New helper ``xclim.indices.generic.season_length_from_boundaries`` takes `season_start` and `season_end` as input and gives `season_length`. This is used when starts and ends are computed with different resampling frequencies: `days_since` are used to compute temporal lengths in this case. (:pull:`2189`).
```

src/xclim/indices/generic.py
```diff
@@ -839,6 +846,42 @@ def season(
...
+    xr.DataArray, [dimensionless]
+        Length of the season.
+
+    Notes
+    -----
+    If `season_start` and `season_end` are computed with different resampling frequencies, the time
+    of `season_start` are selected to write the output. This is useful when season start and end were computed
+    with different resampling frequencies. Otherwise, functions in ``xclim.indices.run_length`` will be appropriate.
+    """
+    days_since_start = doy_to_days_since(season_start, start=dayofyearstr_from_freq(xr.infer_freq(season_start.time)))
```
aulemahal: Wait, isn't the default behaviour with `start=None` exactly what you're doing here ?

Also, to get the `DayOfYearStr` your essentially doing : `date_range(..., freq=infer_freq(...))`, isn't that running in circles ?

src/xclim/indices/generic.py
```diff
@@ -839,6 +846,42 @@ def season(
...
+    days_since_end = doy_to_days_since(season_end, start=dayofyearstr_from_freq(xr.infer_freq(season_end.time)))
+    days_since_end["time"] = days_since_start.time
+    doy_start = season_start.time.dt.dayofyear
+    doy_end = season_end.time.dt.dayofyear
+    # days_since we computed with the respective time arrays of season_start and season_end,
+    # but now we will express the season_length using the times of season_start
+    doy_end["time"] = doy_start.time
+    out = (days_since_end + doy_end - doy_start) - days_since_start
+    out.attrs = season_start.attrs
```
aulemahal:
```suggestion
```
The attributes of `season_start` won't exactly apply to `out ` as it's now a length. I suggest you check that attrs are kept if the xarray option is activated (`xr.set_options(keep_attrs=True)`) and if so, we simply let that decide.

src/xclim/indices/generic.py
```diff
@@ -839,6 +846,42 @@ def season(
...
+    Returns
+    -------
+    xr.DataArray, [dimensionless]
+        Length of the season.
+
+    Notes
+    -----
+    If `season_start` and `season_end` are computed with different resampling frequencies, the time
+    of `season_start` are selected to write the output. This is useful when season start and end were computed
+    with different resampling frequencies. Otherwise, functions in ``xclim.indices.run_length`` will be appropriate.
```
aulemahal:
```suggestion
    The time of `season_start` is selected to write the output. This is only useful when season start and end were computed
    at an annual frequency but with different anchor months. Otherwise, functions in ``xclim.indices.run_length`` will be appropriate.
```
Could we add a check at the beginning of the function to ensure both inputs are annual and raise otherwise ?

aulemahal (commented):
Should we document the potential caveats of this ? Like that the shortest possible season is `doy_end` - `doy_start` and lengths can be over 365. Or that both inputs must have the same length and it is _assumed_ that they align (that year N on each is the same year in the context of the season).

src/xclim/indices/generic.py
coxipi: oups of course

coxipi:
> Should we document the potential caveats of this ? Like that the shortest possible season is `doy_end` - `doy_start` 

Do you mean to say that this could be <0 in principle? I guess so.,

> and lengths can be over 365. 

Sure

> Or that both inputs must have the same length and it is _assumed_ that they align (that year N on each is the same year in the context of the season).

we can require
`start.time.year == end.time.year or start.time.year == end.time.year -1`

aulemahal:
> That would be a case of garbage in garbage out
I think I can see one case it isnt totally the user's fault :
- Data starts on Jan 1st for 3 years : 1990, 1991, 1992
- Season start with freq = "YS" -> output is on Jan 1st  1990, 1991, 1992
- Season end with freq = "YS-APR" -> output is on Apr 1st **1989**, 1990, 1991, 1992

Thus, it breaks because inputs are not aligned!

I suggest we check that : `0 <= end.time[0] - start.time[0]  < 365 days` and raise otherwise ?

coxipi:
can I assume: 

`(season_end.time[0] - season_start.time[0])`

gives nanoseconds?

aulemahal:
I think so. Well, If they allowed `units='s'` for datetimes, maybe you could fall on a `timedelta[s]` case ?

What about  `(season_end.time[0] - season_start.time[0]).dt.days` ? Also makes code more comprehensible ?

aulemahal (approved):

PR #2190, opened 2025-06-05
Add indexing to rolling indicators missing it

### LINKED ISSUE #2187: no indexer for  `max_n_day_precipitation_amount`
### Setup Information

- Xclim version: 0.56.0
- Python version:12
- Operating System: linux

### Description

Not sure if this is a bug or a feature, but why is `indexer` not available for all indicators ?

### Steps To Reproduce

I would like to do:

`xc.atmos.max_n_day_precipitation_amount (ds_meteo.pr, window=5,freq="YS-JAN", date_bounds=[ "06-21", "12-20" ])` 

### Additional context

I want to get the extreme precip in the same period as the hydro indicator Q1EA (crue max annuel été-automne).

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2187 

* Adds the  `**indexer` argument to a few indicators:  `max_n_day_precipitation_amount`, `max_pr_intensity` , `blowing_snow`. These were the only  "rolling" indicators missing it. 
* Modifies `frost_days` to use the Indexing class instead of explicitly handling indexing.
* At first I thought I add to add indexing to `rain_on_frozen_grounds_days` only to realize it was already inheriting it from its class. I generalized the `window` arg along the way.

These are "rolling" indicator as in the indice function does a `rolling` operation and after which the indexing should happen. The indexing feature is still missing the indicators that use `run_length` constructs.

No.

### DIFF
--- a/src/xclim/indicators/atmos/_temperature.py
+++ b/src/xclim/indicators/atmos/_temperature.py
@@ -907,7 +907,7 @@ freshet_start = Temp(
     parameters={"thresh": {"default": "0 degC"}, "window": {"default": 5}},
 )
 
-frost_days = Temp(
+frost_days = TempWithIndexing(
     title="Frost days",
     identifier="frost_days",
     units="days",
--- a/src/xclim/indices/_simple.py
+++ b/src/xclim/indices/_simple.py
@@ -331,7 +331,9 @@ def tx_min(tasmax: xarray.DataArray, freq: str = "YS") -> xarray.DataArray:
 
 @declare_units(tasmin="[temperature]", thresh="[temperature]")
 def frost_days(
-    tasmin: xarray.DataArray, thresh: Quantified = "0 degC", freq: str = "YS", **indexer
+    tasmin: xarray.DataArray,
+    thresh: Quantified = "0 degC",
+    freq: str = "YS",
 ) -> xarray.DataArray:
     r"""
     Frost days index.
@@ -346,9 +348,6 @@ def frost_days(
         Freezing temperature.
     freq : str
         Resampling frequency.
-    **indexer : {dim: indexer}, optional
-        Indexing parameters to compute the frost days on a temporal subset of the data.
-        It accepts the same arguments as :py:func:`xclim.indices.generic.select_time`.
 
     Returns
     -------
@@ -365,8 +364,7 @@ def frost_days(
        TN_{ij} < TT
     """
     frz = convert_units_to(thresh, tasmin)
-    sel = select_time(tasmin, **indexer)
-    out = threshold_count(sel, "<", frz, freq)
+    out = threshold_count(tasmin, "<", frz, freq)
     return to_agg_units(out, tasmin, "count")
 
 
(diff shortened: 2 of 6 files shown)


docs/notebooks/extendxclim.ipynb
```diff
@@ -386,6 +386,7 @@
     "    * A compute function name if given (here it refers to a function in `xclim.indices.generic`).\n",
     "    * Some parameters are injected, the default for `freq` is modified.\n",
     "    * The input variable `data` is mapped to a known variable. Functions in `xclim.indices.generic` are indeed generic. Here we tell xclim that the `data` argument is minimum daily temperature. This will set the proper units check, default value and CF-compliance checks.\n",
+    "- `summer_fd` is similar to `fd` but the `compute` references an index function from which the indicator builder will gather much information. It explicitly specifies a base class that support indexing (which the default `Daily` does not).\n",
```
juliettelavoie: "will not gather much information" ? unclear to mw what this means.

But I think putting an example like this is great!

src/xclim/indicators/atmos/_temperature.py
```diff
@@ -907,7 +907,7 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
     parameters={"thresh": {"default": "0 degC"}, "window": {"default": 5}},
 )
 
-frost_days = Temp(
+frost_days = TempWithIndexing(
```
juliettelavoie, line 910: In the example above, you say that the default base class of frost_days is Daily, but now you are changing it for ResamplingIndicatorWithIndexing. 

The example should be on a class that doesn't usually do indexing.

src/xclim/indices/_simple.py
```diff
@@ -459,6 +459,11 @@ def max_n_day_precipitation_amount(pr: xarray.DataArray, window: int = 1, freq:
         Window size in days.
     freq : str
         Resampling frequency.
+    **indexer : {dim: indexer}, optional
+        Indexing parameters to compute the indicator on a temporal subset of the data.
+        The subset is taken after the N-day sum, thus including data from up to ``window // 2``
+        days before and after the selected period.
```
juliettelavoie: I don't think this is True. By default, the label is on the right edge of the window. Meaning it includes data up to ``window-1`` days before the selected period and nothing after.

juliettelavoie: This default is good for my specific case, but maybe not generally ? idk..

aulemahal: woups, indeed!

src/xclim/indicators/atmos/_temperature.py
aulemahal: On fera mieux alors!

docs/notebooks/extendxclim.ipynb
aulemahal: Paragraph rewritten and the weird confusion between spring, summer and winter has been fixed...

src/xclim/indices/_simple.py
```diff
@@ -495,6 +501,11 @@ def max_pr_intensity(pr: xarray.DataArray, window: int = 1, freq: str = "YS") ->
         Window size in hours.
     freq : str
         Resampling frequency.
+    **indexer : {dim: indexer}, optional
+        Indexing parameters to compute the indicator on a temporal subset of the data.
+        The subset is taken after the N-hour average, thus including data from up to ``window // 2``
```
juliettelavoie: I think the same correction (window-1) is needed here also.

juliettelavoie (approved):
great!

PR #2206, opened 2025-07-08
Smooth transition between ice and water formulas for saturation vapour pressure + Dewpoint indicator

### LINKED ISSUE #2165: Relative humidity computation with interpolation between ice and water (like ERA5)
### Addressing a Problem?

Empirical formulations for relative humidity (or more precisely for the saturation vapour pressure) often have two variants : with reference to liquid water or to ice. Not all models or observational systems use the same convention to decide when to use each. 

For example, the WMO convention is to always compute with respect to liquid water. Others, like the CRCM5 use a threshold of 0°C. Finally, ERA5 does a more complex quadratic interpolation between -23°C et 0°C. [See here](https://www.ecmwf.int/sites/default/files/elibrary/2016/17117-part-iv-physical-processes.pdf#subsection.7.4.2).

xclim allows the simple threshold but does not implement an interpolation like ERA5.

I think it could be interesting ?

### Potential Solution

May be the `ice_thresh` argument could also accept a string like "interp" ? And we could add a new argument to give the interpolation bounds to implement the equations 7.92 and 7.94 of the ERA5 document linked above.

This would probably benefit from some rewriting of the `saturation_vapor_pressure` function to streamline the idea and allow other interpolation types to be implemented if needed.

### Additional context

Does this make sense ?

Pinging people that might want this : @juliettelavoie @coxipi  @mccrayc

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2165

* New arguments `interp_power` and `water_thresh` for `saturation_vapor_pressure` and all related functions. When given, `interp_power` controls the shape of the transition between ice and water as reference for the saturation vapour pressure formulas. 
     + Deprecated 4 letters methods, but only from the doc. A conversion is done to support them anyway.
* Rewrite of the e_sat function to make this simpler (?) and extensible
* New methods "buck81" and "aerk96".
* New method "ECMWF" that combines both.
* New indicators:
   + `vapor_pressure` : To compute the partial pressure of water vapor.
   + `dewpoint_from_specific_humidity` : To get `tdps` from `huss` and `ps`.

~No.~

I changed the value for $\epsilon$ in `specific_humidity` and  `specific_humidity_from_dewpoint` so all three uses of the same constant are now identical. The change is on the 5th decimal, I don't think it really "breaks".

I also changed the "var_name" of indicator `specific_humidity_from_dewpoint` to `huss`, which makes more sense!

### DIFF
--- a/src/xclim/indices/_conversion.py
+++ b/src/xclim/indices/_conversion.py
@@ -141,6 +141,7 @@ def humidex(
     elif hurs is not None:
         # Convert dry bulb temperature to Celsius
         tasC = convert_units_to(tas, "celsius")
+        hurs = convert_units_to(hurs, "%")
         e = hurs / 100 * 6.112 * 10 ** (7.5 * tasC / (tasC + 237.7))
 
     else:
@@ -592,6 +593,7 @@ def vapor_pressure_deficit(
         tas, ice_thresh=ice_thresh, method=method, interp_power=interp_power, water_thresh=water_thresh
     )
 
+    hurs = convert_units_to(hurs, "%")
     vpd = cast(xr.DataArray, (1 - (hurs / 100)) * svp)
 
     vpd = vpd.assign_attrs(units=svp.attrs["units"])
--- a/src/xclim/indices/_conversion.py
+++ b/src/xclim/indices/_conversion.py
@@ -436,7 +436,7 @@ def _saturation_vapor_pressure_over_ice(tas: xr.DataArray, method: str):
     return e_sat
 
 
-@declare_units(tas="[temperature]", ice_thresh="[temperature]")
+@declare_units(tas="[temperature]", ice_thresh="[temperature]", water_thresh="[temperature]")
 def saturation_vapor_pressure(
     tas: xr.DataArray,
     ice_thresh: Quantified | None = None,
@@ -540,7 +540,7 @@ def saturation_vapor_pressure(
     return e_sat
 
 
-@declare_units(tas="[temperature]", hurs="[]", ice_thresh="[temperature]")
+@declare_units(tas="[temperature]", hurs="[]", ice_thresh="[temperature]", water_thresh="[temperature]")
 def vapor_pressure_deficit(
     tas: xr.DataArray,
     hurs: xr.DataArray,
@@ -598,6 +598,7 @@ def vapor_pressure_deficit(
     huss="[]",
     ps="[pressure]",
     ice_thresh="[temperature]",
+    water_thresh="[temperature]",
 )
 def relative_humidity(
     tas: xr.DataArray,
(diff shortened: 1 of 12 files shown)


src/xclim/indices/_conversion.py
```diff
@@ -355,13 +355,96 @@ def sfcwind_to_uas_vas(
...
+            + -6028.076559 / tas
+            + 19.54263612
+            + -2.737830188e-2 * tas
+            + 1.6261698e-5 * tas**2
+            + 7.0229056e-10 * tas**3
+            + -1.8680009e-13 * tas**4
+            + 2.7150305 * np.log(tas)
+        )
+    else:
+        raise ValueError(f"Method {method} is not in ['sonntag90', 'tetens30', 'goffgratch46', 'wmo08', 'its90']")
```
juliettelavoie:
```suggestion
        raise ValueError(f"Method {method} is not in ['sonntag90', 'tetens30', 'goffgratch46', 'wmo08', 'its90', 'ERA5']")
```

src/xclim/indices/_conversion.py
```diff
@@ -645,15 +713,21 @@ def relative_humidity(
         Rw = (461.5,)
         hurs = 100 * np.exp(-L * (tas - tdps) / (Rw * tas * tdps))  # type: ignore
     elif tdps is not None:
-        e_sat_dt = saturation_vapor_pressure(tas=tdps, ice_thresh=ice_thresh, method=method)
-        e_sat_t = saturation_vapor_pressure(tas=tas, ice_thresh=ice_thresh, method=method)
+        e_sat_dt = saturation_vapor_pressure(
+            tas=tdps, ice_thresh=ice_thresh, method=method, interp_power=interp_power, water_thresh=water_thresh
```
juliettelavoie, line 785: with your last comment on dew point temperature, should we put the switch ERA5 method to wmo method here ?

aulemahal: Hmm, I think you might be mixing up "method" and "interpolation" ?
The WMO Standard is to compute everything in reference to water. So does ECMWF with the dew point temperature. However, the WMO guidelines use the "wmo08" equations while the ECMWF uses the "~era5~ecmwf" equations. Both formulas are similar, but not exactly the same.

juliettelavoie: ok yes, i was mixing them up. but still doesn't this mean that we should only use the ECMWF formula for water here, no interpolation or ice ?

aulemahal: From my understanding, this is indeed what you should do in your call to be consistent ? But xclim can still offer all options.

juliettelavoie (approved):
super! Merci pour l'explication sous Notes, je pense que c'est assez claire malgré le sujet qui est pas clair pentoute!

PR #2207, opened 2025-07-09
Refactoring wine climatology indices

### LINKED ISSUE #2201: xclim.indices.huglin_index and method jones
### Setup Information

- Xclim version: 0.57.0
- Python version: 3.11.3
- Operating System: MacOS 15.5 "Sequoia"

### Description

Calculating the Huglin index using the 'Jones' method and an output frequency = 'MS' (monthly), I don't get comparable values if I use an output frequency = 'YS' (yearly).
I think the problem is that the function indices.huglin_index() performs an unnecessary resampling commentes at line 277 of the screenshot, since another resampling occurs later at line 287.

Image

After these changes, I am able to obtain good results.

### Steps To Reproduce

_No response_

### Additional context

_No response_

### Contribution

### Code of Conduct

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #2201 

* Refactors the latitude correction algorithms into `xc.indices.helpers`
* Adjusts existing wine-climatology call signatures to be more restrictive about allowed time frequencies
* Adjusted the `"jones"`, method for the `huglin_index` to drop latitudes falling within the polar extremes:
  * For cases where all latitude values for a growing season fall below `1.0`, that season is dropped from the array.
    * If no season has values above `1.0`, a `ValueError` is raised.
* Added a `cap_value` for simple latitude coefficient functions to allow users to set an arbitrary value for extreme latitudes (e.g. `np.nan`, `1.0`, `1.08`, etc.).
* Added the option of setting values for polar days and polar nights in `day_lengths`
* `day_lengths` no longer modifies the `lat` DataArray passed to it
* latitude-day-length-coefficient algorithms are now found in `xclim.indices.helpers`:
  * gladstones_day_length_latitude_coefficient
  * huglin_day_length_latitude_coefficient
  * jones_day_length_latitude_coefficient

Yes. The `icclim` method for calculating Heliothermal Index of Huglin (HI) has been renamed the `huglin` method. This algorithm is directly based on the original [Huglin (1978)](https://www-iuem.univ-brest.fr/wapps/letg/adviclim/BDX/PDF/CR_Acad%C3%A9mie_agriculture_1978_64_Huglin.pdf) paper. The `icclim` option is still accepted (reverts to `huglin`) but should be deprecated.

The seasonally-aggregated method for calculating the latitude adjustment factors in HI and BEDD based on [Hall and Jones (2010)](https://onlinelibrary.wiley.com/doi/10.1111/j.1755-0238.2010.00100.x) are now restricted to `YS-JAN` or `YS-JUL` time frequencies. Additionally, the previous version of the `jones` day-length calculation were providing erroneous values for extreme latitudes where day-lengths can extend to either 0 or 24 hours; The new behaviour is to drop these latitudes from the calculation.

The `gladstones` method for calculating `k` in `biologically_effective_degree_days` now uses a dedicated function based on a dynamic `day_length` compared to a reference latitude (40 degrees). The previous implementation of the `"gladstones"` method was based off an approximation based on `Hall and Jones (2010)`; This new version is based directly off of [Gladstones (1992) and (2011)](https://www.wakefieldpress.com.au/product/wine-terroir-and-climate-change/)

### Other information:

The `day_length` function has been modified to accept an `infill_polar_days` argument (default: False) that adds the zero- or 24-hour latitudes as needed (instead of `np.nan`), e.g. `infill_polar_days=False`:
image

and `infill_polar_days=True`:
image

For the `"jones"` methods in HI and BEDD, this is hard-coded to `False`, but is available for other indices to make use of.

### DIFF
--- a/src/xclim/indices/_agro.py
+++ b/src/xclim/indices/_agro.py
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from typing import cast
+import warnings
+from typing import Literal, cast
 
 import numpy as np
 import xarray
@@ -26,7 +27,7 @@ from xclim.indices._threshold import (
     first_day_temperature_below,
 )
 from xclim.indices.generic import aggregate_between_dates, get_zones
-from xclim.indices.helpers import _gather_lat, day_lengths, resample_map
+from xclim.indices.helpers import _gather_lat, day_length_latitude_coefficient, day_lengths, resample_map
 from xclim.indices.stats import standardized_index
 
 # Frequencies : YS: year start, QS-DEC: seasons starting in december, MS: month start
@@ -154,7 +155,7 @@ def huglin_index(
     freq: str = "YS",
 ) -> xarray.DataArray:
     r"""
-    Huglin Heliothermal Index.
+    Heliothermal Index of Huglin.
 
     Growing-degree days with a base of 10°C and adjusted for latitudes between 40°N and 50°N for April-September
     (Northern Hemisphere; October-March in Southern Hemisphere). Originally proposed in :cite:t:`huglin_nouveau_1978`.
@@ -183,7 +184,7 @@ def huglin_index(
     Returns
     -------
     xarray.DataArray, [unitless]
-        Huglin heliothermal index (HI).
+        Heliothermal index of Huglin (HI).
 
     Notes
     -----
@@ -222,12 +223,12 @@ def huglin_index(
                      \end{cases}
 
     A more robust day-length calculation based on latitude, calendar, day-of-year, and obliquity is available with
-    `method="jones"`. See: :py:func:`xclim.indices.generic.day_lengths` or :cite:t:`hall_spatial_2010` for more
-    information.
+    `method="jones"`. The current implementation requires an annual frequency for consistent results.
+    See: :py:func:`xclim.indices.generic.day_lengths` or :cite:t:`hall_spatial_2010` for more information.
 
     References
     ----------
-    :cite:cts:`huglin_nouveau_1978, hall_spatial_2010`
+    :cite:cts:`huglin_nouveau_1978,hall_spatial_2010`
     """
     tas = convert_units_to(tas, "degC")
     tasmax = convert_units_to(tasmax, "degC")
@@ -236,37 +237,17 @@ def huglin_index(
     if lat is None:
         lat = _gather_lat(tas)
 
-    if method.lower() == "smoothed":
-        lat_mask = abs(lat) <= 50
-        lat_coefficient = ((abs(lat) - 40) / 10).clip(min=0) * 0.06
-        k = 1 + xarray.where(lat_mask, lat_coefficient, np.nan)
-        k_aggregated = 1
-    elif method.lower() == "icclim":
-        k_f = [0, 0.02, 0.03, 0.04, 0.05, 0.06]
-
-        k = 1 + xarray.where(
-            abs(lat) <= 40,
-            k_f[0],
-            xarray.where(
-                (40 < abs(lat)) & (abs(lat) <= 42),
-                k_f[1],
-                xarray.where(
-                    (42 < abs(lat)) & (abs(lat) <= 44),
-                    k_f[2],
-                    xarray.where(
-                        (44 < abs(lat)) & (abs(lat) <= 46),
-                        k_f[3],
-                        xarray.where(
-                            (46 < abs(lat)) & (abs(lat) <= 48),
-                            k_f[4],
-                            xarray.where((48 < abs(lat)) & (abs(lat) <= 50), k_f[5], np.nan),
-                        ),
-                    ),
-                ),
-            ),
-        )
+    if method.lower() in ["icclim", "smoothed"]:
+        k = day_length_latitude_coefficient(lat, method="icclim")
         k_aggregated = 1
     elif method.lower() == "jones":
+        if not freq.startswith("YS"):
+            msg = (
+                f"Freq {freq} not supported. Must be `YS` or `YS-*` for method 'jones'. "
+                "An annual frequency is required for the current implementation."
+            )
+            raise NotImplementedError(msg)
+
         day_length = (
             select_time(
                 day_lengths(dates=tas.time, lat=lat, method="simple"),
@@ -279,7 +260,7 @@ def huglin_index(
         k = 1
         k_aggregated = 2.8311e-4 * day_length + 0.30834
     else:
-        raise NotImplementedError(f"'{method}' method is not implemented.")
+        raise NotImplementedError("Method is not implemented. Only 'icclim', 'jones', and 'smoothed' are supported.")
 
     hi: xarray.DataArray = (((tas + tasmax) / 2) - thresh).clip(min=0) * k
     hi = (
@@ -304,7 +285,7 @@ def biologically_effective_degree_days(
     tasmax: xarray.DataArray,
     lat: xarray.DataArray | None = None,
     thresh_tasmin: Quantified = "10 degC",
-    method: str = "gladstones",
+    method: Literal["gladstones", "gladstones_simple", "huglin", "icclim"] = "gladstones",
     low_dtr: Quantified = "10 degC",
     high_dtr: Quantified = "13 degC",
     max_daily_degree_days: Quantified = "9 degC",
@@ -328,15 +309,15 @@ def biologically_effective_degree_days(
         Maximum daily temperature.
     lat : xarray.DataArray, optional
         Latitude coordinate.
-        If None and method in ["gladstones", "icclim"],
-        a CF-conformant "latitude" field must be available within the passed DataArray.
+        If None and method is no "icclim", a CF-conformant "latitude" field must be available within the passed DataArray.
     thresh_tasmin : Quantified
         The minimum temperature threshold.
-    method : {"gladstones", "icclim", "jones"}
+    method : {"gladstones", "gladstones_simple", "huglin", "icclim", "jones"}
         The formula to use for the calculation.
-        The "gladstones" integrates a daily temperature range and latitude coefficient. End_date should be "11-01".
+        The "gladstones" method uses a temperature range adjustment and a latitude coefficient based on :cite:t:`huglin_biologie_1998`. End_date should be "11-01".
+        The "gladstones_simple" method integrates a daily temperature range and latitude coefficient. End_date should be "11-01".
         The "icclim" method ignores daily temperature range and latitude coefficient. End date should be "10-01".
-        The "jones" method integrates axial tilt, latitude, and day-of-year on coefficient. End_date should be "11-01".
+        The "jones" method integrates axial tilt, latitude, and day-of-year on coefficient based on :cite:t:`hall_spatial_2010`. End_date should be "11-01".
     low_dtr : Quantified
         The lower bound for daily temperature range adjustment (default: 10°C).
     high_dtr : Quantified
@@ -357,7 +338,7 @@ def biologically_effective_degree_days(
 
     Warnings
     --------
-    Lat coordinate must be provided if method is "gladstones" or "jones".
+    Lat coordinate must be provided if method is "gladstones", "gladstones_simple", or "huglin".
 
     Notes
     -----
@@ -385,7 +366,7 @@ def biologically_effective_degree_days(
 
        k = f(lat) = 1 + \left(\frac{\left| lat  \right|}{50} * 0.06,  \text{if }40 < ``|lat|`` <50, \text{else } 0\right)
 
-    A second version of the BEDD (`method="icclim"`) does not consider :math:`TR_{adj}` and :math:`k` and employs a
+    An alternative version of the BEDD (`method="icclim"`) does not consider :math:`TR_{adj}` and :math:`k` and employs a
     different end date (30 September) :cite:p:`project_team_eca&d_algorithm_2013`.
     The simplified formula is as follows:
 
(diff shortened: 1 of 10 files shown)


tests/test_indices.py
```diff
@@ -268,79 +268,108 @@ def test_corn_heat_units(self, tasmin_series, tasmax_series):
...
+                if method == "icclim":
+                    # Latitude has no influence on 'icclim' method
+                    np.testing.assert_array_equal(bedd.isel(lat=0), bedd.isel(lat=-1))
+                elif method in ["huglin", "interpolated"]:
+                    # Leap year has no influence on 'huglin' or 'interpolated' method
+                    np.testing.assert_array_equal(bedd.isel(lat=0)[0], bedd.isel(lat=0)[1])
+                else:
+                    # Leap-year has slightly higher values for higher latitudes
+                    np.testing.assert_array_less(bedd[0], bedd[1])
+                    np.testing.assert_array_less(bedd[1], bedd[2])
```
coxipi: starts in 1992 (LY) and ends in 1996 (LY)

Should it not be something like: 

```suggestion
                    np.testing.assert_array_less(bedd[1], bedd[0])
                    np.testing.assert_array_less(bedd[1], bedd[-1])
```

according to your comment

coxipi: hmm latitudes are not selected yet? I'm still confused regardless

Zeitsperre:
@antoniogpepe and @giuliaisnotavailable

This Pull Request adds much more control over the day-length latitude coefficient algorithms and corrects several issues that were previously impacting outputs:
- NaN values are more heavily considered (i.e. latitudes with excessive NaN values for a growing season should not be considered).
- Setting a floor value for some algorithms was needed (i.e. the method from Hall and Jones 2010).
- The namings of several methods were inaccurate/imprecise.

This represents a huge breaking change, but provides values much closer to the implementations found in the source materials. For lower latitudes (below 30 deg) and higher latitudes (above 52 deg) one should expect significant changes.

src/xclim/indices/helpers.py
```diff
@@ -224,7 +242,7 @@ def eccentricity_correction_factor(
 def cosine_of_solar_zenith_angle(
     time: xr.DataArray,
     declination: xr.DataArray,
-    lat: Quantified,
+    lat: Quantified | xr.DataArray | xr.DataTree,
```
coxipi:
```suggestion
    lat: Quantified | xr.DataTree,
```

Quantified already includes DataArray no? it's at least equivalent to str | xr.DataArray

src/xclim/indices/helpers.py
```diff
@@ -445,32 +464,363 @@ def day_lengths(
...
+        The "simple" method uses a simple approximation of the day length based on latitude and time of year.
+        The "spencer" method uses a more complex approximation based on the Fourier series of the solar declination.
+
+    Returns
+    -------
+    xarray.DataArray, [dimensionless]
+        Coefficient for the day length based on latitude.
+    """
+    if not isinstance(lat, xr.DataArray):
+        lat = xr.DataArray(lat, attrs={"units": "degree_north"})
```
coxipi, line 659: it's a bit ambiguous? At the very least, say in the docstring that a float/int should be passed as degree_north

src/xclim/indices/helpers.py
```diff
@@ -142,7 +160,7 @@ def solar_declination(time: xr.DataArray, method="spencer") -> xr.DataArray:
             + 0.001480 * np.sin(3 * da)
         )
     else:
-        raise NotImplementedError(f"Method {method} must be one of 'simple' or 'spencer'")
+        raise NotImplementedError("Method must be one of 'simple' or 'spencer'.")
```
coxipi, line 163: what is the convention or printing the wrong argument? We usually print it no? Should we not? I think it's useful, but I don't mind either way. We should try to be as consistent as possible though?

src/xclim/indices/helpers.py
```diff
@@ -445,32 +464,363 @@ def day_lengths(
...
+        day_length_hours = ((24 / np.pi) * np.arccos(-np.tan(radians) * np.tan(declination))).assign_attrs(units="h")
+
+    if infill_polar_days:
+        lat_broadcast, decl_broadcast = xr.broadcast(lat_deg, declination)
+        # Polar day: sun never sets; Polar night: sun never rises.
+        polar_day = ((lat_broadcast > 66.5) & (decl_broadcast > 0)) | ((lat_broadcast < -66.5) & (decl_broadcast < 0))
+        polar_night = ((lat_broadcast > 66.5) & (decl_broadcast < 0)) | ((lat_broadcast < -66.5) & (decl_broadcast > 0))
+        # Infill polar days with 24 hours and polar nights with 0 hours.
+        valid = ~xr.ufuncs.isnan(day_length_hours)
+        day_length_hours = xr.where(polar_day & ~valid, 24.0, xr.where(polar_night & ~valid, 0.0, day_length_hours))
```
coxipi: I would avoid nested where when it's possible, it's difficult to read. But that's just style, do what you prefer
```suggestion
        day_length_hours = day_length_hours.where(~(polar_day & ~valid), 24.0)
        day_length_hours = day_length_hours.where(~(polar_night & ~valid), 0)
```

coxipi: was it for performance?

src/xclim/indices/helpers.py
```diff
@@ -445,32 +464,363 @@ def day_lengths(
...
+                        k_f[3],
+                        xr.where(
+                            (46 < lat_abs) & (lat_abs <= 48),
+                            k_f[4],
+                            xr.where((48 < lat_abs) & (lat_abs <= 50), k_f[5], _cap_value),
+                        ),
+                    ),
+                ),
+            ),
+        )
```
coxipi: here too. 
```suggestion
        k_f_bounds = [(0,0,40), (0.02,40,42), (0.03,42,44), (0.04,44,46), (0.05,46,48), (0.06,48,50)]
        k = xr.fill_like(lat_abs, _cap_value+1)
        
        for k_f_b in  k_f_bounds:
            cond = (k_f_b[1] < lat_abs)&(lat_abs <= k_f_b[2])
            k = k.where(~cond, 1+k_f_b[0])
```

src/xclim/indices/_agro.py
```diff
@@ -171,122 +179,89 @@ def huglin_index(
...
-            ),
-        )
-        k_aggregated = 1
+    k: int | xarray.DataArray = 1
+    k_aggregated: xarray.DataArray | None = None
+    if method.lower() in ["huglin", "icclim", "interpolated"]:
+        if method == "icclim":
+            warnings.warn("Method 'icclim' is deprecated. Use 'stepwise' instead.", DeprecationWarning)
+            method = "huglin"
+        k = huglin_day_length_latitude_coefficient(lat, method=method, cap_value=cap_value)
```
coxipi: not sure this does what you intended. I suspect you might want this instead: 
```suggestion
    if (method:=method.lower()) in ["huglin", "icclim", "interpolated"]:
        if method == "icclim":
            warnings.warn("Method 'icclim' is deprecated. Use 'stepwise' instead.", DeprecationWarning)
            method = "huglin"
        k = huglin_day_length_latitude_coefficient(lat, method=method, cap_value=cap_value)
```

coxipi: Because you could have `method=="Icclim"` and then this would not be catched in the next if condition because method.lower doesn't modify in-place

coxipi: also, did you mean 'hughlin' instead of 'stepwise'?

src/xclim/indices/_agro.py
```diff
@@ -171,122 +179,89 @@ def huglin_index(
...
-                     1.05, & \text{if } 46 < | lat | <= 48 \\
-                     1.06, & \text{if } 48 < | lat | <= 50 \\
-                     NaN, & \text{if } | lat | > 50 \\
-                     \end{cases}
-
-    A more robust day-length calculation based on latitude, calendar, day-of-year, and obliquity is available with
-    `method="jones"`. See: :py:func:`xclim.indices.generic.day_lengths` or :cite:t:`hall_spatial_2010` for more
-    information.
+    For compatibility with the original ICCLIM implementation :cite:p:`project_team_eca&d_algorithm_2013`,
+    `end_date` should be set to `11-01` with `method="huglin"` or `method="icclim"` (deprecated).
```
coxipi: I'm in favor of removing icclim for the actual name of the method. I'm a bit on the fence on this one. Maybe it's better to add a note to state that hughlin corresponds to icclim in old xclim versions. BUT, we do breaking changes from time to time, and we simply trust that our changelog is sufficient to inform users. So maybe it's not necessary actually

src/xclim/indices/_agro.py
```diff
@@ -383,29 +376,41 @@ def biologically_effective_degree_days(
...
+    :cite:cts:`gladstones_viticulture_1992,huglin_biologie_1998,project_team_eca&d_algorithm_2013,hall_spatial_2010`
     """
-    tasmin = convert_units_to(tasmin, "degC")
-    tasmax = convert_units_to(tasmax, "degC")
-    thresh_tasmin = convert_units_to(thresh_tasmin, "degC")
+    if not isinstance(freq, str):
+        raise TypeError("Freq must be a string.")
+
+    _tasmin = convert_units_to(tasmin, "degC")
+    _tasmax = convert_units_to(tasmax, "degC")
```
coxipi, line 396: curious, is there a reason not to modify in-place tasmax, tasmin here? I don't care that much for this, but FYI it deviates from the style of other xclim functions, if you do care.

coxipi: ah it seems like a mixed bag of both approaches, ignore this

src/xclim/indices/_agro.py
```diff
@@ -523,9 +519,9 @@ def cool_night_index(
         elif lat.lower() == "south":
             month = 3
         else:
-            raise ValueError(f"Latitude value unsupported: {lat}.")
+            raise ValueError(f"Latitude value not implemented: {lat}.")
```
coxipi: should this be a NotImplementedError then?

src/xclim/indices/helpers.py
coxipi: If it's not clear, I meant something like: f"Method must be in XYZ. Got {method}."

src/xclim/indices/helpers.py
coxipi: This is a bit different. my loop implements a series of if conditions, that are complimentary by design. Cases that don't satisfy any of the conditions default to _cap_value+1.  Your implementation has an explicit nested if/else structure, with the ultimate assignment corresponding to the _cap_value, if none of the conditions where passed. The result will be the same, but if you don't want to change this that's all  right

coxipi (approved):
approved, but I left a few comments, you can review them

src/xclim/indices/helpers.py
Zeitsperre: I'll preface this with "It's really a nitpick".

On the benign side, formatting a variable in a logging string can create a lot of noise, especially if for instance they pass a non-string (like a DataArray). There are security-related issues around this as well, but I'll explain those privately.

Essentially, we should be moving away from formatted strings in logging, especially when it comes to formatting user-provided arguments.

src/xclim/indices/helpers.py
Zeitsperre: Nope, I thought it was more readable my way, but I'll go with yours.

src/xclim/indices/helpers.py
Zeitsperre: I'm wondering why use `-np.inf` here when `0` will do the trick? Otherwise, this looks great.

src/xclim/indices/helpers.py
Zeitsperre: Can do

src/xclim/indices/_agro.py
Zeitsperre: I'm thinking that we should be adopting this approach in the library as a whole. I might open an issue about this. The unintended effect of `call_variable = func(call_variable)` is that any objects passed to the function are silently modified in place. This isn't great for users (forces them to pass copies to prevent this behaviour).

aulemahal: AFAIU, the two patterns:
```python3
call_variable = func(call_variable)

_call_variable = func(call_variable)
```
Have exactly the same effect on modifying the input variable or not, only the "pointer" has changed. Example:

```python3
def rename_key(dd):
    return {'new': dd['old']}

d_in = {'old': {'a': 1}}

d_out = rename_key(d_in)

d_out['new']['a'] = 2

print(d_out, d_in)
# {'new': {'a': 2}} {'old': {'a': 2}}
```
i.e. the name assigned to the result of the call doesn't matter : if the process modifies a variable in-place, the input is modified in-place.

My dict example uses basic types, but this sometimes happen with `DataArray` and `Dataset`. What I mean with this comment is that the issue is sadly not fixed by renaming the variable. I know two solutions : 
- deepcopying the input (underperformant, not cool)
- coding with precaution (tedious, but cool)

tests/test_indices.py
Zeitsperre: Good catch!

src/xclim/indices/_agro.py
Zeitsperre: Would this be worth opening an issue about, or is it way too daunting of a task? I feel like we should be doing what we can to prevent these types of changes at our control layer at the very least.

src/xclim/indices/helpers.py
coxipi: sure, I wasn't sure what kind of coordinate (some choice of latitudes can be negative?) Anyways, if it makes sense go for it

coxipi: oups sorry, I should have remove the `k_f` in the original code, this is already covered by `k_f_bounds`

src/xclim/indices/_agro.py
aulemahal: I'm not sure there's a xclim-wide issue to be opened here, what I'm describing is a normal quirk of python that's even more common with xarray objects. More like, devs should be aware that objects are passed by reference and to be careful not to modify them unnecessarily... all the while limiting calls to `copy` to a minimum because it affects performance...

Zeitsperre: If only Python had a better way of ensuring copies and references aren't ever mixed up... :crab:

PR #2208, opened 2025-07-09
Snowfall and rain approximation from Dai 2008

### LINKED ISSUE #1752: Add additionnal method to `snowfall_approximation`
### Addressing a Problem?

Collaborators from PCIC and CCCS expressed interest in potentially adding a `Dai 2008` method for snowfall approximation
  
https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2008GL033295

### Potential Solution

add the method

### Additional context

_No response_

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #1752

* New methods "dai_annual" and "dai_seasonal" for snowfall and rain approximation. 
* New argument `landmask` to control which formulation (land or ocean) we use.

No. Only new arguments added, no default changed.

### Other information:
@Zeitsperre : The `landmask` argument is polymorphic (my favorite, your nemesis (?)), it can either be a DataArray (which is assumed to be of boolean dtype, but no need to check) or a straight boolean. This is not a `Quantified` because boolean have no units. Right now, I put the annotation that resolved into the "optional variable" `InputType`, which means "bool" is not part of it. Should we create a new `InputType` for boolean mask ? It would be similar to `Quantified` in that it can be a scalar or a DataArray, but without the units check.

@tlogan2000 : I implemented the rain approximation as well. However, in opposition to the 3 other methods already implemented, the Dai rain approximation is not equal to `pr - prsn` because it considers some small fractions of the total precip to be of another phase (denoted "sleet" in the article). Thus, to imitate DonnéesClimatiques, ones need to compute `prra` as `pr - prsn` instead of using `rain_approximation`. Is this ok ?

Tests need to be added, I'll try to base my tests off of DonnéesClimatiques' data. At least it would compare xclim with another implementation, if not an "original" one.


Zeitsperre:
> @Zeitsperre : The `landmask` argument is polymorphic (my favorite, your nemesis (?)), it can either be a DataArray (which is assumed to be of boolean dtype, but no need to check) or a straight boolean. 

FTR, I don't hate polymorphism; I don't like when polymorphic variables aren't properly annotated or constrained. Live your best (polymorphic) life.

> This is not a `Quantified` because boolean have no units. Right now, I put the annotation that resolved into the "optional variable" `InputType`, which means "bool" is not part of it. Should we create a new `InputType` for boolean mask ? It would be similar to `Quantified` in that it can be a scalar or a DataArray, but without the units check.

I think that makes the most sense. Do you see this being used elsewhere in the library? If so, try to make that annotation accessible to multiple modules.

Zeitsperre:
@aulemahal 

The merge conflict is relatively annoying given that the conversion indices have been refactored. Let me know if you want me to take over that task.

PR #2217, opened 2025-07-24
Assume daily frequency when inferring units of a count or integral index

### LINKED ISSUE #2215: Inferring frequency in indicator aggregations fail when there are timestamps missing
### Setup Information

- Xclim version: 0.57.0
- Python version: 3.12.
- Operating System: Ubuntu

### Description

This might be totally expected but I noticed that one of my indicator computations using `wetdays` suddenly started failing after I upgraded my xclim version. I think I pinpointed it to this change: https://github.com/Ouranosinc/xclim/commit/e6d86843146f01bb8f10f23928fc5f4c557e8db4

That now raises an explicit error when the frequency cannot be inferred from the original data. I am explicitly using wetdays from xclim.indices and not indicators to avoid having to fix missing days so this is slightly annoying.  In this case, I could probably simply add the missing days by reindexing or similar but I wonder whether a function like wetdays when imported from the indices module should work even when the frequency cannot be inferred?

### Steps To Reproduce

```python
import numpy as np
import xarray as xr
from xclim.indices import wetdays
idx = xr.date_range('2000', '2020')
to_drop = np.random.choice(idx, size=20, replace=False)

new_idx = idx.difference(to_drop)

da = xr.DataArray(10*np.random.randn(new_idx.size), coords=dict(time=('time', new_idx))).assign_attrs(units='mm day-1')

wetdays(da, '1 mm day-1')
```
=> `ValueError: Unable to find the sampling frequency of the data.`
### Additional context

_No response_

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2215

 `deffreq` is passed from the index function to `infer_sampling_units` via `to_agg_units` when it matters.

This allows an index to define a default frequency to assume if the data has missing steps, useful when the units depends on it, when `op='count'` or `op='integral'`. 

This way, indices that are clearly written with daily inputs in mind or that do not make sense for other frequencies anyway can avoid failing on data with a few missing days. 

Indicators still have the data check that will fail on non-inferrable time coordinates, and that can also be disabled.

No, what worked before still works the same. Some cases that failed before don't fail anymore.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Internal changes
 
 Bug fixes
 ^^^^^^^^^
+* Indices relying on ``units.to_agg_units(src, out, 'count')`` will not raise on a non-inferrable frequency and instead use the common default of "D", as their docstring implies. (:issue:`2215`, :pull:`2217`).
 * Increase the tolerance in the tests of ``xclim.indices.standardized_groundwater_index`` (the standardized indices are sensitive to package versions because of the parameter optimization in `scipy`).  (:issue:`2183`, :pull:`2193`).
 * In ``xclim.indices._conversion.humidex`` and ``xclim.indices._conversion.vapor_pressure_deficit``, add a converter to ensure `hurs` has '%' units. (:pull:`2209`).
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,9 +27,9 @@ Internal changes
 
 Bug fixes
 ^^^^^^^^^
-* Indices relying on ``units.to_agg_units(src, out, 'count')`` will not raise on a non-inferrable frequency and instead use the common default of "D", as their docstring implies. (:issue:`2215`, :pull:`2217`).
 * Increase the tolerance in the tests of ``xclim.indices.standardized_groundwater_index`` (the standardized indices are sensitive to package versions because of the parameter optimization in `scipy`).  (:issue:`2183`, :pull:`2193`).
 * In ``xclim.indices._conversion.humidex`` and ``xclim.indices._conversion.vapor_pressure_deficit``, add a converter to ensure `hurs` has '%' units. (:pull:`2209`).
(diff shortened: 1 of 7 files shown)


CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Internal changes
 
 Bug fixes
 ^^^^^^^^^
+* Indices relying on ``units.to_agg_units(src, out, 'count')`` will not raise on a non-inferrable frequency and instead use the common default of "D", as their docstring implies. (:issue:`2215`, :pull:`2217`).
```
Zeitsperre: Should we add newer PRs above or below existing ones? Does it matter?

src/xclim/core/units.py
```diff
@@ -532,7 +532,11 @@ def infer_sampling_units(
     freq = xr.infer_freq(da)
     if freq is None:
         if deffreq is None:
-            raise ValueError("Unable to find the sampling frequency of the data.")
+            raise ValueError(f"Unable to find the sampling frequency of the data along dimension {dim}.")
+        warnings.warn(
+            f"Unable to find the sampling frequency of the data along dimension {dim}. Assuming '{deffreq}' instead.",
+            stacklevel=2,
+        )
```
Zeitsperre:
```suggestion
        msg = f"Unable to find the sampling frequency of the data along dimension {dim}. Assuming '{deffreq}' instead."
        warnings.warn(msg, stacklevel=2)
```

src/xclim/core/units.py
```diff
@@ -636,6 +641,9 @@ def to_agg_units(
         but the units are multiplied by the timestep of the data (requires an inferrable frequency).
     dim : str
         The time dimension along which the aggregation was performed.
+    deffreq : str, optional
+        For ``op`` `count` and `integral`, this gives the default source frequency to assume,
```
Zeitsperre:
```suggestion
        For operations `count` and `integral`, this gives the default source frequency to assume,
```

Zeitsperre (approved):
Nice approach. I can see this being useful for many other helper functions/indices.

PR #2224, opened 2025-08-13
Refactor converters into their own modules

### LINKED ISSUE #1289: Variable converters should become their own module/submodule
The `Converter` indicator class and functions found under `xclim.indices._conversion` are beginning to feel like their own specific type of module. These functions are not reliant on a particular frequency, but are useful for generating the variables used for several indices and indicators. A key benefit of using a `conversion` function is that metadata fields and units remain valid, even if units and metadata of composing variables are not great/consistent.

In #1275, we introduce a few `converter` functions for going between `snow water equivalent in metres` and `snow depth` using `snow density`, but this creates three separate functions with very simple operations. This PR also introduces radiation `converters` with the same issue: lots of duplication of code.

In discussion with @aulemahal, we came to the realization that many conversions can be replaced with one-liner arithmetic operations, possibly even generic functions, similar to `indices.generic.compare`. In order to reduce the number of separate functions needed for converting between variables, perhaps we could outline "relationships" between variables that could be used in a more generalized `convert_to_variable` function, thus reducing the amount of duplicate functions. This could then be used to create `Converter`-based indicators with all appropriate metadata.

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1289 

* Factors out the conversion-focused indices and indicators into their own publicly-facing submodules (`xclim.indices.converters` and `xclim.indicators.convert`)
* The majority of tests under `test_indicators.py` were actually tests for indicators that are converters. The tests have been split off into `test_converters.py`
* Fixed a noisy `RuntimeWarning` emitted by log operations in `numpy`

Yes:
- Indicators that were identified as converters have moved, mostly from `xclim.atmos` with a few from `xclim.land` to a newly-registered `xclim.convert`.

... and no:
- The layout of `xclim.indices` has changed, but the conversion indices remain accessible, all listed under the same module.

### Other information:

The conversion indices are imported twice in `src/xclim/indices/__init__.py`: once as `converters` and again using a *-import. Removing the *-import should be the goal here, but this will be an enormous breaking change.


tests/test_atmos.py
```diff
@@ -1,229 +1,22 @@
...
-    pr = pr_series(np.ones(10))
-    tas = tas_series(np.arange(10) + K2C)
+        # Expected values
+        exp = [2, 9, 72, 24, 0]
 
-    prlp = atmos.rain_approximation(pr, tas=tas, thresh="5 degC", method="binary")
+        lfd = atmos.late_frost_days(tasmin, date_bounds=("04-01", "06-30"))
 
-    np.testing.assert_allclose(prlp, [0, 0, 0, 0, 0, 1, 1, 1, 1, 1], atol=1e-5, rtol=1e-3)
+        np.testing.assert_allclose(lfd.isel(time=0), exp, rtol=1e-03)
```
coxipi, line 19: this is just some test that was already there, but since a lot of stuff is removed, it's highlighted? the implementation has changed a bit?

tests/test_indices.py
```diff
@@ -4358,7 +4360,7 @@ def test_snowfall_frequency(self, prsnd_series, prsn_series):
         # test prsn [kg m-2 s-1]
         prsn = prsn_series(np.array([0, 2, 0.3, 0.2, 4]), units="mm day-1")
         prsn = convert_units_to(prsn, "kg m-2 s-1", context="hydro")
-        out = xci.snowfall_frequency(prsnd)
+        out = xci.snowfall_frequency(prsn)
```
coxipi, line 4366: good catch

tests/test_temperature.py
```diff
@@ -1398,8 +1398,8 @@ def test_corn_heat_units(open_dataset):
         tnC = tn - K2C
         tnC.attrs["units"] = "C"
 
-    chu = atmos.corn_heat_units(tasmin=tn, tasmax=tx, thresh_tasmin="4.44 degC", thresh_tasmax="10 degC")
-    chuC = atmos.corn_heat_units(tasmin=tnC, tasmax=tx, thresh_tasmin="4.44 degC", thresh_tasmax="10 degC")
+    chu = convert.corn_heat_units(tasmin=tn, tasmax=tx, thresh_tasmin="4.44 degC", thresh_tasmax="10 degC")
+    chuC = convert.corn_heat_units(tasmin=tnC, tasmax=tx, thresh_tasmin="4.44 degC", thresh_tasmax="10 degC")
```
coxipi: hum, I think `corn_heat_units` should not be in convert. It has thresholds and all that. It should go back in agro  IMO

tests/test_converters.py
```diff
@@ -0,0 +1,568 @@
+# This File is for testing atmos indicators that fit neither in test_precip or test_temperature
+# Expected values might be the same as for the indices tests, see test_indices comments.
+from __future__ import annotations
```
coxipi, line 3: are these all existing tests that were ported here?

src/xclim/indices/_conversion.py
```diff
@@ -1,32 +1,14 @@
 """Conversion and approximation functions."""
 
-from __future__ import annotations
+import warnings
 
-from typing import cast
+from xclim.indices.converters import *
```
coxipi, line 5: had a heart jump, but it's just temporary for phasing the deprecation, OK

coxipi (approved):
approved! but change corn_heat_units location I think

tests/test_converters.py
Zeitsperre: Precisely. It just so happens that 90% of the tests in `test_indicators.py` were testing conversion tools.

src/xclim/indices/_conversion.py
Zeitsperre: It is, but I'm torn between two ideas:
1. We keep all indices together with star imports just as before.
2. We only expose the `converters` under `xclim.indices.converters`.

For `generic.py` users need to explicitly import that module (though it is called on in various `_submodules`). The second makes sense to me, but is the most jarring change.

aulemahal: I would argue that converters are only a (special) subcategory of indices, where are `generic` are not indices (but tools to build those). Thus, it would mean we should expose all converters in `xclim.indices` as before.

tests/test_temperature.py
Zeitsperre: I agree. I need to see what can be done about the Indicator version of this indice though; It's currently a `Converter` there, using CF-checking. Perhaps that can be changed without breaking anything.

tests/test_atmos.py
Zeitsperre: It's a different test entirely. I think the `diff` is having a moment.

src/xclim/indices/_conversion.py
Zeitsperre: The indices are accessible just as before, but if users wish it, `converters` are now exposed to further identify these as special indices.

src/xclim/core/indicator.py
```diff
@@ -948,10 +948,13 @@ def __call__(self, *args, **kwds):
...
             return out
 
-        # Return a single DataArray in case of single output, otherwise a tuple
+        # Return a single DataArray in case of single output
         if self.n_outs == 1:
             return outs[0]
-        return tuple(outs)
+
+        # Return a NamedTuple for multiple outputs
+        NamedOuts = namedtuple(self.identifier, f"{', '.join([o.name for o in outs])}")
```
aulemahal:
```suggestion
        NamedOuts = namedtuple(self.identifier, [o.name for o in outs])
```
Both work, but this seems clearer, no ?

Zeitsperre: Ah, most examples for `NamedTuple` suggest initializing them with strings (`namedtuple("SomeName", "var1 var2 var3")`). Your suggestion is much clearer.

Issue #1707, opened 2024-04-16
Investigate `numbagg` for performance with Quantiles


Zeitsperre:
Also numbagg added a performant nan_quantile not too long ago (https://github.com/numbagg/numbagg/pull/166), it might be interesting to add it to the comparison.

_Originally posted by @bzah in https://github.com/Ouranosinc/xclim/issues/1694#issuecomment-2025104841_

Zeitsperre:
`numbagg` was added as an optional dependency in #2228

PR #2228, opened 2025-08-19
Fix tests relying on inexact pint 0.24 and avoid numbagg-introduced noise in standardized index testing

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Fix tests that were relying on inexact conversions from pint 0.24.4
    + `pr` data flags where strangely inputting an array full of 1 mm/d and then asserting that the "eq_1_repeating_for_10_or_more_days" tests was _false_. Fixed to True, skipped with older pint.
    + Similarly, the other test had hardcoded inexact kg m-2 s-1 values that converted to 1 in pint 0.24.4, but not pint 0.25.
    + The dry spell tests had a day where the precip amount was just under 3 mm because of inexact conversion, but in theory it should have been 3 (which it is in new pint). The threshold was modified to be less sensitive.
* Avoid `numbagg` for standardized indices. It seems the use (or not) of numbagg changes only slightly the rolling mean values, but this change is enough to throw off the mle-fit-based result of the test. I create a simple fixture that simply deactivates the usage of `numbagg` in the tests where it is added as an arg.

No.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -389,7 +389,8 @@ convention = "numpy"
 [tool.vulture]
 exclude = []
 ignore_decorators = ["@pytest.fixture"]
-ignore_names = []
+# no_numbagg is a simple fixture that activates a xr option, it has no value
+ignore_names = ['no_numbagg']
 min_confidence = 90
 paths = ["src/xclim", "tests"]
 sort_by_size = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,7 +170,7 @@ pep621_dev_dependency_groups = ["all", "dev", "docs"]
 "pyyaml" = "yaml"
 
 [tool.deptry.per_rule_ignores]
-DEP002 = ["bottleneck", "h5netcdf", "pyarrow"]
+DEP002 = ["bottleneck", "h5netcdf", "numbagg", "pyarrow"]
 DEP004 = ["matplotlib", "pooch", "pytest", "pytest_socket"]
 
 [tool.flit.sdist]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,7 +108,7 @@ docs = [
   "sphinxcontrib-bibtex",
   "sphinxcontrib-svg2pdfconverter[Cairosvg]"
 ]
-extras = ["flox >=0.9", "xsdba >=0.4.0"]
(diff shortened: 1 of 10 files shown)


aulemahal:
I agree with adding to the documentation. In both cases, the changes are in the realm of the insignificant noise, but notes reminding users of the imprecision of floating-point arithmetic and of the sensitivity of "ML" fitting would help.

I wouldn't go so far as to emit warnings, but having CI runs with numbagg is a good idea!

pyproject.toml
Zeitsperre, line 1: I'd add `numbagg` to the `dev` recipe as well (with a min version if possible).

tests/test_indices.py
```diff
@@ -1008,18 +1009,13 @@ def test_standardized_precipitation_evapotranspiration_index(
                 ],
             ),
             ("MS", 1, "fisk", "APP", [0.4663, -1.9076, -0.5362, 0.8070, -0.8035], 2e-2),
-            pytest.param(
+            (
```
Zeitsperre, line 1012: Was this failure related to Pint? Why was this failing?

tests/test_indices.py
```diff
@@ -1905,7 +1903,7 @@ def test_perfect_xmas_days(self, snd_series, prsn_series):
         prsnd.loc["2000-12-25"] = 5
         prsnd.loc["2001-12-25"] = 2
         prsnd.loc["2001-12-26"] = 30  # too bad it's Boxing Day
-        prsnd.loc["2002-12-25"] = 1  # not quite enough
+        prsnd.loc["2002-12-25"] = 0.9  # not quite enough
```
Zeitsperre: How close can we go?
```suggestion
        prsnd.loc["2002-12-25"] = 0.995  # not quite enough
```

aulemahal: That would be fine.

tests/test_indices.py
aulemahal: I think it was related to `numbagg`.

pyproject.toml
aulemahal: I would argue this is not needed ? At this point `numbagg` has the same status than `flox` : it is an extra that can help with performance, but not a real dependency of xclim, it is not explicitly called. 

We could wait an addition like discussed #1707 before adding it as a dependency ?

Zeitsperre: For sure. That makes sense.

PR #2233, opened 2025-08-26
Address click deprecation - use h5netcdf in cli tests

### LINKED ISSUE #2212: Address Click deprecation warnings
Simply to keep track of the coming deprecations in Click:

```
DeprecationWarning: 'MultiCommand' is deprecated and will be removed in Click 9.0. Use 'Group' instead.

DeprecationWarning: The '__version__' attribute is deprecated and will be removed in Click 9.1. Use feature detection or 'importlib.metadata.version("click")' instead.
```

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2212

* Address a Click warning about using `Group` instead of `MultiCommand`. I tested and the change is retrocompatible with 8.1 (our lowest click allowed)
* Use `h5netcdf` in the cli tests, hopefully that fixes the worker crashes ?

No

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -94,7 +94,7 @@ def test_normal_computation(tasmin_series, tasmax_series, pr_series, tmp_path, i
 
     ds.to_netcdf(input_file, engine="h5netcdf")
 
-    args = ["-i", str(input_file), "-o", str(output_file), "-v", indicator]
+    args = ["-i", str(input_file), "-o", str(output_file), "-v", "--engine", "h5netcdf", indicator]
     runner = CliRunner()
     results = runner.invoke(cli, args)
     for var_name in var_names:
@@ -126,6 +126,8 @@ def test_multi_input(tas_series, pr_series, tmp_path):
             "-o",
             str(output_file),
             "-v",
+            "--engine",
+            "h5netcdf",
             "solidprcptot",
         ],
     )
@@ -150,6 +152,8 @@ def test_multi_output(tmp_path, open_dataset):
             "-o",
             str(output_file),
             "-v",
+            "--engine",
+            "h5netcdf",
             "wind_speed_from_vector",
         ],
     )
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -94,7 +94,7 @@ def test_normal_computation(tasmin_series, tasmax_series, pr_series, tmp_path, i
 
     ds.to_netcdf(input_file, engine="h5netcdf")
 
-    args = ["-i", str(input_file), "-o", str(output_file), "-v", indicator]
+    args = ["-i", str(input_file), "-o", str(output_file), "-v", "--engine", "h5netcdf", indicator]
```
Zeitsperre, line 97: We could offer the `engine` as an option?

Would this technically work with an OPeNDAP URL? I know that `h5netcdf` doesn't support those. Otherwise, `h5netcdf` is a core dep of `xclim` so this is a good idea.

tests/test_cli.py
```diff
@@ -126,6 +126,8 @@ def test_multi_input(tas_series, pr_series, tmp_path):
             "-o",
             str(output_file),
             "-v",
+            "--engine",
+            "h5netcdf",
```
Zeitsperre, line 130: Does it make sense to join these? i.e. `--engine=h5netcdf`

Zeitsperre (approved):
LGTM!

tests/test_cli.py
Zeitsperre: My mistake, this call clearly shows that we already do offer the engine as an option (did I add that??).

I guess my question is: Is there a way to see if this could work for OPeNDAP/`netcdf4`?

tests/test_cli.py
aulemahal: We didn't do that for the other arguments, so I simply followed the same syntax.

tests/test_cli.py
aulemahal: I think it simply won't work. h5netcdf doesn't support OpenDAP. For users wanting to use OpenDAP the easiest would be not to pass `--engine` and let xarray decide, as usual. I added these options there only so that the CI doesn't use netCDF4, in a try to avoid crashes.

PR #2239, opened 2025-09-05
Fit with PWM method returns NaNs when unique values <= than dist args

### LINKED ISSUE #2235: standardized_index_fit_params wit lmoments fails for all equal values.
### Setup Information

- Xclim version: 0.57.0
- Python version:3.12.4
- Operating System: Linux

### Description

Fitting standardized parameters fails for timeseries that only have a single unique value when using the Gamma distribution from lmoments3/the `PWM` method. 

### Steps To Reproduce

```python
import xarray as xr
from xclim.indices.stats import standardized_index_fit_params
from lmoments3.distr import gam
import numpy as np

time = xr.date_range('2000-01-01', '2010-12-31', freq='D')
da = xr.DataArray(np.ones(time.size), coords=dict(time=('time', time)))

params = standardized_index_fit_params(
    da, window=6, freq="MS", dist=gam, method="PWM", zero_inflated=True
)
```

raises `ValueError: L-Moments invalid` in *lmoments3/distr.py:1207*.

One possible solution is to turn these timeseries to null but it has to happen after the data has been resampled. 
e.g. above works if I perform the resampling, rolling and additionally the nulling of data before calling `standardized_index_fit_params` like this:

```python
da = da.resample(time="MS").mean(keep_attrs=True)
da = da.rolling(time=6).mean(skipna=False, keep_attrs=True)
da = da.where(~(da.diff("time") == 0).all("time"))
params = standardized_index_fit_params(
    da.chunk(time=-1), window=6, freq="MS", dist=gam, method="PWM", zero_inflated=True
)
```

I was thinking maybe these lines could be added to *xclim.indices.stats.preprocess_standardized_index* or - potentially better - to `_fitfunc_1d`  inside the if statement for `PWM`.

Happy to implement one of the two if you think it's useful.

### Additional context

_No response_

### Contribution

### Code of Conduct

saschahofmann:
Since I dont know whether this happens for all distributions (although it would probably make sense if it does) I am wondering whether we should first try and catch the fitting and only turn the parameters to NaN afterwards if the unique value condition is fulfilled @coxipi ?

### Pull Request Checklist:
    - This PR fixes #2235

* As discussed on the issue the function now returns NaNs for cases it would previously fail for by automatically returning NaNs for all parameters when there less unique values in the provided array than the distribution as arguments

No, cases that were previously breaking with an Invalid LMoments ratio error will now pass.

### DIFF
--- a/tests/test_stats.py
+++ b/tests/test_stats.py
@@ -286,7 +286,9 @@ class TestPWMFit:
         lmom = pytest.importorskip("lmoments3.distr")
         lm3dc = getattr(lmom, lm3_dist_map[dist])
         time = xr.date_range("2000-01-01", "2000-12-31", freq="M")
-        unique_values = np.arange(lm3dc.numargs | 1) + 1
+        # Some distributions have no parameter but we want to ensure we have at least one
+        # unique value
+        unique_values = np.arange(lm3dc.numargs or 1)
 
         da = xr.DataArray(np.random.choice(unique_values, time.size), coords=dict(time=("time", time)))
         out = stats.fit(da, dist=lm3dc, method="PWM").compute()


coxipi:
Yes, I think it's a good idea.  In the end, what you want is this: if the error `ValueError: L-Moments invalid` comes up, you could always warn and fill with NaNs. Or you can perform the check you implemented, and if it's not a matter of a few uniques values, then throw an error.  

I would not mind if you simply always fill with nans in any case. You could add to the warning message that a possible cause of this is the presence of only a few unique values in the dataset.

saschahofmann:
For now, I went for the more conservative approach that let's it fail fail and then checks whether the condition is fulfilled to return only NaNs. Still need to check the precision issue i saw

saschahofmann:
Alright, finally I managed to recreate the precision issue:
```python
import xarray as xr
import numpy as np
from lmoments3 import distr
from xclim.indices import stats

values = np.array(
    [
        1.9230799e-18,
        6.6979600e-10,
        -3.7058231e-17,
        -3.7053198e-17,
        -3.7060428e-17,
        -3.7066678e-17,
    ]
)
time = xr.date_range("2000-01-01", "2000-01-29", freq="D")
da = xr.DataArray(values, coords=dict(time=("time", time)))

stats.fit(da, dist=distr.gam, method='PWM')
```

Even with the new check this will raise the ValueError for the gamma distribution

tests/test_stats.py
```diff
@@ -278,6 +278,20 @@ def test_pwm_fit(self, dist, use_dask, random):
...
 
+    @pytest.mark.parametrize("dist", lm3_dist_map.keys())
+    def test_not_enough_unique_values(
+        self,
+        dist,
+    ):
+        lmom = pytest.importorskip("lmoments3.distr")
+        lm3dc = getattr(lmom, lm3_dist_map[dist])
+        time = xr.date_range("2000-01-01", "2000-12-31", freq="M")
+        unique_values = np.arange(lm3dc.numargs | 1) + 1
```
coxipi: this gives either numargs or numargs+1 unique_values, which covers the cases we were considering. Is there a reason why for `np.arange(lm3dc.numargs | 1)` (numargs or numargs+1) instead of simply `np.arange(lm3dc.numargs)` ?  

Why not try both `np.arange(lm3dc.numargs)` AND `np.arange(lm3dc.numargs+1)`? 

Anyways, I think this test is good enough, but I'm not sure if I understand the reason why the bitwise OR is needed in this case.

coxipi: Just to further my point, should I understand this: 
* a 2 params dist will fail with unique-values <= 3
* a 3 params dist will fail with unique-values <= 3
or more generally, 
* a 2k params dist will fail with unique-values <= 2k+1
* a 2k+1 params dist will fail with unique-values <= 2k+1

That seems weird, but I'm not 100% familiar with the inner workings of l-moments, so maybe?

Zeitsperre:
@saschahofmann We're hoping to release a new version in the next day or so. If you have the time to address Éric's comments, we can sneak this last PR into it.

Thanks again for your contributions!

tests/test_stats.py
saschahofmann: Ok so i think there are two cases here:
I added the +1 to avoid 0 which i saw sometimes lead to issues but it doesnt seem to be problematic so I will just remove this.

The other is my bad instead of a bitwise operation I meant to write an or statement. To make sure that we have at least a single uniquevalue (otherwise the np.choice command. fails).

I am correcting it now and leaving a comment

coxipi: Great, we will merge this when the tests pass. Thank you for this contribution!

PR #2251, opened 2025-10-10
Fix history handling

### LINKED ISSUE #2240: Variable history still updated if  `as_dataset` is True
The global xclim option `as_dataset` makes indicator return Datasets instead of (tuples of) DataArrays. One enhancement from this option is that we can put the `history` attribute on the dataset, where the conventions says it should be. However, we forgot to skip writing it to the dataarrays, so results now have two (almost identical) history attributes, one of them in the wrong place.

https://github.com/Ouranosinc/xclim/blob/26c35af66d4d66b679d608d24a0a32c4f412733d/src/xclim/core/indicator.py#L1152-L1156

We simply need to skip that block with `if not OPTIONS[AS_DATASET]`.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2240 

* When `as_dataset=True` is activated, "history" is not added on the dataarray. Adding an history to a variable is not CF-compliant. We are doing it because we wouldn't have this information otherwise. But when in dataset mode, there's no need to do it (it is already added to the dataset).
* The `update_history` function was adding the new history at the end of the string. It should be at the top. The CF convention says "append", the netCDF user guide says "append", but I'm pretty sure that's loose language as all nc utilities I've encountered (`ncks` and the like) update the history by _prepending_. 

As in most cases, this is a gray area.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
         args: [ '--autofix', '--no-ensure-ascii', '--no-sort-keys' ]
         exclude: '.ipynb'
   - repo: https://github.com/pappasam/toml-sort
-    rev: v0.24.2
+    rev: v0.24.3
     hooks:
       - id: toml-sort-fix
         exclude: '.pylintrc.toml'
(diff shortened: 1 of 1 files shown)


src/xclim/core/formatting.py
```diff
@@ -433,11 +433,9 @@ def update_history(
...
         **inputs_kws,
     )
-    if len(merged_history) > 0 and not merged_history.endswith("\n"):
-        merged_history += "\n"
-    merged_history += (
-        f"[{dt.datetime.now():%Y-%m-%d %H:%M:%S}] {new_name or ''}: {hist_str} - xclim version: {__version__}"
-    )
+    merged_history = (
+        f"[{dt.datetime.now():%Y-%m-%d %H:%M:%S}] {new_name or ''}: {hist_str} - xclim version: {__version__}\n"
+    ) + merged_history
```
Zeitsperre, line 438: Not the biggest fan of string math, but it wasn't much better before, haha

CHANGELOG.rst
```diff
@@ -17,6 +17,8 @@ Breaking changes
 Bug fixes
 ^^^^^^^^^
 * Fix dimensions of "prsn" in the variable dictionary. (:pull:`2242`).
+* History is not written the DataArray is the ``as_dataset`` option is activated. (:issue:`2240`, :pull:`2251`).
+* ``xclim.core.formatting.update_history`` puts the new history at the top of the new attribute, not at the bottom. (:pull:`2251`).
```
```suggestion
* ``xclim.core.formatting.update_history`` now places the updated history at the top of the new attribute, not at the bottom. (:pull:`2251`).
```

I'd also move this into `Breaking changes`. Different behaviour from existing functions.

PR #2258, opened 2025-10-28
[new xclim] Major generics refactor

aulemahal:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #2173
- [x] Tests for the changes have been added (for bug fixes / features)
  - [x] (If applicable) Documentation has been added / updated (for bug fixes / features)
- [x] CHANGELOG.rst has been updated (with summary of main changes)
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added

### What kind of change does this PR introduce?
Major refactor of the `xclim.indices.generics` submodule.

This new "generic" submodule has harmonized vocabulary and signatures. Names and words are taken from "clix-meta" when possible. 

New submodules:
- `reducers.py ` : for resampling functions like "doymin" and "doymax". Logic from `to_agg_units` could go here in a further PR.
- `classify.py` : First step for #2192, the MCA-like function we had is now there.
- `helpers.py` : Not new, but with now an extented reach as to include tools like `compare` and `spell_mask`

This way, `generic` almost only contains "index-like" function, i.e. functions ready to be wrapped in a `ResamplingIndicator`.

This PR also introduced `is_temporal_rate` because I needed it to make a clix index function.

Finally, I reverted the decision from #2150 and now `to_agg_units` _always_ assumes the input is daily if it can't find its frequency.

### Does this PR introduce a breaking change?
Of course. I'll try to provide a map of the renamed functions tomorrow.

I tried not to break the other `xclim.indices` and `xclim.indicators` modules. I have not yet measured to full extent of the breaking.

### Suggestion and next steps:
Consider this PR as the first one of a series in preparation of a new xclim, xclim 1.0 or xclim v2026, depending on how we feel. 

- ~I would like to refactor `generic` even more~
- Use this to remove as much as possible from the other `indices` files, i.e. : make official indicators use the generics. 
- Improve `reducers` to include everything `to_agg_units` does in an extendable way. 
- Semi unrelated:
    - A better templating for attributes so that "generic" indicators can output meaningful stuff.
        - Example: "statistics"  with "tasmin" and "mean" could spit out a `long_name` like : "Mean of minimum daily temperature" using the template : "{op} of {da.long_name}".
    - Think more about this `is_temporal_rate` thing and the automatic switch between `sum` and `integral` that has to be done in some cases (summing fluxes (pr) often actually means summing amounts).

Tests will fail as I haven't synced them with the refactor yet.

Issue #11, opened 2018-08-30
Developing (with Python 2/3 Compatibility in mind)


Zeitsperre:
@tlogan2000 and I were talking about this and we're wondering how best to proceed.

As mentioned previously, Python2 will be end-of-life [soon](https://pythonclock.org/). Having the library compatible with both for the time being is feasible but that shouldn't mean that we shy away from using Python3 advancements (e.g. **F-strings** and **AsyncIO**). 

At the moment, we're aiming to use a fair amount of abstraction libraries (xarray, numpy) that will handle most of the errors that can be introduced by Python2/3 math issues (integer division) but we should consider using [Python2/3 compatible syntax](http://python-future.org/compatible_idioms.html) where possible. So long as we follow good practices and make use of libraries that can convert code base from Py2->2/3 and Py3->2/3, we should be okay.

Suggestion for Contribution guidelines:
- **When adding a new module or set of algorithms, creating a branch of the library, open a Pull Release and ask someone to review changes when adding to Master** 
- if the additions are written in Python2, install and run [`python-futurize`](http://python-future.org/futurize.html) (Not `python-modernize` as we want to make use of more Python3 features) or;
- if the additions are written in Python3, install and run [`python-pasteurize`](http://python-future.org/pasteurize.html)
- Try installing/running the library in conda environments for both Py2 and Py3. (**The Travis CI is set to automatically try building the Pull Release when created, so you can opt to simply push changes to your branch after running the Python2/3 compatibility library**)
- Comment/document any known errors in the code or on GitHub so others can jump where needed.

Zeitsperre:
For Unicode conversions between Py2 and Py3, [this video](https://www.youtube.com/watch?v=sgHbC6udIqc) can help demystify the major differences.

huard:
I've given a bad example so far, but now I'm done experimenting. I'll follow the contribution guidelines from now on.

Zeitsperre:
Just a warning, but `python-futurize` will make things look kind of ugly (`future` and `alias` imports) and gets tripped up on strings (e.g. bytes vs. unicode handling). It's not a silver bullet but it handles the hard parts. Thankfully, moving from `futurized` Python2 code to stand-alone Python3 code largely involves simply drop the convoluted imports. We're making things easier down the line.

huard:
Tox and the Travis build make a good job of catching py2 and 3 gotchas. I've switched to developing in a py3 environment and will check it works with py2.

huard:
[...]

PR #2264, opened 2025-10-31
[new xclim] Use generics to remove redundant indices

### LINKED ISSUE #744: Reduce duplicated indicator source code

aulemahal:
### Pull Request Checklist:
    - This PR fixes #744

This is why I wanted #2258.

Use index functions from `indices.generic` where possible in `xclim.indicators` and remove explicit function in `xclim.indices`.

For example : Indicator "wetdays" is only a specific case of `count_occurrences`. So redefine the indicator to use that function, and remove the `xclim.indices._threshold.wetdays` function.

Yes. A lot of methods from  `xclim.indices` are being removed.

I try not to break indicators, but some arguments may have different names and the metadata is modified. Mainly, a lot of indicators are loosing their "notes" and "examples". My idea is that for simple indicators, those are not necessary : the abstract, the generic function's docstring and the "injected" parameters should be enough to understand the calculations.

### Other information:
A large group of "indices" I can't remove for now are the "percentile" ones like "tg10p". They do not map to a generic function yet as they are in between `count_occurrences` and `count_percentile_occurrences`, plus the `bootstrap` component. I might create a new generic function to take care of that.


aulemahal:
Files to check: (under ` indices/`)

- [x] `_threshold.py` (@aulemahal , almost done)
- [x] `_simple.py` (@aulemahal)
- [x] `_multivariate.py` (@aulemahal)
- [x] `_synoptic.py`
- [x]  `_agro.py` (@Zeitsperre)
- [x] `_hydrology.py` (@coxipi)
- [x] `_anuclim.py` (@aulemahal)

coxipi:
I think I'm done with `hydrology` (I also worked on snow indices in `simple` and `threshold`), most indices there have complicated behaviours. The only ones I could see fitting a generic behaviour are `snw_max_doy` and `snd_max_doy`, but I don't see this `at_least_n_valid` use anywhere in generic. So the question is do we want to include such procedures in there.

coxipi:
I don't know if this is in the plans, but maybe having a way to call the indicator like:

`snw_max._func`, a bit like in xsdba where we can avoid using `map_groups` (I think) for a certain functions, would be nice. Here, it would simply ignore all checks of the indicator and really behave like indices used to (but wihtout units check).  I guess there is already a way to do this, but I think this should be mentionned in migration guides if not already

src/xclim/indices/_threshold.py
coxipi, line 2387: The `Literal` information is not contained in my current implementation of the indicator. Is there a way I should code the indicator to retain this information?

aulemahal: Well I didn't think of that indeed, there's no way to inject it in the docstring for now. But I have an idea.

However, I think you need to inject the `constrain` arg as well in this case. At least the function will perform a check on the operator, what's missing is something in the docstring.

aulemahal (commented):
> but I don't see this at_least_n_valid use anywhere in generic

I believe the new `doymax` covers this issue! There's a test on the std that masks points with no variability. See `indices/_reducers.py::doymax`.

> a way to call the indicator like: `snw_max._func`

Well, there is `indicator.compute`, however, that does not include the "injected" arguments, nor the renamed ones. We could add an helper for that, but what about :

```python
with xc.set_options(check_missing='skip', data_validation='log'):
    out = indicator(...)
```
?

src/xclim/indices/_threshold.py
coxipi: ah got it, now I know what constrain is for!

Zeitsperre:
@aulemahal I managed to go through some of the `_anuclim.py` indices, but it ended up having me go down a rabbit hole fixing typing in several different places.

One thing I did notice is that there are many instances where the new `Condition` type is not entirely supported, despite being listed as such. Instances where `!=`/`neq` and `==`/`eq` are technically allowed in an indice but would make little sense are around. The same issue exists for places where `Reducer` is supported, where I don't imagine `std`, `var`, `doymin`, `doymax` would be supported when examining the docstring.

One thing I did start doing was I added the `DataType` to some places (`xr.DataArray | xr.Dataset`). If a `compute` function accepts and returns that in lieu of the unified types, type checking understands that this means they retain type consistency (i.e. `DataArray` -> `DataArray`).

We might just need more type aliases?

aulemahal:
@Zeitsperre Thanks! I reverted one change you did in `units.py` because I think the `dim_order_diff` dict might not contains dimensions in `CF_CONVERSIONS.{item}.dimensionality` ?


I think this is now ready ?

aulemahal:
To answer your question on aliases.

I think `DataType` is a good idea for `xr.DataArray | xr.Dataset`, but this is seldomly used in xclim, no ?

A stronger link between Type Aliases and `xclim.core.utils.InputKind` would be good. However, I think we need to accept that sometimes Type Aliases will refer to a larger set of values than what is actually possible. For example, `Condition` is ok to use even if only `>` and `>=`  are supported. The _type_ is annotated, the docstring explains the rest.

src/xclim/indicators/atmos/_temperature.py
```diff
@@ -150,11 +150,17 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
     description="{freq} number of days where daily minimum temperature exceeds {thresh}.",
     abstract="The number of days with minimum temperature above a given threshold.",
     cell_methods="time: sum over days",
-    compute=indices.tn_days_above,
-    parameters={"op": {"default": ">"}},
+    compute=indices.generic.count_occurrences,
+    input={"data": "tasmin"},
+    parameters={
+        "condition": {"default": ">"},
+        "thresh": {"default": "20 °C"},
```
coxipi, line 157: Did you mean -10 degC here too? Both TG and TX have the same threshold for the above/below indicator (just with opposite conditions), but TN has different thresholds. (T>20 degC and T<-10 degC).

Anyways, just wanted to flag there is a broken symmetry, just in case it's not wanted

coxipi: NVM, you're just following what was already in indices, it's ok

src/xclim/indices/_simple.py
```diff
@@ -175,6 +180,7 @@ def tn_max(tasmin: xarray.DataArray, freq: str = "YS") -> xarray.DataArray:
     return statistics(tasmin, statistic="max", freq=freq)
 
 
+@deprecated("1.0", "atmos.tn_mean")
 @declare_units(tasmin="[temperature]")
 def tn_mean(tasmin: xarray.DataArray, freq: str = "YS") -> xarray.DataArray:
```
coxipi: Should we use `Freq` here? almost everywhere?

src/xclim/indices/_threshold.py
coxipi, line 524: Comparing the use of `Freq` and `Quantified` (this is a bit confusing because `Freq` is not used specifically here, but I didn't want to try and find an example that had both).

Say we use `Quantified` and `Freq` in the argument:
```python
def myfunc(thresh:Quantified, freq:Freq):
...
```
Then, what should we do in the docstring? 

EDIT: `Freq` is really just a short-hand for a lot of Literal, so putting it on the same level as `Quantified` may not be warranted. Using `str` in the docstring when we use `Freq` and using `Quantified` both in the arguments and docstring seems reasonable.

ORIGINAL: 
I think it would be best if we are more explicit. I like how we can use `Literal["string1","string2", ....]` and then use `{"string1", "string2",...}` in the docstring. For `Freq`, there are too much strings, so I think having `Freq` in the arguments of the function and `str` in the docstring is reasonable, it's explicit enough: if a new user is not sure what `Freq` is, they can find the info in the docstring. 

So, with that in mind, should we also have a more explicit docstring for `Quantified` object? Does that defeat the purpose of having a short name to design multiple possibilities?

I guess the opposite point could be made that we can just hone and embrace our special types. In that case, should we instead use `Freq` in the docstring too?

src/xclim/indicators/atmos/_precip.py
```diff
@@ -117,6 +137,25 @@ def cfcheck(pr: DataArray, tas: DataArray):
         cfchecks.check_valid(tas, "standard_name", "air_temperature")
 
 
+class PrTasx(Precip):
```
coxipi: We already use `Temp` as a generic temperature {g,x,n}, why not use this here too? I also think `Precip` is used in other classes.

`PrTasx -> `PrecipTemp`
`PrTasxWithIndexing` -> `PrecipTempWithIndexing`. 

`Tasx` feels ambiguous to me, it feels less like a joker, i .e. `Tas* -> T{g,n,x} or Tas{,max,min}`, and more like something  too close to `tasmax` or `TX`. 

It doesn't really matter, but since we're breaking things, might as well uniformize our conventions, unless there is a good reason to deviate here.

src/xclim/indicators/atmos/_precip.py
```diff
@@ -227,37 +290,68 @@ class DailyPrecipNoResample(Indicator):
...
-    compute=indices.maximum_consecutive_dry_days,
+    input={"data": "pr"},
+    compute=indices.generic.spell_length_statistics,
+    parameters={
+        "thresh": {"default": "1 mm/day"},
+        "window": 1,
+        "window_statistic": "max",
+        "condition": {"default": "<"},
+        "statistic": "max",
+        "constrain": ("<", "<="),
```
coxipi, line 328: `constrain` effectively changes the type of `condition` from `Condition` to `Literal[constrain]`. This is really to be able to use a generic function, but add more railguards for the indicator? 

I have the impression this kind of logic could be used elsewhere. I can imagine an indicator that has `freq` as an argument. The generic function uses a type `Freq`. But we would like to constrain the allowed `freq` for this specific indicator. 

Is there something very specific about 'constrain'? it's just so common to want to use `>` or `>=`?  I feel this could be treated in the indicator, or with a decorator that accomplishes something similar to what happens down at `get_binary_op` 
```python
    constraints = []
    if isinstance(constrain, list | tuple | set):
        constraints.extend([BINARY_OPS[c] for c in constrain])
        constraints.extend(constrain)
    elif isinstance(constrain, str):
        constraints.extend([BINARY_OPS[constrain], constrain])

    if constrain:
        if condition not in constraints:
            raise ValueError(f"Operation `{condition}` not permitted for this indicator.")
```
we could have: 
```python 
maximum_consecutive_dry_days = Precip(
   ...
    parameters={
        ...
        "condition": {"default": "<"},
        ...
},
      constrain={
      "condition": {[">",">="]}
      }
```
and then the indicator class would use `constrain` to perform the checks on the arguments. 

Anyways, it would just feel less constrained (badum-tsss) moving on in the future if we want to apply this logic to different kind of inputs without having to add a trail of special arguments, e.g.
`spell_length_statistics -> _spell_length_statistics -> spell_mask -> compare -> get_binary_op`

or parameters_constraints, or something

src/xclim/indicators/atmos/_precip.py
```diff
@@ -633,92 +790,154 @@ class DailyPrecipNoResample(Indicator):
...
     "over a given time window of days is equal or above a given threshold.",
     units="",
     cell_methods="",
-    compute=indices.wet_spell_frequency,
+    input={"data": "pr"},
+    compute=indices.generic.spell_length_statistics,
+    parameters={
+        "window": {"default": 3},
+        "window_statistic": {"default": "sum"},
+        "thresh": {"default": "1 mm", "description": "An amount of precipitation (not a flux or rate)."},
```
coxipi: Does this give  a docstring-like information? In `spell_length_statistics`, `thresh` is documented as 

```
Threshold to test against.
```

should we imitate this style, but with more details? e.g. keep the threshold part?
```
A threshold amount of precipitation (not a flux or rate)."
```

src/xclim/indicators/atmos/_wind.py
```diff
@@ -32,7 +31,9 @@ class Wind(ResamplingIndicatorWithIndexing):
     description="{freq} number of days with surface wind speed below {thresh}.",
     abstract="Number of days with surface wind speed below threshold.",
     cell_methods="time: sum over days",
-    compute=indices.calm_days,
+    compute=indices.generic.count_occurrences,
+    input={"data": "sfcWind"},
+    parameters={"condition": "<", "thresh": {"default": "2 ms-1"}, "constrain": None, "freq": {"default": "MS"}},
```
coxipi:
```suggestion
    parameters={"condition": "<", "thresh": {"default": "2 m s-1"}, "constrain": None, "freq": {"default": "MS"}},
```

should meters / seconds, not 1/milliseconds right?

src/xclim/indicators/atmos/_wind.py
```diff
@@ -44,7 +45,9 @@ class Wind(ResamplingIndicatorWithIndexing):
     description="{freq} number of days with surface wind speed at or above {thresh}.",
     abstract="Number of days with surface wind speed at or above threshold.",
     cell_methods="time: sum over days",
-    compute=indices.windy_days,
+    compute=indices.generic.count_occurrences,
+    input={"data": "sfcWind"},
+    parameters={"condition": ">=", "thresh": {"default": "10.8 ms-1"}, "constrain": None, "freq": {"default": "MS"}},
```
coxipi:
```suggestion
    parameters={"condition": ">=", "thresh": {"default": "10.8 m s-1"}, "constrain": None, "freq": {"default": "MS"}},
```

src/xclim/indices/_anuclim.py
```diff
@@ -141,15 +135,15 @@ def temperature_seasonality(tas: xarray.DataArray, freq: str = "YS") -> xarray.D
     --------
     The following would compute for each grid cell of file `tas.day.nc` the annual temperature seasonality:
 
-    >>> import xclim.indices as xci
+    >>> import xclim as xc
```
coxipi: I like this shorthand, we should it in notebooks too eventually

src/xclim/indices/_anuclim.py
coxipi: doesn't seem right warmest/coldest vs. wettest/driest

src/xclim/indices/_anuclim.py
```diff
@@ -430,7 +424,7 @@ def prcptot_warmcold_quarter(
     pr_qrt = _to_quarter(pr=pr)
 
     if op not in ["warmest", "coldest"]:
-        raise NotImplementedError(f'op parameter ({op}) may only be one of "warmest", "coldest"')
+        raise NotImplementedError('op parameter may only be one of "warmest", "coldest"')
```
coxipi:
```suggestion
        raise NotImplementedError('op parameter may only be one of "warmest" or "coldest"')
```
use `or` to be consistent with other messages

src/xclim/indices/converters.py
```diff
@@ -1852,17 +1852,18 @@ def fao_allen98(net_radiation, tas, wind, es, ea, delta_svp, gamma, G="0 MJ m-2
...
-    a2 = gamma * 900 / (tasK) * wind * (es - ea)
-    a3 = delta_svp + (gamma * (1 + 0.34 * wind))
+    _net_radiation: xr.DataArray = convert_units_to(net_radiation, "MJ m-2 day-1")
+    _wind: xr.DataArray = convert_units_to(wind, "m s-1")
+    _tasK: xr.DataArray = convert_units_to(tas, "K")
+    _es: xr.DataArray = convert_units_to(es, "kPa")
+    _ea: xr.DataArray = convert_units_to(ea, "kPa")
+    _delta_svp: xr.DataArray = convert_units_to(delta_svp, "kPa degC-1")
+    _gamma: xr.DataArray = convert_units_to(gamma, "kPa degC")
+    _G: xr.DataArray = convert_units_to(G, "MJ m-2 day-1")
```
coxipi, line 1862: why the change? don't we always modify units in-place?

src/xclim/indices/generic.py
```diff
@@ -438,9 +442,14 @@ def bivariate_count_occurrences(
     -----
     Sampling length is derived from `data1`.
     """
+    if thresh2 is None:
+        thresh1 = thresh1
```
coxipi:
```suggestion
        thresh2 = thresh1
```
not on my watch 🤠

src/xclim/indicators/land/_snow.py
```diff
@@ -92,7 +146,18 @@ class SnowWithIndexing(ResamplingIndicatorWithIndexing):
...
     units="",
-    compute=xci.snd_season_end,
+    title="Snow cover end date (depth).",
+    compute=generic.season,
+    input={"data": "snd"},
+    parameters={
+        "thresh": {"default": "2 cm"},
+        "window": {"default": 14},
+        "condition": ">=",
+        "aspect": "end",
```
coxipi, line 156: earlier you had `first/last` denoted by `which` in `generic.day_threshold_reached`. 

```python
    if which == "first":
        func = rl.first_run_after_date
    elif which == "last":
        func = rl.last_run_before_date
    else:
        raise ValueError(f"'which' must be 'first' or 'last'. Got {which}.")
```

feels similar to aspect here: 
```python
    func = {"start": rl.season_start, "end": rl.season_end, "length": rl.season_length}
    map_kwargs = {"window": window, "mid_date": mid_date}

    if aspect in ["start", "end"]:
        map_kwargs["coord"] = "dayofyear"
    out: xr.DataArray = resample_map(cond, "time", freq, func[aspect], map_kwargs=map_kwargs)
```

should we uniformize the notation?

src/xclim/indices/_multivariate.py
```diff
@@ -484,36 +490,30 @@ def multiday_temperature_swing(
...
-        )
+    freeze = compare(tasmin, condition_tasmin, freeze_threshold) * 1
+    thaw = compare(tasmax, condition_tasmax, thaw_threshold) * 1
+    ft = select_time(freeze * thaw, **indexer)
 
+    out = rl.resample_and_rl(
+        ft,
+        resample_before_rl,
+        rl.rle_statistics,
+        reducer=statistic,
```
coxipi: is there a difference between `reducer` in run_length.py and `statistic` in generic.py? Is `reducer` meant to be a subset of possible `statistic` specifially when we use the `Literal` called `Reducer`? Should we call everything `statistic` and have the `Reducer` type given in run_length.py give the indication of a subset, instead of using a the different argument (`reducer`)?

src/xclim/indices/_multivariate.py
```diff
@@ -1782,6 +1802,7 @@ def warm_spell_duration_index(
     return to_agg_units(out, tasmax, "count", deffreq="D")
 
 
+@deprecated("1.0", "atmos.liquid_precip_ratio with indexer")
```
coxipi:
```suggestion
@deprecated("1.0", "atmos.liquid_precip_ratio with indexer: {'month':[12]}")
```
but also, `liquid_precip_ratio` doesn't have `indexer` argument, so I'm not sure this is right

src/xclim/indicators/atmos/_temperature.py
```diff
@@ -974,8 +1180,16 @@ class TempHourlyWithIndexing(ResamplingIndicatorWithIndexing):
...
     cell_methods="",
-    compute=indices.last_spring_frost,
-    parameters={"before_date": {"default": "07-01"}},
+    input={"data": "tasmin"},
+    compute=indices.generic.day_threshold_reached,
+    parameters={
+        "condition": {"default": "<"},
+        "thresh": {"default": "0°C"},
+        "date": {"name": "before_date", "default": "07-01"},
+        "which": "last",
```
coxipi, line 1189: use `aspect` like in season?

src/xclim/indices/_threshold.py
coxipi, line 3011: should be `condition` here and in the function, for consistency

tests/test_indicators.py
```diff
@@ -514,6 +514,36 @@ def test_all_jsonable(official_indicators):
         raise ValueError(f"Indicators {problems} provide problematic json serialization.: {err}")
 
 
+def test_no_constrain(official_indicators):
```
coxipi, line 517: I don't really understand the use of this. I would guess that the compute function chosen for the indicator will already enforce which argument can be given or not?

coxipi: it's to ensure `constrain` is never used anywhere but with `condition`? this gives a soft way of defining special keywords?

I get what `test_no_out_units` is accomplishing, but not sure about this one.

coxipi (commented):
Everything complete except the tests

src/xclim/indicators/atmos/_precip.py
aulemahal: Pas fou!

If we do something like that, I would simply remove any such checks on the low level functions. 

We could augment the `Parameter` class instead of adding an attribute to the Indicator ?
Like:
```python
maximum_consecutive_dry_days = Precip(
   ...
    parameters={
        ...
        "condition": {"default": "<", "constrain": (">", ">=")},
        ...
})
```

We're getting dangerously close to a use case of `pydantic` here. But this goes in the direction of other comments : refactor the typing logic of xclim by syncing the Annotation / `InputKind` / `Parameter` mess.

src/xclim/indicators/atmos/_precip.py
aulemahal: It does indeed and we should.

src/xclim/indicators/land/_snow.py
aulemahal: Well the concept is different no ?
In the season case, the three "aspects" are always defined, but we request only one of those.

In the `day_threshold_reached`, the argument `which` changes the computation. It's not an "aspect" of the concept.

However, I would agree to uniformize the notation if we found a term that is clear for both cases! I just didn't think of any.

src/xclim/indices/_multivariate.py
aulemahal: We should.

src/xclim/indices/_multivariate.py
aulemahal: The indexer is added by the indicator!

src/xclim/indices/_simple.py
aulemahal: I guess. But I didn't make much effort for deprecated functions.

tests/test_precip.py
```diff
@@ -435,17 +435,17 @@ def test_3d_data_with_nans(self, open_dataset):
         x1 = prMM[:31, 0, 0] * 0.0
         x1[5:10] = 10
         x1.attrs["units"] = "mm/day"
-        cwd1 = atmos.maximum_consecutive_wet_days(x1, freq="MS")
+        cwd1 = atmos.wet_spell_max_length(x1, freq="MS")
```
coxipi, line 438: Window statistic always `sum` seems so weird to me, because the resulting length 9 is greater than the number of days with precip (5). Is a wet spell typically more associated with `sum` than `min`?  

Addtionnally, it feels that current default values are not coherent. 

If we morally say that ~1mm is an individual day that is wet, then we should have `thresh` equal to 3mm no? i.e. the sum of three wet days? 
```python
        "window": {"default": 3},
        "window_statistic": {"default": "sum"},
        "thresh": {"default": "1 mm"--->"3 mm", "description": "An amount of precipitation (not a flux or rate)."},
```

I get the impression that the default thresh given here is meant to be used with `min`, i.e. a threshold that is independent of `window`. 

I am not a fan of using a default `window_statistic = sum`, since the threshold will depend on the window length. I think it would be less informative to put 3mm as a default threshold. I would rather have 1mm to indicate that 1 mm is a good typical value for a single wet day.

I would suggest replacing `window_statistic` default by "mean" or "min" to have independence on the window length in the definition of `thresh`. If `sum` is really the accepted norm, then let's just roll with it

src/xclim/indices/_threshold.py
aulemahal: So I understand you suggest using `"xr.DataArray, str or pint.Quantity"` in the docstrings for the `Quantified` annotation ?

I find it long, but indeed, it is a bit more explicit...

aulemahal: No, because it is deprecated.

src/xclim/indices/converters.py
aulemahal: @Zeitsperre

src/xclim/indices/generic.py
aulemahal: oh! merci!

coxipi (approved):
Good job! phew

There are a few small things that must be changed (e.g. the units for wind "ms-1" -> "m s-1"), but overall it all seems pretty clean.

But most comments are editorial, ideas I had for further improvement, proposition to uniformize the notation, etc.

tests/test_indicators.py
aulemahal: The main error this detects is when an indicator is implemented but the developer forgot to _inject_ the `constrain` value. 

My idea here is that for normal indicators, the user should never see `constrain` in the signature.

tests/test_precip.py
aulemahal: I kinda agree with you. I simply didn't want to break this here now, it's been like this since it was implemented!

src/xclim/indices/_threshold.py
coxipi: I'm on the fence after thinking about it. I just think that it would be a bit harder for a new user that knows xarray, but not xclim, to get into the code. But since there is generally a default value which is a string, I think the users can understand basic usage easily, even if they don't know what Quantified is. We can keep Quantified in docstrings, it's all right

src/xclim/indicators/atmos/_precip.py
coxipi: Yes, removing the option at low level is what I had in mind too. 

I didn't check what would be easier to implement in the indicator class.  Your implementation, putting `constrain` in `parameters` sounds good, yes.

src/xclim/core/calendar.py
```diff
@@ -727,7 +724,7 @@ def _interpolate_doy_calendar(source: xr.DataArray, doy_max: int, doy_min: int =
     return filled_na.interp(dayofyear=range(doy_min, doy_max + 1))
 
 
-def adjust_doy_calendar(source: xr.DataArray, target: xr.DataArray | xr.Dataset) -> xr.DataArray:
+def adjust_doy_calendar(source: xr.DataArray, target: DataType) -> DataType:
```
coxipi: can't `source` also be a `DataType` here? quick check of `adjust_doy_calendar` and I don't see why not

docs/notebooks/extendxclim.ipynb
```diff
@@ -99,12 +99,13 @@
     "from IPython.display import Code, display\n",
```
coxipi, line 99: <pre class="ql-syntax" spellcheck="false">Line #11.    def tx_days_compare(tasmax: xr.DataArray, thresh: Quantified = "0 degC", condition: Condition = "&gt;", freq: Freq = "YS"):
</pre><p> </p><p>If I follow the docstring, this should be? : </p><p> </p>

<pre class="ql-syntax" spellcheck="false">condition: Literal["&gt;", "&lt;"]
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Ouranosinc/xclim/pull/2264/discussion/#comment-3639546651'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='3'/>

coxipi: <p>oups nvm, this is just what you explained in another comment. Seems fine.</p>

coxipi:
this is a static type checker, could be useful at some point
https://github.com/beartype/beartype

src/xclim/indicators/land/_snow.py
coxipi: ok, so `which` is more like choosing the method used, and `aspect` chooses three non-competing aspects of the snow season. 

I thought since this could be rewritten in the same way, it would be mostly equivalent
```python
 func = {"first": rl.first_run_after_date, "last": rl.first_run_after_date}
```
but I get that those are conceptually different

docs/notebooks/extendxclim.ipynb
aulemahal: Trevor raised a similar issue above, my point of view is:

>  I think we need to accept that sometimes Type Aliases will refer to a larger set of values than what is actually possible. For example, Condition is ok to use even if only > and >= are supported. The type is annotated, the docstring explains the rest.

As an exaggerated counter-argument : in the same signature, `tasmax` is `xr.DataArray`, but it would be  wrong to pass an array of datetime values for example. Should we annotate with: `xr.DataArray[float]` ?

src/xclim/indices/converters.py
aulemahal: I think this pattern is suggested by one of the linting tools we use. The idea being not to modify the pointers passed as input in a function.

PR #2267, opened 2025-11-03
Hydrological signature indicators

### LINKED ISSUE #2266: Hydrological signature Indicators
### Generic Issue

* xclim version:
* Python version:
* Operating System:

### What I Did

```
$ pip install foo --bar
```

### What I Received

```
Traceback (most recent call last):
 File "/path/to/file/script.py", line 3326, in run_code
   exec(code_obj, self.user_global_ns, self.user_ns)
 File "<ipython-input-2-9e1622b385b6>", line 1, in <module>
   1/0
 ZeroDivisionError: division by zero


### Code of Conduct

### PR DESCRIPTION
**Pull request by e-larose**
### Pull Request Checklist:
    - This PR fixes #2266

*  Many changes relative  #2227 : Fixes and better compliance with our conventions
*  Add tests for the changes in #2227
* Add indicators and french translations for the indices introduced in #2227

Yes. All related to functions added in last hydro PR #2227, see CHANGELOG.

### CODE REVIEW

File: src/xclim/indicators/land/_streamflow.py [hunk 1]
```diff
@@ -169,3 +176,58 @@ def cfcheck(q: DataArray):
...
+
+lag_snowpack_flow_peaks = Streamflow(
+    title="Time lag between maximum snowpack and river high flows",
+    identifier="lag_snowpack_flow_peaks",
+    units="days",
+    long_name="Time lag between maximum snowpack and river high flows",
+    description="Number of days between the annual maximum snowpack, measured by the snow water"
+    "equivalent, and the mean date when river flow exceeds a percentile threshold"
+    "during a given year. If the time lag between maximum snowpack and river high flows is ≤ 50 days,"
+    "the watershed is likely in a nival regime.",
```
huard:
```suggestion
    "during a given year.",
```

src/xclim/indicators/land/_streamflow.py
```diff
@@ -169,3 +176,58 @@ def cfcheck(q: DataArray):
...
+
+runoff_ratio = Streamflow(
+    title="Runoff ratio",
+    identifier="runoff_ratio",
+    units="",
+    long_name="Runoff ratio",
+    description="Ratio of runoff volume measured at the stream to the total precipitation volume over the watershed."
+    "Temporal analysis: Yearly values computed from seasonal daily data and yearly data, depending on chosen frequency."
+    "(e.g., 'YS' for yearly starting Jan, or 'QS-DEC' for seasons,"
+    "'30YS' to compute the value over slices of 30 years from the start of the time series).",
```
huard:
```suggestion
    "Temporal analysis: Yearly values computed from seasonal daily data and yearly data, depending on chosen frequency.",
```

src/xclim/indicators/land/_streamflow.py
```diff
@@ -169,3 +176,58 @@ def cfcheck(q: DataArray):
...
+    keywords="streamflow",
+    compute=runoff_ratio,
+)
+
+sen_slope = Streamflow(
+    title="Sen Slope : Temporal robustness analysis of streamflow.",
+    identifier="sen_slope",
+    units="",
+    long_name="Sen Slope",
+    description=" Computes annual and seasonal Theil–Sen slope estimators and performs the"
```
huard:
```suggestion
    description="Compute annual and seasonal Theil–Sen slope estimators and perform the"
```

huard (changes requested):
Looks good. Please add simple tests for indicators, and the french translation (look in  data/fr.json)

src/xclim/indicators/land/_streamflow.py
```diff
@@ -169,3 +176,55 @@ def cfcheck(q: DataArray):
...
+    compute=lag_snowpack_flow_peaks,
+)
+
+runoff_ratio = Streamflow(
+    title="Runoff ratio",
+    identifier="runoff_ratio",
+    units="",
+    long_name="Runoff ratio",
+    description="Ratio of runoff volume measured at the stream to the total precipitation volume over the watershed."
+    "Temporal analysis: Yearly values computed from seasonal daily data and yearly data, depending on chosen frequency.",
```
Zeitsperre:
```suggestion
    description="Ratio of runoff volume measured at the stream to the total precipitation volume "
    "over the watershed. Temporal analysis: Yearly values computed from seasonal daily data "
    "and yearly data, depending on chosen frequency.",
```

docs/paper/tmp_saved_input_paper.bib
aulemahal, line 1: To remove before merging!

docs/references.bib
```diff
@@ -4216,3 +4216,198 @@ @incollection{ecwmf_physical_2016
...
+    dominated snow cover duration were located systematically at lower elevations, due to
+    the high interannual variability in the occurrence of annual peak SWE (in winter or
+    early spring), yielding highly variable melt rates. However, in colder sites the melt
+    season occurs mostly in spring and it is very fast in most of the years. The results
+    highlight the control that the seasonal precipitation patterns, in combination with
+    temperature, exert on the seasonal snow cover duration by influencing the peak SWE and
+    suggest a future increased importance of melt rates as temperatures increase. Despite
+    the high climatological variability of the Iberian mountain ranges, the results showed
+    a consistent behaviour along the different mountain ranges, indicating that the methods
+    and results may be transferrable to other temperate mountain areas of the world.}
```
aulemahal: Do we need the full abstract here ? For a smaller file and for coherence with the other entires, I think we could remove the abstracts here.

src/xclim/indicators/land/_snow.py
```diff
@@ -280,3 +281,12 @@ class SnowWithIndexing(ResamplingIndicatorWithIndexing):
...
 )
+
+days_with_snowpack = Snow(
+    title="Days with snowpack",
+    identifier="days_with_snowpack",
+    units="days",
+    long_name="Days with snowpack",
+    description="The total number of days with snow water equivalent (SWE) above a given threshold.",
+    compute=xci.days_with_snowpack,
+)
```
aulemahal: This already exists as `snw_days_above`, no ?

src/xclim/indices/_hydrology.py
```diff
@@ -685,3 +700,434 @@ def antecedent_precipitation_index(pr: xarray.DataArray, window: int = 7, p_exp:
...
+    pr_freq = pr.resample(time=freq).sum()
+
+    out = runoff_freq / pr_freq
+    out.attrs["units"] = ""
+    return out
+
+
+@declare_units(swe="[length]", thresh="[length]")
+def days_with_snowpack(
+    swe: xarray.DataArray,
```
aulemahal: Ah I see the nuance now. I would argue that  `swe` is exactly the same as `snw` and the difference in variable doesn't justify a new indicator. Also, `snw` is the CMIP name, which is why we chose it in xclim.

src/xclim/indices/_hydrology.py
```diff
@@ -685,3 +700,434 @@ def antecedent_precipitation_index(pr: xarray.DataArray, window: int = 7, p_exp:
...
+    pet = pet.resample(time=freq).sum()
+    ai = pr / pet
+    ai.attrs["units"] = ""
+
+    return ai
+
+
+@declare_units(swe="[length]", q="[discharge]")
+def lag_snowpack_flow_peaks(
+    swe: xarray.DataArray,
```
aulemahal: Same. Please use `snw`.

tests/conftest.py
```diff
@@ -406,6 +413,33 @@ def remove_data_written_flag():
...
+def swe_series():
+    def _swe_series(values, start="1/1/2000", units="mm"):
+        coords = pd.date_range(start, periods=len(values), freq="D")
+        return xr.DataArray(
+            values,
+            coords=[coords],
+            dims="time",
+            name="swe",
+            attrs={
+                "standard_name": "snow_water_equivalent_in_snow_layer",
```
aulemahal: This standard name doesn't exist : https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

aulemahal (commented):
Sorry if I'm late and opiniated...

tests/conftest.py
Zeitsperre: In the case of a missing standard name, we can simply omit adding it.

.gitignore
```diff
@@ -113,8 +113,8 @@ docs/_dynamic/indicators.json
 docs/variables.json
 
 # autogenerated BibTeX files
-docs/tmp_saved_input_references.bib
 docs/paper/tmp_saved_input_paper.bib
+docs/tmp_saved_input_references.bib
```
coxipi, line 117:
```suggestion
docs/tmp_saved_input_references.bib
docs/paper/tmp_saved_input_paper.bib
```

revert unneeded change

src/xclim/indices/_hydrology.py
coxipi: This makes me think of a PR I co-authored with Ludwig, I could dig this up.

There is the special field [precipitation], which can do automatic conversions. Having this with snw -> swe would be dangerous. What you say is that user should do all their work in snw, then convert to swe at the end, correct? `snw_to_snd` functions already exist, it would be easy to add a new converter with the correct density to have `snw_to_swe`

aulemahal: In the context of xclim, what I think is that we should try to limit the number of non-conventional variables to a minimum. `swe` is so close to `snw` that I don't think it warrants a special treatment.

Also, `days_with_snowpack` and `snw_days_above` are two very different names for the same thing. The `{var}_days_above` construct being the choice made by xclim for clear and systematic naming, unless `days_with_snowpack` is a jargon term refering to a (domain-)specific algorithm.

Finally, if we really need to keep this **indicator**, I would suggest removing the compute function nonetheless and simply creating a `day_with_snowpack` in `indicators/land/_snow.py` instead.

But outside of the context of xclim, in my **very personal** opinion, I find talking about snow in terms of liquid water a bit weird. You don't need to melt it to study it on a computer.

Zeitsperre:
@coxipi I performed a merge of `main`, but I found that there were quite a few conflicts from the previous hydro indice PR (as though this branch was based off an older version before it was merged to `main`).

I may have totally botched the merge, so apologies.

CHANGELOG.rst
```diff
@@ -33,7 +34,7 @@ Bug fixes
 
 v0.60.0 (2026-01-23)
 --------------------
-Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), Juliette Lavoie (:user:`juliettelavoie`), Ève Larose (:user:`e-larose`), Faisal Mahmood (:user:`faimahsho`), David Huard (:user:`huard`), Pascal Bourgault (:user:`aulemahal`).
+Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), Juliette Lavoie (:user:`juliettelavoie`),  Faisal Mahmood (:user:`faimahsho`), David Huard (:user:`huard`), Pascal Bourgault (:user:`aulemahal`).
```
Zeitsperre, line 37: Ève's other PR made it into the last version ;)

```suggestion
Contributors to this version: Éric Dupuis (:user:`coxipi`), Trevor James Smith (:user:`Zeitsperre`), Juliette Lavoie (:user:`juliettelavoie`), Ève Larose (:user:`e-larose`), Faisal Mahmood (:user:`faimahsho`), David Huard (:user:`huard`), Pascal Bourgault (:user:`aulemahal`).
```

tests/test_land.py
Zeitsperre, line 1: Don't forget to remove the `print` calls!

tests/test_snow.py
```diff
@@ -2,6 +2,7 @@
 
 import numpy as np
 import pytest
+import xarray as xr
```
Zeitsperre, line 5:
```suggestion
from xarray import DataArray
```

tests/test_snow.py
```diff
@@ -163,3 +164,20 @@ def test_perfect_xmas_days_simple(self, open_dataset):
...
+    a[50:65] = 20
+    # Year 2: 5 days of SWE = 5 mm
+    a[400:405] = 5
+
+    # Create a daily time index
+    swe = swe_series(a)
+    out = land.days_with_snowpack(swe, freq="YS")
+
+    assert out.attrs["units"] == "days"
+    assert isinstance(out, xr.DataArray)
```
Zeitsperre, line 183:
```suggestion
    assert isinstance(out, DataArray)
```

src/xclim/indices/_hydrology.py
coxipi: I forgot about the distinction and already removed it.

coxipi: remove, we will not support SWE in general (although we have converters)

src/xclim/indices/stats.py
```diff
@@ -292,8 +292,8 @@ def func(x):
     return out
 
 
-# FIXME: xclim-v1 — `parametric_cdf` should be like `parametric_pdf`
-# * The new coordinate should be labeled as `v`
+# FIXME: xclim-v1 — `parametric_cdf` should be like `parametric_pdf`, i.e.:
+# The new coordinate should be labeled as `v` and not `cdf`
```
coxipi, line 296: unrelated, just wanted my comment from other PR to be clearer

src/xclim/indices/_hydrology.py
```diff
@@ -927,132 +919,84 @@ def lag_snowpack_flow_peaks(
...
+        qresY = unstack_dates(q.resample(time="YS").mean(), winter_starts_year=True)
+        qres = xarray.concat([qresS, qresY], dim="season")
+
+        out = xarray.apply_ufunc(
+            lambda q: (lambda mk_output: [mk_output.slope, mk_output.p])(mk.original_test(q)),
+            qres,
+            input_core_dims=[["time"]],
+            output_core_dims=[["var"]],
+            vectorize=True,
+            dask="parallelized",
```
coxipi: @aulemahal  I can work around without adding `unstack_dates` in `xclim`, but I think this is much cleaner now. Specifically, we should not convert things to `pd.DataFrame` like it's done above, unless there was something that only `pd` can do, but `xr` does the job welll. Basically, we just needed to use `xr.apply_ufunc` to interface with `mk`, `pd` is not needed here.

src/xclim/indices/_hydrology.py
```diff
@@ -1094,29 +1032,12 @@ def base_flow_index_seasonal_ratio(
...
-    winter_1_aligned = winter_1.reindex(year=summer_bfi.year)
     epsilon = 1e-3  # To avoid division by zero
-
-    w_s_ratio = (winter_1_aligned / (summer_bfi + epsilon)).sel(year=summer_bfi.year)
-    w_s_ratio = w_s_ratio.isel(year=slice(1, None))
+    w_s_ratio = bfi.sel(season="DJF") / (bfi.sel(season="JJA") + epsilon)
     w_s_ratio.attrs["units"] = ""
-
-    return winter_bfi, spring_bfi, summer_bfi, fall_bfi, w_s_ratio
+    return bfi, w_s_ratio
```
coxipi: I'm not sure what we prefer. A variable per season, or a coordinate for the different seasons

tests/conftest.py
```diff
@@ -416,7 +417,7 @@ def atmosds(nimbus) -> xr.Dataset:
 
 
 @pytest.fixture(scope="session")
-def ensemble_dataset_objects() -> dict[str, str]:
+def ensemble_dataset_objects() -> dict[str, list[str]]:
```
coxipi: not on purpose, don't know if it's ok / why it's here

src/xclim/indicators/land/_streamflow.py
```diff
@@ -169,3 +178,96 @@ def cfcheck(q: DataArray):
...
+        "as well as yearly winter to summer bfi ratio.",
+        "Yearly winter to summer baseflow index ratio, defined as the minimum 7-day average flow divided by the "
+        "mean flow as well as yearly winter to summer bfi ratio.",
+    ],
+    abstract="Yearly base flow index per season, defined as the minimum 7-day average flow divided by the mean flow"
+    "as well as yearly winter to summer bfi ratio.",
+    cell_methods="",
+    keywords="streamflow, seasonal",
+    compute=base_flow_index_seasonal_ratio,
+)
```
coxipi: maybe needs to be changed, depending if we use coords or variables for season indicators

src/xclim/indices/_hydrology.py
```diff
@@ -752,16 +745,19 @@ def runoff_ratio(
...
     :cite:cts:'knoben_2024'
     """
-    q = convert_units_to(q, "m3/s")
-    a = convert_units_to(a, "km2")
+    q = convert_units_to(q, "mm3/hr")
+    a = convert_units_to(a, "mm2")
     pr = convert_units_to(pr, "mm/hr")
 
-    runoff = q * 3.6 / a  # unit conversion for runoff in mm/hr : 3.6[s/hr *km2/m2]
+    runoff = q / a
```
coxipi: we were already using conversions. We should just convert to the right units right from the get-go and avoid the 3.6 factor

src/xclim/indices/_hydrology.py
```diff
@@ -818,14 +815,6 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
     The ratio of total precipitation over potential evapotranspiration.
     Classification based on the Aridity Index (AI).
-    +----------------+----------------+-----------------+
-    | Classification | Aridity Index  | Global land area|
-    +----------------+----------------+-----------------+
-    | Hyperarid      | AI < 0.05      | 7.5%            |
-    | Arid           | 0.05 ≤ AI < 0.20 | 12.1%         |
-    | Semi-arid      | 0.20 ≤ AI < 0.50 | 17.7%         |
-    | Dry subhumid   | 0.50 ≤ AI < 0.65 | 9.9%          |
-    +----------------+----------------+-----------------+
```
coxipi, line 828: this won't render properly. I think the docstring is sufficient without this, and there is a link for more information. Anyways, there is not no threshold involved here, so I don't think we need to invoke categories based on the range of AI

src/xclim/indices/_hydrology.py
```diff
@@ -860,32 +850,33 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
-@declare_units(swe="[length]", q="[discharge]")
+# FIXME: Different name??
+@declare_units(snw="[mass]/[area]", q="[discharge]")
 def lag_snowpack_flow_peaks(
-    swe: xarray.DataArray,
+    snw: xarray.DataArray,
     q: xarray.DataArray,
     freq: str = "YS-OCT",
-    percentile: int = 90,
+    p: float = 0.9,
```
coxipi, line 830: using `p` and a number between [0,1] instead of [0,100]. This respects conventions elsewhere in xclim for streamflow indices

src/xclim/indices/_hydrology.py
```diff
@@ -860,32 +850,33 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
     during a given year.
     If the time lag between maximum snowpack and river high flows is ≤ 50 days,
     the watershed is likely in a nival regime.
 
     Parameters
     ----------
-    swe : xarray.DataArray
-        Surface snow amount as snow water equivalent.
+    snw : xarray.DataArray
+        Surface snow amount.
```
coxipi, line 844: we would rather use `snw` everywhere, we have converters so users can convert their  `swe`

src/xclim/indices/_hydrology.py
```diff
@@ -909,150 +900,89 @@ def lag_snowpack_flow_peaks(
...
-    thresh = q.resample(time="YS-OCT").reduce(
-        np.nanpercentile, q=percentile, dim="time"
-    )  # the second q, equal to percentile, is a keyword in np.nanpercentile, not the flow variable.
-    threshold_for_each_time = thresh.reindex_like(q, method="ffill")
-    q_high = q.where(q >= threshold_for_each_time).dropna(dim="time", how="all")
-
-    # Day of year for high flow peaks
-    doy = q_high.time.dt.dayofyear
-
-    t_q_max = doy.resample(time=freq).reduce(partial(circmean, high=366, low=1), dim="time")
```
coxipi, line 927: as noted by @aulemahal , the computation of `doy` doesn't work as intended. By taking `.time.dt.dayofyear`, we simply have a 1D array. But what was wanted here is a full dimensional array, like the `snw_max_doy` above

src/xclim/indices/_hydrology.py
```diff
@@ -909,150 +900,89 @@ def lag_snowpack_flow_peaks(
...
+    doy_snw_max = snw_max_doy(snw, freq)
 
-    # Compute lag
-    lag = t_q_max - doy_swe_max
+    thresh = q.resample(time=freq).quantile(q=p, dim="time")
+    threshold_for_each_time = thresh.reindex_like(q, method="ffill")
+    doys_over_thresh = q.time.dt.dayofyear.where(q >= threshold_for_each_time)
```
coxipi: This I believe fixes the issue. We broadcast the dayofyear the full dimensions, and we fill with nans where the streamflow is not over the threshold, so it won't be computed in the mean

src/xclim/indices/_hydrology.py
```diff
@@ -909,150 +900,89 @@ def lag_snowpack_flow_peaks(
...
+    t_q_max = days_since_to_doy(
+        doy_to_days_since(doys_over_thresh, "10-01").resample(time=freq).mean(dim="time"), "10-01"
+    )
```
coxipi: this is quite different from the previous implementation, but it should be more dask/flox compatible

src/xclim/indices/_hydrology.py
```diff
@@ -909,150 +900,89 @@ def lag_snowpack_flow_peaks(
...
     References
     ----------
     :cite:cts:`sauquet_2025`
     """
-    seasons = ["DJF", "MAM", "JJA", "SON", "Year"]
-
-    def compute_seasonal_stats(x: xarray.DataArray) -> tuple[list, list]:
-        """
-        Seasonal statistics.
-
```
coxipi, line 979: all this function will not play nice at all with dask, see my alternative

src/xclim/indices/_hydrology.py
```diff
@@ -1094,29 +1018,13 @@ def base_flow_index_seasonal_ratio(
...
-    winter_1_aligned = winter_1.reindex(year=summer_bfi.year)
     epsilon = 1e-3  # To avoid division by zero
-
-    w_s_ratio = (winter_1_aligned / (summer_bfi + epsilon)).sel(year=summer_bfi.year)
-    w_s_ratio = w_s_ratio.isel(year=slice(1, None))
+    w_s_ratio = bfi.sel(season="DJF") / (bfi.sel(season="JJA") + epsilon)
     w_s_ratio.attrs["units"] = ""
-
-    return winter_bfi, spring_bfi, summer_bfi, fall_bfi, w_s_ratio
+    return bfi, w_s_ratio
```
coxipi: Should we return indices with season coordinates, or indicators for each season?

src/xclim/indices/_hydrology.py
```diff
@@ -909,150 +900,89 @@ def lag_snowpack_flow_peaks(
...
 
     Parameters
     ----------
     q : xarray.DataArray
         Observed streamflow vector.
-    qsim : xarray.DataArray
+    qsim : xarray.DataArray, optional
         Simulated streamflow vector.
+    freq : str
+        Resampling frequency.
```
coxipi, line 906: before, 'QS-DEC' and 'YS-DEC' were computed in the same call. We think it would be best to separate this computation, and give `freq` as an argument

tests/test_hydrology.py
```diff
@@ -260,36 +260,36 @@ def test_simple(self, pr_hr_series, evspsblpot_hr_series):
...
-        # Year 2: 35 days of high flows 10 days after max swe
+        # Year 2: 35 days of high flows 10 days after max snw
         b[310:345] = 5
 
         # Create a daily time index
         q = q_series(b)
 
-        out = xci.lag_snowpack_flow_peaks(swe, q)
+        out = xci.lag_snowpack_flow_peaks(snw, q)
         np.testing.assert_allclose(out, [17.0, 27.0], atol=1e-14)
```
coxipi, line 259: all the tests still need actual testing on the values

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1715,255 @@ def _reconstruct_time(_time_as_delta, _start):
         periods.append(da.isel(**{dim: i}, drop=True).isel(time=slc).assign_coords(time=real_time.isel(time=slc)))
 
     return xr.concat(periods, "time")
+
```
coxipi, line 1730: All changes in calendar are functions stolen form xscen, TBD how we want to manage this, but I put them here for now

tests/conftest.py
coxipi:
```suggestion
def ensemble_dataset_objects() -> dict[str, str]:
```

src/xclim/indices/_hydrology.py
```diff
@@ -1092,31 +1033,16 @@ def base_flow_index_seasonal_ratio(
     :cite:cts:`singh_2019`
     :cite:cts:`jaffres_2021`
     """
-    # 7-day minimum of raw daily flow
-    q7min = q.rolling(time=7, center=True).min(skipna=False).resample(time=freq).min()
```
coxipi, line 1096: Note that this was an error in the previous version. 

It was written as a double `min`, but the rolling operation should be `mean`. Using `base_flow_index` corrects this.

Zeitsperre:
@coxipi The CodeQL fix is already in #2346

tests/test_land.py
```diff
@@ -97,3 +97,98 @@ def test_low_flow_frequency(q_series):
...
+    a = area_series(a)
+
+    q = q_series(q, start="2000-07-01")
+    pr = pr_series(pr, units="mm/hr")
+
+    # FIXME: this gives nans, but the index is OK.
+    out = land.runoff_ratio(q, a, pr, freq="YS")
+    assert out.attrs["units"] == "1"
+    assert isinstance(out, xr.DataArray)
+    np.testing.assert_allclose(out.values, 0.5, rtol=1e-6)
```
coxipi: This is weird since `base_flow_index` is similarly a `Streamflow` indicator but doesn't have this problem. Maybe the problem is using `a` which doesn't have time dimension, looking into it

coxipi: `a` was not the culprit.

@aulemahal any idea why this fails? `land.runoff_ratio(q, a, pr, freq="YS")` gives me nans, but not the index `xci.runoff_ratio`. If  I change the class of `land.runoff_ratio` to a NonResampling indicator, I get rid of the nans too

coxipi: Also, should `a` be a Quantified rather than DataArray?

aulemahal: I'll look into this.
But I'll note that a `DataArray` _is_ a `Quantified`, provided it has a valid `units` attribute.

src/xclim/indicators/land/_streamflow.py
```diff
@@ -2,57 +2,67 @@
...
-            The input data array.
-        """
-        check_valid(q, "standard_name", "water_volume_transport_in_river_channel")
+class StreamflowSnw(ResamplingIndicator):
+    """Streamflow class."""
+
+    context = "hydro"
+    src_freq = "D"
+    # FIXME: better keywords
+    keywords = "streamflow snow hydrology"
```
coxipi: not sure this is necessary anymore. When there was a check valid with explicit variables, it seemed more necessary

tests/test_land.py
aulemahal: haha, @coxipi  l'année 2000 comportait 366 jours.

coxipi: :shaking_face:  :facepalm:

aulemahal: image

src/xclim/indicators/land/_streamflow.py
aulemahal: Indeed, it's not different enough from `StreamFlow` declared just above. The only indicator using this class is already adding `snw` as a keyword, it could add `snow` as well.

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +866,100 @@ def lag_snowpack_flow_peaks(
...
-        Dataset containing the following variables:
-
-        - ``Sen_slope`` : Sen's slope estimates for seasonal and yearly averages.
-        - ``p_value`` : Mann–Kendall metric indicating slope tendency.
-        - If simulated flows are provided: ``Sen_slope_sim``, ``p_value_sim``,
-          and the ratio of observed ``Sen_slope`` over simulated ``Sen_slope``.
+        Dataset variables:
+        ``sen_slope`` : Sen's slope estimates for seasonal and yearly averages.
+        ``p_value`` : Mann-Kendall metric indicating slope tendency.
+        ``sen_slope_sim``, ``p_value_sim``, ``sen_slope_ratio`` (optional): Sen's slope estimates and Mann-Kendall
```
coxipi: not a fan of optional outputs, thoughts? @aulemahal

src/xclim/indicators/land/_snow.py
```diff
@@ -258,6 +258,9 @@ class SnowWithIndexing(ResamplingIndicatorWithIndexing):
     compute=xci.snw_days_above,
 )
 
+# FIXME: If we insist on having something like `days_with_snowpack`, we should have an indicator:
+# swe_days_above = (.... compute=xci.snd_days_above)
+
```
coxipi: removed `days_with_snowpack` as it was redundant. We could add a `swe_days_above`, or just transform `snw_days_above` in a more permissive indicator (with the new `snowamount` unit), or just tell users to use `snd_days_above` which is already compatible.

src/xclim/indices/_hydrology.py
```diff
@@ -852,6 +800,7 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
     ----------
     :cite:cts:'zomer_2022'
     """
+    pet = convert_units_to(pet, pr)
```
coxipi: listed the previous absence of the conversion as a bug

src/xclim/indices/_hydrology.py
```diff
@@ -860,32 +809,40 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
     return ai
 
 
-@declare_units(swe="[length]", q="[discharge]")
+def _timemax(da):
+    tmax = da.idxmax("time")
+    std = da.std("time")
+    tmax = tmax.where(std != 0)
+    return tmax
```
coxipi, line 821: this helper works like `doymax` but doesn't convert to `doys`.  I can add it in generic if it seems worthy

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +866,100 @@ def lag_snowpack_flow_peaks(
...
-
-    return ds
+    for var in out.data_vars:
+        out[var].attrs["units"] = ""
+    # FIXME: this is temporary. I'm not sure we should allow a different number of output depending on the presence
+    # or not of `qsim`
+    if qsim is not None:
+        return out.sen_slope, out.p_value, out.sen_slope_sim, out.p_value_sim, out.ratio
+    else:
+        return out.sen_slope, out.p_value
```
coxipi: not a fan of optional outputs, thoughts? @aulemahal

Zeitsperre: I don't think anyone on the dev team likes optional outputs. I'd rather it return a series of `None` if it absolutely must.

src/xclim/indices/_hydrology.py
```diff
@@ -707,37 +706,34 @@ def antecedent_precipitation_index(pr: xarray.DataArray, window: int = 7, p_exp:
     return out
 
 
+# TODO: change a to `area`?
```
huard, line 709: The only CF variable name that roughly matches is `cell_area`, even though here we don't have a cell in the model grid sense. 
I think that would be my proposal, even though it's not a perfect match.

src/xclim/indices/_hydrology.py
huard: It however did help me understand what I was getting out of that indicator. Can we keep it as a comment in the function until we figure out how to render it properly ?

src/xclim/indices/_hydrology.py
```diff
@@ -852,40 +802,49 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
+    std = da.std("time")
+    tmax = tmax.where(std != 0)
+    return tmax
+
+
+# TODO: Change the indicator name? That seems a bit off
+@declare_units(snw="[snowamount]", q="[discharge]")
 def lag_snowpack_flow_peaks(
-    swe: xarray.DataArray,
+    snw: xarray.DataArray,
```
huard, line 827: Why the switch ?

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +868,130 @@ def lag_snowpack_flow_peaks(
...
+    dt_snw_max = resample_map(snw, "time", freq, _timemax) - t0
+    # Convert to float, and ensure a NaT gets converted to a nan
+    dt_snw_max = xarray.where(dt_snw_max.isnull(), np.nan, dt_snw_max.astype("timedelta64[s]").astype(float))
 
-    t_q_max = doy.resample(time=freq).reduce(partial(circmean, high=366, low=1), dim="time")
+    dt_q = (q.time - t0).astype("timedelta64[s]").astype(float)
+    thresh = q.resample(time=freq).quantile(q=p, dim="time")
+    threshold_for_each_time = thresh.reindex_like(q, method="ffill")
+    dt_q_over_thresh = dt_q.where(q >= threshold_for_each_time)
+    dt_high_q = dt_q_over_thresh.resample(time=freq).mean(dim="time")
```
huard: Much cleaner this way, thanks !

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +868,130 @@ def lag_snowpack_flow_peaks(
...
-    for var in ds.data_vars:
-        ds[var].attrs["units"] = ""
+    sen_slope, p_value = _mann_kendall(q, freq)
+    sen_slope.attrs["units"] = ""
+    p_value.attrs["units"] = ""
+    return sen_slope, p_value
 
-    return ds
+
+# FIXME: xclim-v1 — Remove this function. Its only utility is to compute a ratio.
```
huard, line 947: ok

src/xclim/indices/_hydrology.py
huard: indeed, should have caught that.

src/xclim/indices/_hydrology.py
huard: I think your version is more xclimesque.

huard (dismissed):

src/xclim/indices/_hydrology.py
coxipi: In the end I just create two indicators to avoid this

src/xclim/indices/_hydrology.py
coxipi: Yes, I had the same conclusion. The way I was handling it instead is to pass this variable as a `Quantified` rather than a DataArray. Because of this, the name of the variable doesn't trigger errors in the indicator class, it doesn't need to be registered as a variable (it also allows passing a uniform area as a string — e.g. "500 m2" — although I'm not sure it would be necessarily useful). So we don't need to necessarily give a cf-name if there's not really a match. How about just `area` ?

src/xclim/indices/_hydrology.py
coxipi: `snw` and `swe` express the same quantity, but just in different units with an additional conversion (divide snow amount (`snw`) by water density to obtain snow water equivalent `swe`).  I added a new type of unit "[snowamount]", that makes it possible to use a `snw` or `swe` in the function, i.e. there is an automatic conversion like there is for precipitations (between e.g. kg m-2 s-1 and mm d-1). 

There was also a function equivalent to `snw_days_above`, called `days_with_snowpack` (or something like this), where the only difference is that it takes `swe` instead of `snw`. We should probably avoid duplicating functions where the only difference is having `swe` or `snw`, so I removed it. In fact, `snd_days_above` can already be used without any changes to compute the quantity of interest with a `swe` field. We could always add an indicator (i.e. not indices) `swe_days_above` if we want specifically the name `swe` appear in the indicator.

More generally, this is an attempt to standardize the inputs we require in the indicators, and at this point is more a matter of documentation. Again, you can use a field with units of `snw` or `swe` in the function. But the docstring, the name of the variable is `snw`, to have uniformity with the rest of the library. 

I suppose we could decide that we really want to have `swe` in the hydrology functions. But at this point, I will invoke @aulemahal on this particular point, which prefers using `snw`.

src/xclim/indices/_hydrology.py
coxipi: My comment was maybe too affirmative, I want to share my thoughts on this, maybe I'm wrong. 

With the modifications, `sen_slope` now only takes one variable `q`. Then, to compare the ratios in the future, you'd need to do:
```python
senslope_obs, pvalue_obs = sen_slope(qobs)
senslope_sim, pvalue_sim = sen_slope(qsim)
xsdba.properties.ratio(senslope_sim, senslope_obs)
```
So the current `sen_slope_ratio` allows to compute this with one call. But I wonder, how many indicators could be modified to compare two datasets? It just seems conceptually to be different to what `xclim` generally aims to do, so that is why I'm not sure `sen_slope_ratio` should be kept in

src/xclim/indices/_hydrology.py
coxipi: OK!

coxipi:
```rst
    - The range in the aridity index over a spatial domain define different environment categories (the percentage of global land area 
      covered by each category is shown in parenthesis)
        - Hyperarid (7.5%): AI < 0.05 
        - Arid (12.1%):  0.05 ≤ AI < 0.20
        - Semi-Arid (17.7%): 0.20 ≤ AI < 0.50
        - Dry subhumid (9.9%): 0.50 ≤ AI < 0.65
        - Humid (52.8%): AI ≥ 0.65
```

I should probably specify the year in which this classification is established (mainly because the land coverage is specified)

src/xclim/indices/_hydrology.py
huard: That works too !

src/xclim/indices/_hydrology.py
huard: Ok, good for me thanks !

src/xclim/indices/_hydrology.py
huard: Agree to remove stuff that can be replaced with one-liners.

src/xclim/indices/_hydrology.py
aulemahal: With @coxipi's new`snowamount`, you can do, `lag_snow_flow_peaks(snw='swe', ds=ds)` if you really don't want to rename your dataset. But indeed this particular point is about uniformisation and avoiding duplication of indicators, `snw` and `swe` express the same quantity and the former is backed up by a standard (CMIP).

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+
+def add_season_coord(ds: xr.Dataset | xr.DataArray, freq: str) -> xr.DataArray | xr.Dataset:
```
coxipi, line 1732: @aulemahal  it seems to do the job. Now `base_flow_index_seasonal_ratio` should be given any resampling frequency between "MS" and "YS" that divide years evenly, and will separate things in season / time.

But just realized, some edge cases broken. For montly frequencies, things work well when the first time is aligned with the frequency (i.e. starts on january)
```python
q = q_series([1]*365*10, start='2001-01-01')
xc.indices.base_flow_index_seasonal_ratio(q, '2MS', winter="JF",summer="MA")
```

but not this
```python
q = q_series([1]*365*10, start='2001-12-01')
xc.indices.base_flow_index_seasonal_ratio(q, '2MS', winter="JF",summer="MA")
```

src/xclim/indices/_hydrology.py
```diff
@@ -843,49 +793,62 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
 
     Notes
     -----
-    - An aridity index below 0.65 indicates an arid environment,
-      while values above this threshold correspond to more humid environments.
+    - The range in the aridity index define different environment categories (percentage of global land area covered)
```
coxipi, line 795: the docstring here is a bit short because not sure how to break an iten list

src/xclim/core/calendar.py
coxipi: One problem is that "2MS" doesn't specify which months should be grouped. If the first month is Jan, the pairs are JF, MA, ... if the first month of ds is Feb, then FM AM ...

There is no "2MS-JAN" for instance.

src/xclim/indices/_hydrology.py
```diff
@@ -1069,54 +1009,30 @@ def base_flow_index_seasonal_ratio(
         Rate of river discharge.
     freq : str
         Resampling frequency.
+    numerator : str
+        String indicating the season in the numerator of the ratio.
+    denominator : str
+        String indicating the season in the denominator of the ratio.
```
coxipi, line 1015: I think more general names than `summer` and `winter` ought to be used. Thoughts on `numerator` and `denominator`?

src/xclim/indices/_hydrology.py
```diff
@@ -1069,54 +1009,30 @@ def base_flow_index_seasonal_ratio(
...
-    w_s_ratio = (winter_1_aligned / (summer_bfi + epsilon)).sel(year=summer_bfi.year)
-    w_s_ratio = w_s_ratio.isel(year=slice(1, None))
-    w_s_ratio.attrs["units"] = ""
-
-    return winter_bfi, spring_bfi, summer_bfi, fall_bfi, w_s_ratio
+    bfi = split_time_to_season_year(base_flow_index(q, freq), freq)
+    den = bfi.sel(season=denominator)
+    ratio = bfi.sel(season=numerator) / den.where(den > 0)
+    ratio = ratio.assign_attrs({"units": "", "denominator": denominator, "numerator": numerator})
+    return bfi, ratio
```
coxipi, line 1038: I added a `missing = "skip"`, but an alternative could be to only output the ratio?

coxipi (commented):
This is ready, if you wanted to check the PR @aulemahal 

Maybe things to watch: 
* I use [snowamount] almost everywhere, do you agree? 
* The season splitter helper could use a second eye
* I think the rest of my comments will help to see what to watch out for

src/xclim/indices/_hydrology.py
aulemahal: Makes sense to me!

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+    Returns
+    -------
+    xr.DataArray or xr.Dataset
+        Input dataset with season coordinate.
+    """
+    dsr = ds[{d: 0 for d in set(ds.dims) - {"time"}}].resample(time=freq).first()
+    mult, base, isstart, anchor = parse_offset(freq)
+    if base not in "YAQM":
+        raise ValueError(f"Only daily frequencies or coarser are supported. Got: {freq}.")
+    if base == "M" and 12 % mult != 0 or (base in "YA" and mult > 1):
```
aulemahal:
```suggestion
    if (base == "M" and 12 % mult != 0) or (base == "Q" and multi not in [1, 2, 4]) or (base in "YA" and mult > 1):
```
Might be something wrong from my version in xscen!

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+        n = mult * {"M": 1, "Q": 3}[base]
+        seasons = {}
+        for m in dsr.time.dt.month.values:
+            label = "".join(months[np.array(range(m - 1, m + n - 1)) % 12])
+            for i in range(n):
+                seasons[(m - 1 + i) % 12 + 1] = label
+        season_coords = [seasons[m] for m in ds.time.dt.month.values]
+    else:  # M or MS
+        seasons = xr.coding.cftime_offsets._MONTH_ABBREVIATIONS
+        season_coords = [seasons[m] for m in ds.time.dt.month.values]
```
aulemahal:
```suggestion
        seasons = dict(zip(_MONTH_NUMBERS.values(), _MONTH_NUMBERS.keys()))
        season_coords = [seasons[m] for m in ds.time.dt.month.values]
```
Better to use a local variable than a private one from the exterior.

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+    season_length = len(season_coords[0]) if freq not in ["M", "MS"] else 1
+    attrs = dict(mult=mult, base=base, isstart=isstart, anchor=anchor, season_length=season_length)
+    return ds.assign_coords(season=("time", season_coords)).assign_attrs(attrs)
```
aulemahal:
```suggestion
    return ds.assign_coords(season=("time", season_coords, attrs))
```
Maybe these attrs are better added to the season coord ?

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+        seasons = {}
+        for m in dsr.time.dt.month.values:
+            label = "".join(months[np.array(range(m - 1, m + n - 1)) % 12])
+            for i in range(n):
+                seasons[(m - 1 + i) % 12 + 1] = label
+        season_coords = [seasons[m] for m in ds.time.dt.month.values]
+    else:  # M or MS
+        seasons = xr.coding.cftime_offsets._MONTH_ABBREVIATIONS
+        season_coords = [seasons[m] for m in ds.time.dt.month.values]
+    season_length = len(season_coords[0]) if freq not in ["M", "MS"] else 1
```
aulemahal:
```suggestion
    season_length = len(season_coords[0]) if base != "M" else 1
```
"M" doesn't exist, it's "ME".

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+    attrs = dict(mult=mult, base=base, isstart=isstart, anchor=anchor, season_length=season_length)
```
aulemahal:
```suggestion
    attrs = dict(mult=mult, base=base, isstart=isstart, anchor=anchor or "JAN", season_length=season_length)
```

src/xclim/core/units.py
```diff
@@ -1517,7 +1518,7 @@ def infer_context(standard_name: str | None = None, dimension: str | None = None
     Returns
     -------
     str
-        "hydro" if variable is a liquid water flux, otherwise "none".
+        "hydro" if variable is a liquid water flux or snow surface amount, otherwise "none".
```
aulemahal:
```suggestion
        "hydro" if variable refers to liquid water or to a mass of water in any phase, otherwise "none".
```
Even more general ?

src/xclim/indicators/land/_snow.py
aulemahal:
```suggestion
```
I think this comment can go away now, no ?

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+    freq : str
+      Resampling frequency. Must be between "MS" and "YS" and divide a year evenly.
+
+    Returns
+    -------
+    xr.DataArray or xr.Dataset
+        Input dataset with season coordinate and yearly time.
+    """
+    ds = add_season_coord(ds, freq)
+    year_needs_shift = ds.time.dt.month < _MONTH_NUMBERS[ds.attrs["anchor"] or "JAN"]
```
aulemahal:
```suggestion
    year_needs_shift = ds.time.dt.month < _MONTH_NUMBERS[ds.season.attrs["anchor"]]
```
With my two suggestions above.

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+    hydro_years = np.where(year_needs_shift, ds.time.dt.year - 1, ds.time.dt.year)
+    _, idx = np.unique(hydro_years, return_index=True)
+    time_per_year = ds.time.values[idx]
+    out = ds.assign_coords(year=("time", hydro_years)).set_index(time=("year", "season")).unstack("time")
+    return out.assign_coords(year=time_per_year).rename(year="time")
```
aulemahal: Your solution is elegant here, but I think there's a corner case where it fails : if ds begins on a month before the anchor. 

```python3
import xclim as xc
from xclim.testing.helpers import test_timeseries
import numpy as np

ds = test_timeseries(np.ones(120), 'tas', units='°C', freq='QS-SEP', start='2001-09-01')

out = xc.core.calendar.split_time_to_season_year(ds, 'QS-DEC')
```
The first "time" is 2001-09, even though we want it to be 2000-12.


I'm not sure how to cleanly solve this. I had this idea:
```suggestion
    base_month = _MONTH_NUMBERS[ds.season.attrs['anchor']]

    def _get_year(dt):
        y = dt.year = 1 if dt.month < base_month else dt.year
        return dt.replace(year=y, month=base_month, day=0, hour=0, minute=0, second=0, microseconds=0)

    new_time = ds.indexes['time'].map(get_year)
    out = ds.assign_coords(year=("time", new_time)).set_index(time=("year", "season")).unstack("time")
    return out.rename(year="time")
```
(This suggestion is incompatible with the previous one, be careful)

You preserve the previous time axis, while this solution reconstructs it. In a way, we have to since there's at least one value that wasn't there before ? 

In opposition to `unstack_times` from xscen, this function does not preserve the original time. Not sure how problematic this is for xclim users, but it means we won't be able to use it in xscen for `climatological_op`, which is sad ?  A subsequent PR could migrate the rest of the features of `unstack_dates` to xclim, for v1 maybe.

src/xclim/core/calendar.py
```diff
@@ -1711,3 +1729,73 @@ def _reconstruct_time(_time_as_delta, _start):
...
+        Input dataset with season coordinate.
+    """
+    dsr = ds[{d: 0 for d in set(ds.dims) - {"time"}}].resample(time=freq).first()
+    mult, base, isstart, anchor = parse_offset(freq)
+    if base not in "YAQM":
+        raise ValueError(f"Only daily frequencies or coarser are supported. Got: {freq}.")
+    if base == "M" and 12 % mult != 0 or (base in "YA" and mult > 1):
+        raise ValueError(f"Only periods  that divide the year evenly are supported. Got {freq}.")
+    if base in "YA":
+        season_coords = ["JFMAMJJASOND"] * ds.time.size
```
aulemahal: You didn't like `annual-ABBR` ?

src/xclim/indices/_hydrology.py
aulemahal: Agreed!

src/xclim/indices/_hydrology.py
```diff
@@ -752,86 +748,40 @@ def runoff_ratio(
...
+      (e.g., 'YS' for yearly starting Jan, or 'QS-DEC' for seasons, '30YS' to compute the value over slices of 30 years
+      from the start of the time series).
 
     References
     ----------
     :cite:cts:'knoben_2024'
     """
-    q = convert_units_to(q, "m3/s")
-    a = convert_units_to(a, "km2")
+    # # TODO: Could be refactored with generic.select_rolling_resample_op, but `skipa=False` cannot be passed right now.
```
aulemahal: Humm, what do you mean ? Is this a mistaken copy paste ?

src/xclim/indices/_hydrology.py
```diff
@@ -752,86 +748,40 @@ def runoff_ratio(
...
+    q = convert_units_to(q, "mm3/hr")
+    a = convert_units_to(a, "mm2")
     pr = convert_units_to(pr, "mm/hr")
 
-    runoff = q * 3.6 / a  # unit conversion for runoff in mm/hr : 3.6[s/hr *km2/m2]
+    runoff = q / a
     runoff_freq = runoff.resample(time=freq).sum()
     pr_freq = pr.resample(time=freq).sum()
```
aulemahal:
```suggestion
runoff_freq = runoff.resample(time=freq).mean()
pr_freq = pr.resample(time=freq).mean()
```
The sum of a rate doesn't mean anything, no ? Taking the sum could yield weird result if the number of missing values is different between the two, but a mean would alleviate this.

src/xclim/indices/_hydrology.py
```diff
@@ -843,49 +793,62 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
+        - Humid (52.8%): AI ≥ 0.65
     - In North America, higher aridity index values can be associated with colder climates
       due to lower evapotranspiration, even when precipitation is limited or occurring as snow.
 
     References
     ----------
     :cite:cts:'zomer_2022'
     """
+    evspsblpot = convert_units_to(evspsblpot, pr)
     pr = pr.resample(time=freq).sum()
```
aulemahal:
```suggestion
    pr = pr.resample(time=freq).mean()
```
Same as above.

src/xclim/indices/_hydrology.py
```diff
@@ -843,49 +793,62 @@ def aridity_index(pr: xarray.DataArray, pet: xarray.DataArray, freq: str = "YS")
...
-    pet = pet.resample(time=freq).sum()
-    ai = pr / pet
+    evspsblpot = evspsblpot.resample(time=freq).sum()
```
aulemahal:
```suggestion
    evspsblpot = evspsblpot.resample(time=freq).mean()
```
Continued (couldn't do it in one suggestion)

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +872,130 @@ def lag_snowpack_flow_peaks(
...
-    doy = q_high.time.dt.dayofyear
+    # delta time between snw_max and t0
+    dt_snw_max = resample_map(snw, "time", freq, _timemax) - t0
+    # Convert to float, and ensure a NaT gets converted to a nan
+    dt_snw_max = xarray.where(dt_snw_max.isnull(), np.nan, dt_snw_max.astype("timedelta64[s]").astype(float))
 
-    t_q_max = doy.resample(time=freq).reduce(partial(circmean, high=366, low=1), dim="time")
+    dt_q = (q.time - t0).astype("timedelta64[s]").astype(float)
+    thresh = q.resample(time=freq).quantile(q=p, dim="time")
+    threshold_for_each_time = thresh.reindex_like(q, method="ffill")
```
aulemahal: héhé, `ffill` only works if the freq is achored at the start. You'd want `bfill` for "YE" for example, no ?

src/xclim/indices/_hydrology.py
```diff
@@ -909,153 +872,130 @@ def lag_snowpack_flow_peaks(
...
-    else:
-        slopes, p_vals = compute_seasonal_stats(q)
-        # Create labeled xarray
-        ds = xarray.Dataset(
-            data_vars={"Sen_slope": ("season", slopes), "p_value": ("season", p_vals)}, coords={"season": seasons}
+    def _mann_kendall(q, freq):
+        qr = q.resample(time=freq).mean()
+        qr = split_time_to_season_year(qr, freq)
+        out = xarray.apply_ufunc(
+            lambda q: (lambda mk_output: np.array([mk_output.slope, mk_output.p]))(mk.original_test(q)),
```
aulemahal, line 930: wow, nested lambdas, first time I see that!

Neat 📸

CHANGELOG.rst
```diff
@@ -26,6 +26,7 @@ New indicators and features
 * Added two `zero_inflated` arguments to ``xclim.indices.stats.standardized_index`` and ``xclim.indices.standardized_precipitation_index`` to control how zero-precipitation probabilities are handled. (:issue:`2279`, :pull:`2280`).
 * ``xclim.indices.stats.parametric_pdf`` allows to compute PDF distributions with given input parameters and values  (:pull:`2323`).
 * ``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index`` now can accept `genextreme` and `lognorm` as inputs for `dist`. (:issue:`2326`, :pull:`2327`).
+* # TODO: mention (un)stack_dates once we settle on how we want to implement it
```
aulemahal: Indeed, we could refrain from mentionning it here before we make a PR fully transferring it from xscen ?


```suggestion
```

CHANGELOG.rst
```diff
@@ -38,15 +39,28 @@ Internal changes
...
+* ``xclim.indices._hydrology.lag_snowpack_flow_peaks``:
+    * The argument `swe` is renamed to `snw`. (:pull:`2267`);
+    * The argument `percentile` is renamed to `p` and should now be a number in `[0,1]` instead of `[0,100]`, similar to other `xclim` function. (:pull:`2267`).
+* ``xclim.indices._hydrology.base_flow_index_seasonal_ratio`` now expects a yearly frequency whose anchor determines the position of the seasons. A single `bfi` output is given with 4 season coordinates. (:pull:`2267`).
+* ``xclim.indices._hydrology.sen_slope`` does not return both seasonal and annual results: Instead, give an appropriate resampling frequency (quarter or annual). (:pull:`2267`).
+* ``xclim.indices._hydrology.sen_slope`` now only computes the sen slope and p-value of one input and doesn't accept `qsim` as an argument. ``xclim.indices._hydrology.sen_slope_ratio`` replaces ``sen_slope`` for the full computation with two input datasets, but may be removed in the future.  (:pull:`2267`).
+* ``xclim.indices._hydrology.days_with_snowpack`` was removed, use ``xclim.land.snd_days_above`` which has the same functionality. (:pull:`2267`).
```
```suggestion
* ``xclim.indices._hydrology.days_with_snowpack`` was removed, use ``xclim.land.snw_days_above`` which has the same functionality. (:pull:`2267`).
```

aulemahal (commented):
Merci! Grosse job.

src/xclim/indices/_hydrology.py
coxipi: yes

src/xclim/indices/_hydrology.py
coxipi: Sounds good, I didn't catch this one

src/xclim/indices/_hydrology.py
coxipi: thanks

src/xclim/indices/_hydrology.py
coxipi: eh, can we just `.ffill`, then `.bfill`? Maybe there is a better way different than this, I was just building around what was already there

src/xclim/indices/_hydrology.py
coxipi: I do like a good one-liner 😆

coxipi: I can write a proper local function if that's better too, just happy someone saw this 🥲

CHANGELOG.rst
coxipi: Ha! In fact `snd_days_above` can also be used in this case, but I agree to refer to `snw_days_above`

src/xclim/indices/_hydrology.py
coxipi: I've included more steps in the resample call to avoid this issue, no reindexing is needed now

aulemahal: `"ffill" if "S" in freq else "bfill"` ?

If you `ffill` on a "YE" it will be applied to the next period, I don't think you can do both at the same time.

src/xclim/indices/_hydrology.py
```diff
@@ -879,11 +878,12 @@ def lag_snowpack_flow_peaks(
...
-    thresh = q.resample(time=freq).quantile(q=p, dim="time")
-    threshold_for_each_time = thresh.reindex_like(q, method="ffill")
-    dt_q_over_thresh = dt_q.where(q >= threshold_for_each_time)
-    dt_high_q = dt_q_over_thresh.resample(time=freq).mean(dim="time")
+    def _mean_timedelta_over_percentile(q, p):
+        dt_q = (q.time - t0).astype("timedelta64[s]").astype(float)
+        thresh = q.quantile(q=p, dim="time")
+        return dt_q.where(q >= thresh).mean(dim="time")
+
+    dt_high_q = q.resample(time=freq).map(lambda q: _mean_timedelta_over_percentile(q, p))
```
aulemahal: Ah! Also good, but may be use `resample_map`  here if possible ?

coxipi (commented):
..

src/xclim/indices/_hydrology.py
coxipi: done

src/xclim/core/calendar.py
coxipi: ok sounds good. FYI this was also taken from xscen

src/xclim/core/calendar.py
coxipi: oh yeah, I thought this one was a bit silly, let's go for `annual`

src/xclim/core/calendar.py
coxipi: ok dok

src/xclim/core/calendar.py
coxipi: Are you sure? 

Consider timeseries where DEC-2000 is missing and we want to use YS-DEC. Then after that, we have the decembers
      X         JAN-2001, FEB-2001, .... NOV-2001
DEC-2001,JAN-2002, FEB-2002, .... NOV-2002

I still would want the first year to be 2000-12-01, even though the first december is missing? 

I don't see why we should require the first date to be present. If we allow nans to occur, they could be anywhere IMO.

aulemahal: Hmm, maybe we misunderstood ourselves?  I too don't want to require the first date to be present. My example above shows that, justement, there's an issue when it is not present. My solution re-builds the datetime of the new axis so we are sure that all expected dates are present, whether or not they were there in the input.

My comment about `original_time` is unrelated and is about having an extra coordinate on the output that is the 2D version of the input's `time` -> `(time, season)`. It is used by `climatological_op` to be able to "re-stack" back to a single time axis, if needed.  I was refering to this : https://github.com/Ouranosinc/xscen/blob/4533b8f5ed1604851bc64d8fe28300671ca4ee32/src/xscen/utils.py#L128

coxipi: oups, indeed, I read things backwards. 

I see you point. I think that the `add_season_coord` gives us something new, we can more easily compute groupby with any freq of our liking: 

`add_season_coords(ds,'QS-NOV').groupby('season').map(...)`

but the `split_time_to_season_year` is indeed a more limited version of `unstack_dates`. This was my attempt not to use `unstack_dates`, maybe it's getting more complicated for nothing trying to avoid it.

I'll use your solution for now and check it works as intended later.

src/xclim/indices/_hydrology.py
coxipi: resolving this, I avoided the problem (see my last comment)

aulemahal (dismissed):
🙈

Zeitsperre (approved):
:shipit:

PR #2268, opened 2025-11-04
Add hemisphere varying thresholds to documentation

### LINKED ISSUE #1391: Document how to call indicators whose parameters depend on the hemisphere
### Addressing a Problem?

Some indicators should be computed with different parameters for the north and south hemisphere. While it is possible to embed this logic within indicators, I believe it will make the code less readable and more difficult to test and validate.

### Potential Solution

Add examples in the documentation describing the issue and how to resolve it. 
- Call the indicator twice, once for the north and once for the south
- Merge the results

Include discussion on the potential gotchas. 

### Additional context

See #1352 and #1361 

### Contribution

### Code of Conduct

ludwiglierhammer:
### Pull Request Checklist:
    - This PR fixes #1391

* This PR addresses the problem of using hemisphere varying thresholds

* No, this PR does only extend the documentation. 

### Other information:

* Hi guys, are you still interested in this issue?

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -90,7 +90,7 @@
     },
     {
       "name": "Lierhammer, Ludwig",
-      "affiliation": "Helmholtz-Zentrum Hereon, Geesthacht, Germany",
+      "affiliation": "Deutscher Wetterdienst, Hamburg, Germany",
       "orcid": "0000-0002-7207-0003"
     },
     {
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -32,7 +32,7 @@ Contributors
 * Sarah Gammon `@SarahG-579462 <https://github.com/SarahG-579462>`_
 * Tom Keel <thomas.keel.18@ucl.ac.uk> `@Thomasjkeel <https://github.com/Thomasjkeel>`_
 * Marie-Pier Labonté <labonte.marie-pier@ouranos.ca> `@marielabonte <https://github.com/marielabonte>`_
-* Ludwig Lierhammer <ludwig.lierhammer@hereon.de> `@ludwiglierhammer <https://github.com/ludwiglierhammer>`_
+* Ludwig Lierhammer <ludwig.lierhammer@dwd.de> `@ludwiglierhammer <https://github.com/ludwiglierhammer>`_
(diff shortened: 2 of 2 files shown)


ludwiglierhammer:
In example.ipynb box 9 and 19 I got a ValueError:

```
out = xclim.atmos.tx_max(ds2.tasmax, freq="YS")
```

ValueError: zero-size array to reduction operation fmax which has no identity.

This is due to the fact that `len(lat)` in `ds2` is `0`.

aulemahal:
Hi @ludwiglierhammer ! 
I took the liberty to add examples on how to merge the resulting datasets, does that look ok to you ?

Also, I created a "fake" dataset from the one in the notebook. As we are never looking as the data itself, I thought that is doesn't really matter if the data isn't real ? We could also add a "global" dataset to xclim's test data.

Finally, the CI errors we were seeing before my commit were simply due to `pre-commit` not having been run on your side. If you `pre-commit install` in your git repo, some checks will be run each time you commit.

ludwiglierhammer:
Hi @aulemahal, thank you for your additions. I am happy with them. They show the user a good test case and how to solve it.
For this test case we don't need an extra "global" dataset, I think.  

This error still remains when executing the notebook:

ValueError: zero-size array to reduction operation fmax which has no identity.

(I've installed ``pre-commit`` 😉)

aulemahal:
It is intentional. By default, if a period is incomplete, xclim flags it as missing as there are good chances that the result isn't erroneous.

`xclim.set_options(check_missing='skip')` removes the check completely, but there are other options. The "usage" notebook has a short example and the method's doc is here : https://xclim.readthedocs.io/en/stable/checks.html#missing-values-identification.

aulemahal (approved):
I think this is ready !

ludwiglierhammer:
> It is intentional. By default, if a period is incomplete, xclim flags it as missing as there are good chances that the result isn't erroneous.
> 
> `xclim.set_options(check_missing='skip')` removes the check completely, but there are other options. The "usage" notebook has a short example and the method's doc is here : https://xclim.readthedocs.io/en/stable/checks.html#missing-values-identification.

@aulemahal: Thank you for the explanation. That sounds reasonable to me.

PR #2269, opened 2025-11-04
Migrate to `tox.toml` specification, remove Python-coveralls

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Replaces the `tox.ini` file with a `tox.toml` file
* Removes `coveralls` from the `tox.toml`-only dependencies (abandoned software / not supported for Python 3.13 or 3.14)
* Adds a `coverallsapp/github-action` step to PyPI/`tox` builds on CI
* Simplified the `tox` environments significantly
* Adds `lmoments3` to the `extras` recipe (this is long overdue)

Yes. The testing suite has been modified since conditional flags are no longer supported in the new spec(?!)

This means that chaining modifiers with dashes (e.g. `py311-numpy-sbck-extras`) can no longer be used since there is no longer any logic-handling mechanism. I'm still not certain if there is a TOML-based equivalent for this type of feature that the documentation just isn't showcasing or perhaps the string replacement mechanism (https://tox.wiki/en/4.30.2/config.html#string-substitution-references) can be worked to have a similar functionality. This isn't clear to me.

### Other information:

I've added `lmoments3` to the `extras` recipe. This shouldn't cause enormous problems because this is an optional library that we do not explicitly link to. I don't imagine this will create licensing problems.

### DIFF
--- a/src/xclim/core/units.py
+++ b/src/xclim/core/units.py
@@ -663,7 +663,7 @@ def to_agg_units(
     ... )
     >>> cond = tas > 100  # Which days are boiling
     >>> Ndays = cond.sum("time")  # Number of boiling days
-    >>> Ndays.attrs.get("units")
+    >>> Ndays.attrs.get("units")  # modern xarray preserves the units of tas, but these are incorrect here
     'degC'
     >>> Ndays = to_agg_units(Ndays, tas, op="count")
     >>> Ndays.units


src/xclim/core/units.py
```diff
@@ -664,7 +664,7 @@ def to_agg_units(
     >>> cond = tas > 100  # Which days are boiling
     >>> Ndays = cond.sum("time")  # Number of boiling days
     >>> Ndays.attrs.get("units")
```
aulemahal: This change comes from xarray's change on "keep_attrs", but I think the new result makes the example a bit confusing, so I suggest adding a comment.
```suggestion
    >>> Ndays.attrs.get("units")  # modern xarray preserves the units of tas, but these are incorrect here
```

aulemahal (approved):

PR #2280, opened 2025-11-24
Flexible handling of zero-precipitation probabilities in standardized precipitation index

### LINKED ISSUE #2279: Add flexibility to handling probability of zero-precipitation events in Standardized Indices (SPI)
### Addressing a Problem?

The calculation of the Standardized Precipitation Index (SPI) involves transforming a fitted probability distribution (often Gamma) into a standard normal distribution. A critical step in this process is handling "zero-inflated" data (months with no precipitation).

Different methodologies exist for assigning a probability to these zero events before the normal transformation:

- Traditional/McKee Method ("Upper"): Assigns the full cumulative probability of zero, $P(x=0) = p_0$. This is known to introduce a positive bias in the mean SPI, especially in arid regions where $p_0$ is large, as it effectively treats all zero events as "wetter" than they should be.

- Center of Mass Method ("Center"): Assigns the probability $p_0 / 2$. This is recommended by Stagge et al. (2015) to correct the bias and ensure the SPI distribution remains centered at 0.

> Reference: 
> [Stagge et al. (2015) Candidate Distributions for Climatological Drought Indices (SPI and SPEI)](https://rmets.onlinelibrary.wiley.com/doi/10.1002/joc.4267)

### Potential Solution

Adding a `prob_zero_method` parameter to `standardized_index` and `standardized_precipitation_index`, which allow users to specify whether use "upper", "center" or customized callable as the probability corresponding to zero-precipitation.

### Additional context

```
# %%
import xarray as xr
import numpy as np
import scipy.stats
from xclim.indices import standardized_precipitation_index
from xclim.indices.stats import standardized_index_fit_params

# Use cftime to create a NOLEAP calendar (365 days every year)
# This ensures a strict 365-day periodicity.
time = xr.cftime_range("2000-01-01", "2010-12-31", calendar="noleap", freq="D")
pr = xr.DataArray(
    np.random.rand(len(time)) * 10,
    coords={"time": time},
    dims="time",
    name="pr",
    attrs={"units": "mm/day"}
)

# Force DOY 180 to be ZERO for ALL 11 years.
pr[{"time": slice(179, None, 365)}] = 0

# Parameters for fitting
fit_params = dict(
    freq=None,
    window=1,
    dist="gamma",
    method="ML",
    doy_bounds=(180, 180),
)

# 1. Get fitting parameters
params = standardized_index_fit_params(pr, zero_inflated=True, **fit_params)
prob_of_zero = params.prob_of_zero.sel(dayofyear=180).values
print(f"Probability of Zero (p0): {prob_of_zero:.4f}")

# 2. Compute SPI
# Original zero-precipitation probability: "upper" (Biased) -> P = p0
spi_upper = standardized_precipitation_index(
    pr, params=params, **fit_params
)

# Verification: Check the SPI value for the first occurrence (index 179)
t_idx = 179
val_upper = spi_upper.isel(time=t_idx).values

print(f"\nResults for Zero-Precipitation Day (p0=1.0):")
print(f"Upper (p0):      SPI={val_upper:.4f}")
```
This will leads to `Probability of Zero (p0): 1.0000` and `SPI=8.2100`, which may leads to misleading "extreme wet".

### Contribution

### Code of Conduct

Hem-W:
### Pull Request Checklist:
    - This PR fixes #2279

Adds a new argument `prob_zero_method` to `standardized_index` and `standardized_precipitation_index`. This argument allows users to control how the probability of zero-precipitation events ($P(x=0)$) is handled when transforming to the standard normal distribution.

Supported options are:
* "center" (Default): Uses the Center of Mass method ($P = p_0 / 2$) as recommended by Stagge et al. (2015) to avoid bias in the mean SPI.
* "upper": Uses the traditional cumulative probability ($P = p_0$), which replicates the legacy behavior (McKee method).
* Callable: Accepts a custom function (e.g., lambda p: p * 0.75) for experimental use cases.

No.

While the default calculation method has changed from the implicit "upper" method to the unbiased "center" method. The API remains backward compatible, and users who strictly require the old values can set prob_zero_method="upper".

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
       - id: yamllint
         args: [ '--config-file=.yamllint.yaml' ]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.11
+    rev: v0.15.1
     hooks:
       - id: ruff-format
         exclude: '(src/xclim/indices/__init__.py|docs/installation.rst)'
@@ -61,7 +61,7 @@ repos:
     hooks:
       - id: vulture
   - repo: https://github.com/kynan/nbstripout
-    rev: 0.8.2
+    rev: 0.9.0
     hooks:
       - id: nbstripout
         files: '.ipynb'
@@ -109,7 +109,7 @@ repos:
     hooks:
       - id: gitleaks
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.36.0
+    rev: 0.36.2
     hooks:
       - id: check-github-workflows
       - id: check-readthedocs
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,12 +8,12 @@ sphinx:
 #  - pdf
 
 build:
-  os: ubuntu-24.04
+  os: "ubuntu-24.04"
   tools:
-    python: "mambaforge-23.11"
+    python: "miniforge3-25.11"
   jobs:
     pre_create_environment:
-      - sed -i "s/python >=3.11,<3.14/python >=3.11,<3.13/" environment.yml
+      - sed -i "s/python >=3.11,<3.15/python >=3.11,<3.13/" environment.yml
     pre_build:
       - sphinx-apidoc -o docs/apidoc/ --private --module-first src/xclim src/xclim/testing/tests src/xclim/indicators src/xclim/indices
       - rm docs/apidoc/xclim.rst
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 
 v0.61.0 (unreleased)
 --------------------
-Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Hui-Min Wang (:user:`Hem-W`), Éric Dupuis (:user:`coxipi`).
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`), Trevor James Smith (:user:`Zeitsperre`), Hui-Min Wang (:user:`Hem-W`), Éric Dupuis (:user:`coxipi`).
 
 New indicators and features
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -12,6 +12,14 @@ New indicators and features
     + This changes the behaviour for indicators ``land.snw_max_doy``, ``land.snd_max_doy``, ``land.doy_qmin`` and ``land.doy_qmax``.
 * Added two `zero_inflated` arguments to `standardized_index` and `standardized_precipitation_index` to control how zero-precipitation probabilities are handled. (:issue:`2279`, :pull:`2280`).
 
+Internal changes
+^^^^^^^^^^^^^^^^
+* Documentation adjustments. (:pull:`2325`):
+    * Updated the `intersphinx` directives to point at new `xsdba` sphinx documentation targets.
+    * Updated the ReadTheDocs configuration to build docs with an older Python version and a newer conda version.
+    * Updated the `pre-commit` hooks (`check-jsonschema`) in order to accept the latest supported conda version in ReadTheDocs config.
+    * Set ``docs/conf.py`` to ignore `sphinx_autodoc_typehints.guarded_import` errors raised by `xarray` type guarding.
+* Set `SocketBlockedError` to be a subset of the `Exception` class when `pytest-socket` is not installed. (:pull:`2324`).
 
 v0.60.0 (2026-01-23)
 --------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -88,7 +88,7 @@ SDBA Module
 
 .. note::
 
-    For more information, the documentation is available at the following link: :doc:`xsdba API <xsdba:api>`
+    For more information about `xsdba`, the documentation is available at the following link: :doc:`xsdba API <xsdba:apidoc/xsdba>`
 
 .. _spatial-analogues-api:
 
@@ -193,7 +193,7 @@ SDBA Utilities
 
 .. note::
 
-    For more information about `xsdba`, please refer to the documentation at the following link: :doc:`xsdba API <xsdba:api>`
+    For more information about the `xsdba` developer utilities, please refer to the documentation at the following link: :mod:`xsdba.utils <xsdba:xsdba.utils>`.
 
 .. _`spatial-analogues-developer-api`:
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -116,6 +116,8 @@ autodoc_typehints_description_target = "documented_params"
 autosectionlabel_prefix_document = True
 autosectionlabel_maxdepth = 2
 
+suppress_warnings = ["sphinx_autodoc_typehints.guarded_import"]
+
 linkcheck_ignore = [
     # too labourious to fully check
     r"https://github.com/Ouranosinc/xclim/(pull|issue).*",
--- a/docs/sdba.rst
+++ b/docs/sdba.rst
@@ -108,7 +108,7 @@ User API
 
 .. note::
 
-    For more information, the documentation is available at the following link: :doc:`xsdba API <xsdba:api>`
+    For more information about `xsdba`, the documentation is available at the following link: :doc:`xsdba API <xsdba:apidoc/xsdba>`
 
 Notes for Developers
 ====================
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,7 +127,7 @@ xclim = "xclim.cli:cli"
 [tool]
 
 [tool.bumpversion]
-current_version = "0.60.1-dev.2"
+current_version = "0.60.1-dev.4"
 commit = true
 commit_args = "--no-verify --signoff"
 tag = false
--- a/src/xclim/__init__.py
+++ b/src/xclim/__init__.py
@@ -13,7 +13,7 @@ from xclim.indicators import atmos, convert, generic, land, seaIce  # noqa
 
 __author__ = """Travis Logan"""
 __email__ = "logan.travis@ouranos.ca"
-__version__ = "0.60.1-dev.2"
+__version__ = "0.60.1-dev.4"
 
 
 with _resources.as_file(_resources.files("xclim.data")) as _module_data:
--- a/src/xclim/ensembles/_base.py
+++ b/src/xclim/ensembles/_base.py
@@ -320,7 +320,7 @@ def ensemble_percentiles(
             # Smart rechunk on dimension where chunks are the largest
             chk_dim, chks = max(
                 enumerate(ens.chunks),
-                key=lambda kv: (0 if kv[0] == ens.get_axis_num("realization") else max(kv[1])),
+                key=lambda kv: 0 if kv[0] == ens.get_axis_num("realization") else max(kv[1]),
             )
             ens = ens.chunk({"realization": -1, ens.dims[chk_dim]: len(chks) * ens.realization.size})
         else:
--- a/src/xclim/indicators/atmos/_temperature.py
+++ b/src/xclim/indicators/atmos/_temperature.py
@@ -1328,9 +1328,11 @@ degree_days_exceedance_date = Temp(
     units="",
     standard_name="day_of_year",
     long_name="Day of year when the integral of mean daily temperature {op} {thresh} exceeds {sum_thresh}",
-    description=lambda **kws: "Day of year when the integral of degree days (mean daily temperature {op} {thresh}) "
-    "exceeds {sum_thresh}"
-    + (", with the cumulative sum starting from {after_date}." if kws["after_date"] is not None else "."),
+    description=lambda **kws: (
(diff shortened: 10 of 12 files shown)


src/xclim/indices/stats.py
```diff
@@ -1053,6 +1059,18 @@ def reindex_time(_da: xr.DataArray, _da_ref: xr.DataArray, _group: str):  # nump
     # standardized_precipitation_index where values are not negative.
     probs = prob_of_zero + ((1 - prob_of_zero) * dist_probs)
 
+    # User-defined probability of zero values
+    if zero_inflated:
+        if callable(prob_zero_method):
+            zeros_prob = prob_zero_method(prob_of_zero)
+        elif prob_zero_method == "center":
+            zeros_prob = prob_of_zero / 2
```
coxipi: Ok, so the relation in the paper is: 

Original probability of zero

$$p_0 = \frac{n_{p=0}}{n+1}$$

and with the center of mass method: 

$$\bar{p}_0 = \frac{n_{p=0} + 1}{2(n+1)}$$

Currently, we don't have the unbiased definition with the n+1, we just compute $p_0 = n_{p=0} / n$
```python
prob_of_zero = da.groupby(group_handler).map(lambda x: (x == 0).sum("time") / x.notnull().sum("time"))
```

Whether we use $n$ or $n+1$ in the relations above will typically have a small impact, but for cases with only few dry days,using $n_{p=0}$ or $n_{p=0}+1$ in the center-of-mass relation would matter. Do you agree?

Should we always the unbiased definition in any case?

coxipi: All that to say that your `zeros_prob = prob_of_zero / 2` will currently correspond to an alternate definition

$$\bar{p}_0' = n_{p=0} / ( 2 n)$$

src/xclim/indices/stats.py
```diff
@@ -1053,6 +1059,18 @@ def reindex_time(_da: xr.DataArray, _da_ref: xr.DataArray, _group: str):  # nump
...
+        elif prob_zero_method == "upper":
+            zeros_prob = prob_of_zero
+        else:
+            raise ValueError(f"Unknown prob_zero_method: {prob_zero_method}")
```
coxipi: @Zeitsperre 

We don't want to print user-input for security reasons, correct?  It's a shame, it's so convenient to understand errors

coxipi:
Hi! 
Thanks for your contribution, this will be a very nice addition to `xclim`! 

For the `Callable`, did you have some relation in mind that is sometimes used? Since it only uses `prob_of_zero` and not `n`, the total number of observations, it's a bit limited in what it can do, correct? In your test, your example is another factor (0.8 I think). 

If that would be more useful to keep $n_{p=0}$ and $n$ rather than $p_0$, we could do that too.

src/xclim/indices/_agro.py
coxipi: could you add your Stagge-2015 reference here? 

with the bibtex entry in ``docs/references.bib``

src/xclim/indices/stats.py
Zeitsperre: Ideally, we shouldn't. If it's really convenient, we can do this to satisfy the checks:
```python
msg = f"Unknown prob_zero_method: {prob_zero_method}"
raise ValueError(msg)
```

coxipi: it's not more necessary than elsewhere

coxipi:
```suggestion
            raise ValueError(f"Unknown `prob_zero_method`. It should be a string ('upper', 'center') or a `Callable`.")
```

src/xclim/indices/_agro.py
Hem-W: Has been added in the following commit.

src/xclim/indices/stats.py
Hem-W: Thank you for reviewing. Your interpretation is right and I also had this issue when implementing this. I am not sure whether changing `prob_of_zero` is desirable so I just use the center of biased definition following what was originally done in xclim.

I just further looked into the detail about the denominator. I found that both unbiased and biased estimators occur in official programs or literature. 

In the [SPI program](https://drought.unl.edu/monitoring/SPI/SPIProgram.aspx) adopted by [WMO SPI user manual](https://library.wmo.int/records/item/39629-standardized-precipitation-index-user-guide?language_id=13), they used a biased one as the following code, and I examined that the sample results provided are from this biased definition without $+1$.
```vbnet
            /*  compute sums */
            for( int i = 0; i < datarr.Length; i++ )
            {
                if( datarr[i] > 0.0 )
                {
                    sum += datarr[i];
                    sumlog += Math.Log( datarr[i] );
                    nact++;
                }
                else
                {
                    pzero++;
                }
            }

            pzero /= datarr.Length;
```  

For the unbiased version, it is also common in literature with rigorous statistical definition, for example from [this textbook](https://www.academia.edu/26724680/N_T_Kottegoda_R_Rosso_Applied_Statistics_for_Civil_and_Environmental_Engineers_Blackwell_Pub_2008_Copy).

image

So I am not sure which should be implemented here. Maybe expose this as an option to user for their decision? Will this be too trivial?

coxipi: I was about to say, let's just use the unbiased (with a breaking warning).

But spei.R library (https://github.com/sbegueria/SPEI/blob/master/R/spei.R) uses the 1/n and not 1/(n+1), and I think it's popular enough to warrant the fact that we should have an option to implement the same method
```R
      # Probability of zero (pze)
      if (distribution != "log-Logistic") {
        pze <- sum(x.mon == 0) / length(x.mon)
        x.mon <- x.mon[x.mon > 0]
      }
```

We can add the option unbiased = False by default.

coxipi: We could use an additional option `plotting_position`
https://matplotlib.org/mpl-probscale/tutorial/closer_look_at_plot_pos.html

and at least offer 'ecdf' and 'weibull'? which are the biased and unbiased we discussed? 

After that, the prob_zero_method could be in the range:
`[plotting_position_of_rank1,  plotting_position_of_rank_n_{pr}]`
i.e. 
`(1-f)*(pp_rank1) + (f)*(pp_rank_{n}_pr)`

with f=1/2 giving the center of mass, and f=1 giving the upper method. 

Maybe we should give the option to choose $\alpha$ and $\beta$, but only permit a factor for your `prob_zero_method` (or 'center', 'upper' options)?

src/xclim/indices/stats.py
```diff
@@ -1068,7 +1068,7 @@ def reindex_time(_da: xr.DataArray, _da_ref: xr.DataArray, _group: str):  # nump
         elif prob_zero_method == "upper":
             zeros_prob = prob_of_zero
         else:
-            raise ValueError(f"Unknown prob_zero_method: {prob_zero_method}")
+            raise ValueError(f"Unknown `prob_zero_method`. It should be a string ('upper', 'center') or a `Callable`.")
```
Zeitsperre: Could be a more specific error, I suppose.

```suggestion
            raise NotImplementedError(f"Unsupported `prob_zero_method`. It should either be a string ('upper', 'center') or a `Callable`.")
```

Hem-W: Modified accordingly. Thank you.

src/xclim/indices/stats.py
```diff
@@ -986,6 +995,17 @@ def standardized_index(
...
+        if prob_zero_method is not None or rank_method is not None:
+            raise ValueError("If `zero_inflated` is `None`, `prob_zero_method` and `rank_method` must be `None`.")
+    # zero_inflated is not None
+    prob_zero_method = prob_zero_method or "upper"
+    rank_method = rank_method or "ecdf"
+    if isinstance(prob_zero_method, str) and prob_zero_method not in {"center", "upper"}:
+        raise ValueError("Accepted strings for `prob_zero_method` are: ['center', 'upper']")
+    if isinstance(rank_method, str) and rank_method not in {"ecdf", "weibull"}:
+        raise ValueError("Accepted strings for `rank_method` are: ['ecdf', 'weibull']")
+
```
Hem-W: When `rank_method` is not str, it must be a iterable with two elements.

src/xclim/indices/stats.py
```diff
@@ -1059,18 +1098,6 @@ def reindex_time(_da: xr.DataArray, _da_ref: xr.DataArray, _group: str):  # nump
     # standardized_precipitation_index where values are not negative.
     probs = prob_of_zero + ((1 - prob_of_zero) * dist_probs)
```
Hem-W: For those non-zero values, the distribution should always use the "upper" to transform the probability, as the decision to use "center" or "upper" should not change the conversion of the non-zero distribution. That's why I used a post-processing of `zeros_prob` previously.

I am not sure whether`prob_of_zero_rank_n = (number_of_zeros - alpha) / (number_of_notnull + 1 - alpha - beta)` or simply `(number_of_zeros) / (number_of_notnull)` (i.e. fixing alpha and beta whatever user inputs) should be used here. I guess the former one is proper as it is consistent.

Hem-W (commented):
`prob_of_zero` in the conversion of distribution for non-zero values needs to be changed. When `prob_of_zero` is specified to use "center", `prob` still needs to use the  full probability of zeros instead of the center to convert the non-zero distribution.

src/xclim/indices/stats.py
coxipi: Yes, I was pondering on the best way to treat this. @Zeitsperre, sometimes we have some checks on the types, sometimes not, we wait for other internal errors to handle the wrong values. For instance, when the variable is a string, we often check that only the accepted values are passed. But, should I also check that the type is ok? I feel we rarely specifically check the type, so that's why I tend not to check it.

Zeitsperre: I think here you don't need to check the type as the membership check is already doing that. The type of checking that we've been progressively adding in is to ensure that whatever logging system doesn't get inundated with text (e.g. if someone were to pass an enormous string (several MBs in size) to a function that writes to a log file). 

There should be some guidance on what exactly we don't want to be doing.

src/xclim/indices/stats.py
coxipi: Right, thanks for catching this. Following the Stagge paper, I think we should have 

$$
\begin{cases}
(p_0)  + (1 - p_0) F(x_{p>0}, \lambda), \quad x>0 \\
\bar{p}_0
\end{cases}
$$


So we must distinguish between $\bar{p}_0$ and $p_0$ at this point (demarkated by `prob_zero_method` and `zero_factor`), but they can both be computed according to weibull or ecdf (or some custom value by giving alpha,beta as a tuple).

src/xclim/indices/stats.py
```diff
@@ -1056,7 +1054,7 @@ def reindex_time(_da: xr.DataArray, _da_ref: xr.DataArray, _group: str):  # nump
     params = reindex_time(params, da, group)
     dist_probs = dist_method("cdf", params, da, dist=dist)
 
-    zero_inflated = "prob_of_zero" in params.coords
+    zero_inflated = "number_of_zeros" in params.coords
     if zero_inflated:
         number_of_zeros = params["number_of_zeros"]
```
Hem-W, line 1092: `prob_of_zero` is not needed anymore, but removing `prob_of_zero` and adding `number_of_zeros` and `number_of_notnull` will lead to breaking changes for users who rely on the saved params from previous versions.

coxipi: Yes, that's why I had not removed it. Adding the new params is fine. Perhaps you can leave `prob_of_zero` in, and add a comment to say it could be removed in the future.

tests/test_indices.py
```diff
@@ -1298,6 +1298,60 @@ def test_zero_inflated(self, open_dataset):
...
+
+        expected_prob = None
+        number_of_zeros = spi.number_of_zeros.isel(time=365 + 179).values
+        number_of_notnull = spi.number_of_notnull.isel(time=365 + 179).values
+        if prob_zero_method == "center":
+            expected_prob = (1 + number_of_zeros) / (number_of_notnull) / 2
+        elif prob_zero_method == "upper":
+            expected_prob = (number_of_zeros) / (number_of_notnull)
+        else:
+            raise ValueError(f"Unknown method: {prob_zero_method}")
```
coxipi:
```suggestion
```
this is a test and we control what is given, we can do without the warning

tests/test_indices.py
```diff
@@ -1298,6 +1298,60 @@ def test_zero_inflated(self, open_dataset):
             spid[zero_inflated] = spid[zero_inflated].where(spid[zero_inflated].notnull(), drop=True)
         np.testing.assert_equal(np.all(np.not_equal(spid[False].values, spid[True].values)), True)
 
+    @pytest.mark.parametrize(
+        "prob_zero_method",
+        ["center", "upper"],
+    )
```
coxipi, line 1309: Splendid, you already coded a test! Could you add test cases for ecdf/weibull? I think the PR will be good to go after this

src/xclim/indices/stats.py
coxipi: I added it in with a note we may consider dropping it in the future

coxipi (approved):
It's good to go, thanks for this nice contribution!

PR #2307, opened 2026-01-14
Remove useless noqas, do not import matplotlib

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Removes many bare `# noqa` comments sprinkled throughout the codebase.
* Updates the `pylint` config in `pre-commit` to allow for newer versions of system-installed `pylint` to run without errors.
* `matplotlib` is no longer imported when checking for the presence of it in the library (import speed up)

Not really. This technically increases the number of `mypy` warnings, but those need to eventually be dealt with anyway.

The `matplotlib` check prevents us from needing to import `matplotlib` on import, so that should save microseconds off of import time; With enough importing, this will save years. You're welcome.

### Other information:

The initial removal of `noqa` comments was performed with [yesqa](https://github.com/asottile/yesqa).


src/xclim/ensembles/_reduce.py
```diff
@@ -19,13 +20,10 @@
 from sklearn.cluster import KMeans
 
 # Avoid having to include matplotlib in xclim requirements
-try:
-    from matplotlib import pyplot as plt  # noqa
-
+MPL_INSTALLED = False
+matplotlib_installed = _util.find_spec("matplotlib")
+if not matplotlib_installed:
     MPL_INSTALLED = True
```
aulemahal:
```suggestion
MPL_INSTALLED = bool(_util.find_spec("matplotlib"))
```
I think your version inverts the logic ?

Zeitsperre: Yeah, that would be the case. :upside_down_face:

PR #2310, opened 2026-01-22
Fix `time_bnds` and fix for Pandas 3

aulemahal:
### Pull Request Checklist:
- [ ] This PR addresses an already opened issue (for bug fixes / features)
    - This PR fixes #xyz
- [x] Tests for the changes have been added (for bug fixes / features)
  - [x] (If applicable) Documentation has been added / updated (for bug fixes / features)
- [x] CHANGELOG.rst has been updated (with summary of main changes)
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added

#### `time_bnds`
While fixing a Pandas v3 error, I also revisited the `time_bnds` function. I'm not sure what I had in mind when implementing it, but it was not following the CF conventions. 

Example:
```python3
import xclim as xc
import xarray as xr

time = xr.date_range('2000-01-01', freq='MS', periods=3)
xc.core.calendar.time_bnds(time, 'MS')
```
Before: the end is the last timestep (up to a precision) before the next period
```
[['2000-01-01T00:00:00.000000000', '2000-01-31T23:59:59.999999999'],
 ['2000-02-01T00:00:00.000000000', '2000-02-29T23:59:59.999999999'],
 ['2000-03-01T00:00:00.000000000', '2000-03-31T23:59:59.999999999']]
```
After: the end is the same as the start of the next period
```
[['2000-01-01T00:00:00.000000000', '2000-02-01T00:00:00.000000000'],
 ['2000-02-01T00:00:00.000000000', '2000-03-01T00:00:00.000000000'],
 ['2000-03-01T00:00:00.000000000', '2000-04-01T00:00:00.000000000']]
```

The `precision` argument as thus been removed.

#### Pandas v3
We had two other Pandas v3 issues:

In `aggregate_between_dates` we were modifying an array in-place (`days[days < 0] = np.nan`) and this seems not to be possible anymore. However, it always been discouraged by xarray, so I changed it to a `days.where(days >=0)`.

In the partitioning functions of `xclim.ensembles` we were creating new coordinate by passing a `pd.Index`, this seems to be broken now that pandas has implemented a new String dtype with nan support. Numpy does not recognize that.  Xarray should probably fix this by having an automated translation, but in the mean time it was easy to fix by creating the coordinate with `xr.DataArray` instead.

### Does this PR introduce a breaking change?
Yes, the `time_bnds` result is now different, an argument has been removed.

PR #2312, opened 2026-01-22
Remove `keep_attrs` option - follow xarray for preserving attributes

### LINKED ISSUE #2308: Sync with xarray : preserve attributes by default.
### Generic Issue

As of xarray [2025.11](https://docs.xarray.dev/en/stable/whats-new.html#v2025-11-0-nov-17-2025), attributes are preserved by default.

However, in xclim's indicator we call the  `compute` so:
https://github.com/Ouranosinc/xclim/blob/648f85f55036ccb6f040af004988c5daad6dd471/src/xclim/core/indicator.py#L896-L897

I think the `keep_attrs=False` was intended here to ensure the rest of the function had full control of the attributes management. The function is explicitly saving and re-adding the attributes if xarray's option is "True". Previously, the default (`"default"` was equivalent to `False`). Now it is stil `"default"` but with the opposite behaviour, which makes xclim a bit contradictory.

With the change in xarray, users are now _expecting_ the attributes to be preserved (see for example Ouranosinc/xsdba#244). I believe we should follow suite and simply remove this option setting at line 896.

Then, I don't think xclim's `keep_attrs` option would still be useful, it would only be needed when `as_dataset=True` for controlling global attributes, but we could simply base the behaviour on xarray's option (drop if `keep_attrs=False`, preserve otherwise).

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2308

Removes xclim's  `keep_attrs` option.  Attributes for indicators are now implicitly controlled by xarray's option. 

When option `as_dataset` is True and a dataset is given as input, or when a DataTree is given as input, the global attributes are managed explicitly by xclim, but still following xarray's option.

Yes. An option was removed.

### Other information:
The documentation of the `as_dataset` option has been updated. A line has been added to the "usage" notebook.

### DIFF
--- a/tests/test_indicators.py
+++ b/tests/test_indicators.py
@@ -184,6 +184,7 @@ def test_attrs(tas_series):
     ],
 )
 def test_keep_attrs(tasmin_series, tasmax_series, xropt, exp):
+    pytest.importorskip("xarray", minversion="2025.11")
     tx = tasmax_series(np.arange(360.0))
     tn = tasmin_series(np.arange(360.0))
     tx.attrs.update(something="blabla", bing="bang", foo="bar")
@@ -198,7 +199,6 @@ def test_keep_attrs(tasmin_series, tasmax_series, xropt, exp):
 
 @pytest.mark.parametrize("xrkeep", [True, False])
 def test_as_dataset(tasmax_series, tasmin_series, xrkeep):
-    pytest.importorskip("xarray", minversion="2025.11")
     tx = tasmax_series(np.arange(360.0))
     tn = tasmin_series(np.arange(360.0))
     tx.attrs.update(something="blabla", bing="bang", foo="bar")


docs/notebooks/partitioning.ipynb
```diff
@@ -93,7 +93,7 @@
     "ens_mon = xr.combine_by_coords(data)[\"world\"]\n",
     "\n",
     "# Then resample the monthly time series at the annual frequency\n",
-    "ens = ens_mon.resample(time=\"Y\").mean()\n",
+    "ens = ens_mon.resample(time=\"YE\").mean()\n",
```
aulemahal, line 96: This is what fixes the CI issue following #2310.

src/xclim/core/formatting.py
```diff
@@ -773,28 +772,3 @@ def get_percentile_metadata(data: xr.DataArray, prefix: str) -> dict[str, str]:
         f"{prefix}_window": data.attrs.get("window", "<unknown window>"),
         f"{prefix}_period": clim_bounds,
     }
-
-
-# Adapted from xarray.structure.merge_attrs
-def _merge_attrs_drop_conflicts(*objs):
```
juliettelavoie, line 779: on l'appelait dans xscen... tristesse, ça nous apprendra à appeler des fonctions cachées.

aulemahal: oh oups! en même temps c'est pour ça qu'on pinne xclim dans xscen!

PR #2314, opened 2026-01-23
Doymin, doymax :  nan when all the same

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Rewrite of `doymin` and `doymax`. Addition of a check to put NaN when all the values are the same.

Also, I don't know since where this is possible, but using `idxmin` is much simpler than `argmin` with `lazy_indexing`! I think the `dt` accessor wasn't always available with dask.

I want to remove the special treatment in snow indicators where we first check is any value is above 0. With this new behaviour of `doymax` , the special treatment is unnecessary for `snd_max_doy`, for example.

This PR also has the side effect of 
- allowing all-nan slices
- return a float instead of an integer

Before a slice where all values where the same would return the doy of the first timestamp. I think this is problematic.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -178,6 +178,11 @@
       "name": "Hofmann, Armin",
       "affiliation": "Julius-Maximilians-Universität, Würzburg, Germany",
       "orcid": "0009-0005-9580-3256"
+    },
+    {
+      "name": "Vandelanotte, Kobe",
+      "affiliation": "Royal Meteorological Institute of Belgium (RMI), Brussels, Belgium",
+      "orcid": "0009-0001-1252-7315"
     }
   ],
   "keywords": [
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -53,3 +53,4 @@ Contributors
 * Jens de Bruijn <j.a.debruijn@outlook.com> `@jensdebruijn <https://github.com/jensdebruijn>`_
 * Armin Hofmann `@HofmannGeo <https://github.com/HofmannGeo>`_
 * Eve Larose <eve.larose.r@gmail.com> `@e-larose <https://github.com/e-larose>`_
+* Kobe Vandelanotte <kobe.ande@gmail.com> `@kobebryant432 <https://github.com/kobebryant432>`_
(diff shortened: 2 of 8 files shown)


src/xclim/indices/generic.py
```diff
@@ -188,15 +187,13 @@ def doymax(da: xr.DataArray) -> xr.DataArray:
...
-    if uses_dask(da):
-        out = lazy_indexing(doy, i, "time").astype(doy.dtype)
-    else:
-        out = doy.isel(time=i)
-    return to_agg_units(out, da, "doymax")
+    tmax = da.idxmax("time")
+    std = da.std("time")
+    tmax = tmax.where(std != 0)
+    doy = tmax.dt.dayofyear
+    return to_agg_units(doy, da, "doymax")
```
coxipi, line 196: `idxmax` doesn't require lazy indexing? 

Also, I think it's hard to find an appropriate way to handle things when a dataset is problematic like this. I don't like the idea of giving a NaN since this behaviour deviates from the usual behaviour of `idxmax`, but it's probably the best thing to do.

CHANGELOG.rst
```diff
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+v0.61.0 (unreleased)
+--------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+New indicators and features
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* ``xclim.indices.generic.doymin`` and ``xclim.indices.generic.doymax`` will now return `nan` if all values along the time axis are the same. They now also support all-nan arrays.
```
coxipi: don't forget the pull number

src/xclim/indices/generic.py
aulemahal: Indeed it does!
They do it here : https://github.com/pydata/xarray/blob/a24c8c73b79771e0442687b0ba29c894dc0a7387/xarray/computation/computation.py#L1000
That's fewer lines and looks simpler than our implementation.

aulemahal: @RondeauG  opinions about my change to doymin where a series without any variation returns NaN instead of the first doy ?

RondeauG: Yes, this is fine to me.

PR #2316, opened 2026-01-26
[new xclim] Adopt PEP-735 (Dependency groups)

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Splits the `lint` dependencies into a `dependency-group` in `pyproject.toml`.
* Adjusts the Makefile to silently install linting dependencies if the `lint` command is called.
* Updates the `tox` configuration to the latest version supporting `dependency-groups`.
* Addresses a small `yamllint` warning concerning line length.
* Adds `--help` alias to the `xclim` CLI.

Yes. The development dependencies no longer install the linting dependencies by default. `make lint` will now install the linting dependencies only if explicitly called; Otherwise, the `pre-commit` step will handle the running of linting steps. `tox` has been adjusted in order to refer to the "lint" `dependency-group` (removing redundancy).

### Other information:

`dependency-groups` are a way to ensure that development-time dependencies are now published with distributions of packages, which is great for packages that do not bundle code, or for preventing end-users from installing tools they will never require when using the package.

Example invocation:
```
$ pip install --group lint,test
```

A `dependency-group` is only available when examining the source code of the library and allows developers to only install the dependencies without needing to install the library itself (great for linting). It also supports installing the library or nesting of other `dependency-groups`/`optional-dependencies`, if desired.

Testing and other miscellaneous development libraries are also suggested to be listed under `dependency-groups`; Is this something that people would be OK with? We could split these dependencies out such that the only additional libraries would be listed under `xclim[extras]`.

https://peps.python.org/pep-0735/

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,6 +19,7 @@ on:
   pull_request:
     branches:
       - main
+      - dev-v099
   schedule:
     - cron: '30 23 * * 5'
 


pyproject.toml
```diff
@@ -2,6 +2,20 @@
 requires = ["flit_core >=3.11.0,<4"]
 build-backend = "flit_core.buildapi"
 
+[dependency-groups]
```
aulemahal, line 5: Maybe this section can be put after the the main `[project]` ? So that a human reading this file would start with the most important ?

Zeitsperre: The section order is alphabetically ordered by `toml-sort`. I could see if there's a way to override placement of a particular section though.

aulemahal: oh! in that case you can forget that.

PR #2323, opened 2026-02-17
Parametric pdf

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Add ``xclim.indices.stats.parametric_pdf``


src/xclim/indices/stats.py
```diff
@@ -354,6 +355,72 @@ def func(x):  # numpydoc ignore=GL08
     return out
 
 
+def parametric_pdf(
+    p: xr.DataArray,
+    x: xr.DataArray | float | Sequence[float],
```
coxipi: I think a DataArray makes sense here, so I can assign units with it. Do you agree? I could also change parametric_cdf in this way

src/xclim/indices/stats.py
```diff
@@ -354,6 +355,72 @@ def func(x):  # numpydoc ignore=GL08
...
+        da_x = xr.DataArray(x, dims=["x"]).assign_coords(x=x)
+    else:
+        if len(x.dims) > 1:
+            raise ValueError("`x` must be one-dimensional.")
+        da_x = x
+
+    dist = get_dist(dist or p.attrs["scipy_dist"])
+
+    data = xr.apply_ufunc(
+        lambda x, p: dist.pdf(x, *p),
```
coxipi: what do we think of in-line lambda functions? I'm guessing this would not pass on the council ...

Zeitsperre: We can make exceptions.

Simple functions are better than more complex ones; If it's easily read/understood, I don't see a problem.

src/xclim/indices/stats.py
Zeitsperre: So long as both `pdf` and `cdf` are consistent, that should be fine. Are there cases where a user would just want to send in a `numpy` array?

Zeitsperre (approved):
I'm no expert on this, but the API and logic seem good to me!

src/xclim/indices/stats.py
```diff
@@ -315,16 +316,18 @@ def parametric_cdf(
...
-    # Create a lambda function to facilitate passing arguments to dask. There is probably a better way to do this.
-    def func(x):  # numpydoc ignore=GL08
-        return dist.cdf(v, *x)
-
     data = xr.apply_ufunc(
-        func,
+        lambda v, p: dist.pdf(v, *p),
         p,
         input_core_dims=[["dparams"]],
         output_core_dims=[["cdf"]],
```
coxipi: BTW, I don't like how the coordinate is called `cdf`. It should be `v` the value. And then, CDF is the variable associated to this value.

Probably not worth changing. But I don't want to put "pdf" in `parametric_pdf` even if it would be more cohesive. 

Another minor nitpick: I think `x` is a better variable name in `parametric_pdf` than the one used in `parametric_cdf`. In general, we should probably aim to conform better with scipy. 

For instance: 
```python
sp.stats.quantile(x, p, ...)
```
but we define: 
```python
parametric_quantile(p, q, ...)
```

where `q` corresponds to `p` in scipy, and `p` are just the params ... I'm remembering that xarray is responsible for this confusion. *sigh*

Zeitsperre: The only thing that pops into my mind is I wouldn't want to confuse a variable `x` with a coordinate `x` (longitude). As far as I'm concerned, if we think that it makes better sense to rename this variable, add a warning now and break it in `xclim` v1.0.

coxipi: Ah yeah, that's fair. The `x` for longitude is probably too widespread, so maybe I should avoid this. 

I still don't like naming the coordinate `pdf` or `cdf` though. I think the coordinate should be `v` in this case (or whatever name we choose to represent the variable values we input).

CHANGELOG.rst
```diff
@@ -11,7 +11,11 @@ New indicators and features
 * ``xclim.indices.generic.doymin`` and ``xclim.indices.generic.doymax`` will now return `nan` if all values along the time axis are the same. They now also support all-nan arrays (:pull:`2314`).
     + This changes the behaviour for indicators ``land.snw_max_doy``, ``land.snd_max_doy``, ``land.doy_qmin`` and ``land.doy_qmax``.
 * Added two `zero_inflated` arguments to `standardized_index` and `standardized_precipitation_index` to control how zero-precipitation probabilities are handled. (:issue:`2279`, :pull:`2280`).
+<<<<<<< add_parametric_pdf
 * ``xclim.indices.stats.parametric_pdf`` allows to compute PDF distributions with given input parameters and values  (:pull:`2323`).
+=======
+* ``xclim.indices.standardized_precipitation_index`` and ``xclim.indices.standardized_precipitation_evapotranspiration_index`` now can accept `genextreme` and `lognorm` as inputs for `dist`. (:issue:`2326`, :pull:`2327`).
+>>>>>>> main
```
Zeitsperre: Woops?

CHANGELOG.rst
```diff
@@ -11,7 +11,11 @@ New indicators and features
 * ``xclim.indices.generic.doymin`` and ``xclim.indices.generic.doymax`` will now return `nan` if all values along the time axis are the same. They now also support all-nan arrays (:pull:`2314`).
     + This changes the behaviour for indicators ``land.snw_max_doy``, ``land.snd_max_doy``, ``land.doy_qmin`` and ``land.doy_qmax``.
 * Added two `zero_inflated` arguments to `standardized_index` and `standardized_precipitation_index` to control how zero-precipitation probabilities are handled. (:issue:`2279`, :pull:`2280`).
+<<<<<<< add_parametric_pdf
+* ``xclim.indices.stats.parametric_pdf`` allows to compute PDF distributions with given input parameters and values  (:pull:`2323`).
+=======
```
coxipi:
```suggestion
* ``xclim.indices.stats.parametric_pdf`` allows to compute PDF distributions with given input parameters and values  (:pull:`2323`).
```

CHANGELOG.rst
coxipi: for the record, I'm git-blaming Github for this error, it was all buggy when I tried to commit the changes

src/xclim/indices/stats.py
```diff
@@ -315,33 +318,36 @@ def parametric_cdf(
...
-    # Assign quantile coordinates and transpose to preserve original dimension order
-    dims = [d if d != "dparams" else "cdf" for d in p.dims]
-    out = data.assign_coords(cdf=v).transpose(*dims)
+    # Assign value coordinates and transpose to preserve original dimension order
+    out = data.transpose(*(d if d != "dparams" else "cdf" for d in p.dims))
     out.attrs = unprefix_attrs(p.attrs, ["units", "standard_name"], "original_")
 
     attrs = {
-        "long_name": f"{dist.name} cdf",
+        "long_name": f"{dist.name} CDF",
```
coxipi:
```suggestion
        "long_name": f"{dist.name} cdf",
```

PR #2329, opened 2026-03-04
Validating call signature types

### LINKED ISSUE #2328: Validating call signature types
### Addressing a Problem?

Hi guys,

I was thinking about a check that tests whether all call signature types are valid. 
This check could be a simple function or even better a decorator which runs before the actual method.

We could see this as an addition to https://github.com/Ouranosinc/xclim/issues/1719.

I would appreciate your interest.

Thank you.

### Potential Solution

In https://github.com/glamod/marine_qc/pull/128, I implemented these validation checks for another project in which I am already in contact with @Zeitsperre.  

### Additional context

_No response_

### Contribution

### Code of Conduct

ludwiglierhammer:
### Pull Request Checklist:
    - This PR fixes #2328

* [ ] add simple type validation functions in module `xclim.core.validations`
* [ ] add decorator that validates function arguments against type hints `xclim.core.validations.validate_parameters`

* new  decorator that validates function arguments against type hints

### Questions

- Do you want me to adjust the new decorator to some functions?
- How can we take other decorators into account? Do other decorators convert the type?

### DIFF
--- a/src/xclim/core/validations.py
+++ b/src/xclim/core/validations.py
@@ -1,0 +1,406 @@
+"""Module containing base QC which call multiple QC functions and could be applied on a DataBundle."""
+
+from __future__ import annotations
+
+import collections.abc as abc
+import inspect
+from collections.abc import Callable, Iterable, Mapping, Sequence
+from functools import wraps
+from types import UnionType
+from typing import (
+    Annotated,
+    Any,
+    Literal,
+    Union,
+    get_args,
+    get_origin,
+    get_type_hints,
+)
+
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+
+
+def validate_non_generic(value: Any, expected: Any) -> bool:
+    """
+    Validate a non-generic type (str, int, float, etc.).
+
+    Parameters
+    ----------
+    value : Any
+        The value to validate.
+    expected : Any
+        The expected type.
+
+    Returns
+    -------
+    bool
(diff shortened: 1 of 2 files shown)


Zeitsperre:
@ludwiglierhammer I finally had some time/headspace to look all of this over.

I agree with @aulemahal on the fact that the call signature typing/hinting that we use is fairly customized for the indicator code and with the upcoming changes, we're advertising that the new `compute` module should be used by more capable/"power" users who likely don't need this level of checking.

`mypy`/`ty` do indeed check the call signature types, but not at the user-level. We still need to finish working on that compliance, but we have a few funky types that we haven't yet figured out how to best rewrite.

I like the idea of working some of this into `finch`; We are planning on doing a revamp/rewrite of that service to support the modern `xclim` release, and we don't currently have much logging/verbosity when it comes to reporting user error. This could definitely be helpful there.

As an aside:
> I think we should remove the
> 
> ```python
>     if not isinstance(freq, str):
>         raise TypeError("Freq must be a string.")
> ```
I do think we should be fine removing some of these checks. There are a handful of instances where `mypy`/`ty` would ask for us to redefine an object to remove the `optional` quality from its potential type e.g.
```python
def some_func(freq: str | None)
    if isinstance(freq, str):
        _freq = str(freq)
    ...
```
But that isn't the case with those specific checks.

@ludwiglierhammer Thanks for taking the time to propose this, but it doesn't seem to work with our intended approach at this time.

Issue #1726, opened 2024-04-22
Either support passing of all `Quantified`-like objects to all functions supporting them or create a new `TypeVar`


Zeitsperre:
* xclim version: v0.48.3-dev
* Python version: Any
* Operating System: All

### Description

We currently use the `Quantified` type alias in almost all the threshold-based indicators to specify that this variable should conform to something resembling a string with units (`str | pint.Quantity`), but this alias also includes `xarray.DataArray`. Does this mean that we can pass (or would like to enable passing) DataArrays of grids denoting thresholds? No, we currently don't support that behaviour, but maybe we would want to?

If not, we may need to create a new, more constrained custom variable (`str2pint` often takes variables of this type but does not support `pint.Quantity` or `xarray.DataArray`). Solving this issue alone will resolve at least 1/4 of all errors associated with `mypy` typing (~300+).

See also: #1719 and #1721

PR #2355, opened 2026-05-13
Drop Python 3.10, adjust typing, and other breaking changes

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Firmly drops support for Python3.10.
* Replaces `mypy` with `ty` and associated configurations (mostly deactivated).
* Raises several dependency minimum versions:
  * `numba` (>=0.60.0)
  * `numpy` (>=2.0)
  * `pip` (>=26.1)
  * `scikit-learn` (>=1.5.0)
  * `tox` (>=4.52.0)
  * `xarray` (>=2024.6.0,!=2024.10.0)
* Replaces `pre-commit` with `prek` (Rust-built reimplementation).
* Removes several previously-deprecated indicators and removed associated tests:
  * `xclim.indicators.land.snd_to_snow`
  * `xclim.indicators.land.snow_to_snd`
  * `xclim.indicators.convert.tg`
* Adjusts several type definitions within code base, based on `ty` suggestions. Many `cast()` calls have been removed.
* Upgraded `xclim` development status to `Production/Stable`.
 
Absotively-posolutely.

### Other information:

Previously, Python 3.10 support was "_until such a time that it would be blocking development_". Python 3.10 will be losing security patches soon; Continuing to support that version is no longer a good idea.


aulemahal (approved):
Looks good!

Lots of renaming with an underscore, not sure I see the benefit of this linting rule.  I understand that it does not reassign to a name from the arguments, but I don't understand what it protects us from. Anyway, this is also a personal opinion and too minor to block this.

Zeitsperre:
> Lots of renaming with an underscore, not sure I see the benefit of this linting rule.

It's not a linting thing but more of a typing approach; Some functions will transform an existing variable from one type (or set of possible types) to another. If the original variable is being rewritten, this means that the code for that variable must support all types that it could be, both those that it was assigned in the call signature and those being returned by the transformation/function. Renaming with an underscore is a cheap/easy way to create a new variable with less need for types it should no longer be/support.

I can't say whether it helps speed anything up, but it does help catch some typing issues that are deep in the code base, beyond the biggest offenders. Getting ever closer to closing #1719 and #1726.

aulemahal:
Ah! I see it now. Indeed, for example, in the case of `Quantified`, the first move is usually a  `convert_units_to` which modifies the potential types from `DataArray | str` to `DataArray | float`.

PR #2357, opened 2026-05-15
Fix rate2amount error of 1000 when sampling inferred from cftime coord

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

In `rate2amount` and `amount2rate`,  when `sampling_rate_from_coord` is True, which is implicit when the sampling freq is monthly or coarser, we compute the length of the sampling period by taking the diff of the time coordinate, we then convert the `timedelta` objects in seconds.

Previously, both time types (`numpy.datetime64` and  `cftime`) yielded a `timedelta64[ns]`  dtype when subtracted. But, when pandas introduced support for other units, xarray followed and now it will often return `timedelta64[us]` in the cftime case. 

This PR changes the conversion to be explicit, not simply assuming nanoseconds.

No, it was broken before.

### Other information:
I'm afraid the changes in pandas and xarray are not so recent, so this bug has been there for a while!!

@Zeitsperre I think this is a patch-release level seriousness (but perhaps not urgent enough for releasing on a friday).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+v0.61.1 (unreleased)
+--------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Fix conversion error with ``xc.units.rate2amount`` and ``xc.units.amount2rate`` when ``sampling_rate_from_coord=True`` or sampling frequency is monthly or coarser and time coordinate is cftime-backed. Previous results were 1000x to small. (:pull:`2357`).
+
 v0.61.0 (2026-05-07)
 --------------------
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+v0.61.1 (unreleased)
+--------------------
+Contributors to this version: Pascal Bourgault (:user:`aulemahal`).
+
+Bug fixes
+^^^^^^^^^
+* Fix conversion error with ``xc.units.rate2amount`` and ``xc.units.amount2rate`` when ``sampling_rate_from_coord=True`` or sampling frequency is monthly or coarser and time coordinate is cftime-backed. Previous results were 1000x to small. (:pull:`2357`).
```
```suggestion
* Fix conversion error with ``xc.units.rate2amount`` and ``xc.units.amount2rate`` when ``sampling_rate_from_coord=True`` or sampling frequency is monthly or coarser and time coordinate is `cftime`-based. Previous results were 1000x too small. (:pull:`2357`).
```

Zeitsperre (approved):
Good catch. Definitely deserves a patch release. Let's aim for Monday.

PR #2358, opened 2026-05-15
Fix Indicator documentation for Sphinx 9

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

Turns out sphinx 9 became the one installed in the RTD env somewhere between xclim 0.60 and 0.61. Sphinx 9 has completely rewritten the `autodoc` module and our custom indicator documenter is not supported in the new version. 

Recall that an indicator is an instance of the `Indicator` class, so not one of the objects `autodoc` knows how to document. We have a custom sphinx extension that registers an `IndicatorDocumenter` which inherits `FunctionDocumenter` and allows documenting indicators as if they were functions.

The "legacy" method is still usable with a new option, which this PR activates.

It looks like extending the new `autodoc` in a similar way is simply impossible for the moment.  There's an issue open on sphinx about this. For the record : sphinx's doc has an example for custom documenters and no where in it does it mentions that it's using a "legacy" mechanism and there's no explanation on how to do this with the new version yet.

The changes to `autodoc_indicator.py` are mainly esthetic, remnants from a few experiments trying to make it work before finding the solution.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,6 +112,7 @@ extensions = [
 autodoc_typehints = "description"
 autodoc_typehints_format = "fully-qualified"
 autodoc_typehints_description_target = "documented_params"
+autodoc_use_legacy_class_based = True
 
 autosectionlabel_prefix_document = True
 autosectionlabel_maxdepth = 2
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,7 +112,7 @@ extensions = [
 autodoc_typehints = "description"
 autodoc_typehints_format = "fully-qualified"
 autodoc_typehints_description_target = "documented_params"
-autodoc_use_legacy_class_based = True
+autodoc_use_legacy_class_based = True  # required for Indicators API with sphinx v9.0
 
 autosectionlabel_prefix_document = True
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -112,6 +112,7 @@
 autodoc_typehints = "description"
 autodoc_typehints_format = "fully-qualified"
 autodoc_typehints_description_target = "documented_params"
+autodoc_use_legacy_class_based = True
```
Zeitsperre:
```suggestion
autodoc_use_legacy_class_based = True  # required for Indicators API with sphinx v9.0
```

Zeitsperre (approved):
Looks good. Just add a Changelog entry for it!

Zeitsperre:
Very random question, not blocking the PR: Would it be better for us to just have subclasses for every indicator? Would there be any potential gain beyond being "more Pythonic"?

aulemahal:
Not sure I understand the question? Each indicator is currently already it's own subclass. 

But the `Documenter` works on all subclasses inheriting from `Indicator` and documents them according to their `__call__`, as this is the principal entry point of the class. A normal user should see an indicator as a function.

PR #2372, opened 2026-06-08
Default values for dc0 and dmc0 when dry start is activated

### LINKED ISSUE #2371: Bug: `dry_start="CFS"` with `overwintering=True` causes NaN DMC (and therefore FWI) for always-on grid points and at first season startup
**Version:** xclim 0.61.1

When using `cffwis_indices` with both `overwintering=True` and `dry_start="CFS"` (or `"GFWED"`), FWI is all NaN for:
1. The entire time series for always-on grid points (e.g. tropical/arid regions like North Africa)
2. The entire first fire season at any grid point where no previous season exists

**Root cause:** `ow_DMC` is initialized to `dmc0.copy()`, which is NaN when no previous season is provided. During winter days before the first `start_up`, the CFS accumulation adds to NaN:

```python
ow_DMC[winter & ~wetpts] = ow_DMC[winter & ~wetpts] + params["dmc_dry_factor"]
```

NaN + number = NaN. At `start_up`, there is no NaN fallback for DMC:

```python
ind_prevs["DMC"][start_up] = ow_DMC[start_up]  # NaN → poisons BUI and FWI
```

Unlike DC, which has an explicit NaN fallback in the overwintering block:

```python
ind_prevs["DC"][start_up] = np.where(
    np.isnan(dc0),
    params["dc_start"],  # ← correct fallback
    _overwintering_drought_code(...)
)
```

DMC has no equivalent fallback when `dry_start` is active.

**Why always-on cells are affected for the entire time series:** For always-on grid points (season mask always `True`), the season never shuts down, so `ow_DMC[shut_down] = params["dmc_start"]` never fires. `ow_DMC` stays NaN forever. On day 1, `start_up=True` (due to `initial_start_up=True`), so `ind_prevs["DMC"]` is set to NaN and propagates NaN through the daily DMC computation for the entire time series. There is no recovery.

**Why later seasons work for cells with genuine shutdowns:** At shutdown, `ow_DMC[shut_down] = params["dmc_start"]` resets `ow_DMC` to a finite value. Subsequent winter accumulation therefore starts from a finite value, and the next `start_up` produces a correct DMC. The bug only persists for cells that never shut down.

**Proposed fix:** Replace:
```python
ind_prevs["DMC"][start_up] = ow_DMC[start_up]
```

With:
```python
ind_prevs["DMC"][start_up] = np.where(
    np.isnan(ow_DMC[start_up]),
    params["dmc_start"],
    ow_DMC[start_up],
)
```

**Workaround:** Use `dry_start=None` with `overwintering=True`.

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2371

* Adds a default value to `DC` and `DMC` when the dry start is activated. The idea of dry start is to increase the starting values of these during dry winter days, so it didn't make sense to start at NaN and only perform the dry start mechanism after the first season.

Yes, values have changed, but this is quite niche.

### Other information:
The test is very simple, but I don't think we have canonical data for this specific case.


src/xclim/indices/fire/_cffwis.py
```diff
@@ -925,9 +925,9 @@ def fire_weather_ufunc(  # noqa: C901 # numpydoc ignore=PR01,PR02
     lat : xr.DataArray, optional
         Latitude in °N, not needed for FFMC or ISI.
     dc0 : xr.DataArray, optional
-        Previous DC map, see Notes. Defaults to NaN.
+        Previous DC map, see Notes. Defaults to NaN, or `dc_start` is `dry_start` is not none.
```
SarahG-579462:
```suggestion
        Previous DC map, see Notes. Defaults to NaN, or `dc_start` if `dry_start` is not none.
```

src/xclim/indices/fire/_cffwis.py
```diff
@@ -925,9 +925,9 @@ def fire_weather_ufunc(  # noqa: C901 # numpydoc ignore=PR01,PR02
...
     dmc0 : xr.DataArray, optional
-        Previous DMC map, see Notes. Defaults to NaN.
+        Previous DMC map, see Notes. Defaults to NaN, or `dmc_start` is `dry_start` is not none.
```
SarahG-579462:
```suggestion
        Previous DMC map, see Notes. Defaults to NaN, or `dmc_start` if `dry_start` is not none.
```

SarahG-579462 (commented):
Urgh these functions are such 🍝 ... it seems ok? docs should definitely be made clearer on the startups for these functions, I find them quite hard to read.

I think I ran into this issue in my project, my solution was just to skip dry_start on the first iteration... I thought I just misunderstood it 🙄 

> If overwintering is activated, this mechanism is only used for the DMC. When dry start is activated, `dc0` and `dmc0` default to `dc_start` and `dmc_start` respectively.

implies `dc0 = dc_start` even with overwintering (this seems to only happens at NaNs in the code)

>  If overwintering is activated, it overrides this parameter and only DMC is handled through the dry start mechanism.

implies `dc0 != dc_start` with overwintering

aulemahal:
@SarahG-579462  I think I worded it better now!

SarahG-579462 (approved):
Looks good!

PR #2374, opened 2026-06-17
Allow open bounds in select_time

### LINKED ISSUE #2365: Possibility of using open bounds doy in select_time
### Addressing a Problem?

Right now, using open ``doy_bounds`` are not supported in ``xclim.core.calendar.select_time`` while sometime it can be useful to select a variable from the start of the year until a given doy, or from a given doy to the end of the year. This is can be easily done when using ``dayofyear`` as ``int`` (e.g., ``doy_bounds=(1, 100)`` to select from the start of the year to the 100th doy or ``doy_bounds=(150, 366)`` for from the 150th doy to the end of the year), however, there isn't an easy way to do it when using ``xr.DataArray`` ``doy_bounds`` with a frequency other than ``"YS"``.

Let's say we want to select the time from a doy, based on the degree days exceedance, until the end of the year. If we use a ``dayofyear`` ``xr.DataArray`` calculated using ``freq="YS"`` is still quite easy:
```python
doy = degree_days_exceedance_date(tas, thresh="10 degC", sum_thresh="650 K days", freq="YS")

# from the doy to end of year
select_time(tas, doy_bounds=(doy, 366))
# from the start of the year to the doy
select_time(tas, doy_bounds=(1, doy))
```
However, if the ``dayofyear`` are computed using another frequency, for example ``YS-JUL`` that is common when working on the South Hemisphere, it's less straighforward. The exemple above with ``freq="YS-"JUL"`` gives something like that:

Image

You can see with the figure on the right that the ``doy`` occurs around january, so this is why we don't have value on the left figure because ``doy`` < 366. Although, the output are "corrects" as the function returns exactly what was asked, it does not return what we want (i.e., data from/to the start/end of the year). To do so, we need to update the 1 and 366 values with the values corresponding to our resampling frequency. Here, the start and end of the resampling are the 07-01 and 06-30, corresponding to the 182th and 183th day of year (except for leap year where it's +1). I can update the code above:
```python
doy = degree_days_exceedance_date(tas, thresh="10 degC", sum_thresh="650 K days", freq="YS")

# from the doy to end of year
select_time(tas, doy_bounds=(doy, 181))
# from the start of the year to the doy
select_time(tas, doy_bounds=(182, doy))
```

Image

There isn't any issue for non-leap data but with data with leap years (like in this case), there are two issues. While the left figure looks good, the last day of the year is missing for leap years because we should have used 182. On the right, we don't have data for the leap year because the bound we used is before the start of the ``YS-JUL`` which is 183 for leap year. If we use the day of year values for leap years (i.e., 183 and 182):

Image

There are still the two same issues, but this time it's the opposite. On the left, we only have data for leap year and on the right the first day of the year is missing, except for leap years.

Currently, the only way to address these issues is to use the "leap safe" doy as in the exemple above and thus intentionally drop the first and last days of year of the resampling frequency and I don't think it's a good way of doing it. Moreover, while everyone knows the start and end doy for ``YS``, it's required more time to figure out what are the appropriate value for other frequencies, still feasible though. For these reasons above, I think ``xclim`` should be able to handdle the open doy bound to automatically deal with this.

### Potential Solution

I've already fixed that in one of our project by tweaking ``mask_between_doy``. You can have a look at what the function (``_mask_between_doys``) looks like [here](https://forge.inrae.fr/agroclim/Indicators/OutilsPourIndicateurs/fonctionspython/pynar/-/blob/main/pynar/helpers.py?ref_type=heads).

Basically, the function would allow this: ``select_time(da, doy_bounds=(doy, None))``.  A ``NaN`` handling strategy is also implemented to determine how to deal with year with ``NaN`` in the input doy ``DataArray``.
I'm happy to submit a PR if you think it worth it !

### Additional context

In addition to the above issue, I've just realised that the name ``mask_between_doys`` is quite confusing and does not reflect what the function does. I think the function should be rename ``mask_outside_doys`` or ``select_between_doys`` (I have a preference for the second one which is consistent with ``select_time``). I can open another issue for this!

### Contribution

### Code of Conduct

baptistehamon:
### Pull Request Checklist:
    - This PR fixes #2365 

* Rename ``xclim.core.calendar.mask_between_doys`` to ``xclim.core.calendar.select_between_doys``
* Allow open ``date_bounds`` and ``doy_bounds`` (i.e., ``None``) in ``xclim.core.calendar.select_time``
* Add ``include_doy_bounds_nans`` and ``bounds_freq`` parameters to ``select_time`` to specify how NaN should be handle in ``doy_bounds`` and the start of the period, respectively.

The minor introduced breaking change is that array-like ``doy_bounds`` without a ``time`` dimension must now be consecutive according to the frequency (it was already the case for ``DataArray`` with ``time``).


src/xclim/core/calendar.py
```diff
@@ -1147,41 +1147,74 @@ def _get_doys(start: int, end: int, inclusive: tuple[bool, bool]):
     return doys
 
 
-def mask_between_doys(
-    da: xr.DataArray,
+def select_between_doys(
+    da: xr.DataArray | xr.Dataset,
     doy_bounds: tuple[int | xr.DataArray, int | xr.DataArray],
```
Zeitsperre: The signature and logic needs to handle the case for `None` if the docstring says it does (`doy_bounds = None`).

src/xclim/core/calendar.py
```diff
@@ -1147,41 +1147,74 @@ def _get_doys(start: int, end: int, inclusive: tuple[bool, bool]):
...
     """
-    Mask the data outside the day of year bounds.
+    Select data between day of year bounds.
 
     Parameters
     ----------
     da : xr.DataArray or xr.Dataset
         Input data. It must have a time coordinate.
-    doy_bounds : 2-tuple of integers or DataArray
+    doy_bounds : 2-tuple of integers or DataArray or None
```
Zeitsperre:
```suggestion
    doy_bounds : 2-tuple of integers or DataArray, optional
```

src/xclim/core/calendar.py
```diff
@@ -1147,41 +1147,74 @@ def _get_doys(start: int, end: int, inclusive: tuple[bool, bool]):
...
+        Whether the bounds of `doy_bounds` should be inclusive or not. Default is True (inclusive).
+    include_nans : bool, optional
+        Whether to include values associated with NaN in `doy_bounds`. If True (default), missing values (NaN) in
+        the start and end bounds are replaced by the start and end of the period, respectively.
+    freq : str, optional
+        The yearly frequency (e.g. "YS", "YS-JUL") used to determine the open bounds (start and end of the period)
+        with array-like `doy_bounds` without a `time` dimension (Default "YS"). If `doy_bounds` have a `time`
+        dimension, the frequency is first tried to be inferred from the time coordinate of the bounds; if it cannot
+        be inferred, the frequency must be passed explicitly.
+    drop : bool, optional
```
Zeitsperre: `False` is not `None`
```suggestion
    drop : bool
```

Zeitsperre (commented):
I'll let others weigh in on this, but (correct me if I'm wrong) I think the logic tells me the changes may run into problems dealing with 360-day calendars. Is that possible?

src/xclim/core/calendar.py
baptistehamon: Thanks for the suggetions. Just to clarify, `doy_bounds` can't be `None` but the bounds can, so `doy_bounds=None` won't work, but `doy_bounds=(150, None)`, for example, will. I can easily improve the signature as follow:
```python
    doy_bounds: tuple[int | xr.DataArray | None, int | xr.DataArray | None],
```
And I agree the docstring might be confusing, I'm not sure how to maje it clear though. It could be:
```
    doy_bounds : 2-tuple of integers, DataArray or None
    # or
    doy_bounds : 2-tuple of optional integers or DataArray
```
Let me know what you prefer

baptistehamon:
You're right, the logic does not work with a 360-day calendar. However, it is not the result of this PR changes as I just realised that the current behaviour of xclim's `select_time` (and underlying `mask_between_doys`) does not handle 360-day `xr.DataArray` doy bounds neither. Is it an acceptable issue with this PR that can be fixed later, or do you want this PR to fix that?

src/xclim/core/calendar.py
Zeitsperre: It's a bit of a complicated docstring but I prefer `doy_bounds : 2-tuple of optional integers or DataArray`. This makes the most sense to me.

Zeitsperre:
@baptistehamon I didn't realize that that functionality wasn't already there. My suggestion would be to open an issue at most. There may be a justification as to why that might be the case, but I'd let my colleague @aulemahal weigh in there as he wrote much of the calendar-handling functionality.

Perhaps for here if you want to focus solely on ensuring that the feature supports all existing functionality, that would be enough.

Most the team is currently on vacation, but some of them should be back by next week. Thanks again!

baptistehamon:
I'm open an issue (#2393) regarding the error with 360_days calendar.

src/xclim/core/calendar.py
baptistehamon: I improved the signature and docstrings in one of the last commit

baptistehamon:
I've edited the code to fix small issues and to minimise breaking changes. I added back the old behaviour for closed ``date_bounds`` in ``select_time`` to avoid an issue with the 360_days calendar; the new behaviour applies only to open bounds (i.e., ``None``). As a result, the PR introduces only minor breaking changes (detailed in the PR description), and I updated the relevant tests. I'm going to add more tests to cover the new functionality (open bounds) in the coming couple of days, but otherwise the PR is almost ready :)

src/xclim/core/calendar.py
```diff
@@ -1212,58 +1249,72 @@ def mask_between_doys(
...
-        else:  # Only "Spatial" dims, we can't constrain as in days since, so there are two cases
-            doys = da.time.dt.dayofyear  # for readability
-            # Any missing value is replaced with the min/max of possible values
-            start = start.fillna(1)
+        # add time dimension if not present, with bounds given by freq
+        if "time" not in start.dims:
+            freq = freq if freq is not None else "YS"
+            bnds = time_bnds(da.resample(time=freq))
+            start = start.expand_dims(time=bnds.time)
+            end = end.expand_dims(time=bnds.time)
```
aulemahal, line 1257: Is there a case where `"time" not in start.dims` but `"time" in end.dims` ? Should we handle this too ?

src/xclim/core/calendar.py
```diff
@@ -1212,58 +1249,72 @@ def mask_between_doys(
...
+
+        # Convert the doy bounds to a duration since the beginning of each period defined
+        # in the bound's time coordinate.
+        # Also ensures the bounds share the same time calendar as the input.
+        calkws = {"calendar": da.time.dt.calendar, "use_cftime": (da.time.dtype == "O")}
+        start = doy_to_days_since(start.convert_calendar(**calkws))
+        end = doy_to_days_since(end.convert_calendar(**calkws))
+
+        # Fill missing values in start and end bounds
+        if include_nans or start.isnull().all():
```
aulemahal: Wouldn't there be a case where `start` was passed as an array full of NaN and `include_nans=False` ? I feel like the second part of this condition should test if  `start` was passed a None explicitly.

Also, as far as I understand, I think this is the only compute-triggering line in the function ? Avoiding `start.isnull().all()` would allow `start` to be a dask array, no ?

src/xclim/core/calendar.py
```diff
@@ -1370,6 +1436,27 @@ def select_time(
...
+                """Convert MM-DD string to day of year, for each year in time."""
+                doys = [doy_from_string(date_str, year, cal) for year in time.dt.year]
+                return xr.DataArray(doys, coords={"time": time}, dims="time", name="dayofyear")
+
+            if get_calendar(da) == "360_day":
+                raise NotImplementedError(
+                    "Passing open `date_bounds` (i.e., with None) is not supported for 360_day calendars."
+                )
+
+            bnds = time_bnds(da.time.resample(time=bounds_freq if bounds_freq is not None else "YS"))
```
aulemahal:
```suggestion
            bnds = time_bnds(da.time.resample(time=bounds_freq or "YS"))
```

src/xclim/core/calendar.py
```diff
@@ -1370,6 +1436,27 @@ def select_time(
...
+            cal = get_calendar(da)
```
aulemahal:
```suggestion
            cal = da.time.dt.calendar
```

aulemahal:
Thanks @baptistehamon ! I made a first pass, nothing major to say here!

src/xclim/core/calendar.py
baptistehamon: It's possible indeed, but the ``start`` and ``end`` ``xr.DataArray`` are broadcasted against each other just before:
```python
        # Ensure they both have the same dims
        # align join='exact' will fail on common but different coords, broadcast will add missing coords
        start, end = xr.broadcast(*xr.align(start, end, join="exact"))
```
So correct me if I'm wrong but this should already handled that case.

src/xclim/core/calendar.py
baptistehamon: You're right, I haven't thought about this, but I can implement it

src/xclim/core/calendar.py
aulemahal: oh! indeed, my bad.

src/xclim/core/calendar.py
baptistehamon: Implemented in one of the last commit.

baptistehamon:
Should be ready now! In my opinion, there are two last things that might be worth considering before the merging though:

1. Making the use of ``doy_bounds`` possible with 360_days calendar and from what I've seen, this can be easily fix (see #2393).
2. Maybe I missed something but I think the indicator time indexing is only cover in the  _Basic Usage_ notebook in the documentation and I feel like this is under-documentated, especially since the implemetation of the array-like ``doy_bounds``. I think it is worth getting a _Indicator Time Indexing_ example notebook providing more detailed about this feature and the different options. I'm happy to do it but I think it's going to take a bit of time, so depending of when you want to release the v1 of xclim, it could be part of a later PR.

aulemahal (approved):
I pushed a fix to #2393 directly here! 

Thanks @baptistehamon !

Zeitsperre (approved):
Looking good! Reviewing in order to trigger additional tests

baptistehamon:
Thanks @aulemahal for fixing #2393! I just noticed you forgot to remove the parentheses specifying “360-day calendar are not supported” in the ``select_between_doys`` and ``select_time`` docstrings.

Zeitsperre:
Thanks again, @baptistehamon!

PR #2386, opened 2026-07-22
Add solar interpolation module

SarahG-579462:
### Pull Request Checklist:
    - This PR fixes #xyz

* Adds a module for interpolating solar noon (and potentially other interesting solar times)

* No

### Other information:
Todo:
- [x] Module code
- [x] Optional dependencies
- [ ] Tests
- [ ] Documenting with a notebook?


src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+    """
+    Calculate solar noon with astral library
+
+    Parameters
+    ----------
+    t : np.ndarray[datetime64]
+        day datetime to calculate solar noon at.
+
+    lon : np.ndarray[float]
+        longitudes to calculate solar noon for.
```
Zeitsperre: Parameters can't be separated by a blank line:
```suggestion
    t : np.ndarray[datetime64]
        day datetime to calculate solar noon at.
    lon : np.ndarray[float]
        longitudes to calculate solar noon for.
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+for lib in ["pvlib", "astral", "ephem"]:
+    if importlib.util.find_spec(lib):
+        default_method = lib
+        if lib != "pvlib":
+            warnings.warn(f"pvlib library not found, default solar calculations will be performed with {lib}")
+        break
+
+
+def _solar_noon_astral_calc(t, lon):
```
Zeitsperre:
```suggestion
def _solar_noon_astral_calc(t: np.ndarray, lon: np.ndarray):
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+        dask="parallelized",
+        dask_gufunc_kwargs=dict(),
+    )
+
+    # solar_noon_timedelta is in fractions of hour, as a float, for dask compatibility.
+    # We convert to int (timedelta64[s]), to not give better impressions on our accuracy.
+    return ds.time + (solar_noon_timedelta * 3600).astype("timedelta64[s]")
+
+
+def _solar_noon_ephem_calc(lat, lon, day, sun, elev=0):
```
Zeitsperre: Not so sure about the `ephem.Sun` since you can't use that type before importing it, but the rest should be good.
```suggestion
def _solar_noon_ephem_calc(lat: float, lon: float, day: datetime.date, sun: "ephem.Sun", elev: float = 0.0):
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+    elev : float, optional
+        elevation of observer
+
+    Returns
+    -------
+    pd.Timestamp
+        time of solar noon
+    """
+    import ephem
```
Zeitsperre: I think it would be good to guard these with warnings on import. We do that for other libraries as well so that the documentation doesn't freak out (we don't really mock missing libraries).

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+    Requires ds.time, ds.lon, and ds.lat.
+
+    Requires one of 3 libraries: pvlib, PyEphem, or astral.
+
+    Parameters
+    ----------
+    ds : xr.DataArray or xr.Dataset
+        Dataset with variables ds.time, ds.lon, and ds.lat.
+    method : str, optional
```
Zeitsperre: The `optional` here is from the C/C++ convention, in that a Null/None is permitted. This is allowed:
```suggestion
    method : {"pvlib", "astral", "ephem"}
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+        kwargs={"numthreads": 1},
+        input_core_dims=[["time"], [], [], ["time"]],
+        output_core_dims=[["time"], ["time"], ["time"]],
+        # output_dtypes=['datetime64[s]','datetime64[s]','datetime64[s]'],
+        vectorize=True,
+    )
+    return transit.astype("datetime64[s]")
+
+
+def solar_noon(ds, method=default_method):
```
Zeitsperre:
```suggestion
def solar_noon(ds, method: Literal["pvlib", "astral", "ephem"] = default_method):
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+
+def get_dt(freq):
+    """
+    Get the time delta, in seconds for a given pandas frequency.
+
+    Parameters
+    ----------
+        freq : str
+            Pandas time frequency.
```
Zeitsperre:
```suggestion
    freq : str
        Pandas time frequency.
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+    Returns
+    -------
+    float:
```
Zeitsperre:
```suggestion
    float
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+    d_tilprev = sel_with_nans(da_cum, "time", prev_fl - pd.Timedelta(dt, "s"))
+    d_prev = sel_with_nans(da, "time", prev_fl)
+    da_accum = (d_tilcurr + curr_ratio * d_curr) - (d_tilprev + prev_ratio * d_prev)
+
+    return da_accum
+
+
+def interpolate_to_time(ds, var, freq, curr_time):
+    """
+    Interpolate ds to the given time DataArray (such as Solar noon times).
```
Zeitsperre:
```suggestion
    Interpolate Dataset to the given time DataArray (such as Solar noon times).
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+    da_interp = (1 - curr_ratio) * d_curr_fl + curr_ratio * d_curr_cl
+    return da_interp
+
+
+def interpolate_to_solar_noon(da, method="interpolate", solar_method=default_method):
+    """
+    Interpolate (or accumulate) da to solar noon.
+
+    If da is precipitation data, and is accumulated, then the output units will be converted to mm/d.
```
Zeitsperre:
```suggestion
    If DataArray is precipitation data, and is accumulated, then the output units will be converted to mm/d.
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+    ----------
+        da : xr.Dataset or xr.DataArray
+            Data to interpolate.
+        method : str, optional
+            Either `interpolate` at solar noon, or `accumulate` between solar noons.
+            Defaults to 'interpolate'.
+        solar_method : str, optional
+            Python library to use to perform solar noon calculations.
+            Defaults to first available library from ['pvlib', 'astral', 'ephem'].
+
```
Zeitsperre: Over-indented

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+    ----------
+    ds : xr.Dataset or xr.DataArray
+        Dataset with time, lat and lon variables when to calculate solar noon.
+
+    Returns
+    -------
+    xr.DataArray
+        Times when solar noon is expected to occur.
+    """
+    import pvlib
```
Zeitsperre: Same as for `ephem`

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+
+    deltat = xr.DataArray(
+        pvlib.spa.calculate_deltat(ds.time.dt.year, ds.time.dt.month), coords={"time": ds.time}, dims=["time"]
+    )
+    (
+        transit,
+        sunrise,
+        sunset,
```
Zeitsperre: If these aren't being returned, affix an underline the them.

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+def sel_with_nans(da, dim, sel, label="tmp_time", fill=np.nan, lazy=True):
+    """
+    Select from da on dimension dim, with DataArray from the *sorted* da[dim] index.
+
+    This is similar to xc.core.utils.lazy_indexing, but allows for labelled indexing,
+    and fills locations with `fill` if not available.
+    It is similar to xr.reindex, but allows for multi-dimensional reindexing.
+    It is also similar to xr.sel, but allows for lazy evaluation and filling for unavailable selections.
+
+    Note: this can be quite memory intensive. Optimizing chunking may help.
```
Zeitsperre:
```suggestion
    Warnings
    --------
    This function can be quite memory intensive. Optimizing chunking may help.
```

src/xclim/solar.py
```diff
@@ -0,0 +1,458 @@
...
+    ----------
+    da : xr.DataArray
+        DataArray to select from. Requires `dim` dimension.
+    dim : str
+        Dimension over which to select.
+    sel : xr.DataArray
+        DataArray with which to select. Requires `dim` dimension.
+    label : str, optional
+        Label to rename dim in `da`, by default "tmp_time".
+    fill : float, optional
```
Zeitsperre: Is `None` permitted?

pyproject.toml
```diff
@@ -108,7 +108,16 @@ docs = [
...
-extras = ["flox >=0.9", "lmoments3 >=1.0.7", "numbagg >=0.8", "xsdba >=0.4.0", "pymannkendall >=1.4.0"]
+extras = [
+  "flox >=0.9",
+  "lmoments3 >=1.0.7",
+  "numbagg >=0.8",
+  "xsdba >=0.4.0",
+  "pymannkendall >=1.4.0",
+  "astral >=3.2.0",
+  "ephem >=4.2.1",
+  "pvlib >=0.15.2"
```
Zeitsperre: It might be better to add a new "extra" entirely specific for solar.

Zeitsperre (changes requested):
Neat additions!

I think you should rebase on the `dev-v099` branch since we're so close to the release for it. If we do put out a final `v0.x` version, it shouldn't have any brand new modules.

The v1 candidate branch has more subdivided extra dependencies, so it might make sense to add a new one (`xclim[solar]`?).

src/xclim/solar.py
SarahG-579462: I feel there should be a better way of stating directly in the docs that this parameter has a default, than within the text description...

src/xclim/solar.py
SarahG-579462: Not sure what you mean here, why warn that you're importing a library? Feels a bit unnecessary, but I can...

SarahG-579462: These are also within ufuncs, so they would warn on every dask processor if we did so. I'll put the warning in the higher-level function

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        default_method = lib
+        if lib != "pvlib":
+            warnings.warn(f"pvlib library not found, default solar calculations will be performed with {lib}")
+        break
+
+
+def _solar_noon_internal(ds: xr.Dataset | xr.DataArray) -> xr.DataArray:
+    """
+    Approximate solar noon values, using the internal methods available in xclim
+    accuracy is only correct on the order of ±180s, and gets worse in the latter half of the century.
```
aulemahal: Of which century ?

pyproject.toml
```diff
@@ -108,8 +108,18 @@ docs = [
   "sphinxcontrib-bibtex",
   "sphinxcontrib-svg2pdfconverter[Cairosvg]"
 ]
-extras = ["flox >=0.9", "lmoments3 >=1.0.7", "numbagg >=0.8", "xsdba >=0.4.0", "pymannkendall >=1.4.0"]
-all = ["xclim[dev]", "xclim[docs]", "xclim[extras]"]
+solar = [
+  "astral >=3.2.0",
+  "pvlib >=0.15.2"
+]
```
aulemahal, line 114: Both are needed ? Cool that we support multiple ones, but is there one that is better ? Could we default to that, not needing to install the other option ?

src/xclim/solar.py
aulemahal, line 1: Should we reunite all "solar" helpers in the same submodule ?  I think it would make sense to put the following methods of `xclim.indices.helpers` here to:
```
cosine_of_solar_zenith_angle
day_angle
day_lengths
distance_from_sun
eccentricity_correction_factor
extraterrestrial_solar_radiation
solar_declination
time_correction_for_solar_angle
```
And what about `xclim.compute.solar` ? With the renaming, `indices -> compute`, that submodule gains a larger scope.

If you really need to merge this before v1, I see no issue in making this refactor as a second PR.

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        day datetime to calculate solar noon at.
+    lon : np.ndarray[float]
+        longitudes to calculate solar noon for.
+
+    Returns
+    -------
+    np.ndarray[float]
+        time of solar noon, as a fraction of hour away from the date.
+    """
+    import astral.sun
```
aulemahal: Importing within the function wrapper in `apply_ufunc` would trigger the import on each dask worker after the compute is started, wouldn't it ?

Not sure about what I'm suggesting here, but what about the following pattern:
```
default_method = 'internal'
pvlib = None
astral = None
if importlib.util.find_spec('pvlib'):
    default_method = 'pvlib'
    import pvlib
if importlib.util.find_spec('astral'):
    import astral
    if pvlib is not None:
        default_method = 'astral'
         warnings.warn(f"pvlib library not found, default solar calculations will be performed with astral")
```
?

src/xclim/solar.py
aulemahal: I think that would mean moving the warning about external libraries to the `solar_noon` function instead of upon import.

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        do_calc = _solar_noon_astral
+    elif method == "internal":
+        do_calc = _solar_noon_internal
+    else:
+        errmsg = f"Method does not exist: {method}"
+        raise ValueError(errmsg)
+    return do_calc(ds)
+
+
+def get_dt(freq: str):
```
aulemahal:
```suggestion
def _get_dt(freq: str):
```
Unless I am mistaken, this function is only valid for `freq <= 'D'` in the way it is used in this module ?

I suggest making it private and clarifying the constrain in the docstring, maybe even raising on invalid frequencies ?

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+    Returns
+    -------
+    float
+        Total seconds between two timestamps with this frequency.
+    """
+    return pd.date_range(freq=freq, periods=2, start="2000-01-01").diff()[1].total_seconds()
+
+
+def accumulate_between_times(
+    ds: xr.Dataset, var: str, freq: str, prev_time: xr.DataArray, curr_time: xr.DataArray
```
aulemahal:
```suggestion
    da: xr.DataArray, freq: str = None, prev_time: xr.DataArray, curr_time: xr.DataArray
```
Is there a reason to work on a dataset but return a dataarray ? Elsewhere in xclim, we usually preserve the type, so here it would mean taking an DataArray in, no ?

Can we default to inferring `freq` with `xr.infer_freq` ?

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+
+    Returns
+    -------
+    float
+        Total seconds between two timestamps with this frequency.
+    """
+    return pd.date_range(freq=freq, periods=2, start="2000-01-01").diff()[1].total_seconds()
+
+
+def accumulate_between_times(
```
aulemahal: This feels out of place. `xclim.compute.helpers` maybe ?

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        Defaults to first available library from ['pvlib', 'astral', 'internal'].
+
+    Returns
+    -------
+        xr.Dataset or xr.DataArray:
+            Data interpolated/accumulated to solar noon.
+    """
+    if isinstance(da, xr.DataArray):
+        ds = da.to_dataset()
+        var = da.name
```
aulemahal:
```suggestion
        ds = da._to_temp_dataset()
        var = da.name
```
Otherwise this would fail on a unnamed array.

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+    elif isinstance(da, xr.Dataset):
+        output = []
+        for var in da.data_vars:
+            output.append(interpolate_to_solar_noon(da[var], method=method, solar_method=solar_method))
+        ds_out = xr.merge(output)
+        ds_out.attrs = da.attrs.copy()
```
aulemahal: What about
```suggestion
        ds_out = da.map(interpolate_to_solar_noon, method=method, solar_method=solar_method)
```
?

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        raise NotImplementedError("interpolate_to_solar_noon is not implemented for DataTrees")
+    else:
+        raise ValueError("da must be an instance of xr.DataArray or xr.Dataset.")
+
+    if method not in ["accumulate", "interpolate"]:
+        raise ValueError("Unknown method passed to interpolate_to_solar_noon, need one of [accumulate, interpolate]")
+
+    if method == "accumulate" and var == "pr":
+        ds["pr"] = xc.units.convert_units_to(ds["pr"], "mm/hr")
+
```
aulemahal: This is not yours to do I believe.
```suggestion

```

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+    # set metadata
+    # if precip, set proper units.
+    da_out.attrs = ds[var].attrs.copy()
+    if method == "interpolate":
+        da_out.attrs["cell_methods"] = "time: point (comment: at solar noon)"
+        da_out.attrs["comments"] = f"Interpolated date to solar noons, calculated with {default_method} library."
+    elif method == "accumulate":
+        if var == "pr":
+            da_out.attrs["units"] = "mm/d"
```
aulemahal:
```suggestion
        da_out = to_agg_units(da_out, da, "sum")
```

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+        da_out.attrs["cell_methods"] = "time: sum (interval: 24 hr comment: since solar noon)"
+        da_out.attrs["comments"] = f"Accumulated between solar noons, calculated with {default_method} library."
+    da_out.name = var
```
aulemahal: Is it not preserved ?

src/xclim/solar.py
```diff
@@ -0,0 +1,377 @@
...
+    do_calc = interpolate_to_time
+    if method == "accumulate":
+        do_calc = accumulate_between_times
+        kwargs["prev_time"] = ds_days.noon_yst
+    da_out = do_calc(ds, **kwargs)
+    da_out = da_out.assign_coords(noon=ds_days.noon)
+
+    # set metadata
+    # if precip, set proper units.
+    da_out.attrs = ds[var].attrs.copy()
```
aulemahal: I would think that `interpolate_to_time` and `accumulate_between_times` would preserve attributes ?

aulemahal (commented):
First pass before leaving for two weeks so I don't need to explain it any further.

src/xclim/solar.py
SarahG-579462: Yes, that's what I had initially but Trevor suggested the current approach instead. I'll change it back unless @Zeitsperre you argue against it?

src/xclim/solar.py
SarahG-579462: Doesn't seem so on my end, probably due to multiplying by the ratio DataArray which doesn't have metadata nor a name.

src/xclim/solar.py
SarahG-579462: They don't now, I'll ensure they do.

src/xclim/solar.py
coxipi:
```python
   out = func_that_kills_your_attrs(ds)
```

pyproject.toml
SarahG-579462: astral is faster, because it works in a vectorized fashion, but pvlib is more accurate. pvlib is also much more general than what I would normally install by default, so I'd err towards using astral by default.

src/xclim/solar.py
SarahG-579462: It would probably work for 'W' as well, but done!

Zeitsperre (commented):
This is looking really good. I'm going to restate though that this should be added to the first `v1` release, given that it adds a bunch of new features and refactors several existing functions into a new (albeit much more sensible) location.

Is there a delivery date involved here or are things OK to stay as a PR for the next month or two?

SarahG-579462:
> This is looking really good. I'm going to restate though that this should be added to the first `v1` release, given that it adds a bunch of new features and refactors several existing functions into a new (albeit much more sensible) location.
> 
> Is there a delivery date involved here or are things OK to stay as a PR for the next month or two?

This is fine, I'll just point towards this branch and ask them to install locally.

src/xclim/indices/solar.py
```diff
@@ -0,0 +1,1194 @@
+"""The solar module offers functions for interpolating and accumulating variables to solar noon."""
```
aulemahal, line 1:
```suggestion
"""The solar module offers helper functions for astronomical calculations related to the sun, its position in the sky, its revolution and to the length of day."""
```

src/xclim/indices/solar.py
```diff
@@ -0,0 +1,1194 @@
...
+else:
+    XR2409 = False
+
+try:
+    from flox.xarray import rechunk_for_blockwise
+
+    flox_err = None
+except ImportError:
+    rechunk_for_blockwise = None
+
```
aulemahal, line 27:
```suggestion
```
I believe this was copied from the helpers, but is not actually needed ? I think it's for `resample_map` which still resides in the other file.

aulemahal (approved):

PR #2398, opened 2026-08-20
Add Canadian Hardiness Zones

### LINKED ISSUE #1290: Add Hardiness Zones and generic functions for calculating climatologies

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #1290

* Adds the `canadian_hardiness_zones` compute function.
* Update the `hardiness_zones` function to add the `flag_values` and `flag_meanings` for zones.

No. This adds more attribute information to existing hardiness zones indicators.

### Other information:

The changes here were adapted from previous work in the now-deleted `plant-hardiness-zones` branch.

The zones follow the `flag` CF-Conventions (https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#flags); This only satisfies a portion of the CF-Conventions pertaining to variable attributes, which is fine/an improvement.

### Questions

This indicator is one of those that should _technically_ be returning a categorization, rather than an index value as the compute function is currently coded:
- _Would it be more accurate to rename this function to return an `*_index` with the index values?_
  - No.
- _Should the indicator version return the zones based on the index values?_
  - Yes, now returned as flags

There is also the issue that this zoning categorization is meant to be performed on series' of variables spanning at least 30 years. The current tests rely on the `atmosds` (four years of real-world data).
- Should I be adding more tests based on artificial time series data?

### DIFF
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -4111,7 +4111,7 @@
   year          = {1967},
   doi           = {10.4141/cjps67-064},
   url           = {https://doi.org/10.4141/cjps67-064},
-  eprint        = {https://doi.org/10.4141/cjps67-064},
+  eprint        = {https://doi.org/10.4141/cjps67-064}
 }
 @article{ouellet_hardiness_1967c,
   author        = {Ouellet, C. E. and Sherk, L. C.},
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -4112,23 +4112,6 @@
   doi           = {10.4141/cjps67-064},
   url           = {https://doi.org/10.4141/cjps67-064},
   eprint        = {https://doi.org/10.4141/cjps67-064},
-  abstract      = {
-    A method has been developed to estimate the relative suitability index for
-    the winter survival of ornamental trees and shrubs at any locality in
-    Canada from which average climatic data are available. It was based
-    essentially on hardiness data obtained from 108 stations for 174 species
-    and cultivars. Of nine regression equations studied, one gave estimates of
-    suitability indices in best agreement with field experience. This equation
-    was developed with 40 actual indices chosen for their geographic
-    representation. These actual indices are based on the collected hardiness
-    data and plant hardiness indices previously determined. Seven
-    meteorological variables, that is, low winter temperatures, frost-free
-    period, summer and winter rainfall, summer high temperatures, snow depth,
-    and wind speed were involved in this equation. The multiple correlation
-    coefficient obtained was 0.98, and the coefficient of variation of the
-    indices estimated from regression equalled 9.6 percent of the mean, which
-    was 49.
-  }
 }
 @article{ouellet_hardiness_1967c,
   author        = {Ouellet, C. E. and Sherk, L. C.},
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -4074,3 +4074,55 @@
   year          = {2022},
   publisher     = {Nature Publishing Group UK London}
 }
+@article{ouellet_hardiness_1967b,
+  author        = {Ouellet, C. E. and Sherk, L. C.},
+  title         = {WOODY ORNAMENTAL PLANT ZONATION: II. SUITABILITY INDICES OF LOCALITIES},
+  journal       = {Canadian Journal of Plant Science},
+  volume        = {47},
+  number        = {4},
+  pages         = {339--349},
+  year          = {1967},
+  doi           = {10.4141/cjps67-064},
+  url           = {https://doi.org/10.4141/cjps67-064},
+  eprint        = {https://doi.org/10.4141/cjps67-064},
+  abstract      = {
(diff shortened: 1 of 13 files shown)


docs/references.bib
```diff
@@ -4101,3 +4101,97 @@ @article{zomer_2022
...
+    was developed with 40 actual indices chosen for their geographic
+    representation. These actual indices are based on the collected hardiness
+    data and plant hardiness indices previously determined. Seven
+    meteorological variables, that is, low winter temperatures, frost-free
+    period, summer and winter rainfall, summer high temperatures, snow depth,
+    and wind speed were involved in this equation. The multiple correlation
+    coefficient obtained was 0.98, and the coefficient of variation of the
+    indices estimated from regression equalled 9.6 percent of the mean, which
+    was 49.
+  }
```
aulemahal:
```suggestion
```
To keep this file reasonable, should we remove the abstracts ?

src/xclim/compute/_agro.py
```diff
@@ -1593,6 +1594,113 @@ def chill_units(tas: xarray.DataArray, positive_only: bool = False, freq: Freq =
     return cu.resample(time=freq).sum().assign_attrs(units="")
 
 
+# TODO: declare_units()
+def canadian_hardiness_zones(
```
aulemahal, line 1602: For the review, I went back in the references you added and I think I was wrong yesterday. It seems indeed that the statistic over the 30 years (`resample(time=freq)`) is done independently for each variable but includes interannual variability. I mean, for example, that `X1` might be: 
```
x1 = statistics(_tasmin, statistic="mean", freq="MS").resample(time=freq).min()
```
i.e. the mean of daily minimum temperature of the coldest month _of the whole period_, and not as I suggested, the average over the 30 coldest months of each year.

However, I feel like the articles aren't really clear about that. Maybe we could write to someone? Daniel McKenney ?

src/xclim/compute/_agro.py
```diff
@@ -1593,6 +1594,113 @@ def chill_units(tas: xarray.DataArray, positive_only: bool = False, freq: Freq =
...
+    ----------
+    tasmin : xarray.DataArray
+        Minimum daily temperature.
+    tasmax : xarray.DataArray
+        Maximum daily temperature.
+    pr : xarray.DataArray
+        Precipitation.
+    snd : xarray.DataArray
+        Snow depth.
+    sfcWindmax : xarray.DataArray
```
aulemahal: The 2025 article talks about "maximum wind gust". Wind gust is CMIP's `wsg` and "maximum wind gust" is  `wsgsmax` in cordex, or `wsgmax10m` in cmip6.

docs/references.bib
Zeitsperre: They're not available in the generated bibliographies, so yes. People can search them out via DOI.

src/xclim/compute/_agro.py
Zeitsperre: I've read from a few different sources but I'm with you that it isn't 100% clear in `xarray`-like terms how this should be calculated. I'll write to his team tomorrow about this.

src/xclim/indicators/atmos/_multivariate.py
```diff
@@ -0,0 +1,36 @@
...
+canadian_hardiness_zones = MultivariateIndicator(
+    keywords="temperature precipitation speed",
+    title="Canadian hardiness zones",
+    identifier="canadian_hardiness_zones",
+    standard_name="",
+    units="",
+    long_name="Canadian hardiness zones",
+    description="A climate index based on a {freq} climatology of the annual average of maximum and minimum "
+    "monthly temperatures, seasonal precipitation, average annual maximum snow depth, and maximum wind gust "
+    "experienced over the entire period. Developed specifically to aid in determining plant suitability of Canadian "
```
Zeitsperre, line 24: @aulemahal How complicated would it be to support adding scalars to `freq`, e.g. `nYS` -> `n-year`? Very, right?

Zeitsperre:
FYI, this PR is just waiting on some clarifications from the team at NRCan.

PR #2399, opened 2026-08-20
Refactor the Indicator class, move modules to IndicatorCollection

### LINKED ISSUE #2397: [xclim v1] Proposition : Indicator derivation via copying instances - Refactor OOP
An idea I have been considering for a while now. I think the v1 work is the now-or-never moment.

## Problem
The `Indicator` class's internals are complex and spaghetti-adjacent. It works, but does a bit of so-called "magic". I feel like this an obstacle for developers to improve it, even for people inside the Ouranos core team.

## How it is now
Creating an `Indicator` creates a new class for each instance. The `__new__` method is long and complex, but somewhat monolithic in nature (not using inheritance or encapsulation). The idea of a class for each instance was that deriving a new indicator from an existing one would be similar to creating a child inheriting from it.

## Proposition
Here are the main changes my idea contains. Some might not be directly related to the problem above, but are now feasible/logic with the refactor.
- Divide the `Indicator` class into a few parent classes for clearer code:
    + `IndexWrapper` : Parses a compute function into the metadata : `title`, Parameters, outputs, etc. Generates the signature and docstring. Formats the outputs.
    + `TranslatableIndexWrapper` : Adds translation support to the previous.
    + `Indicator` : Allows overriding of metadata from parent `IndexWrapper`, orchestrates the computation.
    + Other subclasses already existing would work as they do now.
- The same way we have a `Parameter` class, create a `Output` one for metadata of the outputs. And then a `TranslatableOutput` for handling translation. `[Translatable]IndexWrapper` classes would use encapsulation.
- Derivation of a new indicator from an existing instance would use a new `Indicator.copy(**new metadata)` method. This function would first do a `deepcopy(self)` and then update the fields as passed in the call. Some fields would not be updatable (`compute` for example)
- Registration of indicators into a central dict would still be useful. The two existing registries would be kept, but `xc.core.indicator.registry` would contain instances, not classes.
- In the yaml module I would change the `base` field so it could accept either the indicator `identifier` or its full python path, avoiding the current ambiguity. The module creator would detect if an instance or a class is referenced and create the new indicator accordingly.

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2397

Finally, changing indicator derivation from subclassing to copying made things more complicated, I abandoned the idea. I did pursue the idea of refactoring `indicator.py` so that it was less of a monolith.

* The `Indicator` class is now the child of a hierarchy of smaller classes.
    - `IndexWrapper` : Wraps a compute function, extracting its metadata. Docstring parsing and generation has been removed from `formatting`, put here.
    - `IndicatorBase` : Handles the fancy subclassing, with modification of parameters, outputs (etc) from user input. It also defines the structure of `__call__` and the related methods subclasses are supposed to use as entry points in the computation.
    - Then a stack of classes incrementing the features of the indicator:
        + `_DatasetIO` : Support for dataset as input and output.
        + `_DataTreeIterator` : Support for DataTree as input and output
        + `_MetadataFormatter` : Format output attributes with our template system, add them to the output(s).
        + `_LocaleMetadataFormatter` : Adds translation support.
        + `_DeprecationWarner ` : Adds possibility to deprecate an indicator.
        + `_InputChecker` : Defines the checks performed on the inputs.
        + `_Compatibility` : Some support for input arguments I broke and here is the support for passing output attributes directly in the constructor.
        + `_Registrer` : Adds the indicator to the registry.
     - `Indicator` : Inherits from all the rest. Defines an `__init__` in the sole goal of having a nice docstring.
     - The other subclasses are the same as before, mostly untouched.
* New `Output` object to store output metadata. It acts as a dictionary of attributes, that also exposes `var_name`, `dimensionality`, `units` and `units_metadata` explicitly. 
* New `xclim.core.collection.IndicatorCollection`, replacing module monkeypatching when creating from Yaml. I aimed to reduce unneeded magic. I feel this opens the door to adding more features to indicator modules in the future.
* All stuff related to `InputKind` now in `xclim.core._types` with other typing related constants.
* New dependency  : [docstring_parser](https://github.com/rr-/docstring_parser) to read numpydoc style docstrings into usable structures, replacing some custom stuff we had. It seems well maintained and used.

Apart from the structural changes listed above:

- The basic `cfcheck` now only looks at the `standard_name`. IMHO, cell methods are almost always broken except on pristine datasets. Warning about this is just noise. Also, it allowed a lot of pruning.
- Renamed two indicator attributes:
    + `cf_attrs` -> `attrs` : Make explicit the fact that not all of this is 100% CF-compliant. This comes with changes that allow the inclusion of more attributes outside the "cf_names" defined earlier.

And finally:
- `xclim.core.indicator.registry` is now a registry of Indicator instances, _not classes_. And keys are the identifiers as-is, no more uppercase weirdness.

### Other information:
~As of writing, this needs a lot of changes to the documentation and tests.~

One goal of moving the "virtual submodules" to an `IndicatorCollection` is is to open possibilities, like adding features, maybe via subclassing ? I'm thinking of the derived variable registry of intake-esm. We'll see.


src/xclim/compute/_hydrology.py
```diff
@@ -903,10 +903,11 @@ def sen_slope(q: xarray.DataArray, freq: Freq = "YS") -> tuple[xarray.DataArray,
 
     Returns
     -------
-    tuple[xarray.DataArray, xarray.DataArray]
-        Returns Sen slope and p-value of in the input
-        ``sen_slope`` : Sen's slope estimates.
-        ``p_value`` : Mann-Kendall metric indicating slope tendency.
+    sen_slope : xarray.DataArray, [dimensionless]
+        Sen's slope estimates.
+
```
Zeitsperre:
```suggestion
```

src/xclim/compute/generic.py
Zeitsperre:
```suggestion
condition1: Condition,
condition2: Condition | None,
thresh1: Quantified,
thresh2: Quantified | None,
```

src/xclim/core/_types.py
```diff
@@ -125,3 +131,225 @@
   The method names are keys and values are dicts of keyword arguments to pass
   (an empty dict if there's nothing to configure).
 """
+
+
+class InputKind(IntEnum):
+    """
+    Constants for input parameter kinds.
+
+    For use by external parses to determine what kind of data the indicator expects.
```
Zeitsperre:
```suggestion
    For use by external parsers to determine what kind of data the indicator expects.
```

src/xclim/core/_types.py
```diff
@@ -125,3 +131,225 @@
...
+    On the creation of an indicator, the appropriate constant is stored in
+    :py:attr:`xclim.core.indicator.Indicator.parameters`. The integer value is what gets stored in the output
+    of :py:meth:`xclim.core.indicator.Indicator.json`.
+
+    For developers : for each constant, the docstring specifies the annotation a parameter of a compute function
```
Zeitsperre:
```suggestion
    For developers: For each constant, the docstring specifies the annotation a parameter of a compute function
```

src/xclim/core/collection.py
```diff
@@ -0,0 +1,339 @@
...
+        yield from self.items()
+
+    @classmethod
+    def from_yaml(  # noqa: C90
+        cls,
+        filename: PathLike,
+        name: str | None = None,
+        computes: dict[str, Callable] | ModuleType | PathLike | None = None,
+        translations: dict[str, dict | PathLike] | None = None,
+        mode: str = "raise",
```
Zeitsperre:
```suggestion
        mode: Literal["raise", "warn", "ignore"] = "raise",
```

src/xclim/core/collection.py
```diff
@@ -0,0 +1,339 @@
...
+        translations : Mapping of dicts or path, optional
+            Translated metadata for the new indicators. Keys of the mapping must be two-character language tags.
+            Values can be translations dictionaries as defined in :ref:`internationalization:Internationalization`.
+            They can also be a path to a JSON file defining the translations.
+        mode : {'raise', 'warn', 'ignore'}
+            How to deal with broken indicator definitions.
+        encoding : str
+            The encoding used to open the `.yaml` and `.json` files.
+            It defaults to UTF-8, overriding python's mechanism which is machine dependent.
+        validate : bool or path
```
Zeitsperre:
```suggestion
        validate : bool or PathLike
```

src/xclim/core/collection.py
```diff
@@ -0,0 +1,339 @@
...
+            - `example.py` : defining a few compute functions.
+            - `example.fr.json` : French translations
+        """
+        filepath = Path(filename)
+
+        if not filepath.suffix:
+            # A stem was passed, try to load files
+            yml_path = filepath.with_suffix(".yml")
+        else:
+            yml_path = filepath
```
Zeitsperre: I would instead check that the suffix is either `yml` or `yaml`. Some ~sociopaths~ individuals add periods into filenames.

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+        if "choices" in self:  # A set is stored, convert to list
+            out["choices"] = list(self.choices)
+        if self.default is not _empty_default:
+            out["default"] = self.default
+        for field in ["annotation", "compute_name", "units"]:
+            if field in self:
+                out[field] = str(getattr(self, field))
+        return out
+
+    def _gen_doc(self, name: str = None) -> str:
```
Zeitsperre:
```suggestion
    def _gen_doc(self, name: str | None = None) -> str:
```

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+        """
+        Generate an item of a numpydoc style Parameters section for this parameter.
+
+        Parameters
+        ----------
+        name: str, optional
```
Zeitsperre:
```suggestion
        name : str, optional
```

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+            default = f" Default: {self.default!r}."
+        elif self.kind == InputKind.KWARGS:
+            default = ""
+        unit = ""
+        if self.units is not _empty:
+            unit = f" [Required units : {self.units}]"
+
+        return f"{name or self.compute_name} : {annot}\n {self.description}{default}{unit}"
+
+    def _gen_signature(self, name: str = None) -> inspect.Parameter:
```
Zeitsperre:
```suggestion
    def _gen_signature(self, name: str | None = None) -> inspect.Parameter:
```

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+
+    def __init__(self, var_name=None, dimensionality=None, units=None, units_metadata=None, **kwargs):
+        """
+        Create an output attributes dictionary.
+
+        Parameters
+        ----------
+        var_name : str, optional
+            The name of the output variable.
+        dimensionality, str, optional
```
Zeitsperre:
```suggestion
        dimensionality : str, optional
```

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+    dimensionality: str | None
+    """Dimensionality specification, similar but not necessarily compatible with pint."""
+
+    units: str | None
+    """Units of the output."""
+
+    units_metadata: str | None
+    """Additional CF metadata for the units."""
+
+    def __init__(self, var_name=None, dimensionality=None, units=None, units_metadata=None, **kwargs):
```
Zeitsperre:
```suggestion
    def __init__(self, var_name: str | None = None, dimensionality: str | None = None, units: str | None = None, units_metadata: str | None = None, **kwargs):
```

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
+            name = f"{self.var_name} : "
+        dim = ""
+        if self.units:
+            dim = f", [{self.units}]"
+        elif self.dimensionality:
+            dim = f", {self.dimensionality}"
+        sname = "" if self.get("standard_name") is None else f"{self['standard_name']}, "
+        add = ""
+        if other := (set(self.keys()) - {"standard_name", "long_name"}):
+            add = f". With additional attributes: {', '.join([f'**{k}**: ``{self[k]}``' for k in other])}"
```
Zeitsperre, line 329: `else: add = "."`?

src/xclim/core/indicator.py
```diff
@@ -255,392 +141,567 @@ def __contains__(self, key) -> bool:
...
         if "compute" in kwds:
-            # Parsed parameters and metadata override parent's params entirely.
-            parameters, docmeta = cls._parse_compute(kwds["compute"], kwds.get("parameters", {}))
-            for name, value in docmeta.items():
-                # title, abstract, references, notes, long_name
-                kwds.setdefault(name, value)
+            if cls.compute is not IndexWrapper.compute:
+                raise TypeError(
+                    "Can't change the compute function of an indicator, create from a base class instead."
+                    f" Indicator {identifier} got {kwds['compute']} but already has {cls.compute}"
```
Zeitsperre:
```suggestion
                    "Can't change the compute function of an indicator, create from a base class instead. "
                    f"Indicator {identifier} got {kwds['compute']} but already has {cls.compute}."
```

src/xclim/core/indicator.py
```diff
@@ -899,74 +852,35 @@ def __call__(self, *args, **kwds):
...
-        ba = self.__signature__.bind(*args, **kwds)
-        ba.apply_defaults()
-
-        # Assign inputs passed as strings from ds.
-        self._assign_named_args(ba)
-
         # Extract variables + inject injected
-        das = OrderedDict()
-        params = ba.arguments.copy()
+        das = dict()
```
Zeitsperre:
```suggestion
        das = {}
```

src/xclim/core/indicator.py
```diff
@@ -975,269 +889,342 @@ def _parse_variables_from_call(self, args, kwds) -> tuple[OrderedDict, OrderedDi
...
+        list
+            Same as `outs`.
+        dict
+            Same as `meta`, potentially modified.
+        """
+        return outs, meta
 
-        In use case #1, we have two compute functions with arguments in a different order:
-            `func1(tasmin, tasmax)` and `func2(tasmax, tasmin)`
+    def _finalize(self, outs: list[DataArray], das: dict[str, DataArray], params: dict[str, Any], meta: dict[str, Any]):
```
Zeitsperre:
```suggestion
    def _finalize(self, outs: list[DataArray], das: dict[str, DataArray], params: dict[str, Any], meta: dict[str, Any]) -> Any:
```

src/xclim/core/indicator.py
```diff
@@ -975,269 +889,342 @@ def _parse_variables_from_call(self, args, kwds) -> tuple[OrderedDict, OrderedDi
...
         ----------
-        locale : str or sequence of str
-            The POSIX name of the locale or a tuple of a locale name and a path to a json file defining translations.
-            See :py:mod:`xclim.locale` for details.
-        fill_missing : bool
-            If True (default) fill the missing attributes by their english values.
+        attrs : Output or dict[str, str]
+            The attributes to format and update. All will be formatted except `units` or `units_metadata`,
+            which were already handled at computation time.
+        fmtargs: dict[str, Any]
```
Zeitsperre:
```suggestion
        fmtargs : dict[str, Any]
```

src/xclim/core/indicator.py
```diff
@@ -1248,113 +1235,120 @@ def json(cls, args: dict | None = None) -> dict:
...
-                mba[k] = v
-        out = {}
-        for key, val in attrs.items():
-            if callable(val):
-                val = val(**mba)
-
-            out[key] = formatter.format(val, **mba)
+                vv = self._version_deprecated
+            warnings.warn(
+                f"`{self.title}` is deprecated as of `xclim` v{vv} and will be removed in a future release."
```
Zeitsperre:
```suggestion
                f"`{self.title}` is deprecated as of `xclim` v{vv} and will be removed in a future release. "
```

src/xclim/core/indicator.py
```diff
@@ -1248,113 +1235,120 @@ def json(cls, args: dict | None = None) -> dict:
...
+                f"{alternative} See the `xclim` release notes for more information: "
+                "https://xclim.readthedocs.io/en/stable/history.html",
```
Zeitsperre:
```suggestion
                "https://xclim.readthedocs.io/en/stable/changelog.html",
```

src/xclim/core/indicator.py
```diff
@@ -1411,79 +1403,190 @@ def datacheck(self, **das) -> None:
...
+                    atts[name] = val
+        kwargs["attrs"] = attrs
+
+        module = kwargs.get("module", cls.__module__)
+        # Infer realm for built-in xclim instances, handle module
+        xclim_realm = None
+        # If the module is xclim.indicators.XYZ, we assume an "official" indicator, realm is XYZ, module set
+        # otherwise everything is up to the caller
+        if module.startswith("xclim.indicators."):
+            xclim_realm = module.split(".")[2]
```
Zeitsperre, line 1466: Is there any possibility of nested submodules? If so, `[-1]`?

src/xclim/core/indicator.py
```diff
@@ -1411,79 +1403,190 @@ def datacheck(self, **das) -> None:
...
+    """Register the indicator in the xclim registry."""
 
-    @property
-    def parameters(self) -> dict:
-        """
-        Create a dictionary of controllable parameters.
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.identifier in registry:
+            warnings.warn(f"Indicator {self.identifier} already exists and will be overwritten.", stacklevel=1)
```
Zeitsperre: Quick question: What `stacklevel` should we be using? I imagine it depends largely on the degree of nestedness of code portions. I tend to use `2` by default for library commands.

src/xclim/core/indicator.py
```diff
@@ -1411,79 +1403,190 @@ def datacheck(self, **das) -> None:
...
+        input : dict, optional
+            Mapping from input variable name in the compute function to known variable name.
+            Useful for transforming generic compute functions into variable-specific indicator.
+            The new variables names must be defined in :py:data:`xclim.core.VARIABLES`.
+        parameters: dict, optional
+            Overrides for the parameters. Either value to "inject", removing that parameter from the call signature,
+            or dictionaries of properties to override the ones parsed from the docstring.
+            See :py:class:`~xclim.core.indicator.Parameter` for valid properties. Additionally,
+            `name` can be passed to change the name of the argument in the call signature.
+        attrs : list of dicts
```
Zeitsperre:
```suggestion
        attrs : list of dict
```

src/xclim/core/indicator.py
```diff
@@ -1411,79 +1403,190 @@ def datacheck(self, **das) -> None:
...
+            Attributes to be formatted and added to the computation's output.
+            Any attribute are accepted, but `var_name` is required for multi-output indicators.
+            The list must be the same length as the number of outputs of the compute function.
+        context : str
+            A `pint` unit context enabled during the computation of this indicator.
+            For example use 'hydro' to allow conversion from 'kg m-2 s-1' to 'mm/day' for all inputs an outputs.
+        src_freq : str, sequence of strings, optional
```
Zeitsperre:
```suggestion
        src_freq : str or sequence of str, optional
```

src/xclim/core/indicator.py
```diff
@@ -1594,23 +1685,14 @@ class ResamplingIndicator(CheckMissingIndicator):  # numpydoc ignore=PR02
...
-        Arguments to pass to the `missing` function.
-        If None, this will be determined by the global configuration.
-    allowed_periods : Sequence[str], optional
-        A list of allowed periods, i.e. base parts of the `freq` parameter.
-        For example, indicators meant to be computed annually only will have `allowed_periods=["Y"]`.
-        `None` means "any period" or that the indicator doesn't take a `freq` argument.
     """
 
-    allowed_periods: list[str] | None = None
+    allowed_periods: list[str] = None
```
Zeitsperre, line 1699: Why not `allowed_periods: list[str]`?

src/xclim/core/locales.py
```diff
@@ -221,13 +223,13 @@ def get_local_formatter(
...
     """
-    _, loc_dict = get_local_dict(locale)
+    loc_name, loc_dict = get_local_dict(locale)
     if "attrs_mapping" in loc_dict:
         attrs_mapping = loc_dict["attrs_mapping"].copy()
         mods = attrs_mapping.pop("modifiers")
         return AttrFormatter(attrs_mapping, mods)
 
-    warnings.warn("No `attrs_mapping` entry found for locale {loc_name}, using default (english) formatter.")
+    warnings.warn(f"No `attrs_mapping` entry found for locale {loc_name}, using default (english) formatter.")
```
Zeitsperre:
```suggestion
    warnings.warn(f"No `attrs_mapping` entry found for locale {loc_name}, using default (English) formatter.")
```

src/xclim/indicators/atmos/_precip.py
```diff
@@ -1021,7 +992,7 @@ class DailyPrecipNoResample(Indicator):
         "The start of the season is on the last day of the first sequence i) and must be between {date_min_start} and "
         "{date_max_start}.",
         "Last day in a dry sequence after the start of the season, i.e. a sequence of {window_dry_end} days "
-        "with at least {thresh_dry_end} {method_dry_end}. It must be between {date_min_end} and {date_max_end}. ",
+        "with at least {thresh_dry_end} {method_dry_end}. It must be between {date_min_end} and {date_max_end}.",
```
Zeitsperre, line 996: Why remove the space here? There's another sentence that follows.

tests/test_formatting.py
```diff
@@ -23,26 +25,15 @@ def test_prefix_attrs():
...
-    assert doc[7] == "Keywords : temperature health,."
-    assert doc[13] == "    Default : `ds.tasmin`. [Required units : [temperature]]"
-    assert (
-        doc[45]
-        == "    Total number of series of at least {window} consecutive days with daily minimum temperature above "
-        "{thresh_tasmin} and daily maximum temperature above {thresh_tasmax}, "
-        "with additional attributes: **description**: {freq} number of heat wave events within a given period. "
-        "A heat wave occurs when daily minimum and maximum temperatures exceed {thresh_tasmin} and {thresh_tasmax}, "
-        "respectively, over at least {window} days."
-    )
```
Zeitsperre, line 42: Good riddance. Hated updating this test.

tests/test_indicators.py
```diff
@@ -143,25 +134,21 @@ def multioptvar_compute(
...
 
 def test_attrs(tas_series):
-    import datetime as dt
 
     a = tas_series(np.arange(360.0))
     txm = uniIndTemp(a, thresh="5 degC", freq="YS")
     assert txm.cell_methods == "time: mean time: mean within years"
-    assert f"{dt.datetime.now():%Y-%m-%d %H}" in txm.attrs["history"]
-    assert "TMIN(da=tas, thresh='5 degC', freq='YS') with options check_missing=any" in txm.attrs["history"]
-    assert f"xclim version: {__version__}" in txm.attrs["history"]
```
Zeitsperre, line 161: We no longer record the version in the metadata?

tests/test_indicators.py
```diff
@@ -648,55 +600,21 @@ def test_parsed_doc():
...
+    assert (
+        doc[6] == "With injected parameters: window_statistic=min, statistic=count, min_gap=1, constrain=('>', '>=')."
+    )
+    assert doc[11] == " Minimum surface temperature. Default: 'tasmin'. [Required units : [temperature]]"
+    assert doc[34] == (
+        "  Total number of series of at least {window} consecutive days with daily minimum temperature above "
+        "{thresh_tasmin} and daily maximum temperature above {thresh_tasmax}. With additional attributes: "
+        "**description**: ``{freq} number of heat wave events within a given period. A heat wave occurs when daily "
+        "minimum and maximum temperatures exceed {thresh_tasmin} and {thresh_tasmax}, respectively, over at least "
+        "{window} days.``"
```
Zeitsperre, line 616: Oh no, the test is just here now. 🥲

Zeitsperre (approved):
Enormous amount of work here but so much cleaner. This is great, IMO. Well done @aulemahal.

src/xclim/core/indicator.py
aulemahal: Nope! The realm is always the third word. In fact, the "private" submodules are in the last position.

For example, `tg_mean` has module `xclim.indicators.atmos._temperature`, and we want `atmos` here.

src/xclim/core/indicator.py
aulemahal: Because it needed a "default " value. With the way we handle class creation in the `__new__`, putting a value in the class declaration acts the same as a default value in the constructor.  If it wasn't set here, we'd need to set it explicitly somewhere.

Normal class construction would use an override of `__init__` for this, and we could implement a `__new__`, but it seemed verbose too me.

src/xclim/indicators/atmos/_precip.py
aulemahal: The next sentence is another element of the list, the description of the third output!

The extra space was messing the `_gen_docstring` function, because the values of attributes are now wrapped in double backticks in the generated docstrings. Look how it is rendered here : https://xclim--2399.org.readthedocs.build/en/2399/api_indicators.html#xclim.indicators.atmos.rain_season

tests/test_indicators.py
aulemahal: We no longer add a `history` attribute to DataArray outputs, only to datasets. The CF conventions only define `history` on the global level. 

This reminds me I wanted to suggest to flip the default to `as_dataset=True` ! Because indeed, the current default no longer adds the version.

src/xclim/core/indicator.py
Zeitsperre: We could simply set that default to `[]` (falsy)? No need to handle `None` in that case.

src/xclim/core/collection.py
aulemahal: As the code is using `Path.with_suffix` to generate file names with other extensions, passing a period in a filename is simply not compatible with that logic. My last commit adds a note in the doc and small changes in the code, but for now these _individuals_ will need to open a PR to be able to pursue their sociopathic behaviour in xclim.

src/xclim/core/indicator.py
aulemahal: After a few tests, it seems `stacklevel=4` gives the best information for a normal indicator. See:

```python3
import xclim as xc

xc.atmos.tg_mean.__class__(identifier=xc.atmos.tg_mean.identifier, parameters={'freq': 'MS'})
<ipython-input-5-66a66a693078>:1: UserWarning: Indicator tg_mean already exists and will be overwritten.
  xc.atmos.tg_mean.__class__(identifier=xc.atmos.tg_mean.identifier, parameters={'freq': 'MS'})
```

PR #2405, opened 2026-08-26
Resample not mandatory in generic functions

coxipi:
### Pull Request Checklist:
    - This PR fixes #xyz

* Allows to pass `freq=None` to signify no resampling should occur.

No


src/xclim/compute/run_length.py
```diff
@@ -428,10 +425,7 @@ def windowed_run_events(
         else:
             d = rle(da, dim=dim, index=index)
             d = xr.where(d >= window, 1, 0)
-        if freq is not None:
-            d = d.resample({dim: freq})
-        out = d.sum(dim=dim)
-
+        out = resample_map(d, dim, freq, lambda da: da.sum(dim=dim))
```
coxipi: Could I just use `xr.DataArray.sum` everywhere and it's OK, no danger with Datasets? 

instead of my lambda function

coxipi:
not sure if I should change the CHANGELOG in this case

added a few tests, tell me if it seems more are needed

src/xclim/compute/generic.py
```diff
@@ -521,10 +525,9 @@ def count_percentile_occurrences(
     data = select_time(data, **indexer)
 
     @percentile_bootstrap
-    def _count_percentile_occurrences(data, per, freq, bootstrap, op):
+    def _count_percentile_occurrences(data, per, freq, bootstrap, condition):
         thresh = resample_doy(clim, data)
-        cond = compare(data, op, thresh, constrain)
-        out = cond.resample(time=freq).sum()
+        out = thresholded_statistics(data, condition, thresh, "sum", freq, constrain)
```
coxipi: oups, need to revert this, was doubly buggy

src/xclim/compute/helpers.py
```diff
@@ -1149,8 +1151,9 @@ def resample_map(
         Dimension over which to resample.
     freq : str
         Resampling frequency along `dim`.
-    func : callable
-        Function to map on each resampled group.
+    func : callable or str
+        Function to map on each resampled group. If a string is given, it is converted to a callable from
+        a given registry.
```
Zeitsperre:
```suggestion
        Function to map on each resampled group. 
        If str, it is converted to a callable from a given registry.
```

src/xclim/compute/run_length.py
```diff
@@ -1902,6 +1891,8 @@ def find_events(
     ds = xr.Dataset({"da_start": condition, "da_stop": condition_stop})
     if data is not None:
         ds = ds.assign(data=data)
-    return ds.resample(time=freq).map(
-        lambda grp: _find_events(grp.da_start, grp.da_stop, grp.get("data", None), window, window_stop)
-    )
+
+    def _func(grp):
+        return _find_events(grp.da_start, grp.da_stop, grp.get("data", None), window, window_stop)
```
Zeitsperre: Should this be passing `window` and `window_stop` as well?

Zeitsperre (approved):
LGTM, but I'd wait on Pascal for validation on the algorithm changes,

Zeitsperre:
> not sure if I should change the CHANGELOG in this case

It's almost always a good idea to add to the CHANGELOG.

coxipi:
Opted for the uniform notation `per` for percentile since it was already used in many places. If we ever want to change this, I would suggest: `pctl`

src/xclim/compute/run_length.py
coxipi: good point. I did the same in other functions of run_length.py. Also, some check (if freq is None or not) were not necessary anymore with the new resample_map

src/xclim/compute/generic.py
```diff
@@ -95,8 +95,9 @@ def statistics(
         Input data.
     statistic : {"min", "max", "mean", "std", "var", 'count', 'sum', 'integral', 'doymax', 'doymin'} or Callable
         Reducing operation. It can either be a DataArray method or a function that can be applied to a DataArray.
-    freq : str
+    freq : str, optional
         Resampling frequency defining the periods as defined in :ref:`timeseries.resampling`.
+        If None, no resampling is applied.
```
aulemahal, line 100:
```suggestion
        If None, the time dimension is completely reduced.
```
haha, not the same! I think the same sentence is here multiple times.

src/xclim/compute/clix.py
aulemahal, line 1: This module aims to replicate the clix-meta definitions exactly, so I think we need to add a note about renaming the "percentile" arguments to "per" in the top docstring here.

tests/test_generic.py
```diff
@@ -26,6 +26,11 @@ def test_season(self, q_series):
         o = generic.statistics(q, "count", freq="YS-DEC", season="DJF")
         assert o[0] == 31 + 29
 
+    def test_noresampling(self, q_series):
+        q = q_series(np.arange(1000))
+        o = generic.statistics(q, "count", freq=None)
+        assert o == 1000
+
```
aulemahal, line 33:
```suggestion
        assert o == 1000
        assert 'time' not in o.dims

```

CHANGELOG.rst
```diff
@@ -1845,12 +1845,14 @@ New features and enhancements
 * A convenience function (``xclim.core.dataflags.ecad_compliant``) is also offered as a method for asserting that data adheres to all relevant ECAD/ICCLIM checks. For more information on usage, consult the docstring/documentation.
 * A new utility "``dataflags``" is also available for performing fast quality control checks from the command-line (`xclim dataflags --help`). See the CLI documentation page for usage examples.
 * Added missing typed call signatures, expected returns and docstrings for many ``xclim.core.calendar`` functions.
+* Generic functions ``xclim.generic.percentile``, ``xclim.generic.count_occurences`` now accept `freq=None`: This simply skips the resampling part, and performs the reduction on the full time series. This also applies to functions in ``xclim.compute.run_length``.
```
aulemahal, line 1848:
```suggestion
* Generic compute functions now accept `freq=None`: This simply performs the reduction on the full time series instead of resampling. This also applies to functions in ``xclim.compute.run_length``.
```

src/xclim/compute/generic.py
```diff
@@ -115,9 +116,9 @@ def statistics(
     if statistic == "sum" and is_temporal_rate(data):
         statistic = "integral"
     if isinstance(statistic, str):
-        out: xr.DataArray = getattr(data.resample(time=freq), statistic.replace("integral", "sum"))(
-            dim="time", keep_attrs=True
-        )
+        out: xr.DataArray = getattr(
+            data if freq is None else data.resample(time=freq), statistic.replace("integral", "sum")
+        )(dim="time", keep_attrs=True)
```
aulemahal, line 121: Why not use `resample_map` here like in the other cases below?

src/xclim/compute/helpers.py
```diff
@@ -1171,6 +1173,16 @@ def resample_map(
     """
     resample_kwargs = resample_kwargs or {}
     map_kwargs = map_kwargs or {}
+    if isinstance(func, str):
```
aulemahal, line 1176: What do you think about injecting `map_kwargs={'dim': 'time'}` whenever `func` is a string ? It would make the calls to `resample_map` less verbose in the other files ?

I think this would require adding the `dim` argument to the functions in `xclim.compute.reducers`, but that would be better anyway.

aulemahal (commented):
First pass, I'll be sur place la semaine prochaine si on veut en jaser!

src/xclim/compute/clix.py
coxipi: I should just leave things as they are then. It's an oversight. We should maintain compatibility

src/xclim/compute/generic.py
```diff
@@ -114,13 +115,8 @@ def statistics(
         statistic = XCLIM_OPS.get(statistic, statistic)
     if statistic == "sum" and is_temporal_rate(data):
         statistic = "integral"
```
coxipi: Why not do this after, and avoid the replace below?

src/xclim/compute/generic.py
```diff
@@ -114,13 +115,8 @@ def statistics(
...
-    if isinstance(statistic, str):
-        out: xr.DataArray = getattr(data.resample(time=freq), statistic.replace("integral", "sum"))(
-            dim="time", keep_attrs=True
-        )
-    else:
-        with xr.set_options(keep_attrs=True):
-            out: xr.DataArray = resample_map(data, "time", freq, statistic)
+    with xr.set_options(keep_attrs=True):
+        out: xr.DataArray = resample_map(data, "time", freq, statistic.replace("integral", "sum"))
```
coxipi: wrong, fixing it

src/xclim/compute/helpers.py
coxipi: Ok. Ideally we could also have  `dim` in `statistics` so that I can use it with `group.apply` (xsdba) without a weird implementation

coxipi: proposition: have `dim`  in `statistics` with `"time"` default

aulemahal: Good idea!

src/xclim/compute/clix.py
aulemahal: The function's name is still `percentile` with a `percentile` argument, so one of those has to change...

coxipi: ok ok, I added something at top-level

Zeitsperre:
Tests here are failing for precision differences below 10e-5. We probably need to relax some assertions.

aulemahal:
This might be a consequence of dtype switch. I think we should check that we are not breaking dtype preservation. if we had it before.

tests/test_wind.py
```diff
@@ -45,8 +49,8 @@ class TestSfcWindMax:
     def test_sfcWindmax(self, open_dataset, metric):
         with open_dataset(self.test_data) as ds:
             sfcWind, _ = convert.wind_speed_from_vector(ds.uas, ds.vas)
-            sfcWindmax_calculated = getattr(atmos, f"sfcWindmax_{metric}")(sfcWind)
-
             resample = sfcWind.resample(time="YS")
-            c = getattr(resample, metric)()
+            func = getattr(xr.DataArray, metric)
+            c = resample.map(func, dim="time")
+            sfcWindmax_calculated = resample_map(sfcWind, "time", "YS", metric)
```
coxipi: @aulemahal small rounding errors when comparing `da.resample.mean` and `da.resample.map(mean)`, it is what it is. I modified the test to match things with the new behaviour of `statistics` that uses `resample_map`

src/xclim/compute/helpers.py
coxipi: done

src/xclim/compute/generic.py
coxipi: done

src/xclim/compute/generic.py
coxipi: done

coxipi:
@aulemahal  in my last push I have `dim` as an argument on everything that also has `freq`. It's a lot of redundancy, but I feel the logic is the same if we want consistent function calls

src/xclim/compute/generic.py
```diff
@@ -1208,30 +1269,33 @@ def statistics_between_dates(
...
 
     if freq is None:
         frequencies = []
         for bound in [start, end]:
             try:
-                frequencies.append(xr.infer_freq(bound.time))
-            except AttributeError:
+                frequencies.append(xr.infer_freq(bound[dim]))
+            # FIXME:  had to change error type.  xr.infer_freq(string) yields TypeError, not AttributeError. Normal?
+            except TypeError:
```
coxipi, line 1240: had to change this error type, previously AttributeError, now TypeError. Weird?

tests/test_indicators.py
```diff
@@ -616,7 +616,8 @@ def test_indicator_docstring():
         doc[6] == "With injected parameters: window_statistic=min, statistic=count, min_gap=1, constrain=('>', '>=')."
     )
     assert doc[11] == " Minimum surface temperature. Default: 'tasmin'. [Required units : [temperature]]"
-    assert doc[34] == (
+    # FIXME: Changed 34 -> 36. Why does this bug suddenly? is that OK?
+    assert doc[36] == (
```
coxipi: I guess when the new signature of generic functions might have changed this?

src/xclim/compute/generic.py
```diff
@@ -153,9 +154,11 @@ def running_statistics(
         Operation to apply to the rolling window.
     statistic : {"min", "max", "mean", "std", "var", "count", "sum", "integral", "doymax", "doymin"} or Callable
         Reducing operation. Can either be a DataArray method or a function that can be applied to a DataArray.
-    freq : str
+    freq : str, optional
         Resampling frequency defining the periods as defined in :ref:`timeseries.resampling`.
-        Resampling is done after the running statistic.
+        Resampling is done after the running statistic. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling is done after the running statistic. If None, the dimension provided is completely reduced.
```
The same change as in `statistics`.

src/xclim/compute/generic.py
```diff
@@ -260,9 +268,11 @@ def thresholded_running_statistics(
...
```
aulemahal:
```suggestion
        Resampling is done after the running statistic. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
```diff
@@ -489,15 +512,17 @@ def count_percentile_occurrences(
...
         The percentile to compute on the reference period, between 0 and 100.
     condition :  {">", "gt", "<", "lt", ">=", "ge", "<=", "le", "==", "eq", "!=", "ne"}
         Logical comparison operator. Computed as  ``data[i] {condition} climatology[doy(i)]``.
     reference_period : tuple of two dates
         Start and end of the period used to compute the percentiles. Dates should be given as YYYY-MM-DD.
-    freq : str
+    freq : str, optional
         Resampling frequency defining the periods as defined in :ref:`timeseries.resampling`.
-        This function only makes sense with annual frequencies.
+        This function only makes sense with annual frequencies. If None, no resampling is applied.
```
aulemahal:
```suggestion
        This function only makes sense with annual frequencies. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
```diff
@@ -562,15 +588,17 @@ def count_thresholded_percentile_occurrences(
...
         The percentile to compute on the reference period, between 0 and 100.
     condition :  {">", "gt", "<", "lt", ">=", "ge", "<=", "le", "==", "eq", "!=", "ne"}
         Logical comparison operator to find occurrences. Computed as  ``data[i] {condition} climatology[doy(i)]``.
     reference_period : tuple of two dates
         Start and end of the period used to compute the percentiles. Dates should be given as YYYY-MM-DD.
-    freq : str
+    freq : str, optional
         Resampling frequency defining the periods as defined in :ref:`timeseries.resampling`.
-        This function only makes sense with annual frequencies.
+        This function only makes sense with annual frequencies. If None, no resampling is applied.
```
aulemahal:
```suggestion
        This function only makes sense with annual frequencies. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
```diff
@@ -689,8 +721,10 @@ def spell_length_statistics(
         Threshold to test against.
     statistic : {'max', 'sum', 'count'} or sequence of str
         Statistic on the spell lengths. If a list, multiple statistics are computed.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
```diff
@@ -801,8 +837,10 @@ def bivariate_spell_length_statistics(
         Threshold to test against for data2.
     statistic : {'max', 'sum', 'count'} or sequence of str
         Statistic on the spell lengths. If a list, multiple statistics are computed.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
```diff
@@ -875,8 +915,10 @@ def season(
         Minimum number of days that the condition must be met / not met for the start / end of the season.
     aspect : {'start', 'end', 'length'}, or a list of those
         Which season aspect(s) to return. If a list, this function returns a tuple in the same order as this argument.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/generic.py
aulemahal: I think it was an attribute error because `time` didn't exist on `bound`. If you do `getattr(bound, dim)` you should get back to an attribute error. Maybe not the cleanest, but keeps the same behaviour...

src/xclim/compute/run_length.py
```diff
@@ -52,8 +52,8 @@ def use_ufunc(
         Input array.
     dim : str
         The dimension along which to find runs.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -108,8 +108,8 @@ def resample_and_rl(
         Run length function to apply.
     *args : Any
         Positional arguments needed in `compute`.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -295,8 +295,8 @@ def rle_statistics(
         Minimal length of consecutive runs to be included in the statistics.
     dim : str
         Dimension along which to calculate consecutive run; Default: 'time'.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -352,8 +349,8 @@ def longest_run(
         N-dimensional array (boolean).
     dim : str
         Dimension along which to calculate consecutive run; Default: 'time'.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -399,8 +396,8 @@ def windowed_run_events(
         When equal to 1, an optimized version of the algorithm is used.
     dim : str
         Dimension along which to calculate consecutive run (default: 'time').
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -455,8 +449,8 @@ def windowed_run_count(
...
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -508,8 +500,8 @@ def windowed_max_run_sum(
...
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -562,8 +551,8 @@ def _boundary_run(
         When equal to 1, an optimized version of the algorithm is used.
     dim : str
         Dimension along which to calculate consecutive run.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -665,8 +648,8 @@ def first_run(
         When equal to 1, an optimized version of the algorithm is used.
     dim : str
         Dimension along which to calculate consecutive run (default: 'time').
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -715,8 +698,8 @@ def last_run(
...
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

src/xclim/compute/run_length.py
```diff
@@ -818,8 +801,8 @@ def keep_longest_run(da: xr.DataArray, dim: str = "time", freq: Freq | None = No
         Boolean array.
     dim : str
         Dimension along which to check for the longest run.
-    freq : str
-        Resampling frequency.
+    freq : str, optional
+        Resampling frequency. If None, no resampling is applied.
```
aulemahal:
```suggestion
        Resampling frequency. If None, the dimension provided is completely reduced.
```

tests/test_collections.py
```diff
@@ -126,7 +126,7 @@ def test_indicator_module_input_mapping(atmosds):
     prveg = atmosds.pr.rename("prveg").assign_attrs(standard_name="precipitation_flux_onto_canopy")
     with set_options(as_dataset=True):
         out = ex.RX5day_canopy(prveg=prveg)
-    assert "RX5day_canopy(prveg=prveg)" in out.attrs["history"]
+    assert "RX5day_canopy(prveg=prveg, dim='time')" in out.attrs["history"]
```
aulemahal: I'd argue we don't want that ? The `dim` argument shouldn't appear in an indicator created from a generic function ?

tests/test_indicators.py
```diff
@@ -616,7 +616,8 @@ def test_indicator_docstring():
         doc[6] == "With injected parameters: window_statistic=min, statistic=count, min_gap=1, constrain=('>', '>=')."
     )
     assert doc[11] == " Minimum surface temperature. Default: 'tasmin'. [Required units : [temperature]]"
-    assert doc[34] == (
+    # FIXME: Changed 34 -> 36. Why does this bug suddenly? is that OK?
```
aulemahal: New parameter `time` in the docstring inserted two lines. But, see comment above, I'm not sure that's what we want.

src/xclim/compute/helpers.py
```diff
@@ -1171,6 +1173,17 @@ def resample_map(
...
+        # Get function for xclim-implemented statistics
+        func = XCLIM_OPS.get(func, func)
+        func = func if not isinstance(func, str) else getattr(type(obj), func)
```
aulemahal: I'm sorry.

I think we need to rethink this. Look at that:
```
import xarray as xr
ds = open_dataset("ERA5/daily_surface_cancities_1990-1993.nc")
da = ds.tas.chunk()

dr_mean= d.resample(time='YS').mean()

dr_map = d.resample(time='YS').map(xr.DataArray.mean)

len(dr_map.__dask_graph__().keys())
# 22
len(dr_mean.__dask_graph__().keys())
# 9
```

This is because the second has a special pathway for using flox. This seems to be usable only when calling the `mean()` method of the Resample object directly...

I can try my hand at something here.

tests/test_indicators.py
coxipi: reverted

src/xclim/compute/helpers.py
coxipi: `resample_map` now uses built-in functions if possible

src/xclim/compute/generic.py
coxipi: But should I force this to be an AttributeError? All tests pass now with TypeError, things should be fine as-is?

aulemahal: Can't make a suggestion, I don't know why. But rereading this code, I think the best pattern would be to move away from a try-except. The idea here is to take the `dim` coordinate when it exists, which requires `bound` to be a DataArray. 

The previous code was taking `bound.time` which failed with an `AttributeError` in two cases: `bound` is a string and `bound` is a DataArray with no `time` coordinate.

I suggest we do:

```python
if isinstance(bound, DataArray) and dim in bound.coords:
    frequencies.append(xr.infer_freq(bound[dim]))
else:
    frequencies.append(None)
```

coxipi: Agreed for if/else. When should we maintain implicit compatibility with Dataset? this would break this.

aulemahal: The rest of the function (and docstring!) is DataArray-specific here, so no worry.  I would say that in xclim, compute functions are expecting DataArrays. 

The issue in supporting both types is more for general methods, like in  `xclim.core.units` and `xclim.core.calendar`.

PR #2409, opened 2026-08-28
Pylint adjustments

Zeitsperre:
### Pull Request Checklist:
    - This PR fixes #xyz

* Addresses many small issues raised by `pylint`
* Updates `pylint` to v4.0
* Replaces existing `pylint: disable=` statements using error codes to use error titles (`R0912` -> `too-many-branches`).

No.

### Other information:

Most issues now can either be addressed via refactoring (too many conditionals/variables/lines/etc.). `pylint` also catches all `TODO` and `FIXME` comments (of which there are *many*). Perhaps it may be worth it to try to bring this number down before v1.0 release?

### DIFF
--- a/src/xclim/core/indicator.py
+++ b/src/xclim/core/indicator.py
@@ -1413,7 +1413,7 @@ class _InputChecker(_DeprecationWarner):
             )
 
 
-class _Convenience(_InputChecker):  # pylint: disable=too-many-ancestors
+class _Convenience(_InputChecker):
     """
     Adds pre-processing to the constructor arguments so it can accept some v0 names
     and CF attributes passed by name instead of within `attrs`.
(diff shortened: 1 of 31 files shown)


src/xclim/core/indicator.py
```diff
@@ -1413,7 +1413,7 @@ def datacheck(self, **das) -> None:
             )
 
 
-class _Convenience(_InputChecker):
+class _Convenience(_InputChecker):  # pylint: disable=too-many-ancestors
```
aulemahal: Could we deactivate `too-many-ancestors` at the module-level ? (in `pyproject`?). `indicator.py` is the only actual OOP structure in xclim, so it would only apply to here anyway, but it would remove noise from the file.

aulemahal (approved):
Nice, thanks.

Suggestion about a particular warning.

PR #2411, opened 2026-08-31
Add multimember test

### LINKED ISSUE #2406: potential mistake in ipcc-ar6-c robustness test
The [xclim code](https://github.com/Ouranosinc/xclim/blob/879dcf2482bd6ba3bca6653bcb45f21c707524bb/src/xclim/ensembles/_robustness.py#L654) and the [IPCC](https://www.ipcc.ch/report/ar6/wg1/chapter/atlas/) both define `gamma = np.sqrt(2 / 20) * 1.645 * ref_detrended.std("time")`.

However, my understanding of the 20 is that this comes from the number of year in the period.

In the IPCC atlas, the period is always 20 years. This is not the case with xclim and Portraits Climatiques (where it is 30 years).

I think it should probably be `gamma = np.sqrt(2 / len(refy.time)) * 1.645 * ref_detrended.std("time")`

@martinleduc, can you please confirm that I understand the equation correctly ?

juliettelavoie:
### Pull Request Checklist:
    - This PR fixes #2406 

* Add multimember test for robustness_fractions.
  * the input is the same a before ( a realization dim with everything). `stack_ensemble_member` finds the members pattern and the rest of the id (which I call sub_id) is considered the new realization. 
  * Previous test only change computation inside of the test, but if this test is used, it also affects the computation of `delta` and `valid` which are now first reduce on the member dimension.
* fix #2406 
* Change a pre-commit url for original version instead of mirror

yes, for ipcc-ar6-c test the previous results were wrong

### Other information:
math showing that multimember is the same as ipcc-ar6-c but with less assumptions.
Capture d’écran, le 2026-08-31 à 10 38 41


src/xclim/ensembles/_robustness.py
```diff
@@ -68,6 +71,68 @@ def significance_test(func: Callable) -> Callable:
...
+
+        ["MIROC6_r1i1p1f1_hist", "MIROC6_r2i1p1f1_hist", "CanESM5_r1i1p2f1_hist"]
+
+    the result has realization coordinates ``["MIROC6_hist", "CanESM5_hist"]`` and
+    member coordinates ``[0, 1]``.
+    """
+    if "realization" not in ens.dims:
+        raise ValueError("Input array must have a 'realization' dimension.")
+
+    pattern = re.compile(r"_(r\d+i\d+p\d+f\d+)_?")  # rXiYpZfW anywhere, optional trailing underscore
```
aulemahal: Would it be possible to also support CMIP5 here ? I.e. make the `f\d+` part optional.

CMIP7 seems to use the same `variant_label` as CMIP6.

src/xclim/ensembles/_robustness.py
```diff
@@ -194,6 +261,11 @@ def robustness_fractions(
     >>> ref = tgmean.sel(time=slice("1990", "2020"))
     >>> fractions = ensembles.robustness_fractions(fut, ref, test="ttest")
     """
+    if test == "multimember":
+        fut = stack_ensemble_member(fut)
+        if ref is not None:
+            ref = stack_ensemble_member(ref)
```
aulemahal: Is it possible to avoid doing this inside the function ? I mean that I'd prefer not to assume anything about the `realization` dimension inside the principal functions. Rather the user could be pointed to `stack_ensemble_member` and this function would detect the presence of an additional `member` dimension to act accordingly.

src/xclim/ensembles/_robustness.py
```diff
@@ -673,6 +750,48 @@ def _gen_test_entry(namefunc):
     return entry
 
 
+@significance_test
+def _multimember(
```
aulemahal: If this the same a `ipcc-ar6-c`, I think a better approach would be to handle the multimember case in that function directly.

aulemahal (commented):
Nice, thanks!

However, I feel like we are introducing a new concept here that merits more attention. It feels like the whole submodule should support multimember ensembles.

Also, I don't think I understand the mathematical derivation ? I'm not sure I see why $$\sigma_p = \sigma_f$$ _only in the single-member case_ ?

juliettelavoie:
> Also, I don't think I understand the mathematical derivation ? I'm not sure I see why \sigma_p = \sigma_f only in the single-member case ?

fair question. It is not because of the number of member. The IPCC makes the assumption that the variability will not change in time. Martin (and I) disagree with that assumption. We could have a single-member test that uses \sigma_f  also

src/xclim/ensembles/_robustness.py
juliettelavoie: it is not exactly the same, because we also remove the assumption of stationnarity of variance.

We could have a test for ipcc-ar6-c (single and multimember) and another one to-be-named (single and multimember)

src/xclim/ensembles/_robustness.py
juliettelavoie: sure. I hesitated on that one. I thought it might be better to always have the same input, but I am ok asking the user to stack first. Gives more liberty to stack how you want also.

aulemahal: I think your PR introduces the need to more clearly define what `xclim.ensembles` requires and I think we could extend it to cover the two cases: `realization` only, `realization`+ `member`.

src/xclim/ensembles/_robustness.py
juliettelavoie: pretty sure it's doable

src/xclim/ensembles/_robustness.py
juliettelavoie: ok I'll work on generalizing the approach.

src/xclim/ensembles/_robustness.py
juliettelavoie: I take back the idea of ipcc-ar6-c multimember because that is not defined in the ipcc, so I don't want to invent something here.

but we can have both options for more general tests.

src/xclim/ensembles/_robustness.py
juliettelavoie: does this last commit match what you had in mind ? or did you want to go broader than robustness ? I can see an argument for adding it to the ensemble statistics, but that would be equivalent to only the simplest case of  `xs.generate_weight`. This is making me think that maybe `unstack_ensemble_member` is too reliant on ids to be xclim. Should it be in xscen instead ?

@RondeauG thoughts ?

Also, I am wondering if it is worth it to generalize all the tests ? If so, I am not sure what is the best way to do that.. maybe unstacking the time and member dimension to compare larger distributions ?

RondeauG: It's not a trivial question... There's already a `weight` argument elsewhere that should work for most cases, but it's true that some tests (mostly in `robustness`) or specific computations (`std`, maybe ?) might benefit from stacking members together.

I'd be in favor of maybe adding a helper function similar to xclim.ensembles.reduce.make_criteria to create a multidimensional `weights` object with a new stacked dimension ?

src/xclim/ensembles/_robustness.py
```diff
@@ -661,6 +763,52 @@ def _ipcc_ar6_c(fut, ref, *, ref_pi=None):
...
+    fut_detrended = detrend(futy, dim="time", deg=1)
+
+    if "member" in ref_detrended.dims:
+        ref_var = ref_detrended.var(["time", "member"])
+        fut_var = fut_detrended.var(["time", "member"])
+    else:
+        ref_var = ref_detrended.var("time")
+        fut_var = fut_detrended.var("time")
+
+    gamma = 1.645 * np.sqrt((ref_var / nref) + (fut_var / nfut))
```
aulemahal: I suggest making the 1.645 value an argument so it can be modified ?

Could also be expressed in terms of confidence interval ? Like:
```
def _signal_to_noise(fut, ref, confidence=0.9):
     fact = scipy.stats.norm.ppf(confidence / 2 +  0.5)
```

To me "confidence interval" is more meaningful information, but 1.645 is the actual value, so its more explicit...

aulemahal (commented):
Yep that's what I was thinking about!

Let's implement this new thing in  `robustness_fractions` only for this PR.

src/xclim/ensembles/_robustness.py
juliettelavoie: i like it

juliettelavoie:
@martinleduc voici la PR à réviser!
@aulemahal  et @RondeauG aussi ça serait prêt pour une revue please.

Pour le RdT, peut-être que juste relancer marcherait, mais je n'ai pas la permission.

aulemahal (commented):
Nice! Thanks for implementing the multi-member case to the other tests.

I have two preoccupations with this new thing:
- I don't think `unstack_ensemble_member` is xclim-esque. This looks like the kind of thing `xscen` manages, there is no standard I know of for "ensemble" axes in CF/xarray, so we can't really assume anything about it. I would suggest removing it from this PR and move to `xscen` for our usage.
- Similarly, the word `realization` was generic enough, but it's always been a bit out-of-scope to prescribe a non-standard term like this. I find `realization` and `member` to be very ambiguous terms for the concepts here. I would suggest adding a `dim` argument to the function. If `dim` is a sequence of length 2, then we activate the "multi-member" cases using the second element.

Also, we probably need to add tests for the multi-member cases of the simple statistical tests.

src/xclim/ensembles/_robustness.py
```diff
@@ -661,6 +792,51 @@ def _ipcc_ar6_c(fut, ref, *, ref_pi=None):
...
+    This is similar to the "ipcc-ar6-c" test, but there is no assumption that the
+    variance will be the same in the past and in the future. Further, this test can handle
+    multimember ensembles, while the "ipcc-ar6-c" test assumes a single dimension.
+
+    """
+    z = spstats.norm.ppf(confidence / 2 + 0.5)
+    # Ensure annual
+    refy = ref.resample(time="YS").mean()
+    futy = fut.resample(time="YS").mean()
+    nref = len(refy.time) * len(ref[pool_dim]) if pool_dim in refy.dims else len(refy.time)
```
juliettelavoie, line 818: @martinleduc pour info, c'est ici que je répare le n.

src/xclim/ensembles/_robustness.py
```diff
@@ -83,27 +103,41 @@ def robustness_fractions(
...
-        Reference period values along realization' and 'time'  (..., nr, nt2).
+        Reference period values along 'dim'(s) and 'time'  (..., nr, nt2).
         The size of the 'time' axis does not need to match the one of `fut`.
-        But their 'realization' axes must be identical and the other coordinates should be the same.
+        But their dim(s) axes must be identical and the other coordinates should be the same.
         If `None` (default), values of `fut` are assumed to be deltas instead of
         a distribution across the future period.
+    dim : str or list of str, optional
+        If str or list of length 1, the realization dimension along which to compute the fractions.
+        If list of length 2, the first element is the realization dimensionand the second element is the pool dimension.
```
aulemahal:
```suggestion
        If list of length 2, the first element is the realization dimension and the second element is the pool dimension.
```

src/xclim/ensembles/_robustness.py
```diff
@@ -634,24 +776,29 @@ def _brownforsythe_test(fut, ref, *, p_change=0.05):
...
     Change is considered significant if the delta exceeds a threshold related to the internal variability.
     If pre-industrial data is given in argument `ref_pi`, the threshold is defined as
     :math:`\sqrt{2}*1.645*\sigma_{20yr}`, where :math:`\sigma_{20yr}` is the standard deviation of 20-year
-    means computed from non-overlapping periods after detrending with a quadratic fit.
+    means computed from non-overlapping periods after detrending with a quadratic fit and
+    1.645 is the z-score for the 90% confidence interval of the two-sided test .
     Otherwise, when such pre-industrial control data is not available, the threshold is defined in relation to
-    the historical data (`ref`) as :math:`\sqrt{\frac{2}{20}}*1.645*\sigma_{1yr}, where :math:`\sigma_{1yr}`
-    is the inter-annual standard deviation measured after linearly detrending the data.
+    the historical data (`ref`) as :math:`\sqrt{\frac{2}{n}}*1.645*\sigma_{1yr}, where :math:`\sigma_{1yr}`
```
aulemahal:
```suggestion
    the historical data (`ref`) as :math:`\sqrt{\frac{2}{n}}*1.645*\sigma_{1yr}`, where :math:`\sigma_{1yr}`
```

src/xclim/ensembles/_robustness.py
```diff
@@ -634,24 +776,29 @@ def _brownforsythe_test(fut, ref, *, p_change=0.05):
...
+    is the inter-annual standard deviation measured after linearly detrending the data
+    and n is the number of years in the reference period.
+    This test is only for ensembles with a single dimension.
```
aulemahal:
```suggestion

    This test does not work with ensembles with a pool dimension.
```
Nitpick, but using the "pool" word felt more coherent with the others ?

src/xclim/ensembles/_robustness.py
```diff
@@ -68,12 +70,30 @@ def significance_test(func: Callable) -> Callable:
     return func
 
 
+def _unpack_dim(dim):
+    """Unpack the dim argument of robustness_fractions into realization_dim and pool_dim."""
+    if isinstance(dim, str):
+        realization_dim = dim
+        pool_dim = "doesnotexist"
```
aulemahal:
```suggestion
        pool_dim = None
```
Why not None ? `None` already can't be a dimension name.

src/xclim/ensembles/_robustness.py
```diff
@@ -193,55 +230,76 @@ def robustness_fractions(
...
-    # Realization dimension name
-    realization = "realization"
+    # unpack dims
+    realization_dim, pool_dim = _unpack_dim(dim)
+    if "member" in fut.dims and pool_dim != "member":
+        warnings.warn(
+            'We noticed that the "member" dimension is present in fut,'
+            ' but not in the second element of "dim". Make sure to pass the correct '
+            'dimensions to "dim" if you want to pool members for each realization.'
+        )
```
aulemahal: I see where you're coming from, but that tests seems quite SSC-specific ? I would argue having `member` as `realization_dim` could make sense in some context, like a single-model multi-member ensemble.
```suggestion
```

src/xclim/ensembles/_robustness.py
```diff
@@ -193,55 +230,76 @@ def robustness_fractions(
...
+        ref = ref.assign_coords({realization_dim: "dummy"})
+        ref = ref.expand_dims(realization_dim)
 
     # Get dummy weights to simplify code
     if weights is not None:
+        if pool_dim in weights.dims:
+            raise ValueError(
+                f"Weights cannot have a {pool_dim} (second element of dim) dimension,"
+                f"only a {realization_dim} (first element of dim) dimension."
+            )
```
aulemahal, line 253:
```suggestion
                f"Weights cannot have the {pool_dim} (second element of dim) dimension,"
                f"only the {realization_dim} (first element of dim) dimension."
```

src/xclim/ensembles/_robustness.py
```diff
@@ -250,24 +308,30 @@ def robustness_fractions(
...
 
-        changed, pvals = test_func(fut, ref, **test_params)
+        changed, pvals = test_func(fut, ref, pool_dim, **test_params)
     else:
         raise ValueError(f"Statistical test {test} must be one of {', '.join(SIGNIFICANCE_TESTS.keys())}.")
 
-    valid_frac = valid.weighted(w).sum(realization) / fut[realization].size
-    n_valid = valid.weighted(w).sum(realization)
-    change_frac = changed.where(valid).weighted(w).sum(realization) / n_valid
+    # at this point all member dimension should have been collapsed,
```
aulemahal:
```suggestion
    # at this point all pool dimension should have been collapsed,
```

src/xclim/ensembles/_robustness.py
```diff
@@ -250,24 +308,30 @@ def robustness_fractions(
...
+    # so we can compute the fractions along realization_dim
+
+    # TODO: reviewer please confirm that changing the denominator makes sense.
+    # this change is not because of my new feature. I think it was an error in the past.
+    # as we do the weighted sum, we should divide by the sum of weights to get a fraction that make sense I think
```
aulemahal:
```suggestion
```
You are right! Weights are not normalized by xarray, so we need to divide by the sum indeed!

src/xclim/ensembles/_robustness.py
```diff
@@ -634,24 +776,29 @@ def _brownforsythe_test(fut, ref, *, p_change=0.05):
...
-    the historical data (`ref`) as :math:`\sqrt{\frac{2}{20}}*1.645*\sigma_{1yr}, where :math:`\sigma_{1yr}`
-    is the inter-annual standard deviation measured after linearly detrending the data.
+    the historical data (`ref`) as :math:`\sqrt{\frac{2}{n}}*1.645*\sigma_{1yr}, where :math:`\sigma_{1yr}`
+    is the inter-annual standard deviation measured after linearly detrending the data
+    and n is the number of years in the reference period.
+    This test is only for ensembles with a single dimension.
     See notebook :ref:`notebooks/ensembles:Ensembles` for more details.
     """
+    if pool_dim in fut.dims:
+        raise ValueError("This test does not support a second dimension. Only give one element or a string to dim.")
```
aulemahal:
```suggestion
    if pool_dim is not None:
        raise ValueError("This test does not support a second dimension. Only give one element or a string to dim.")
```
I would prefer testing against the sentinel value here and not combine this condition with a test for existence. I mean : this should fail if `pool_dim` is given, even if it doesn't exist on `fut`.

src/xclim/ensembles/_robustness.py
```diff
@@ -661,6 +808,51 @@ def _ipcc_ar6_c(fut, ref, *, ref_pi=None):
     return changed, None
 
 
+@significance_test
+def _signal_to_noise(fut, ref, pool_dim, confidence=0.9):
```
aulemahal:
```suggestion
def _signal_to_noise(fut, ref, pool_dim, *, confidence=0.9):
```

src/xclim/ensembles/_robustness.py
```diff
@@ -661,6 +808,51 @@ def _ipcc_ar6_c(fut, ref, *, ref_pi=None):
...
+    r"""
+    Robustness test using the signal-to-noise ratio.
+
+
+    Change is considered significant if the delta exceeds a threshold related to the
+    internal variability (noise).
+    The threshold is defined as :math:`z*\sqrt{\frac{\tilde{\sigma}^2_{ref}}{n_{ref}}
+    + \frac{\tilde{\sigma}^2_{fut}}{n_{fut}}}, where :math:`\tilde{\sigma}}`
```
aulemahal:
```suggestion
    + \frac{\tilde{\sigma}^2_{fut}}{n_{fut}}}`, where :math:`\tilde{\sigma}}`
```

src/xclim/ensembles/_robustness.py
```diff
@@ -661,6 +808,51 @@ def _ipcc_ar6_c(fut, ref, *, ref_pi=None):
...
+    is the pooled interannual variance across members and time measured after linearly
```
aulemahal:
```suggestion
    is the pooled interannual variance across the pool dimension and time measured after linearly
```

tests/test_ensembles.py
```diff
@@ -767,3 +775,291 @@ def test_robustness_coefficient():
...
+        (
+            None,
+            [1, 1, 1, 1],
+            [1, 0.5, 1, 1],
+            [],
+            {},
+        ),
+    ],
+)
+def test_robustness_fractions_2d_dask(robust_data, test, exp_chng_frac, exp_pos_frac, exp_changed, kws):
```
aulemahal: Is this the same test as the non-dask version ?

Could we:

```suggestion
@pytest.mark.parametrize("use_dask", [True, False])
def test_robustness_fractions_2d_dask(robust_data, test, exp_chng_frac, exp_pos_frac, exp_changed, kws, use_dask):
```
instead ?  (with the appropriate if-else below).

CHANGELOG.rst
```diff
@@ -60,6 +61,7 @@ Breaking changes
     * For array-like ``doy_bounds`` without ``time`` dimension, the start and end bounds must now be consecutive according to the frequency (default: ``freq="YS"``). Otherwise, the indexing is invalid and no data are selected.
 * Variable `q` has been renamed to `rivo` in hydrological indicators (``xclim.indicators.land``) to follow modern naming conventions. ``xclim.land.doy_q{min|max}`` are renamed to ``xclim.land.rivo_{min|max}_doy``. (:issue:`2407`, :pull:`2408`).
 * Translation: ``xclim.core.locales.get_local_attrs`` has been rewritten and only accepts a single "locale" now. Locale dictionaries are now case-insensitive. (:pull:`2397`).
+* Replace hard-coded length of period to actually use the length of the input in ``xclim.ensembles.robustness_fractions``. (:pull:`2411`, :issue:`2406`).
```
```suggestion
* Replace hard-coded length of period to actually use the length of the input in test `ipcc-ar6-c` of ``xclim.ensembles.robustness_fractions``. (:pull:`2411`, :issue:`2406`).
```

aulemahal (approved):
Merci full!
This is a cool new functionality and test.

src/xclim/ensembles/_robustness.py
juliettelavoie: there was a reason.. I'll go back to check I don't remember

src/xclim/ensembles/_robustness.py
juliettelavoie: ok that's true

tests/test_ensembles.py
juliettelavoie: well we already parametrize to iterate through all the tests. can you do 2 levels of parametrization ? or I have to rewrite each test  twicebut with a new arg use dask (this option felt the same level of repetitive a having two tests..)

aulemahal: You can simply add another level, another decorator entirely. Pytest will run on the « product » of both.

juliettelavoie: ah ben good to know!

src/xclim/ensembles/_robustness.py
juliettelavoie: well i seems to work with None now

PR #2414, opened 2026-09-04
Indicators return Datasets - flip option `as_dataset` to True

### LINKED ISSUE #2410: Proposition : Flip default to output datasets
My proposition is to return `Datasets` object by default in `xclim v1` when computing an indicator.

My arguments:
- Cleaner multi-variable support
- The `history` attribute, our principal "provenance" attribute is defined as a global attribute in CF. Before #2399, we were adding it to the DataArrays, which often resulted in broken provenance tracks (lost information). That PR simply dropped this non-CF behaviour.
- Better potential support for other features (auxiliary coordinates, other global attributes)

Specifically, my proposition is to flip this line to True. : https://github.com/Ouranosinc/xclim/blob/4d08949ad7cb2210bd56d10979d4236218557067/src/xclim/core/options.py#L37

What says the crowd ? 
@Ouranosinc/xclim-core

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2410 

* Flips the option `as_dataset` to True, i.e. all indicators now return Datasets by default.

Oh my it does!

### Other information:
@Ouranosinc/xclim-core This seems one of the biggest breaking change in v1, just look at the size of this single commit!

I think this is better, but I could be convinced that breaking worflows this much is not worth it.

To go back to the previous situation, one simply needs to do `xclim.set_options(as_dataset=False)`, in a context or even globally.


docs/notebooks/example.ipynb
```diff
@@ -222,17 +222,23 @@
...
     "\n",
-    "The second layer are **Indicators**, class instances that you'll find organized by *realm*. So far, there are three realms available in `xclim.indicators.atmos`, `xclim.indicators.seaIce` and `xclim.indicators.land`, the first one being the most exhaustive. Before running computations, these classes check if the input data is a daily average of the expected variable:\n",
+    "The second layer are **Indicators**, class instances that you'll find organized by *realm*. So far, there are three realms available in `xclim.indicators.atmos`, `xclim.indicators.seaIce` and `xclim.indicators.land`, the first one being the most exhaustive. Two other submodules are available : `xclim.indicators.generic` for indicators not specific to an input variable and `xclim.indicators.convert` for indicators that do not resample along the time dimension. \n",
```
Zeitsperre:
```suggestion
    "The second layer are **Indicators**, class instances that you'll find organized by *realm*. So far, there are three realms available in `xclim.indicators.atmos`, `xclim.indicators.seaIce` and `xclim.indicators.land`, the first one being the most exhaustive. Two other submodules are available: `xclim.indicators.generic` for indicators not specific to an input variable and `xclim.indicators.convert` for indicators that do not resample along the time dimension. \n",
```

docs/notebooks/example.ipynb
```diff
@@ -384,7 +389,7 @@
    "source": [
     "## *Lazy* computation - Nothing has been computed so far !\n",
     "\n",
-    "If you look at the output of those operations, they're identified as `dask.array` objects. What happens is that `dask` creates a chain of operations that, when executed, will yield the values we want. We have thus far only created a schedule of tasks with a small preview and not done any actual computations, except when making figures. You can trigger computations by using the `load` or `compute` method, or writing the output to disk via `to_netcdf`. Of course, calling `.plot()` will also trigger the computation."
+    "Except for the last subsection, if you look at the output of those operations, they're identified as `dask.array` objects. What happens is that `dask` creates a chain of operations that, when executed, will yield the values we want. We have thus far only created a schedule of tasks with a small preview and not done any actual computations, except when making figures. You can trigger computations by using the `load` or `compute` method, or writing the output to disk via `to_netcdf`. Of course, calling `.plot()` will also trigger the computation."
```
Zeitsperre:
```suggestion
    "Except for the last subsection, if you look at the output of those operations, they're identified as `dask.array` objects; `dask` creates a chain of operations that, when executed, will yield the values we want. We have thus far only created a schedule of tasks with a small preview and not done any actual computations, except when making figures. You can trigger computations by using the `load` or `compute` method, or by writing the output to disk via `to_netcdf`. Of course, calling `.plot()` will also trigger the computation."
```

docs/notebooks/usage.ipynb
```diff
@@ -225,7 +225,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`xclim` also allows us to call indicators using datasets and variable names."
+    "`xclim` also allows us to call indicators using datasets and variable names. xarray's `keep_attrs` option is used to decide if dataset (or node) attributes should be preserved and passed to the output."
```
Zeitsperre:
```suggestion
    "`xclim` also allows us to call indicators using datasets and variable names. `xarray`'s `keep_attrs` option is used to decide if dataset (or node) attributes should be preserved and passed to the output."
```

docs/notebooks/usage.ipynb
```diff
@@ -247,7 +247,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can also get datasets as an output with the `as_dataset` option."
+    "We can also get dataarrays as an output by disabling the `as_dataset` option. Note that some meaningful metadata is usually appended in the global `history` attribute. When returning DataArrays, this is lost."
```
Zeitsperre:
```suggestion
    "We can also get DataArrays as an output by disabling the `as_dataset` option. Note that some meaningful metadata is usually appended in the global `history` attribute. When returning DataArrays, this is lost."
```

docs/indicators.rst
```diff
@@ -3,9 +3,8 @@ Climate Indicators
 ==================
 
 :py:class:`xclim.core.indicator.Indicator` instances essentially perform the same computations as the functions
-found in the :mod:`xclim.compute` library, but also run a number of health checks on input data
-and assign attributes to the output arrays. So for example, if there are missing values in
-a time series, compute functions won't notice, but indicators will return NaNs for periods with missing
+found in the :mod:`xclim.compute` library, but also run a number of health checks on input data, assign attributes to the output arrays and return a :py:class:`xarray.Dataset`.
+So for example, if there are missing values in a time series, compute functions won't notice, but indicators will return NaNs for periods with missing
```
Zeitsperre:
```suggestion
For example, if there are missing values in a time series, compute functions will ignore them, but indicators will return `NaN` for periods with missing
```

src/xclim/core/indicator.py
```diff
@@ -1512,7 +1512,8 @@ class Indicator(_Registrer):  # numpydoc ignore=PR01
     or a :py:class:`xarray.DataTree` as input, with the added argument `ds` in the call signature.
     All arguments that were indicated by the compute function to be variables (DataArrays) through
     annotations will be promoted to also accept strings that correspond to variable names
-    in the `ds` dataset (or on each DataTree nodes).
+    in the `ds` dataset (or on each DataTree nodes). Also, indicators return Datasets by default,
```
Zeitsperre:
```suggestion
    in the `ds` Dataset (or on each DataTree nodes). Also, indicators return Datasets by default,
```

tests/test_indicators.py
```diff
@@ -157,7 +156,8 @@ def test_attrs(tas_series):
         attrs={"long_name": "A thresh", "units": "degC"},
         name="TT",
     )
-    txm = uniIndTemp(a, thresh=thresh, freq="YS")
+    with xclim.set_options(as_dataset=False):
```
Zeitsperre, line 159: Just wondering but in the case of conlficting but nested `set_options`, such as a script calling a function with a context not unlike this:
```python
with xclim.set_options(as_dataset=False):
    some_script()

def some_script():
    with xclim.set_options(as_dataset=True):
        ds = xclim.indicator(...)
```
Which option would take precedence?

Zeitsperre (approved):
This looks great. I think we were long overdue to make this change in order to better follow CF-Conventions.

tests/test_indicators.py
aulemahal: The last one in execution order I think.

`set_options` modifies a global dictionary that xclim reads when needed, so the last  `set_options.__init__` to be executed overrides the previous ones.

Zeitsperre:
@aulemahal Do you think you could add a temporary warning to the top of the README.rst that says something to the effect of:

> The `xclim` 'main' branch is currently undergoing significant changes and no longer behaves like the previous 'v0' releases. For more information on changes, be sure to consult https://xclim.readthedocs.io/en/latest/changelog.html {|| _page-with-explanations?_} for more information.

aulemahal:
ah isn't there something very similar already there ?

Zeitsperre:
> ah isn't there something very similar already there ?

I think I need another coffee. Disregard.

PR #2415, opened 2026-09-04
Updates to IndicatorCollection

### LINKED ISSUE #2423: Building yaml with wind_speed_from_vector breaks with xclim v1.0
### Setup Information

- Xclim version: 0.99.0-dev.25
- Python version: 3.14.7
- Operating System: linux

### Description

We used to be able to have yaml with this wind_speed_from_vector without specifying the var_name but now it fails with 

Now this fails:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
ValueError: Constructing wind_speed_from_vector failed with ValueError('Output #1 of example.wind_speed_from_vector is missing a var_name.')

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In[1], line 14
      9 vas.attrs["units"] = "m s-1"
     11 wind, wind_dir = xc.convert.wind_speed_from_vector(uas=uas, vas=vas)
---> 14 xc.IndicatorCollection.from_yaml('/home/jlavoie/xclim/docs/notebooks/example/example.yml')

File ~/xclim/src/xclim/core/collection.py:295, in IndicatorCollection.from_yaml(cls, filename, name, computes, translations, mode, encoding, validate)
    292         mapping[identifier] = base(identifier=f"{coll_name}.{identifier}", module=coll_name, **data)
    294     except Exception as err:  # pylint: disable=broad-except
--> 295         raise_warn_or_log(err, mode, msg=f"Constructing {identifier} failed with {err!r}")
    297 coll = cls(mapping, name=coll_name, doc=doc)
    298 # If there are translations, load them

File ~/xclim/src/xclim/core/_exceptions.py:57, in raise_warn_or_log(err, mode, msg, err_type, stacklevel)
     55     warnings.warn(message, stacklevel=stacklevel + 1)
     56 else:  # mode == "raise"
---> 57     raise err from err_type(message)

File ~/xclim/src/xclim/core/collection.py:291, in IndicatorCollection.from_yaml(cls, filename, name, computes, translations, mode, encoding, validate)
    289     data.setdefault("realm", defkwargs.get("realm"))
    290     print('base')
--> 291     print(base(identifier=f"{coll_name}.{identifier}", module=coll_name, **data))
    292     mapping[identifier] = base(identifier=f"{coll_name}.{identifier}", module=coll_name, **data)
    294 except Exception as err:  # pylint: disable=broad-except

File ~/xclim/src/xclim/core/indicator.py:1481, in _Convenience.__new__(cls, **kwargs)
   1479 # Priority given to passed realm -> parent's realm -> location of class declaration (xclim indicators only)
   1480 kwargs.setdefault("realm", cls.realm or xclim_realm)
-> 1481 return super().__new__(cls, **kwargs)

File ~/xclim/src/xclim/core/indicator.py:668, in IndicatorBase.__new__(cls, **kwds)
    666 print('2')
    667 print(attrs)
--> 668 attrs = cls._ensure_correct_attrs(attrs, identifier)
    669 kwds["attrs"] = attrs
    671 # Other metadata
    672 # If these fields were not given, set them from the parsed docstring

File ~/xclim/src/xclim/core/indicator.py:847, in IndicatorBase._ensure_correct_attrs(cls, attrs, identifier)
    845 for i, atts in enumerate(attrs, 1):
    846     if atts.var_name is None:
--> 847         raise ValueError(f"Output #{i} of {identifier} is missing a var_name.")
    848 return attrs

ValueError: Output #1 of example.wind_speed_from_vector is missing a var_name.
```

### Steps To Reproduce

```
xc.IndicatorCollection.from_yaml('example.yml')
```

with yaml:
```
realm: atmos
indicators:
  wind_speed_from_vector: 
    base: wind_speed_from_vector
```

### Additional context

Yes, I could add the var_name by hand in the yaml, but this fails later in xscen.

I think the issue is in `indicators._update_attrs`. For this case, `attrs= [<Output wind (...) >, <Output wind_from_dir (...}' >]` and `new_attrs=[]`. When they are merged with ` fillvalue=Output()`, the output is `[<Output [unnamed] (...)>, <Output [unnamed] (..).'>]`
 Notice the "unnamed" which then fails when we check the attrs.

I can get this to work by adding:

```
        if len(attrs) ==2 and attrs[0].var_name is not None:
            attrs[0]['var_name']=attrs[0].var_name
            attrs[1]['var_name']=attrs[1].var_name
```

here https://github.com/Ouranosinc/xclim/blob/62b932f4d42482a1af8455f9fcae281656d8ec28/src/xclim/core/indicator.py#L808 

but this feels like cheating, might break something else and doesn't really fix the underlying issue.

@aulemahal  Let me know what you think and what would be the best way to adress this.

### Contribution

### Code of Conduct

aulemahal:
### Pull Request Checklist:
    - This PR fixes #2423

* Registering an indicator to `xclim.core.indicator.registry` is now optional
* Indicators created in the context of the Collection from a yaml are not registered by default.

My goal was to uncouple Collections from the main xclim code. Extra variables are still shared in `xclim.core.VARIABLES`, but this is less of an issue I think.

One change this will do is fewer warnings about overwriting indicators when loading a collection. You know things like:
```
UserWarning: Class conversions.WIND_SPEED_FROM_VECTOR already exists and will be overwritten.
```

EDIT: I added two other changes here:
- Possibility to use python qualified names in the yaml for `base` and `compute`. Ex: `atmos.precip_accumulation` instead of `prcptot`.
- New `bases` section in the Yaml for indicator _class_ definition, i.e. base classes to reuse when defining indicators.

Yes, but easily worked around by passing `register=True` to  `IndicatorCollection.from_yaml`. I'm guessing the registering behaviour was not used a lot for virtual submodules.


docs/notebooks/extendxclim.ipynb
```diff
@@ -309,15 +309,20 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`example.yml` created a module of 7 indicators.\n",
+    "`example.yml` created a module of 7 indicators and 1 base indicator class.\n",
     "\n",
     "<div class=\"alert alert-info\">\n",
     "\n",
-    "Values of the `base` arguments are the **identifier** of the associated indicators, and those can be different from their name within the Python modules. For example, `xclim.atmos.relative_humidity` has `hurs` as identifier. One can always access `xclim.atmos.relative_humidity.identifier` to get the correct name to use. The `base` argument also accepts generic base classes which are registered in `xc.core.indicator.base_registry`. The indicator registry is case-insensitive.\n",
+    "Indicators are always derived from a subclass of `xclim.core.indicator.Indicator`, which one to use is set through the `base` argument. This can have multiple forms :\n",
```
Zeitsperre:
```suggestion
    "Indicators are always derived from a subclass of `xclim.core.indicator.Indicator`. The decision of which one to use is set through the `base` argument. This can have multiple forms:\n",
```

docs/notebooks/extendxclim.ipynb
```diff
@@ -309,15 +309,20 @@
...
+    "\n",
+    "- A name referring to the identifier in the indicator registry `xclim.core.indicator.registry` or to a name in the base class registry `xclim.core.indicator.base_registry`\n",
+    "- A qualified name relative to `xclim.indicators` (see first indicator below) or a full qualified name of something to import dynamically.\n",
+    "- The name of another indicator or class declared within the current collection.\n",
+    "- When unset, the `base` defaults to the one defined collection-wide or to `xclim.core.indicator.Daily`.\n",
     "\n",
     "</div>\n",
     "\n",
-    "- `RX1day` is as  `registry['rx1day']`, but with an updated `long_name` and an injected argument : its `indexer` arg is now set to only compute over may to september.\n",
+    "- `RX1day` is derived from `xclim.indicators.atmos.max_1day_precipitation_amount`, with an updated `long_name` and an injected argument : its `indexer` arg is now set to only compute over may to september.\n",
```
Zeitsperre:
```suggestion
    "- `RX1day` is derived from `xclim.indicators.atmos.max_1day_precipitation_amount`, with an updated `long_name` and an injected argument; Its `indexer` arg is now set to only compute over May to September.\n",
```

docs/notebooks/extendxclim.ipynb
```diff
@@ -326,9 +331,7 @@
...
-    "- `R95p` is similar to `first_frost_day` but here the `compute` is not defined in `xclim` but rather in  `example.py`. Also, the custom function returns two outputs, so the `output` section is a list of mappings rather than a mapping directly.\n",
-    "- `R99p` is the same as `R95p` but changes the injected value. In order to avoid rewriting the output metadata, and allowed periods, we based it on `R95p` : as the latter was defined within the current YAML file, the identifier is prefixed by a dot (.).\n",
-    "\n",
+    "- `R95p` and `R99p` are both small variations of the base indicator class defined in the `bases` section : `RXXp`. The latter is similar to `first_frost_day` but here the `compute` is not defined in `xclim` but rather in  `example.py`. Also, the custom function returns two outputs, so the `output` section is a list of mappings rather than a mapping directly. The two variations only change the variable names and inject a different `perc` value.\n",
```
Zeitsperre:
```suggestion
    "- `R95p` and `R99p` are both small variations of the base indicator class defined in the `bases` section: `RXXp`. The latter is similar to `first_frost_day` but here the `compute` is not defined in `xclim` but rather in  `example.py`. Also, the custom function returns two outputs, so the `output` section is a list of mappings rather than a mapping directly. The two variations only change the variable names and inject a different `perc` value.\n",
```

src/xclim/core/collection.py
```diff
@@ -271,64 +319,113 @@ def from_yaml(  # noqa: C90
...
+                if name.count(".") == 1 and name.split(".")[0] in [
+                    "atmos",
+                    "land",
+                    "seaIce",
+                    "convert",
+                    "generic",
+                    "cf",
+                    "anuclim",
+                    "icclim",
+                ]:
```
Zeitsperre: Do we want to maybe have One True List :tm: of the various indicator-domain modules that could be imprted and reused here? Who know if we'll add more.

src/xclim/core/collection.py
```diff
@@ -271,64 +319,113 @@ def from_yaml(  # noqa: C90
...
                 func = getattr(xclim.compute.generic, name, getattr(xclim.compute, name, None))
+        if func is None and "." in name:
+            modname, funcname = name.rsplit(".", 1)
+            try:
+                func = getattr(import_module(modname), funcname)
+            except ModuleNotFoundError as err:
+                raise_warn_or_log(err, "log", msg=f"Failed importing {funcname} from {modname}.")
         if func is None:
-            raise ValueError(f"Can't find requested compute function {name}.")
+            raise ValueError(f"Can't find compute function {name}.")
```
Zeitsperre:
```suggestion
            raise ValueError(f"Can't find compute function '{name}'.")
```

src/xclim/core/indicator.py
```diff
@@ -1540,7 +1542,9 @@ def __init__(
         identifier : str
             Unique ID for this indicator. Single-output indicator will use this as their output variable
             name if no `var_name`is passed to the first element of `attrs`.
-            All indicators are registered to :py:data:`xclim.core.indicator.registry`, which is case-insensitive.
+            Unless ``register`` is False, indicators are registered to :py:data:`xclim.core.indicator.registry`,
+            which is case-insensitive.
+            This field is required and can't be None.
```
Zeitsperre, line 1551: Is it only allowed to be `None` in the case of `register==False`?

Zeitsperre (approved):
LGTM!

Enormous nitpick, but I'm a bigger fan of `e.g.` (_exempli gratia_) over `Ex`/`ex` for examples. Not blocking, no pressure to adopt it. I might push that change for the entire library later.

src/xclim/core/collection.py
aulemahal: indeed, in fact I should try reading them dynamically!

src/xclim/core/indicator.py
aulemahal: No. This might be a bit strict indeed.

The identifier is still used when looking for translations. And for inferring the output variable name.

In both case, we could raise an error instead.

aulemahal:
I fixed #2423 in here as well, the merging/inheritance mechanism for `Output` objects has been fixed.

PR #2424, opened 2026-09-11
Transition page, API indexing and refactor

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* New page in the doc to document the change from v0 to v1.
* Indicator derivation with `Indicator.copy(` (instead of `obj.__class__`)
* Re-implement `Indicator.from_dict`, but deprecated.

No.

### DIFF
--- a/docs/notebooks/extendxclim.ipynb
+++ b/docs/notebooks/extendxclim.ipynb
@@ -212,7 +212,7 @@
     "\n",
     "The first method is best when defining indicators in scripts or external modules and is explained here. The second is explained further down and in the [submodule doc](../api.rst#indicator-collections).\n",
     "\n",
-    "Creating a new indicator that simply modifies some metadata output of an existing one is a simple call like:"
+    "Creating a new indicator that simply modifies some metadata output of an existing one is a simple call to the parent's `copy`:"
    ]
   },
   {
@@ -224,7 +224,7 @@
    "outputs": [],
    "source": [
     "# An indicator based on tg_mean, but returning Celsius and fixed on annual resampling\n",
-    "tg_mean_c = xclim.atmos.tg_mean.__class__(\n",
+    "tg_mean_c = xclim.atmos.tg_mean.copy(\n",
     "    identifier=\"tg_mean_c\",\n",
     "    units=\"degC\",\n",
     "    title=\"Mean daily mean temperature but in degC\",\n",
--- a/docs/transition_v1.rst
+++ b/docs/transition_v1.rst
@@ -1,0 +1,33 @@
+Differences between v0 and v1
+=============================
+In september 2026, a major overhaul of xclim was implemented. While most of the functionalities were preserved, the internals were modified in many breaking ways. The goal was to modernize the lower-level API, aiming to ease development and favor contributions to xclim by the larger community. As of xclim v1.0, not all the planned changes are implemented, but only non-breaking ones remain.
+
+This page aims to summarize the largest and most breaking changes, to help transition to v1.
+
+Indicators calculations return Datasets
+---------------------------------------
+The ``as_dataset`` option of :py:func:`xclim.set_options` was changed to True, meaning that indicator calculations now return :py:class:`~xarray.Dataset` objects by default instead of :py:class:`~xarray.DataArray` objects or tuples of those. Re-enabling the previous behaviour is simply done with ``xclim.set_options(as_dataset=False)``, either globally or as a context.
+
+Renamed `indices` to `compute`
+------------------------------
+The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the compute.
+
+Major reimplementation of the generic compute functions
+-------------------------------------------------------
+The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this github comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
+
+The names of the indicators in :py:mod:`xclim.indicators` were largerly unchanged, but some arguments might have been renamed.
+
+Indicator constructor and "virtual submodules"
+----------------------------------------------
+The :py:mod:`xclim.core.indicator` submodule was refactored in hopes of making the internals of the :py:class:`~xclim.core.indicator.Indicator` object easier to maintain and extend. In addition, the "virtual submodules" concept was rewritten as the :py:class:`~xclim.core.collection.IndicatorCollection` object. These are dict-like structure holding a collection of indicators, usually created from a YAML file, like before.
+
+The main breaking changes are:
+
+	- Attribute `cf_attrs` was renamed `attrs` and is now a list of :py:class:`~xclim.core.indicator.Output` objects. These act like dictionaries of attributes, but have also 4 properties : `var_name`, `units`, `units_metadata` and `dimensionality`. This name change is to be done in the indicator's constructor and in the YAML files.
+	- Function ``xclim.build_indicator_module_from_yaml`` is changed to :py:meth:`~xclim.core.collection.IndicatorCollection.from_yaml`.
+	- Indicators defined within a collection are not registered by default into the indicators registry, the collection is not added as module under ``xclim.indicators``.
+	- When creating an indicator from an existing one, the ``var_name`` of the output of an indicator with a single output now only defaults to the indicator's identifier if the parent does not have a ``var_name``. Previously, the identifier would always be used as the ``var_name``, regardless of the parent's attributes.
+	- The :py:meth:`~xclim.core.indicator.Indicator.from_dict` method is deprecated. Use :py:meth:`~xclim.core.indicator.Indicator.copy` instead, calling it on the indicator object you want to subclass/copy. This new method doesn't parse ``compute`` function names given as string, instead pass the function directly.
+
+See the updated :ref:`Extending xclim` page for more details.
--- a/docs/transition_v1.rst
+++ b/docs/transition_v1.rst
@@ -1,26 +1,26 @@
 Differences between v0 and v1
 =============================
-In september 2026, a major overhaul of xclim was implemented. While most of the functionalities were preserved, the internals were modified in many breaking ways. The goal was to modernize the lower-level API, aiming to ease development and favor contributions to xclim by the larger community. As of xclim v1.0, not all the planned changes are implemented, but only non-breaking ones remain.
+In September 2026, a major overhaul of `xclim` was implemented. While most of the functionalities were preserved, the internals were modified in many breaking ways. The goal was to modernize the lower-level API, aiming to ease development and favour contributions to `xclim` by the larger community. As of `xclim` v1.0, not all planned changes have been performed, but only non-breaking changes remain.
 
 This page aims to summarize the largest and most breaking changes, to help transition to v1.
 
 Indicators calculations return Datasets
 ---------------------------------------
-The ``as_dataset`` option of :py:func:`xclim.set_options` was changed to True, meaning that indicator calculations now return :py:class:`~xarray.Dataset` objects by default instead of :py:class:`~xarray.DataArray` objects or tuples of those. Re-enabling the previous behaviour is simply done with ``xclim.set_options(as_dataset=False)``, either globally or as a context.
+The ``as_dataset`` option of :py:func:`xclim.set_options` was changed to `True`, meaning that indicator calculations now return :py:class:`~xarray.Dataset` objects by default instead of :py:class:`~xarray.DataArray` objects or tuples of those. Re-enabling the previous behaviour is simply done with ``xclim.set_options(as_dataset=False)``, either globally or as a context.
 
 Renamed `indices` to `compute`
 ------------------------------
-The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the compute.
+The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the computation.
 
 Major reimplementation of the generic compute functions
 -------------------------------------------------------
-The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this github comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
+The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this GitHub comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
 
 The names of the indicators in :py:mod:`xclim.indicators` were largerly unchanged, but some arguments might have been renamed.
 
 Indicator constructor and "virtual submodules"
 ----------------------------------------------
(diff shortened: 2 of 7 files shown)


docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
+Differences between v0 and v1
+=============================
+In september 2026, a major overhaul of xclim was implemented. While most of the functionalities were preserved, the internals were modified in many breaking ways. The goal was to modernize the lower-level API, aiming to ease development and favor contributions to xclim by the larger community. As of xclim v1.0, not all the planned changes are implemented, but only non-breaking ones remain.
```
Zeitsperre:
```suggestion
In September 2026, a major overhaul of `xclim` was implemented. While most of the functionalities were preserved, the internals were modified in many breaking ways. The goal was to modernize the lower-level API, aiming to ease development and favour contributions to `xclim` by the larger community. As of `xclim` v1.0, not all planned changes have been performed, but only non-breaking changes remain.
```
🇨🇦

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+
+This page aims to summarize the largest and most breaking changes, to help transition to v1.
+
+Indicators calculations return Datasets
+---------------------------------------
+The ``as_dataset`` option of :py:func:`xclim.set_options` was changed to True, meaning that indicator calculations now return :py:class:`~xarray.Dataset` objects by default instead of :py:class:`~xarray.DataArray` objects or tuples of those. Re-enabling the previous behaviour is simply done with ``xclim.set_options(as_dataset=False)``, either globally or as a context.
```
Zeitsperre:
```suggestion
The ``as_dataset`` option of :py:func:`xclim.set_options` was changed to `True`, meaning that indicator calculations now return :py:class:`~xarray.Dataset` objects by default instead of :py:class:`~xarray.DataArray` objects or tuples of those. Re-enabling the previous behaviour is simply done with ``xclim.set_options(as_dataset=False)``, either globally or as a context.
```

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Renamed `indices` to `compute`
+------------------------------
+The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the compute.
```
Zeitsperre:
```suggestion
The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the computation.
```

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+-------------------------------------------------------
+The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this github comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
+
+The names of the indicators in :py:mod:`xclim.indicators` were largerly unchanged, but some arguments might have been renamed.
+
+Indicator constructor and "virtual submodules"
+----------------------------------------------
+The :py:mod:`xclim.core.indicator` submodule was refactored in hopes of making the internals of the :py:class:`~xclim.core.indicator.Indicator` object easier to maintain and extend. In addition, the "virtual submodules" concept was rewritten as the :py:class:`~xclim.core.collection.IndicatorCollection` object. These are dict-like structure holding a collection of indicators, usually created from a YAML file, like before.
```
Zeitsperre:
```suggestion
The :py:mod:`xclim.core.indicator` submodule was refactored in hopes of making the internals of the :py:class:`~xclim.core.indicator.Indicator` object easier to maintain and extend. In addition, the "virtual submodules" concept was rewritten as the :py:class:`~xclim.core.collection.IndicatorCollection` object. These are dictionary-like in structure, holding a collection of indicators and typically created from a YAML file, similar to the previous implementation.
```

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Renamed `indices` to `compute`
+------------------------------
+The ``xclim.indices`` module was renamed as ``xclim.compute``. This was part of a general move to stop using the words "index" or "indices" which were always quite amibugous with "indicator". In the documentation, functions that perform the actual computation, which were previously named "indices", are now usually named "compute functions" or "index-like compute functions". "Indicator" still denotes the larger object that performs all the checks and metadata formatting in adititon to the compute.
+
+Major reimplementation of the generic compute functions
+-------------------------------------------------------
+The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this github comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
```
Zeitsperre:
```suggestion
The :py:mod:`xclim.compute.generic` submodule holds compute functions (previously "indices") that are not variable-specific and can be reused in many indicators. All functions and their arguments were modified and renamed to follow a more systematic naming convention, heavily inspired by the work of `clix-meta <github.com/clix-meta/clix-meta/>`_. See `this GitHub comment <https://github.com/Ouranosinc/xclim/pull/2258#issuecomment-3473430173>`_ for an exhaustive list of the changes.
```

Zeitsperre (approved):
Looks great, thank you!

docs/notebooks/extendxclim.ipynb
```diff
@@ -224,7 +224,7 @@
    "outputs": [],
    "source": [
     "# An indicator based on tg_mean, but returning Celsius and fixed on annual resampling\n",
-    "tg_mean_c = xclim.atmos.tg_mean.__class__(\n",
+    "tg_mean_c = xclim.atmos.tg_mean.copy(\n",
```
juliettelavoie, line 227: i might add precision here that the var_name of the parent will be used

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+
+The names of the indicators in :py:mod:`xclim.indicators` were largerly unchanged, but some arguments might have been renamed.
+
+Indicator constructor and "virtual submodules"
+----------------------------------------------
+The :py:mod:`xclim.core.indicator` submodule was refactored in hopes of making the internals of the :py:class:`~xclim.core.indicator.Indicator` object easier to maintain and extend. In addition, the "virtual submodules" concept was rewritten as the :py:class:`~xclim.core.collection.IndicatorCollection` object. These are dictionary-like in structure, holding a collection of indicators and typically created from a YAML file, similar to the previous implementation.
+
+The main breaking changes are:
+
+	- Attribute `cf_attrs` was renamed `attrs` and is now a list of :py:class:`~xclim.core.indicator.Output` objects. These act like dictionaries of attributes, but have also 4 properties : `var_name`, `units`, `units_metadata` and `dimensionality`. This name change is to be done in the indicator's constructor and in the YAML files.
```
juliettelavoie: this is still a bit unclear to me.

attrs is a list of Output that hold a dict of attrs  (again?) and also 4 properties ?

maybe the top level should have another name than attrs ?

it is a little bit weird that in the yaml we can pass`var_name` to `attrs` and but that won't actually be in the attributes of the output dataset.

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+	- Function ``xclim.build_indicator_module_from_yaml`` is changed to :py:meth:`~xclim.core.collection.IndicatorCollection.from_yaml`.
+	- Indicators defined within a collection are not registered by default into the indicators registry, the collection is not added as module under ``xclim.indicators``.
```
juliettelavoie: maybe add something about the yaml no longer creating a module that can be called under xc.indicators.<MODULENAME>

docs/transition_v1.rst
juliettelavoie: also related and I don't know it this is too technical but it was a bit hard to figure out. now you have to do  `ind.attrs.var_name` instead of`ind.attrs['var_name']`

docs/transition_v1.rst
juliettelavoie: maybe  be explicit  that you can't recover the collection (like you could with the module) any other way than by keeping the output of `xclim.core.collection.IndicatorCollection.from_yaml`

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+	- When creating an indicator from an existing one, the ``var_name`` of the output of an indicator with a single output now only defaults to the indicator's identifier if the parent does not have a ``var_name``. Previously, the identifier would always be used as the ``var_name``, regardless of the parent's attributes.
+	- The :py:meth:`~xclim.core.indicator.Indicator.from_dict` method is deprecated. Use :py:meth:`~xclim.core.indicator.Indicator.copy` instead, calling it on the indicator object you want to subclass/copy. This new method doesn't parse ``compute`` function names given as string, instead pass the function directly.
+
```
juliettelavoie, line 43: in the xscen test, I made a few replacement that I am not 100% sure are correct. maybe these replacement should be make explicit here.

1. https://github.com/Ouranosinc/xscen/pull/764/changes#diff-c651c3c4d3347d798570b0e8d47af7daa0d6342c88fde2edf059c62dc5e029f2L352
2. https://github.com/Ouranosinc/xscen/pull/764/changes#diff-a6980326f4904b3ed1c4720f1aacbb99319b926ce4c66a319f3b9720d9903ec7L190

docs/transition_v1.rst
```diff
@@ -0,0 +1,33 @@
...
+	- Function ``xclim.build_indicator_module_from_yaml`` is changed to :py:meth:`~xclim.core.collection.IndicatorCollection.from_yaml`.
+	- Indicators defined within a collection are not registered by default into the indicators registry, the collection is not added as module under ``xclim.indicators``.
+	- When creating an indicator from an existing one, the ``var_name`` of the output of an indicator with a single output now only defaults to the indicator's identifier if the parent does not have a ``var_name``. Previously, the identifier would always be used as the ``var_name``, regardless of the parent's attributes.
+	- The :py:meth:`~xclim.core.indicator.Indicator.from_dict` method is deprecated. Use :py:meth:`~xclim.core.indicator.Indicator.copy` instead, calling it on the indicator object you want to subclass/copy. This new method doesn't parse ``compute`` function names given as string, instead pass the function directly.
```
juliettelavoie: the argument module is not longer needed

src/xclim/core/locales.py
```diff
@@ -28,10 +28,10 @@
         # ... and so on for other indicators...
     }
 
-Indicators are named by subclass identifier, the same as in the indicator registry (`xclim.core.indicators.registry`),
-but which can differ from the callable name. In this case, the indicator is called through
+Indicators are named by their identifier, the same as in the indicator registry (`xclim.core.indicators.registry`),
+but which can differ from the callable name. In the example above, the indicator is called through
 `atmos.daily_temperature_range_variability`, but its identifier is `DTRVAR`.
-Use the `ind.__class__.__name__` accessor to get its registry name.
+Use the `ind.identifier` accessor to get its registry name.
```
juliettelavoie: not a v1 change, but it is not obvious to me why callable name and identifier are different. could be cool to explain it somewhere. sorry if it already is an i missed it.

juliettelavoie (commented):
cool merci! a few comments to improve the documentation

docs/transition_v1.rst
aulemahal: Yeah... I thought naming this `outputs`, do you think it would make more sense ?

And `.attrs` could be a property of `Output`, so you would have: `ind.outputs[0].var_name` and  `ind.outputs[0].attrs`.  

`ind.outputs[0]` would still not be a dictionary itself, but it could act so, making `ind.outputs[0]['var_name']` work, if we wished. Indeed, this behaviour would be consistent with the fact we can initialize it with a single dict, even though the underlying object is slightly more complex.

juliettelavoie: yes I think naming it `ouputs` is clearer.

It makes sense to me that only `.var_name` works (like only `.attrs` and `.dimensionality` work.) I would just warn people that they should change that.

docs/transition_v1.rst
aulemahal: Commented on your first element and one below in the PR.

The second one indeed merits a note on this page.

docs/transition_v1.rst
aulemahal: It was never needed, ~but I can put a note indeed.~ 

Hmm, I think behaviour of `module` is mainly unchanged, except that virtual submodules are not a thing anymore, making it much less useful.

src/xclim/core/locales.py
aulemahal: Well, it is explained there and in the IndicatorCollection doc. 

The reasons they are different are:
- Some indicators have explicit names, but standardized identifiers (i.e. short names similar to some external source), the quoted example being one of them
- Indicators created dynamically don't have a "callable name"
- "callable" names are hard to get programmatically and fragile , see:
```python
température_moyenne = xc.atmos.tg_mean
```

I can add something in the docstring for `identifier` ?  With xclim v1, this issue is really only vital for the translations, as you can now pass "callable" names in the `base` field of a YAML definition.

aulemahal:
@Zeitsperre  I noticed just now that the automated API doc lists all parents classes of `Indicator` explicitly in its rendering of `xclim.core.indicator`: https://xclim--2424.org.readthedocs.build/en/2424/apidoc/xclim.core.html#module-xclim.core.indicator

That makes it much harder to understand I think... When I implemented it, I configured the User API page to list imported members and only added `Indicator` and none of it's parent classes : https://xclim--2424.org.readthedocs.build/en/2424/api.html#indicator-utilities

The issue here being that all internal links in the doc bring to the automated page, instead of the carefully (well erm,... you know what I mean) built one. I know this is an old issue, but couldn't we put `noindex` in the automated apidoc and link to the custom one ?

docs/transition_v1.rst
juliettelavoie: maybe needed is the wrong word.
in xscen, I was looking to replace things like `indicator_qs = xclim.core.indicator.Indicator.from_dict(
            data={"base": "tg_min", "parameters": {"freq": "QS-DEC"}},
            identifier="tg_min_qs",
            module="tests",
        )` and I was wondering where to put the `module="tests"` bit (even if it was, I realize now, actually useless). I guess we say that virtual modules don't exist so maybe there is no need for more guidance.

src/xclim/core/locales.py
juliettelavoie: I understand why we need both concept but, I guess my question is more why is the function not just called `atmos.dtrvar`, but I guess this might just be a question of taste and maybe. you could argue, clarity?

personnally, I can never remender the name of callable that goes with prcptot 😅 that might just be a me-problem

docs/transition_v1.rst
aulemahal: Ok yeah, I see that. I'll add a note!

src/xclim/core/locales.py
aulemahal: Yeah a question of clarity and now some kind of legacy I guess.

We didn't really tackle the indicator names issue in v1 finally. Because there are multiple incoherence... Like  I would rename `precip_accumulation` as `pr_sum` if I was to write it now, for example.

docs/transition_v1.rst
aulemahal: In #2426

Zeitsperre:
> @Zeitsperre I noticed just now that the automated API doc lists all parents classes of `Indicator` explicitly in its rendering of `xclim.core.indicator` ... That makes it much harder to understand I think ... When I implemented it, I configured the User API page to list imported members and only added `Indicator` and none of it's parent classes.

I think that can be adjusted. We already employ a two-step apidoc-generating process, so changing/remaking a particular module docs page is easily done.
 
> The issue here being that all internal links in the doc bring to the automated page, instead of the carefully (well erm,... you know what I mean) built one. I know this is an old issue, but couldn't we put `noindex` in the automated apidoc and link to the custom one ?

The issue we run into here is that there will be no way to directly reference (via intersphinx or URL) the autogenerated module documentation for "Indicator". If that's fine for us, I think we can easily make the change.

Zeitsperre:
OK here`s what I can think of for solving this:
- We move `xclim.core.indicator.py` into a new submodule (`xclim/core/indicator/__init__.py`) .

Followed by either:
- Specifying via sphinx-apidoc to only show specific classes (this is sounding a lot like a catered API page).
- Manually adding the `Indicator` class to the catered API page and ensure that that any indexed references for Indicator.

I'm a fan of keeping the full module-level mapping of `Indicator`, even if it seems confusing. Normal users shouldn't be going anywhere near that page; only developers who want to dive deep should care.

Does this make sense?

aulemahal:
> Manually adding the Indicator class to the catered API page and ensure that that any indexed references for Indicator.

If we do this, is it really needed to move `indicator` to it's own sub-submodule ? Also, that's already what I did, except for the "indexing" part. As the generated apidoc enables indexing, we disabled it on the catered API.  Could we flip this ? Enable indexing on the catered pages and disable it in the automated ones ?

Zeitsperre:
> Could we flip this ? Enable indexing on the catered pages and disable it in the automated ones ?

Not really a fan of this idea, but let's try it out.

Zeitsperre:
This seems to be working as intended now. We can no longer make references to anything under `apidoc`, beyond the pages themselves, as members are no longer indexed. If we want to talk about a specific function/class, it needs to be added to `api.rst` or somewhere else manually before it can be targeted.

aulemahal:
Nice thanks! I'll try my hand a reorganizing the "User API" section, it's a bit messy I find.

juliettelavoie (approved):
super!

aulemahal:
@Zeitsperre  commit [1822fb2](https://github.com/Ouranosinc/xclim/pull/2424/commits/1822fb2702195bf8695c3ce19af1a5ac5cca9ccf) moves things around a lot to better divide the "user api" page that was quite long and make its TOC appear cleaner in the left panel. Also played a bit with headers and moved some pages to the api: "checks" and "internationalization".

Zeitsperre:
@aulemahal The new User API section is so much better organized, thanks for your work on this!

PR #2425, opened 2026-09-14
Enforce mypy hook

### LINKED ISSUE #1719: Adopt `mypy` and stronger enforcement of `pylint`

ludwiglierhammer:
### Pull Request Checklist:
    - This PR fixes #1719

* This PR is a attempt to enforce the mypy hook for strict type checking

Hopefully not 😄 
### Other information:

Hey guys, 

I've created this PR to try and get the mypy hook enforced. There is still lots of work to be done. I just wanted to let you know that I'm trying to work on it. I hope you find my approach interesting.

Let's give it a try 🚀!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,6 +63,12 @@ repos:
     rev: v2.16
     hooks:
       - id: vulture
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v2.3.0
+    hooks:
+      - id: mypy
+        exclude: ^tests/|^docs
+        additional_dependencies: [ 'numpy>=1.21' ]      
   - repo: https://github.com/kynan/nbstripout
     rev: 0.9.1
     hooks:
--- a/src/xclim/compute/_agro.py
+++ b/src/xclim/compute/_agro.py
@@ -230,18 +230,26 @@ def huglin_index(
     if not isinstance(freq, str):
         raise TypeError("Freq must be a string.")
 
-    _tas = convert_units_to(tas, "degC")
-    _tasmax = convert_units_to(tasmax, "degC")
-    _thresh = convert_units_to(thresh, "degC")
+    _tas: xarray.DataArray = convert_units_to(tas, "degC")
+    _tasmax: xarray.DataArray = convert_units_to(tasmax, "degC")
+    _thresh: float = convert_units_to(thresh, "degC")
 
     if lat is None:
         lat = _gather_lat(tas)
 
     k: int | xarray.DataArray = 1
     k_aggregated: xarray.DataArray | None = None
-    if (method := method.lower()) in ["huglin", "interpolated"]:
+    
+    method = method.lower()
+    
+    if method not in ["huglin", "interpolated", "jones"]:
+        raise NotImplementedError(
+            "Method is not implemented. Only 'huglin', 'interpolated', and 'jones' are supported."
+        )        
+    
+    if method in ["huglin", "interpolated"]:
         k = huglin_day_length_latitude_coefficient(lat, method=method, cap_value=cap_value)
-    elif method.lower() == "jones":
+    elif method == "jones":
         k_aggregated = jones_day_length_latitude_coefficient(
             dates=tas.time,
             lat=lat,
@@ -249,10 +257,6 @@ def huglin_index(
             start_date=start_date,
             end_date=end_date,
             freq=freq,
-        )
-    else:
-        raise NotImplementedError(
-            "Method is not implemented. Only 'huglin', 'interpolated', and 'jones' are supported."
         )
 
     hi: xarray.DataArray = (((_tas + _tasmax) / 2) - _thresh).clip(min=0) * k
@@ -391,10 +395,10 @@ def biologically_effective_degree_days(
     if not isinstance(freq, str):
         raise TypeError("Freq must be a string.")
 
-    _tasmin = convert_units_to(tasmin, "degC")
-    _tasmax = convert_units_to(tasmax, "degC")
-    thresh_tasmin_deg = convert_units_to(thresh_tasmin, "degC")
-    max_daily_degree_days = convert_units_to(max_daily_degree_days, "degC")
+    _tasmin: xarray.DataArray = convert_units_to(tasmin, "degC")
+    _tasmax: xarray.DataArray = convert_units_to(tasmax, "degC")
+    _thresh_tasmin: float = convert_units_to(thresh_tasmin, "degC")
+    _max_daily_degree_days: float = convert_units_to(max_daily_degree_days, "degC")
 
     k: int | xarray.DataArray = 1
     k_aggregated: xarray.DataArray | None = None
@@ -438,8 +442,8 @@ def biologically_effective_degree_days(
             "Method is not implemented. Only 'gladstones', 'huglin', 'icclim', 'interpolated', and 'jones' are supported."
         )
 
-    bedd: xarray.DataArray = ((((_tasmin + _tasmax) / 2) - thresh_tasmin_deg).clip(min=0) * k + tr_adj).clip(
-        max=max_daily_degree_days
+    bedd: xarray.DataArray = ((((_tasmin + _tasmax) / 2) - _thresh_tasmin).clip(min=0) * k + tr_adj).clip(
+        max=_max_daily_degree_days
     )
     bedd = select_time(bedd, date_bounds=(start_date, end_date), include_bounds=(True, False)).resample(time=freq).sum()
     if k_aggregated is not None:
@@ -643,8 +647,8 @@ def dryness_index(  # numpydoc ignore=SS05
 
     # Resample all variables to monthly totals in mm units.
     evspsblpot = amount2lwethickness(rate2amount(evspsblpot), out_units="mm").resample(time="MS").sum()
-    pr = amount2lwethickness(rate2amount(pr), out_units="mm").resample(time="MS").sum()
-    wo = convert_units_to(wo, "mm")
+    _pr: xarray.DataArray = amount2lwethickness(rate2amount(pr), out_units="mm").resample(time="MS").sum()
+    _wo: float = convert_units_to(wo, "mm")
 
     # Different potential evapotranspiration rates for northern hemisphere and southern hemisphere.
     # wo_adjustment is the initial soil moisture rate at beginning of season.
@@ -661,7 +665,7 @@ def dryness_index(  # numpydoc ignore=SS05
 
     has_north, has_south = False, False
     if lat is None:
-        lat = _gather_lat(pr)
+        lat = _gather_lat(_pr)
     if isinstance(lat, xarray.DataArray):
         if (lat >= 0).any():
             has_north = True
@@ -689,7 +693,7 @@ def dryness_index(  # numpydoc ignore=SS05
     k = adjustment.sel(month=evspsblpot.time.dt.month)
 
     # Drop all pr outside seasonal bounds
-    pr_masked = (k > 0) * pr
+    pr_masked = (k > 0) * _pr
 
     # Potential transpiration of the vineyard
     t_v = evspsblpot * k
@@ -701,13 +705,13 @@ def dryness_index(  # numpydoc ignore=SS05
         * (pr_masked / 5).clip(max=evspsblpot.time.dt.daysinmonth)
     )
 
-    di_north: xarray.DataArray | None = None
-    di_south: xarray.DataArray | None = None
+    di_north: xarray.DataArray #| None = None
+    di_south: xarray.DataArray #| None = None
     # Dryness index
     if has_north:
-        di_north = wo + (pr_masked - t_v - e_s).resample(time="YS-JAN").sum()
+        di_north = _wo + (pr_masked - t_v - e_s).resample(time="YS-JAN").sum()
     if has_south:
-        di_south = wo + (pr_masked - t_v - e_s).resample(time="YS-JUL").sum()
+        di_south = _wo + (pr_masked - t_v - e_s).resample(time="YS-JUL").sum()
         # Shift time for Southern Hemisphere to allow for concatenation with Northern Hemisphere
         di_south = di_south.shift(time=1).isel(time=slice(1, None))
         di_south["time"] = di_south.indexes["time"].shift(-6, "MS")
@@ -892,9 +896,9 @@ def rain_season(
     """
     # Unit conversion.
     pram = rate2amount(pr, out_units="mm")
-    thresh_wet_start = convert_units_to(thresh_wet_start, pram)
-    thresh_dry_start = convert_units_to(thresh_dry_start, pram)
-    thresh_dry_end = convert_units_to(thresh_dry_end, pram)
+    _thresh_wet_start: float = convert_units_to(thresh_wet_start, pram)
+    _thresh_dry_start: float = convert_units_to(thresh_dry_start, pram)
+    _thresh_dry_end: float = convert_units_to(thresh_dry_end, pram)
 
     # should we flag date_min_end  < date_max_start?
     def _get_first_run(run_positions, start_date, end_date):
@@ -908,14 +912,14 @@ def rain_season(
         _pram = select_time(_pram, date_bounds=(date_min_start, last_doy))
 
         # First condition: Start with enough precipitation
-        da_start = _pram.rolling({"time": window_wet_start}).sum() >= thresh_wet_start
+        da_start = _pram.rolling({"time": window_wet_start}).sum() >= _thresh_wet_start
 
         # Second condition: No dry period after
         if method_dry_start == "per_day":
-            da_stop = _pram <= thresh_dry_start
+            da_stop = _pram <= _thresh_dry_start
             window_dry = window_dry_start
         elif method_dry_start == "total":
-            da_stop = _pram.rolling({"time": window_dry_start}).sum() <= thresh_dry_start
+            da_stop = _pram.rolling({"time": window_dry_start}).sum() <= _thresh_dry_start
             # equivalent to rolling forward in time instead, i.e. end date will be at beginning of dry run
             da_stop = da_stop.shift({"time": -(window_dry_start - 1)}, fill_value=False)
             window_dry = 1
@@ -932,10 +936,10 @@ def rain_season(
     # FIXME: This function mixes local and parent-level variables. It should be refactored.
     def _get_first_run_end(_pram):
         if method_dry_end == "per_day":
-            da_stop = _pram <= thresh_dry_end
+            da_stop = _pram <= _thresh_dry_end
             run_positions = rl.rle(da_stop) >= window_dry_end
         elif method_dry_end == "total":
-            run_positions = _pram.rolling({"time": window_dry_end}).sum() <= thresh_dry_end
+            run_positions = _pram.rolling({"time": window_dry_end}).sum() <= _thresh_dry_end
         else:
             raise ValueError(f"Unknown method_dry_end: {method_dry_end}.")
         return _get_first_run(run_positions, date_min_end, date_max_end)
@@ -1350,9 +1354,9 @@ def effective_growing_degree_days(
     ----------
     :cite:cts:`bootsma_impacts_2005`
     """
-    _tasmax = convert_units_to(tasmax, "degC")
-    _tasmin = convert_units_to(tasmin, "degC")
-    _thresh = convert_units_to(thresh, "degC")
+    _tasmax: xarray.DataArray = convert_units_to(tasmax, "degC")
+    _tasmin: xarray.DataArray = convert_units_to(tasmin, "degC")
+    _thresh: float = convert_units_to(thresh, "degC")
 
     tas = (_tasmin + _tasmax) / 2
     tas.attrs["units"] = "degC"
--- a/src/xclim/compute/_anuclim.py
+++ b/src/xclim/compute/_anuclim.py
@@ -429,9 +429,9 @@ def prcptot(pr: xarray.DataArray, thresh: Quantified = "0 mm/d", freq: Freq = "Y
     xarray.DataArray, [length]
        Total {freq} precipitation.
     """
-    thresh = convert_units_to(thresh, pr, context="hydro")
-    pram: xarray.DataArray = rate2amount(pr.where(pr >= thresh, 0))
-    pram = pram.resample(time=freq).sum().assign_attrs(units=pram.units)
+    _thresh: float = convert_units_to(thresh, pr, context="hydro")
+    _pr: xarray.DataArray = rate2amount(pr.where(pr >= thresh, 0))
+    pram = _pr.resample(time=freq).sum().assign_attrs(units=_pr.units)
     return pram
 
 
@@ -565,7 +565,8 @@ def _to_quarter(
             # Ensure units are back to a "rate" for rate2amount below
             pram = rate2amount(ts_var)
             ts_var = statistics(pram, statistic="sum", freq="7D")
-            ts_var = convert_units_to(ts_var, "mm", context="hydro").assign_attrs(units="mm/week")
+            ts_var = convert_units_to(ts_var, "mm", context="hydro")
+            ts_var = ts_var.assign_attrs(units="mm/week")
         freq_upper = "W"
     if freq_upper.startswith("W"):
         window = 13
--- a/src/xclim/compute/_multivariate.py
+++ b/src/xclim/compute/_multivariate.py
@@ -689,11 +689,11 @@ def heat_wave_frequency(
     ----------
     :cite:cts:`casati_regional_2013,robinson_definition_2001`
     """
-    thresh_tasmax = convert_units_to(thresh_tasmax, tasmax)
-    thresh_tasmin = convert_units_to(thresh_tasmin, tasmin)
+    _thresh_tasmax: float = convert_units_to(thresh_tasmax, tasmax)
+    _thresh_tasmin: float = convert_units_to(thresh_tasmin, tasmin)
 
     constrain = (">", ">=")
-    cond = (compare(tasmin, op, thresh_tasmin, constrain)) & (compare(tasmax, op, thresh_tasmax, constrain))
+    cond = (compare(tasmin, op, _thresh_tasmin, constrain)) & (compare(tasmax, op, _thresh_tasmax, constrain))
 
     out = rl.resample_and_rl(
         cond,
@@ -770,11 +770,11 @@ def heat_wave_max_length(
     ----------
     :cite:cts:`casati_regional_2013,robinson_definition_2001`
     """
-    thresh_tasmax = convert_units_to(thresh_tasmax, tasmax)
-    thresh_tasmin = convert_units_to(thresh_tasmin, tasmin)
+    _thresh_tasmax: float = convert_units_to(thresh_tasmax, tasmax)
+    _thresh_tasmin: float = convert_units_to(thresh_tasmin, tasmin)
 
     constrain = (">", ">=")
-    cond = (compare(tasmin, op, thresh_tasmin, constrain)) & (compare(tasmax, op, thresh_tasmax, constrain))
+    cond = (compare(tasmin, op, _thresh_tasmin, constrain)) & (compare(tasmax, op, _thresh_tasmax, constrain))
     out = rl.resample_and_rl(
         cond,
         resample_before_rl,
@@ -839,11 +839,11 @@ def heat_wave_total_length(
     -----
     See notes and references of `heat_wave_max_length`
     """
-    thresh_tasmax = convert_units_to(thresh_tasmax, tasmax)
-    thresh_tasmin = convert_units_to(thresh_tasmin, tasmin)
+    _thresh_tasmax: float = convert_units_to(thresh_tasmax, tasmax)
+    _thresh_tasmin: float = convert_units_to(thresh_tasmin, tasmin)
 
     constrain = (">", ">=")
-    cond = compare(tasmin, op, thresh_tasmin, constrain) & compare(tasmax, op, thresh_tasmax, constrain)
+    cond = compare(tasmin, op, _thresh_tasmin, constrain) & compare(tasmax, op, _thresh_tasmax, constrain)
     out = rl.resample_and_rl(
         cond,
         resample_before_rl,
@@ -920,6 +920,7 @@ def liquid_precip_ratio(
 
        R_j = \frac{PR^{\mathrm{liquid}}_{j}}{PR_j}
     """
+    rain: xarray.DataArray
     if prra is not None:
         rain = convert_units_to(prra, pr)
     elif prsn is not None:
@@ -1174,10 +1175,10 @@ def high_precip_low_temp(
     >>> tasmin = xr.open_dataset(path_to_tasmin_file).tasmin
     >>> hplt = high_precip_low_temp(pr, tas=tasmin, pr_thresh="10 mm/d", tas_thresh="-0.2 degC")
     """
-    pr_thresh = convert_units_to(pr_thresh, pr, context="hydro")
-    tas_thresh = convert_units_to(tas_thresh, tas)
+    _pr_thresh: float = convert_units_to(pr_thresh, pr, context="hydro")
+    _tas_thresh: float = convert_units_to(tas_thresh, tas)
 
-    cond = (pr >= pr_thresh) * (tas < tas_thresh) * 1
+    cond = (pr >= _pr_thresh) * (tas < _tas_thresh) * 1
     out = cond.resample(time=freq).sum(dim="time")
     return to_agg_units(out, pr, "count", deffreq="D")
 
@@ -1231,8 +1232,8 @@ def days_over_precip_thresh(
     >>> p75 = pr.quantile(0.75, dim="time", keep_attrs=True)
     >>> r75p = days_over_precip_thresh(pr, p75)
     """
-    _pr_per = convert_units_to(pr_per, pr, context="hydro")
-    _thresh = convert_units_to(thresh, pr, context="hydro")
+    _pr_per: xarray.DataArray = convert_units_to(pr_per, pr, context="hydro")
+    _thresh: float = convert_units_to(thresh, pr, context="hydro")
 
     tp = _pr_per.where(_pr_per > _thresh, _thresh)
     if "dayofyear" in pr_per.coords:
@@ -1285,11 +1286,11 @@ def fraction_over_precip_thresh(
     xarray.DataArray, [dimensionless]
         Fraction of precipitation over threshold during wet days.
     """
-    pr_per = convert_units_to(pr_per, pr, context="hydro")
-    thresh = convert_units_to(thresh, pr, context="hydro")
+    _pr_per: xarray.DataArray = convert_units_to(pr_per, pr, context="hydro")
+    _thresh: float = convert_units_to(thresh, pr, context="hydro")
 
-    tp = pr_per.where(pr_per > thresh, thresh)
-    if "dayofyear" in pr_per.coords:
+    tp = _pr_per.where(_pr_per > _thresh, _thresh)
+    if "dayofyear" in _pr_per.coords:
         # Create time series out of doy values.
         tp = resample_doy(tp, pr)
 
@@ -1370,7 +1371,7 @@ def tg10p(
     tas_per: xarray.DataArray,
     freq: Freq = "YS",
     bootstrap: bool = False,
-    condition: Literal[">", ">=", "gt", "ge"] = "<",
+    condition: Literal["<", "<=", "lt", "le"] = "<",
 ) -> xarray.DataArray:
     r"""
     Number of days with daily mean temperature below the 10th percentile.
@@ -1544,7 +1545,7 @@ def tx90p(
     tasmax_per: xarray.DataArray,
     freq: Freq = "YS",
     bootstrap: bool = False,
-    condition: Literal["<", "<=", "lt", "le"] = ">",
+    condition: Literal[">", ">=", "gt", "ge"] = ">",
 ) -> xarray.DataArray:
     r"""
     Number of days with daily maximum temperature over the 90th percentile.
@@ -1879,8 +1880,8 @@ def blowing_snow(
     xarray.DataArray
         Number of days when snowfall and wind speeds are above respective thresholds.
     """
-    snd_thresh = convert_units_to(snd_thresh, snd)
-    sfcWind_thresh = convert_units_to(sfcWind_thresh, sfcWind)
+    _snd_thresh: float = convert_units_to(snd_thresh, snd)
+    _sfcWind_thresh: float = convert_units_to(sfcWind_thresh, sfcWind)
 
     # Net snow accumulation over the last `window` days
     snow = snd.diff(dim="time").rolling(time=window, center=False).sum()
@@ -1888,7 +1889,7 @@ def blowing_snow(
     sfcWind = select_time(sfcWind, **indexer)
 
     # Blowing snow conditions
-    cond = (snow >= snd_thresh) * (sfcWind >= sfcWind_thresh) * 1
+    cond = (snow >= _snd_thresh) * (sfcWind >= _sfcWind_thresh) * 1
 
     out = cond.resample(time=freq).sum(dim="time")
     out = out.assign_attrs(units=to_agg_units(out, snd, "count", deffreq="D"))
--- a/src/xclim/compute/_threshold.py
+++ b/src/xclim/compute/_threshold.py
@@ -290,9 +290,9 @@ def cold_spell_max_length(
     xarray.DataArray, [days]
         The {freq} longest spell in cold periods of minimum {window} days.
     """
-    thresh = convert_units_to(thresh, tas)
+    _thresh: float = convert_units_to(thresh, tas)
 
-    cond = compare(tas, op, thresh, constrain=("<", "<="))
+    cond = compare(tas, op, _thresh, constrain=("<", "<="))
     max_l = rl.resample_and_rl(
         cond,
         resample_before_rl,
@@ -341,9 +341,9 @@ def cold_spell_total_length(
     xarray.DataArray, [days]
         The {freq} total number of days in cold periods of minimum {window} days.
     """
-    thresh = convert_units_to(thresh, tas)
+    _thresh: float = convert_units_to(thresh, tas)
 
-    cond = compare(tas, op, thresh, constrain=("<", "<="))
+    cond = compare(tas, op, _thresh, constrain=("<", "<="))
     out = rl.resample_and_rl(
         cond,
         resample_before_rl,
@@ -859,21 +859,21 @@ def degree_days_above_approximation(
     ----------
     :cite:cts:`spinoni_2018`
     """
-    thresh = convert_units_to(thresh, tas)
-    tasmax = convert_units_to(tasmax, tas)
-    tasmin = convert_units_to(tasmin, tas)
+    _thresh: float = convert_units_to(thresh, tas)
+    _tasmax: xarray.DataArray = convert_units_to(tasmax, tas)
+    _tasmin: xarray.DataArray = convert_units_to(tasmin, tas)
 
     cdd = xarray.where(
-        tasmax < thresh,
+        _tasmax < _thresh,
         0,
         xarray.where(
-            tasmin < thresh,
+            _tasmin < _thresh,
             xarray.where(
-                tas <= thresh,
-                (tasmax - thresh) / 4,
-                (tasmax - thresh) / 2 - (thresh - tasmin) / 4,
+                tas <= _thresh,
+                (_tasmax - _thresh) / 4,
+                (_tasmax - _thresh) / 2 - (_thresh - _tasmin) / 4,
             ),
-            tas - thresh,
+            tas - _thresh,
         ),
     )
     cdd = cdd.resample(time=freq).sum(dim="time")
@@ -1535,9 +1535,9 @@ def frost_free_spell_max_length(
     xarray.DataArray, [days]
         The {freq} longest spell in frost-free periods of minimum {window} days.
     """
-    thresh = convert_units_to(thresh, tasmin)
+    _thresh: float = convert_units_to(thresh, tasmin)
 
-    cond = compare(tasmin, op, thresh, constrain=(">", ">="))
+    cond = compare(tasmin, op, _thresh, constrain=(">", ">="))
     max_l = rl.resample_and_rl(
         cond,
         resample_before_rl,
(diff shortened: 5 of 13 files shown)


Zeitsperre:
One major hurdle to this is that we need to re-examine how to handle the `Quantified` complex type: https://github.com/Ouranosinc/xclim/issues/1726.

From my understanding, if we were to be using generic types in the call signatures and returns, this would work fine:

```python
Quantified = TypeVar("Quantified", xr.DataArray, str, Quantity)

def some_func(da_or_str: Quantified) -> Quantified:

    ...

    out: Quantified = _something(Quantified)
    return out
```
V1.0 is definitely a great time to start on this.

src/xclim/compute/generic.py
```diff
@@ -849,7 +849,7 @@ def season(
     condition: Condition,
     thresh: Quantified,
     window: int,
-    aspect: Literal["start", "end", "length"] | Sequence[Literal["start", "end", "length"]],
+    aspect: Literal["start", "end", "length"],  # | Sequence[Literal["start", "end", "length"]],
```
ludwiglierhammer, line 854: I do not think this could be a Sequence.

ludwiglierhammer:
Why did you remove threshold test from `ensembles._robustness.robustness_fractions`:

```
    elif test == "threshold":
        abs_thresh = kwargs.get("abs_thresh")
        rel_thresh = kwargs.get("rel_thresh")
        if abs_thresh is not None and rel_thresh is None:
            changed = abs(delta) > abs_thresh
            test_params = {"abs_thresh": abs_thresh}
        elif rel_thresh is not None and abs_thresh is None and isinstance(ref, xr.DataArray):
            changed = abs(delta / ref.mean("time")) > rel_thresh
            test_params = {"rel_thresh": rel_thresh}
        else:
            raise ValueError("One and only one of abs_thresh or rel_thresh must be given if test='threshold'.")
        pvals = None
```

ludwiglierhammer:
I've fixed the numpydoc-validation hook. I've also fixed a lot of the mypy issues. I'll probably need some help with the remaining ones. Unfortunately, I've broken the test suite; I'll have another look at it tommorow.

src/xclim/compute/generic.py
aulemahal: You are right, I must have broken that feature in v1 but forgot to update the signature and doc!

aulemahal:
> Why did you remove threshold test from ensembles._robustness.robustness_fractions:

It was moved to its own function in #2411, see here : https://github.com/Ouranosinc/xclim/blob/a1c014efcee68fb6e475be2cd2c95df521b7c633/src/xclim/ensembles/_robustness.py#L576

ludwiglierhammer:
> > Why did you remove threshold test from ensembles._robustness.robustness_fractions:
> 
> It was moved to its own function in #2411, see here :
> 
> https://github.com/Ouranosinc/xclim/blob/a1c014efcee68fb6e475be2cd2c95df521b7c633/src/xclim/ensembles/_robustness.py#L576

Sorry, my bad 🙈. I should have read all of the code.

ludwiglierhammer:
I've almost fixed the test suite. There are still two errors left. What they have in common is that something went wrong when formatting the percentiles for the NetCDF attributes:

```
      def test_days_over_precip_doy_thresh(open_dataset):
          pr = open_dataset("ERA5/daily_surface_cancities_1990-1993.nc").pr
          per = percentile_doy(pr, window=5, per=80)
      
          out1 = atmos.days_over_precip_doy_thresh(pr, per)
          np.testing.assert_array_equal(out1.days_over_precip_doy_thresh[1, :, 0], np.array([81, 60, 68, 78]))
      
          out2 = atmos.days_over_precip_doy_thresh(pr, per, thresh="2 mm/d")
          np.testing.assert_array_equal(out2.days_over_precip_doy_thresh[1, :, 0], np.array([80, 59, 66, 78]))
      
          assert "Only days with at least 2 mm/d are counted." in out2.days_over_precip_doy_thresh.description
  >       assert "[80]th percentile" in out2.days_over_precip_doy_thresh.attrs["description"]
  E       assert '[80]th percentile' in "Annual number of days with precipitation above the [80.]th daily percentile. Only days with at least 2 mm/d are count...entered on each calendar day in the ['1990-01-01', '1993-12-31'] period, is used to compute the [80.]th percentile(s)."
```

I would be happy if you could have a look at it. Thanks!

ludwiglierhammer:
I would be happy about some help fixing mypy issues in the following files:

- src\xclim\core\utils.py
- src\xclim\core\bootstrapping.py
- src\xclim\core\missing.py
- src\xclim\core\indicator\_indicator.py
- src\xclim\__init__.py

Thank you!

ludwiglierhammer:
I have managed to resolve all the mypy issues and fix the test suite. I look forward to your review.

ludwiglierhammer:
> One major hurdle to this is that we need to re-examine how to handle the `Quantified` complex type: #1726.
> 
> From my understanding, if we were to be using generic types in the call signatures and returns, this would work fine:
> 
> ```python
> Quantified = TypeVar("Quantified", xr.DataArray, str, Quantity)
> 
> def some_func(da_or_str: Quantified) -> Quantified:
> 
>     ...
> 
>     out: Quantified = _something(Quantified)
>     return out
> ```
> 
> V1.0 is definitely a great time to start on this.

I could have a look at this next week. I hope you all have a nice weekend.

aulemahal:
Vielen dank ! I will look at this next week, but it sounds like lots of great work.

src/xclim/compute/_multivariate.py
```diff
@@ -1370,7 +1371,7 @@ def tg10p(
     tas_per: xarray.DataArray,
     freq: Freq = "YS",
     bootstrap: bool = False,
-    condition: Literal[">", ">=", "gt", "ge"] = "<",
+    condition: Literal["<", "<=", "lt", "le"] = "<",
```
Zeitsperre, line 1374: Good catch!

src/xclim/compute/_agro.py
```diff
@@ -1296,7 +1324,7 @@ def effective_growing_degree_days(
     *,
     thresh: Quantified = "5 degC",
     method: Literal["bootsma", "qian"] = "bootsma",
-    after_date: DayOfYearStr = "07-01",
+    after_date: DayOfYearStr | None = None,
```
Zeitsperre, line 1327: I like the `default` option you used elsewhere:
```suggestion
    after_date: DayOfYearStr | Literal["default"] = "default",
```

src/xclim/compute/_threshold.py
```diff
@@ -1226,7 +1232,7 @@ def frost_season_length(
         Minimum daily temperature.
     window : int
         Minimum number of days with temperature below threshold to mark the beginning and end of frost season.
-    mid_date : str, optional
+    mid_date : str, optional, defaults to '01-01'
         The date must be included in the season. It is the earliest the end of the season can be.
         ``None`` removes that constraint.
```
Zeitsperre, line 1237:
```suggestion
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1311,7 +1319,7 @@ def frost_free_season_start(
         Threshold temperature on which to base evaluation.
     window : int
         Minimum number of days with temperature above/under the threshold to start/end the season.
-    mid_date : DayOfYearStr, optional
+    mid_date : DayOfYearStr, optional, defaults to '07-01'
         A date that must be included in the season. `None` removes that constraint.
```
Zeitsperre, line 1323:
```suggestion
        A date that must be included in the season. It is the earliest the end of the season can be.
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1373,7 +1383,7 @@ def frost_free_season_end(
         Threshold temperature on which to base evaluation.
     window : int
         Minimum number of days with temperature above/under the threshold to start/end the season.
-    mid_date : DayOfYearStr, optional
+    mid_date : DayOfYearStr, optional, defaults to '07-01'
         A date what must be included in the season. `None` removes that constraint.
```
Zeitsperre, line 1387:
```suggestion
        A date that must be included in the season. It is the earliest the end of the season can be.
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1442,7 +1454,7 @@ def frost_free_season_length(
...
```
Zeitsperre, line 1458:
```suggestion
        A date that must be included in the season. It is the earliest the end of the season can be.
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1572,7 +1586,7 @@ def last_spring_frost(
         Threshold temperature on which to base evaluation.
     op : {"<", "lt", "<=", "le"}
         Comparison operation. Default: "<".
-    before_date : str,
+    before_date : str, optional, defaults to '07-01'
         Date of the year before which to look for the final frost event. Should have the format '%m-%d'.
```
Zeitsperre, line 1590:
```suggestion
        Date of the year before which to look for the final frost event. Should have the format '%m-%d'.
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1632,7 +1649,7 @@ def first_day_temperature_below(
         Threshold temperature on which to base evaluation.
     op : {"<", "lt", "<=", "le"}
         Comparison operation. Default: ">".
-    after_date : str
+    after_date : str, optional, defaults to '07-01'
         Date of the year after which to look for the first event. Should have the format '%m-%d'.
```
Zeitsperre, line 1653:
```suggestion
        Date of the year after which to look for the first event. Should have the format '%m-%d'.
        Setting `None` removes that constraint.
```

src/xclim/compute/_threshold.py
```diff
@@ -1686,7 +1705,7 @@ def first_day_temperature_above(
         Threshold temperature on which to base evaluation.
     op : {">", "gt", ">=", "ge"}
         Comparison operation. Default: ">".
-    after_date : str
+    after_date : str, optional, default to '01-01'
         Date of the year after which to look for the first event. Should have the format '%m-%d'.
```
Zeitsperre, line 1709:
```suggestion
        Date of the year after which to look for the first event. Should have the format '%m-%d'.
        Setting `None` removes that constraint.
```

src/xclim/compute/clix.py
Zeitsperre, line 400:
```suggestion
       Latest day the occurrence can be found.
       Added to definition to support "lsf" without indexing.
       Setting `None` removes this constraint.
```

src/xclim/compute/converters.py
```diff
@@ -1816,7 +1817,16 @@ def _get_D_from_M(time):  # noqa: N802
...
-def fao_allen98(net_radiation, tas, wind, es, ea, delta_svp, gamma, G="0 MJ m-2 day-1"):
+def fao_allen98(
+    net_radiation: xr.DataArray,
+    tas: xr.DataArray,
+    wind: xr.DataArray,
+    es: xr.DataArray,
+    ea: xr.DataArray,
+    delta_svp: xr.DataArray,
+    gamma: xr.DataArray | str,
+    G: str | None = "0 MJ m-2 day-1",
```
Zeitsperre, line 1828: I don't think `G` can be `None` here. It shouldn't be optional as the docstring suggests.

src/xclim/compute/generic.py
```diff
@@ -1236,7 +1236,7 @@ def _get_days(_bound, _group, _base_time):
     if freq is None:
         frequencies = []
         for bound in [start, end]:
-            if isinstance(bound, xr.DataArray) and "time" in bound.coords:
+            if isinstance(bound, xr.DataArray) and hasattr(bound, "time"):
```
Zeitsperre, line 1239: Must `time` explicitly be a coordinate or is it enough that the variable for it is present?

src/xclim/compute/helpers.py
```diff
@@ -196,7 +196,9 @@ def spell_mask(
     # Checks
     if not isinstance(data, xr.DataArray):
         # thus a sequence
-        if np.isscalar(thresh) or isinstance(thresh, xr.DataArray) or len(data) != len(thresh):
+        if isinstance(thresh, (float, int)) or isinstance(thresh, xr.DataArray):
```
Zeitsperre, line 199: This could be condensed, no?
```suggestion
        if isinstance(thresh, (float, int, xr.DataArray)) or (len(data) != len(thresh)):
```

src/xclim/compute/run_length.py
```diff
@@ -1149,7 +1150,7 @@ def run_end_after_date(
         Input N-dimensional DataArray (boolean).
     window : int
         Minimum duration of consecutive run to accumulate values.
-    date : str
+    date : str, optional, defaults to '07-01'
         The date after which to look for the end of a run.
```
Zeitsperre, line 1154:
```suggestion
        The date after which to look for the end of a run.
        Setting `None` removes this constraint.
```

src/xclim/compute/run_length.py
```diff
@@ -1203,7 +1207,7 @@ def first_run_after_date(
         Input N-dimensional DataArray (boolean).
     window : int
         Minimum duration of consecutive run to accumulate values.
-    date : DayOfYearStr, optional
+    date : DayOfYearStr, optional, defaults to '07-01'
         The date after which to look for the run.
```
Zeitsperre, line 1211:
```suggestion
        The date after which to look for the run.
        Setting `None` removes this constraint.
```

src/xclim/compute/run_length.py
```diff
@@ -1246,7 +1253,7 @@ def last_run_before_date(
         Input N-dimensional DataArray (boolean).
     window : int
         Minimum duration of consecutive run to accumulate values.
-    date : DayOfYearStr
+    date : DayOfYearStr, optional, defaults to '07-01'
         The date before which to look for the last event.
```
Zeitsperre, line 1257:
```suggestion
        The date before which to look for the last event.
        Setting `None` removes this constraint.
```

src/xclim/compute/run_length.py
Zeitsperre, line 1300:
```suggestion
        The date before which to look for the run.
        Setting `None` removes this constraint.
```

src/xclim/compute/stats.py
```diff
@@ -232,7 +232,7 @@ def parametric_quantile(
         Distribution parameters returned by the `fit` function.
         The array should have dimension `dparams` storing the distribution parameters,
         and attribute `scipy_dist`, storing the name of the distribution.
-    q : float or Sequence of float
+    q : float, Sequence of float or np.ndarray of float
```
Zeitsperre, line 235: Huge nitpick, but I'm not sure about placing commas in the type description; Makes me think it might be a tuple. 
```suggestion
    q : float or Sequence of float or np.ndarray of float
```

src/xclim/core/utils.py
```diff
@@ -186,6 +173,10 @@ def load_module(path: os.PathLike, name: str | None = None) -> ModuleType:
     """
     path = Path(path)
     spec = importlib.util.spec_from_file_location(name or path.stem, path)
+    if spec is None:
+        raise ValueError("spec is not a valid ModuleSpec type but None.")
+    if spec.loader is None:
+        raise ValueError("spec.loader is not a valid ModuleSpec type but None.")
```
Zeitsperre, line 179:
```suggestion
    if spec is None:
        raise ValueError("'spec' is not a valid ModuleSpec type but 'None'.")
    if spec.loader is None:
        raise ValueError("'spec.loader' is not a valid ModuleSpec type but 'None'.")
```

src/xclim/ensembles/_base.py
Zeitsperre, line 405:
```suggestion
   calendar : str, optional
       The calendar of the time coordinate of the ensemble.
       For conversions involving '360_day', the align_on='date' option is used.
       See :py:func:`xclim.core.calendar.convert_calendar`.
       'default' is the standard calendar using np.datetime64 objects.
```

src/xclim/ensembles/_robustness.py
```diff
@@ -140,12 +141,12 @@ def robustness_fractions(
         is missing data, the realization is invalid.
         Not used if only deltas are passed as `fut`.
     strict_sign : bool
-        Whether to include zeros When determining the sign of change. True (default) does not include
+        Whether to include zeros when determining the sign of change. True (default) does not include
```
Zeitsperre: Need to promote Canadian English where possible ;)
```suggestion
        Whether to include zeroes when determining the sign of change. True (default) does not include
```

src/xclim/ensembles/_robustness.py
```diff
@@ -564,8 +565,8 @@ def diff_cdf_sq_area_int(x1, x2):
     R.attrs.update(
         name="R",
         long_name="Ensemble robustness coefficient",
-        description="Ensemble robustness coefficient as defined by Knutti and Sedláček (2013).",
-        reference="Knutti, R. and Sedláček, J. (2013) Robustness and uncertainties in the new CMIP5 climate "
+        description="Ensemble robustness coefficient as defined by Knutti and Sedlacek (2013).",
+        reference="Knutti, R. and Sedlacek, J. (2013) Robustness and uncertainties in the new CMIP5 climate "
         "model projections. Nat. Clim. Change.",
```
Zeitsperre, line 571: Does the UTF-8 encoding mess up when writing this to the DataArray?

.pre-commit-config.yaml
```diff
@@ -63,6 +63,12 @@ repos:
     rev: v2.16
     hooks:
       - id: vulture
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v2.3.0
+    hooks:
+      - id: mypy
+        exclude: ^tests/|^docs
+        additional_dependencies: [ 'numpy>=1.21' ]
```
Zeitsperre, line 71: We could probably bump this to 2.1, no?
```suggestion
        additional_dependencies: [ 'numpy>=2.1' ]
```

Zeitsperre (commented):
First pass at this. Thanks so much for this enormous amount of work. This has been on the backburner for years.

src/xclim/compute/_agro.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/_threshold.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/clix.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/run_length.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/run_length.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/run_length.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/run_length.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/4b4c70ea951523d5ef999bdedcf9ea3f23f96938.

src/xclim/compute/converters.py
ludwiglierhammer: You are right, Resolved in https://github.com/Ouranosinc/xclim/pull/2425/commits/c05fda7a7ab583555f7d2b5b4b7c992226929d58.

src/xclim/compute/helpers.py
ludwiglierhammer: Absolutely. Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/09a743f0cd2f9978414eac6c6275555f254ccb22.

src/xclim/compute/stats.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/09a743f0cd2f9978414eac6c6275555f254ccb22.

src/xclim/core/utils.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/09a743f0cd2f9978414eac6c6275555f254ccb22.

src/xclim/ensembles/_base.py
ludwiglierhammer: Resolved in https://github.com/Ouranosinc/xclim/pull/2425/changes/09a743f0cd2f9978414eac6c6275555f254ccb22.

src/xclim/compute/generic.py
ludwiglierhammer: I am not quiet sure. I mostly did this to resolve the mypy issues.

src/xclim/ensembles/_robustness.py
ludwiglierhammer: I changed this to resolve the numpydoc-validation issue.

.pre-commit-config.yaml
ludwiglierhammer: I bumped the version to 2.1 in https://github.com/Ouranosinc/xclim/pull/2425/changes/f3548a34ac38ba385003a3b2b58514dc0666a5cf.

src/xclim/compute/generic.py
Zeitsperre: @aulemahal Any thoughts?

src/xclim/ensembles/_robustness.py
Zeitsperre: I'm going to look into this now. It shouldn't be complaining over a valid use of UTF-8.

ludwiglierhammer:
In addtion to https://github.com/ludwiglierhammer/xclim/pull/3, I'll create a new PR where I'll have a look at the "Quantified" isse.

src/xclim/compute/fire/_cffwis.py
```diff
@@ -587,7 +596,7 @@ def _fire_season(
     tas : array_like
         Temperature [degC], the time axis on the last position.
     snd : array_like, optional
-        Snow depth [m], time axis on the last position, used with method == 'LA08'.
+        Snow depth [m], time axis on the last position, used with method in ['LA08', 'GFWED'].
```
aulemahal, line 599:
```suggestion
        Snow depth [m], time axis on the last position, used with methods "LA08" and "GFWED".
```
Nitpick, but docstring are meant for humans, this seemed clearer.

src/xclim/compute/fire/_cffwis.py
```diff
@@ -1687,7 +1698,8 @@ def _apply_fire_season(ds, **kwargs):
 
         return season_mask
 
-    ds = convert_units_to(tas, "degC").rename("tas").to_dataset()
+    da: xr.DataArray = convert_units_to(tas, "degC")
```
aulemahal, line 1701: Ugh. I understand where this is coming from, but I find the `name: TYPE = ...` pattern to be quite verbose and heavy. Is this only due to the polymorphism of `convert_units_to` ?

Xarray uses `overload` , I think to solve this issue, is this something we could explore ? It concentrates the type-checker-related code to a single function, not to every call of it.

src/xclim/compute/fire/_ffdi.py
```diff
@@ -42,7 +42,7 @@
     nopython=True,
     cache=True,
 )
-def _keetch_byram_drought_index(p, t, pa, kbdi0, kbdi: float):  # pragma: no cover
+def _keetch_byram_drought_index(p, t, pa, kbdi0, kbdi: dict[float, float]):  # pragma: no cover
```
aulemahal, line 45: `kdbi` is the output array I believe, should be a numpy array, no ?

src/xclim/compute/_agro.py
```diff
@@ -155,8 +155,8 @@ def huglin_index(
     thresh: Quantified = "10 degC",
     method: Literal["huglin", "interpolated", "jones"] = "huglin",
     cap_value: float = 1.0,
-    start_date: str | DayOfYearStr = "04-01",
-    end_date: str | DayOfYearStr = "10-01",
+    start_date: DayOfYearStr | Literal["default"] = "default",
```
aulemahal, line 158: What was the issue with "04-01" ? Isn't it a valid `DayOfYearStr` ?

Reading below, I see that it seems a `DayOfYearStr` needs to be explicitly created ? I would argue that if this is the case, the whole idea here is to be rethought. The goal was to include information about how the string is structured in the annotation, but without making implementing indices heavy and code hard to read (for humans).

Is there a way to have an annotation that mypy simply understands a an alternative name for "str"?
Or at least, can't we put `DayOfYearStr("04-01")` as the default directly in the signature ?

Otherwise, if the "casting" is necessary for mypy, then I would prefer reverting to a simpler syntax, even if it means decreasing precision of the indicator parameters system. Like: 
```python3
def func(date: str):
    """
    A func

    Parameters
    ----------
    date : str
        A date, given as a day-of-year string in MM-DD format.
    """
    ...
```
i.e. don't bother with fancy annotations.

src/xclim/compute/_agro.py
```diff
@@ -227,33 +227,60 @@ def huglin_index(
...
+        ----------
+        value : str
+            Method name to validate.
+
+        Returns
+        -------
+        bool
+            True if `value` is one of `"huglin"`, `"interpolated"`, or `"jones"`, otherwise ``False``.
+        """
+        return value in {"huglin", "interpolated", "jones"}
```
aulemahal, line 246: This is really verbose just for a simple check, what was the issue with  simply doing `if method in {"huglin", "interpolated", "jones"}` ?

aulemahal (commented):
First pass. I only looked at the first 3 files, but I raised concerns for me.

To be fully honest (and @Zeitsperre knows my personal opinion on the matter), I've always found that polymorphism and duck-typing was a strength of Python. I understand concerns about potential errors being uncaught because of type mismatch, but xclim is also not a bank backend. Furthermore, I usually aim to write code that humans can read (with relative success, but that's not the point ;) ), and the interpreter of course. My vision is usually for the code to be editable by another human, and not for it to be readable by a linter.

This said, I can see how this PR can increase overall code quality for xclim, and I am really thankful for all the work you put in @ludwiglierhammer !  So first a philosophical question :  what is the goal of adapting the code to mypy ? What does this linter tool being successful brings to user's quality-of-life ?

I think the issues I raise in this review show what kind of patterns I am questioning here. One of my least favorite pattern in this kind of endeavour is doing things like:
```python
def func(pr: xr.DataArray):
    pr = method(pr)
# being rewritten as :
def func(pr: xr.DataArray):
    _pr : xarray.DataArray = method(pr)
```
I don't see why "overwriting" the "pr" pointer is an error, seems to me a valid use of python's local scope. And I believe the type of `pr` is not that relevant here, as long as the code is contextually clear and tested.

src/xclim/compute/fire/_cffwis.py
Zeitsperre: It is specifically because of the polymorphism of `convert_units_to`. If we were change the retrun type to something that reflects the same type as the inputs, e.g.
```python
def convert_units_to(
    source: Quantified_or_Dataset
    target: Quantified | pint.Unit | dict,
    context: Literal["infer", "hydro", "none"] | None = None,
) -> Quantified_or_Dataset:
```
This would likely solve this issue. A str being passed would expect to then return a str. So long as the variable being passed is clear (xr.DataArray), the return would be clear as well.

src/xclim/compute/_agro.py
Zeitsperre: I've wondered about this as well. I know there are ways to validate a str-like variable in other languages using length or regex, but I don't think there's a comparison in Python. Having `DayOfYearStr` in the call signature gives the impression that we expect users to pass those. Perhaps we could use that internally?

I like the idea of validating within the function using regex (`^\d{2}\-\d{2}$`) instead of defining a new str type. This could be turned into a `DayOfYearStr` when passing it to internal functions?

aulemahal: I would argue that the "validation" is the fact that it makes the parsing code fail or not, here, in the case of `DateStr` and `DayOfYearStr` : https://github.com/Ouranosinc/xclim/blob/10a10ccd2d277b0b5cb916389e83676fef1243db/src/xclim/compute/run_length.py#L1640-L1646

You're right that using a special type as annotation might not be the best way to achieve the goal. In my view, the first goal is to inform the user about the required form that the string could take. If this is better achieved by the docstring, than maybe that's enough ?

The second goal was to inform the indicator constructor so that is can expose a meaningful value as the `Indicator.parameters[argname].kind`. But xclim never really makes use of that information (it only uses the `kind` to differentiate variables, optional variables and the rest), it was meant for "external" wrappers, mainly Finch. Those could probably live with a `str` annotation, or we could probably have other ways the indicator can guess this, if it can lower the need for adding type-checker code ?

ludwiglierhammer:
> First pass. I only looked at the first 3 files, but I raised concerns for me.
> 
> To be fully honest (and @Zeitsperre knows my personal opinion on the matter), I've always found that polymorphism and duck-typing was a strength of Python. I understand concerns about potential errors being uncaught because of type mismatch, but xclim is also not a bank backend. Furthermore, I usually aim to write code that humans can read (with relative success, but that's not the point ;) ), and the interpreter of course. My vision is usually for the code to be editable by another human, and not for it to be readable by a linter.
> 
> This said, I can see how this PR can increase overall code quality for xclim, and I am really thankful for all the work you put in @ludwiglierhammer ! So first a philosophical question : what is the goal of adapting the code to mypy ? What does this linter tool being successful brings to user's quality-of-life ?
> 
> I think the issues I raise in this review show what kind of patterns I am questioning here. One of my least favorite pattern in this kind of endeavour is doing things like (the same for `DateStr`):
> 
> ```python
> def func(pr: xr.DataArray):
>     pr = method(pr)
> # being rewritten as :
> def func(pr: xr.DataArray):
>     _pr : xarray.DataArray = method(pr)
> ```
> 
> I don't see why "overwriting" the "pr" pointer is an error, seems to me a valid use of python's local scope. And I believe the type of `pr` is not that relevant here, as long as the code is contextually clear and tested.

Thanks @aulemahal for your honest review. From my view, it is quiet simple. I saw #1719 and thought I might be able to help. My first goal was to fix the mypy issues (which I managed to do). However, as you have desribed, some fixes have led to other issues. As you mentioned in your review, there are a few parts in the code that are directly related to this issue and that we could still update or improve:

* Make the docstrings more human readable: https://github.com/ludwiglierhammer/xclim/pull/3
* Think about `Quantified` and `convert_units_to`  which hopefully covers the example in your post and some other issues that were disscussed here.
* How to use `DayOfYearStr`? The workaround with defaults is not that satisfying. I'd prefer something like this:
  ```python
  DayOfYearStr: TypeAlias = str

  def func(date: DayOfYearStr | None = "01-01"):
  ```
  Additionally, we could implement a decorator that verifies the format (here is a pretty simple example):
  ```python
  def check_day_of_year(func):
      @wraps(func)
      def wrapper(day: DayOfYearStr = "01-01"):
          datetime.strptime(day, "%m-%d")
          return func(day)

      return wrapper
      
  @check_day_of_year
  def func(date: DayOfYearStr | None = "01-01"):
  ```
* This is a very good point. I can try to sort out this workaround to resolve the mpyp issues, and then revert to the previous version.
  ```python
    def func(pr: xr.DataArray):
       pr = method(pr)
   # being rewritten as :
   def func(pr: xr.DataArray):
       _pr : xarray.DataArray = method(pr)
   ```
The question is now if we want to solve all of theses issues in this PR? I really hope we find a solution that everyone is happy with.

src/xclim/compute/fire/_cffwis.py
ludwiglierhammer: As mentioned earlier we should change something in the logic of `Quantified` and `convert_units_to`. That's on my bucket list!

src/xclim/compute/fire/_cffwis.py
ludwiglierhammer: I try to cover this in https://github.com/ludwiglierhammer/xclim/pull/3.

src/xclim/compute/fire/_ffdi.py
ludwiglierhammer: I fixed this in https://github.com/ludwiglierhammer/xclim/pull/3.

src/xclim/compute/_agro.py
ludwiglierhammer: You are absolutely right that is not a nice solution. I fixed this in https://github.com/ludwiglierhammer/xclim/pull/3 using `typing.cast`.

Zeitsperre:
> * Make the docstrings more human readable: [Update docstrings and keep them both mypy and numpydoc-validation compliant ludwiglierhammer/xclim#3](https://github.com/ludwiglierhammer/xclim/pull/3)

I like what I'm seeing in that PR. @aulemahal, you should give it a look if you have the time. 

> * Think about `Quantified` and `convert_units_to`  which hopefully covers the example in your post and some other issues that were disscussed here.

Solving this issue would remove a lot of the explicit typing that we're running into issues with, I imagine. I'm a lot less hesitant to integrate the changes as is, seeing as we're still in the "breaking things" stage when it comes to xclim v1. If we think we can solve this problem in the next few months, we can even hold off on the release (we currently don't expect to have a full release until maybe November/December at the latest).

> * How to use `DayOfYearStr`?

We were discussing this off GH, and one idea I had was that this could be something used within internal functions. Building on what @aulemahal mentioned above, we could simply pass this type of object around specifically targeted for indicators and other internal functions.

I do like the idea of having a check that is similar to the variable annotation; We adopted a similar pattern with decorators in `CLISOPS` for validating lon/lat coordinates. Perhaps that could be a reasonable approach!

> * This is a very good point. I can try to sort out this workaround to resolve the mpyp issues, and then revert to the previous version.
>   ```python
>    def func(pr: xr.DataArray):
>        pr = method(pr)
>    # being rewritten as :
>    def func(pr: xr.DataArray):
>        _pr : xarray.DataArray = method(pr)
>   ```

The only places where I believe the underline variable with annotation is absolutely needed is when we pass an `Option` to a function, e.g. (`pr: xr.DataArray | None`). For most functions, if we're comfortable with changing the value of the object passed to the function (debatable), we can probably get away with just reassigning (`pr = method(pr)`).

> The question is now if we want to solve all of theses issues in this PR? I really hope we find a solution that everyone is happy with.

This is an immense and super helpful PR. I think we're close to having something mergeable!

src/xclim/ensembles/_reduce.py
```diff
@@ -379,16 +379,16 @@ def kmeans_reduce_ensemble(
...
         else:
-            argmax = 0
+            argmax = np.argmax(0)
 
         r_clust = r[clusters == i]  # index of the cluster simulations within the full ensemble
 
         out[i] = r_clust[argmax]
 
-    out = sorted(out.astype(int))
+    output = sorted(out.astype(int))
```
coxipi: so we can't overwrite `out` here because we change the type?

src/xclim/ensembles/_reduce.py
```diff
@@ -379,16 +379,16 @@ def kmeans_reduce_ensemble(
             argmax = np.argmax(like)  # index of the maximum likelihood
 
         else:
-            argmax = 0
+            argmax = np.argmax(0)
```
coxipi: We don't want polymorphism for intermediary variables, is that it?

coxipi: i.e. `int` vs. `np.int64` in this case

src/xclim/ensembles/_robustness.py
```diff
@@ -31,7 +32,7 @@
 ]
 
 
-SIGNIFICANCE_TESTS = {}
+SIGNIFICANCE_TESTS: dict[str, Callable] = {}
```
coxipi: So empty lists and dictionaries should be declare what will fill them.

src/xclim/ensembles/_reduce.py
```diff
@@ -307,11 +307,11 @@ def kmeans_reduce_ensemble(
         )
     """
     if make_graph:
-        fig_data = {}
+        fig_data: dict[str, Any] = {}
         if max_clusters is not None:
             fig_data["max_clusters"] = max_clusters
     else:
-        fig_data = None
+        fig_data = {}
```
coxipi: It feels like this breaks a rule, based on what I gathered reading the PR so far. Earlier I saw we declare `np.argmax(0)` instead of `0` presumably to have consistent types. But, depending on the branch, `fig_data` will either be `dict[str, Any]` or `dict[]`, which is not the same. Should this be illegal under mypy? 

And if we say this is not a problem, those types are equivalent, then I believe the second line should be: 

``` python
fig_data: dict[str, Any] = {}
```

src/xclim/compute/_agro.py
```diff
@@ -227,33 +227,60 @@ def huglin_index(
...
+
     if not isinstance(freq, str):
         raise TypeError("Freq must be a string.")
 
-    _tas = convert_units_to(tas, "degC")
-    _tasmax = convert_units_to(tasmax, "degC")
-    _thresh = convert_units_to(thresh, "degC")
+    _tas: xarray.DataArray = convert_units_to(tas, "degC")
+    _tasmax: xarray.DataArray = convert_units_to(tasmax, "degC")
+    _thresh: float = convert_units_to(thresh, "degC")
```
coxipi: If I understood correctly, we change the variable name if the type is changed. But, `tas` and `tasmax` are still the same type in this case. Should they not keep the same name? If we follow this rule, adding an underscore for changed types, then I feel that `tas` and `tasmax` should keep the same name, otherwise it would create more confusion (I would wonder: what? has the type changed?)

It is an object related to the same class, I don't think changing an attribute should count for adding an underscore? Or is the underscore really used for anything that is being overwritten, even if the type is the same??

src/xclim/compute/_agro.py
```diff
@@ -661,15 +693,16 @@ def dryness_index(  # numpydoc ignore=SS05
 
     has_north, has_south = False, False
     if lat is None:
-        lat = _gather_lat(pr)
+        lat = _gather_lat(_pr)
```
coxipi: In this case, `lat` would be changed? So should this not be renamed `_lat`?

src/xclim/compute/_agro.py
```diff
@@ -1296,7 +1342,7 @@ def effective_growing_degree_days(
     *,
     thresh: Quantified = "5 degC",
     method: Literal["bootsma", "qian"] = "bootsma",
-    after_date: DayOfYearStr = "07-01",
+    after_date: DayOfYearStr | Literal["default"] | None = "default",
```
coxipi: 1. I prefer "07-01" than something generic like "default". Any string argument could be defaulted to "default" and changed inside the function, but why do this when we can set the default directly in the signature?

2. Is that related to an interaction between `mypy` and `DayOfYearStr`? If so, I don't know what will be the correct way to handle this, but this current solution doesn't feel right.

src/xclim/compute/_agro.py
```diff
@@ -1319,8 +1365,9 @@ def effective_growing_degree_days(
         For "bootsma", the start date is defined as 10 days after the average temperature exceeds a threshold.
         For "qian", the start date is based on a weighted 5-day rolling average,
         based on :py:func`qian_weighted_mean_average`.
-    after_date : str
+    after_date : DayOfYearStr, optional, defaults to '07-01'
```
coxipi: Is this something new, saying what the default is on the typing line, is it on purpose? This information would usually go on the next line, description of the variable, no?

src/xclim/compute/_anuclim.py
coxipi: Here `_np_ops` is not modified anywhere. There is no ambiguity here, the RHS is there, clearly a `Dict[str,str]`. Am I understanding that any variable we declare should come with a type? 

Should I declare?
```python
i : int = 0 
```

Somewhere else, we define: 
```python
has_north = some_func(...)
```
by this logic, should we declare this as: 
```python
has_north: bool = some_func(...)
```
or we say that the function is sufficient to indicate this is a bool?

src/xclim/compute/classify.py
```diff
@@ -102,20 +102,23 @@ def get_zones(
...
-        bins = sorted([convert_units_to(b, da) for b in bins])
+    if bins is None:
+        zone_bins = _get_zone_bins(zone_min, zone_max, zone_step)
+    else:
+        zone_bins = bins
+    if isinstance(zone_bins, list):
+        zone_bins = sorted([convert_units_to(b, da) for b in zone_bins])
     else:
-        bins = convert_units_to(bins, da)
+        zone_bins = convert_units_to(zone_bins, da)
```
coxipi: previous we changed the name of `tas` to `_tas`. Should we also change the name of `_zone_bins` here?

coxipi: It's also a polymorphous function. Sometimes we would write
```python
_tas: xr.DataArray = convert_units_to(tas, da)
```
Should we also specify the typing here?

src/xclim/compute/converters.py
```diff
@@ -2038,25 +2050,25 @@ def potential_evapotranspiration(  # pylint: disable=too-many-statements
...
         ab = tr - 0.0123 * _pr
-        pet = 0.0013 * ra * (_tas + 17.0) * ab**0.76
+        pet = 0.0013 * _ra * (_tas + 17.0) * ab**0.76
         pet = xr.where(np.isnan(ab**0.76), 0, pet)
         pet = pet.clip(0)  # mm/month
 
     elif method in ["mcguinnessbordne05", "MB05"]:
         if tas is None:
             _tasmin = convert_units_to(tasmin, "degC")
             _tasmax = convert_units_to(tasmax, "degC")
```
coxipi: specify typing?

coxipi:
Hi, thanks for this contribution! 

* It would be good if we can avoid the typing `xr.DataArray` specification with `convert_units_to`
* Do we really need to rename variable when the type is unchanged? 
* If I declare a dictionnary like this
```python
_np_ops = {'a':'1'}
```
I don't think I should need to declare the type `dict[str, str]`. It's already explicit. Are there some mypy requirements that are more optional than others? The use I can see for this is indeed for empty collections to be filled:
```python
_np_ops: Dict[str,str] = {}
```
Or, if we want to allow an argument to change type: 
`thresh: str | float = " 1 mm d-1"`
so we could have 
```python
thresh = convert_units_to(thresh, units)
```
although we never really define `thresh` in the function, it's a user input

coxipi:
Overall, I think this adds a mental cost of parsing through code written like this, but it should diminish with time. Do we still have gains if we are not 100% mypy compliant? I get the impression that we will not start writing simple variables as:
```python
i: int = 0
```
so where do we draw the line?

PR #2429, opened 2026-09-17
Fix indicator copy with some defaults - document registries

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* My bad, the test verifying that either the identifier is given or register is False was not assuming that `register` is True by default, this is fixed here.
* Also moved `copy` to `Indicator` and added default values for `identifier` and `register`, properly documenting that at least one must be given.
* Fixed the test accordingly.

Also, I realized the registries were not documented. This was harder than I though because sphinx can't properly use the `autodata` directive if the object is not referred by its exact path. So in order to use it, we would need to refer the `registry` in the doc as `xclim.core.indicator._indicator.registry`, although we want users to use `xclim.core.indicator.regisry`.  This is an old issue : https://redirect.github.com/sphinx-doc/sphinx/issues/6495

The alternative was to put the documentation directly in the RST file.

No. It fixes one.


Zeitsperre (approved):
I don't fully get the changes, but LGTM

src/xclim/core/indicator/_indicator.py
```diff
@@ -1584,7 +1569,7 @@ class _Registrer(_Convenience):
     """Register the indicator in the xclim registry."""
 
     def __new__(cls, **kwargs):
-        if kwargs.get("identifier") is None and kwargs.get("register") is True:
+        if kwargs.get("identifier") is None and kwargs.get("register", True) is True:
```
aulemahal, line 1572: @Zeitsperre  the main issue was here.

This test was using `kwargs.get("register")` for testing argument validity, but the actual registering below was using `kwargs.get("register", True)`, which changes the behaviour when `register` is not passed!

The other main change is to add default arguments to `copy`, so it's better documented. This way it's also less likely that an indicator is created without `register` being passed to `__new__`.

PR #2431, opened 2026-09-21
Put history back on DataArray outputs as `xclim_description` - Fix units metadata handling

aulemahal:
### Pull Request Checklist:
    - This PR fixes #xyz

* Puts a `history` on DataArray outputs, as before. Dataset outputs only get one global history, their variables do not get one. DataArray outputs do not get a "merge" of inputs' history, only the xclim part.
* Moves the history -editing methods to `_MetadataFormatter`
* Adds a new `base_variables` attribute to all indicator outputs, listing xclim variables names of the inputs.

No.

### Other information:
@RondeauG @juliettelavoie and anyone else. This currently adds a space-separated list (common in cf) of input variables, using the indicator's argument names. This means it's `da` for generic indicator, but also means it's `tas` even if the input was named `température_de_lair_à_2m`.

Is this sufficient ? I also thought of doing something like : `{arg_name}={input name}`. So a generic could look like `da=tas` and a funky one like my example : `tas=température_de_lair_à_2m`. Opinions ?


RondeauG:
I suggested using the `name` property to try to circumvent the prevalence of `da` in generic indicators, so I think that I'd be in favour of your 2nd option? Not a strong opinion however. AFAIK, the old `history` attribute would not have been better in that regard.

juliettelavoie:
Given that we have general indicators with da, which is not very helpful, I like the second option also. but maybe would call it `input_variables` as `base` is used for soemthing else in the yamls

coxipi:
I'd go with the information that can be deduced from the signature? 
e.g.
```python
def heat_wave_frequency(
    tasmin: xarray.DataArray,
    tasmax: xarray.DataArray,
    ...
```

would yield `'tasmax tasmin'`. I think this is what your `das.keys()` accomplishes? I like the simplicitly of it. 

Oops, didn't realize I had opposite opinion from xscen power users. Since this is something that will be used in xscen I think their opinion should prevail

aulemahal:
So here is my latest suggestion : attribute `xclim_description` that is essentially the main part of the history string. So the "call" string and the options. It is added to all output variables.

It is not using the `history` name and not tracking any parent's history, so the « not CF compliant » argument I made before is satisfied. But there's no loss of information with xclim v0.

aulemahal:
And sorry for the unrelated business, I fixed something in the temperature handling as I realized the metadata set by the compute function was not following through in the indicator when I edited the usage notebook.

So this PR also handles the `delta_` temperature units more explicitly by looking for the 3 known flavors (°C, °F, °Re) when detecting metadata automatically. Also, if the target units have "unknown" metadata but the source is known, the result of `convert_units_to` preserves the source's metadata.

src/xclim/core/units.py
```diff
@@ -233,28 +241,32 @@ def pint2cfattrs(value: pint.Quantity | pint.Unit, is_difference=None) -> dict[s
         Input unit.
     is_difference : bool
         Whether the value represent a difference in temperature, which is ambiguous in the case of absolute
-        temperature scales like Kelvin or Rankine. It will automatically be set to True if units are "delta_*"
-        units.
+        temperature scales like Kelvin or Rankine. Default is to set to True if units contains difference units,
+        or to False if the units are the on-scale units (°C, °F or °Re). Otherwise, if `value` has temperature
+        dimensions, "unknown" is given as the "units metadata" (this usually happens with K or °R).
```
juliettelavoie: I don't understand this description. what is the difference between being "the on-scale units" and "have temperature units"?

juliettelavoie (approved):
I think this is a good solution!

src/xclim/core/units.py
aulemahal: `K d` has temperature dimensions but is not a "on-scale unit" and does not "contain difference units", in this case, units metadata are therefore "unknown". I.e., this mostly applies to composed units.

haha, I had difficulties writing this sentence and I see it's still not very clear, I'll try again.

src/xclim/core/units.py
```diff
@@ -697,7 +711,9 @@ def to_agg_units(
         statistic = statistic.__name__
 
     is_difference = (
-        True if statistic in ["std", "var"] or "difference" in orig.attrs.get("units_metadata", "") else None
+        True
+        if statistic in ["std", "var", "integral"] or "difference" in orig.attrs.get("units_metadata", "")
+        else None
     )
```
Zeitsperre, line 723: Why not just make this a `bool` instead of `bool | None`?

src/xclim/core/units.py
juliettelavoie: ah! I think adding the example would be helpful

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Major changes
     * The ``xclim.core.indicator.registry`` now holds ``Indicator`` _instances_ (not classes) and is case-insensitive.
     * "Virtual submodules" were transformed into ``xclim.core.collection.IndicatorCollection`` instances (and not actual python modules). Indicators created this way automatically have the collection's name prepended to their identifier. Indicators created in the context of a collection are not registered by default. (:pull:`2415`).
     * Copying/Subclassing an existing indicator is now done through ``Indicator.copy``, method ``Indicator.from_dict`` is deprecated. (:pull:`2424`).
+    * Indicators will now add a ``xclim_description`` attribute to all their outputs, This is similar to the ``history`` that is now only put on output datasets, but excludes datetime, xclim's version and does not append any input metadata. (:pull:`2431`).
```
```suggestion
    * Indicators will now add an ``xclim_description`` attribute to all their outputs, This is similar to the ``history`` that is now only appended to output datasets, but excludes datetime and `xclim`'s version. It does not append any input metadata. (:pull:`2431`).
```

docs/notebooks/usage.ipynb
```diff
@@ -77,6 +77,8 @@
     "\n",
     "Finally, like almost all indicators, the function takes a `freq` argument to specify over what time period it is computed. These are called \"Offset Aliases\" by Pandas and are the same as the resampling string arguments (in `xr.DataArray.resample`). Valid arguments are detailed in [pandas docs](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases) (note that aliases involving \"business\" notions are not supported by `xarray` and thus could raise issues in xclim).\n",
     "\n",
+    "**Attributes:** xclim added a global `history` attribute to the dataset, as well as other attributes to the output variable : `standard_name`, `cell_methods`, `long_name` , `description`, the latter two formatted to include some information about the inputs. Of course, xclim updated the `units`, but also added `units_metadata`, required by the CF conventions to precise how temperature units should be interpreted. Finally, it added a `xclim_description` attribute that summarizes the computation done, similar to the history but without all the datetime, version and provenance tracking.\n",
```
Zeitsperre:
```suggestion
    "**Attributes:** `xclim` added a global `history` attribute to the dataset, as well as other attributes to the output variable: `standard_name`, `cell_methods`, `long_name` , `description`, with the latter two formatted to include some information about the inputs. Of course, `xclim` updated the `units`, but also added `units_metadata`, required by the CF-Conventions to specify how temperature units should be interpreted. Finally, it added an `xclim_description` attribute that summarizes the computation done, similar to the `history` but without all the datetime, version, and provenance tracking.\n",
```

CHANGELOG.rst
```diff
@@ -40,6 +41,12 @@ New indicators and features
     * New ``bases`` section for defining indicator classes to reuse in defining the ones in the ``indicators`` section (:pull:`2415`).
     * Fields ``base`` and ``compute`` can now be a "qualified name" of an object to be imported dynamically, or the qualified name of an indicator with its submodule/realm. Ex: ``atmos.precip_accumulation``, which was previously referred by its identifier ``prcptot`` (:pull:`2415`).
 
+Bug fixes
+^^^^^^^^^
+* Handling of ``units_metadata`` has been updated (:pull:`2431`).
```
```suggestion
* Handling of ``units_metadata`` has been updated. (:pull:`2431`).
```

Zeitsperre (approved):

src/xclim/core/units.py
aulemahal: None means « we can't know from the statistic requested, let `pint2cfattrs` guess it ».

